PR #38, opened 2020-05-30
Shared prefs ktx shorthand

jakobmoellerdev:


htooisap (dismissed):

timbrueggenthies:
Thanks @jakobmoellersap for working on the issue. 👍 
I just noticed that previously the edits on the SharedPrefs were comleted with the `commit` call, which writes the changes synchronously to the disk. The KTX edit function uses `apply` by default, which synchronously writes the changes to the in-memory SharedPrefs, and then writes it async to the disk. [More info](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()) on the difference.
If it matters, it could be switched back to like this:
```kotlin
getSharedPreferenceInstance().edit(commit = true) {
    putString("key", value)
}
```

harambasicluka (dismissed):

jakobmoellersap:
Will recheck regarding commit.

jakobmoellersap:
I checked this again, and I think it makes sense to use commit to make sure that we persist correctly. We have no timely critical application usage here and some ms diff on a commit will not kill us. A inconsistency after a crash before the commit happens will.

harambasicluka (dismissed):

pwoessner (dismissed):

marcmuschko (approved):

htooisap (approved):

timbrueggenthies:
Sorry to bring this up again, but some changes are missing the removal of the additional call to `.commit()` (e.g. [here](https://github.com/corona-warn-app/cwa-app-android/blob/0302bf9da70d63babf8fc3a30dbaa20332f157ab/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt#L117) or [here](https://github.com/corona-warn-app/cwa-app-android/blob/0302bf9da70d63babf8fc3a30dbaa20332f157ab/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt#L148)). There are more places in this file where the call was left.
This shouldn't result in an error, but an unnecessary call is happening.

jakobmoellersap:
Thanks for pointing it out, I will go over this again :)

jakobmoellersap:
Please reference the latest PR #132  where I removed all double commits that I was aware of in LocalData

PR #46, opened 2020-05-31
Changed text type for TAN input to textPassword (closes #43)

### LINKED ISSUE #43: Use android:inputType=textVisiblePassword for tan input edittext
## Feature description
Currently,  `android:inputType=text` is used for the tan input edittext. This has the consequence that the user's soft keyboard shows suggestions and may save the TAN to a dictionary.

With `android:inputType=textVisiblePassword`, no suggestions are shown and the keyboard shouldn't save it in its dictionary.

## Problem and motivation
The autocompletion of the softkeyboard may alter the TAN the user has typed in.

## Is this something you're interested in working on

No

ironjan:
* Closes #43 
 * TAN input was using text type which provided the user with input suggestions
 * Set the input type to `textPassword` instead of `textVisiblePassword` since the view is not visible per se (`width = heigth = 1dp`, `alpha=0`)


ironjan:
Here are two screenshots to show the behavioral difference. Old is left, new is right.

Unbenannt

m1ga:
looking at those last screenshots: the background color in dark-mode needs to be changed fix: https://github.com/corona-warn-app/cwa-app-android/pull/51 :smile:

kolyaopahle (dismissed):
looks good to me 👍

ironjan:
FYI: @kolyaopahle I rebased on current master and your review got invalidated.

kolyaopahle (dismissed):

harambasicluka (approved):

PR #74, opened 2020-06-01
Add persistance of timestamp and whether user is allowed to submit keys

tklingbeil:
* This timestamp documents when the device has been paired (i.e. when a GUID/teleTAN has been used to retrieve the registration token)
* Store whether user is allowed to submit keys, either due to a positive test result or because a teleTAN has been entered
* Naming fixes


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/SubmissionRepository.kt
```diff
@@ -19,12 +19,14 @@ object SubmissionRepository {
...
-
+        if (testResult == TestResult.POSITIVE) {
+            LocalData.isAllowedToSubmitDiagnosisKeys(true)
+        }
         val initialTestResultReceivedTimestamp = LocalData.inititalTestResultReceivedTimestamp()
 
         if (initialTestResultReceivedTimestamp == null) {
             val currentTime = System.currentTimeMillis()
-            LocalData.initialTracingActivationTimestamp(currentTime)
+            LocalData.inititalTestResultReceivedTimestamp(currentTime)
```
kolyaopahle, line 29: Good catch!

kolyaopahle (approved):
Looking Good, lets merge this 👍

jakobmoellerdev (approved):
Looks good 👍

PR #41, opened 2020-05-30
Add FLAG_SECURE to prevent screenshots (closes #9)

m1ga:
## Checklist

* [x] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [x] If this PR comes from a fork, please [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
* [x] Set a speaking title. Format: {task_name} (closes #{issue_number}). For example: Use logger (closes #41)
* [x] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)
* [ ] Create Work In Progress [WIP] pull requests only if you need clarification or an explicit review before you can continue your work item.
* [x] Make sure that your PR is not introducing _unncessary_ reformatting (e.g., introduced by on-save hooks in your IDE)
* [x] Check our [Contribution Guidelines](https://github.com/corona-warn-app/cwa-app-android/blob/master/CONTRIBUTING.md)

## Description

Adding FLAG_SECURE to block screenshots from being made:

Before:
screen_0

After:
screen_1

Fixes https://github.com/corona-warn-app/cwa-app-android/issues/9

PR #84, opened 2020-06-01
Update Gradle Wrapper files to match defined version

otormaigh:
## Checklist

* [ ] Set a speaking title. Format: {task_name} (closes #{issue_number}). For example: Use logger (closes #41)
* [ ] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)
* [x] Make sure that your PR is not introducing _unncessary_ reformatting (e.g., introduced by on-save hooks in your IDE)
* [x] Check our [Contribution Guidelines](https://github.com/corona-warn-app/cwa-app-android/blob/master/CONTRIBUTING.md)

## Description
Keeping the local Gradle Wrapper script and JAR file upto date with the set version (5.6.4 as of this PR) is imperative when building the project to ensure that the build system is using the expected version of things.

This can be achieve by either using the long form `wrapper` task (`./gradlew wrapper --gradle-version=5.6.4 --distribution-type=all`) or by extending this task and adding it to the root `build.gradle` file which will allow you to run `./gradlew wrapper` and it will pick up the version and distro type from there.

Whichever way is used, this wrapper task should always be done when changing the Gradle version in the project. 

Looks like there have been a few changes to the Gradle version bumping it up and down. This PR keeps the version as the currently defined one, but just updates any other relvant files to match this version.


jakobmoellersap:
We are currently in the thought of switching to Gradle 6.1.1 along with the latest Android Studio Update and will thus keep things like that for a while. We make sure of this by using the integrated "update" feature in Android Studio. I do not think we need to use dedicated hardcoding of the gradle wrapper like you suggest here, but we will keep the option in mind. 

The reason this (referring to the new gradle version) is not in here is because we want to make sure our release builds are also working correctly with an updated Gradle and the original update was thus reverted in order to ensure stability. I put it in by mistake when I updated Android Studio and missed it in my commit log.

If we update to 6.1.1 I will recheck on this PR again and see if it makes sense for us. As long as we do not have the conversion and or the decision wether we stay or upgrade, I do not think we will invest here.

otormaigh:
The hard coded task in the root ‘build.gradle’ is more of a convenience task and not required so it can be removed if necessary.

Either way, when updating the Gradle version, the wrapper task should always be used to make sure all relevant files are updated. Using the Android Studio update process doesn’t do a full update of these files, it’ll just bump the version in the ‘gradle-wrapper.properties‘ file (as can be seen from the other changes in this PR, these changes were made after running the task against the defined Gradle version).

jakobmoellersap:
Solved in #128

PR #88, opened 2020-06-01
Updated strings, introduced more details on app-information screens

marcmuschko:
Updated several strings mostly regarding app-information, risk card and risk details
Updated app-information contact and app-information about


harambasicluka (approved):
Awesome work! :)

PR #96, opened 2020-06-02
Circle progress bar

harambasicluka:
Added the following view:

image


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CircleProgress.kt
```diff
@@ -0,0 +1,96 @@
...
+            strokeWidth = progressWidth
+            isAntiAlias = true
+        }
+        progressPaint =  Paint().apply {
+            color = progressColor
+            style = Paint.Style.STROKE
+            strokeWidth = progressWidth
+            isAntiAlias = true
+        }
+        styleAttrs.recycle()
```
Fabian-K, line 64: There is a little helper method `context.withStyledAttributes` which might be nice in this case

marcmuschko (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CircleProgress.kt
harambasicluka: I'll add this later, thanks for the input :)

PR #115, opened 2020-06-02
Feature: Bluetooth callback

harambasicluka:
Added bluetooth callback and bound it to the ui via the settings repository and settings view model.


marcmuschko (approved):
Nice, finally we got the auto update working !

PR #117, opened 2020-06-02
okhttp stack and cert pinning setup

### LINKED ISSUE #89: Replacing Volley based network stack with a OkHttp based one.
## Current Implementation
The current network stack is based on Volley using a requestQueue. For every request a complex class is defined which declares how the request should be parsed and some other boilerplate code. As far as i analyzed the code none of this requests is used in a asynchronous way. The WebRequestBuilder object do synchronize the requests so they can be used in a serial programming style using coroutines. Btw. a little bit misleading to call the functions e.g.: WebRequestBuilder.asyncGetKeyFilesFromServer because it is not a async call it is a synchronized call. ;-)

## Suggested Enhancement
Use OkHttp to do the network calls with a small layer of serialization helper (Kind of OkHttp rest class to make it easy describe requests in a simple function call) for ProtoBuf and json (Why do you mix JSON and ProtoBuf is this really necessary?).
I use this approach in many projects and the OkHttp helper class will only take about 100 loc. Maybe a little bit more supporting both JSON and ProtBuf and implementing a general error classification system. Than the WebRequestBuilder could be a class which defines a function for each call. We do not need a class for every request anymore.
Sample for this OkHttp helper class is here: https://gitlab.com/timod/compose-playground/-/blob/master/app/src/main/java/de/appsonair/network/RestJsonClient.kt

## Expected Benefits
- Reducing lines of code (Reducing boilerplate code for request definitions).
- Reducing complexity (We do not need to do asnychrouse calls and than synchronize them with suspendCoroutine continuation any more)
- The refactoring will not change the interface ob the WebRequestBuilder

jakobmoellerdev:
Fixes #89


jakobmoellersap:
fixes #89

AlexanderAlferov (approved):

PR #144, opened 2020-06-03
Verification process error handling

tklingbeil:
* Show a dialog if users try to submit their diagnosis keys, even though the exposure notification framework is disabled
* Texts are still WiP


tklingbeil:
I see this has already been handled. Removing the PR.

PR #149, opened 2020-06-03
Replace Volley with Retrofit

jakobmoellerdev:
Complete Layer Replacement of the http package from a Volley-Based Stack towards Retrofit


pwoessner (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/RetryInterceptor.kt
```diff
@@ -0,0 +1,23 @@
+package de.rki.coronawarnapp.http
+
+import android.util.Log
+import okhttp3.Interceptor
+import okhttp3.Request
+import okhttp3.Response
+
+class RetryInterceptor : Interceptor {
```
R4md4c, line 8: AFAIK [OkHttp](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-ok-http-client/-builder/retry-on-connection-failure/) has implicit retry policy, so no need for that interceptor (unless you disable the built-in retry policy).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/ServiceFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+import retrofit2.converter.protobuf.ProtoConverterFactory
+import java.io.File
+import java.util.concurrent.TimeUnit
+
+class ServiceFactory {
+    companion object {
+        /**
+         * 10 MiB
+         */
+        private const val HTTP_CACHE_SIZE = 50L * 1024L * 1024L
```
R4md4c, line 25: That's 50MB no?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/ServiceFactory.kt
```diff
@@ -0,0 +1,121 @@
...
+         * Cache file name
+         */
+        private const val HTTP_CACHE_NAME = "http_cache"
+    }
+
+    /**
+     * List of interceptors, e.g. logging
+     */
+    private val mInterceptors: List<Interceptor> = listOf(
+        HttpLoggingInterceptor().also {
```
R4md4c, line 37: I'd say wrap it around `if (BuildConfig.DEBUG)`, otherwise I think that's a security vulnerability since requests will be logged during the release build and anyone can attach to Logcat to read the output.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/OfflineCacheInterceptor.kt
```diff
@@ -0,0 +1,48 @@
...
+
+class OfflineCacheInterceptor(private val context: Context) : Interceptor {
+    companion object {
+        private const val MAX_AGE = 5
+        private const val MAX_STALE_DAYS = 1
+        private const val MAX_STALE = 60 * 60 * 24 * MAX_STALE_DAYS
+    }
+    override fun intercept(chain: Interceptor.Chain): Response {
+        // Get the request from the chain.
+        var request = chain.request()
```
R4md4c, line 16: OkHttp has a more fluent api for cache control perhaps it's better to use?
https://square.github.io/okhttp/4.x/okhttp/okhttp3/-cache-control/

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/RetryInterceptor.kt
jakobmoellersap: This is an interceptor for every request failure that is not catches by OkHTTP:
Unreachable IP addresses. Stale pooled connections. Unreachable proxy servers.

This means Server Errors are not covered.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/OfflineCacheInterceptor.kt
jakobmoellersap: I will check it out and adapt in the next release if we decide to invest.

pwoessner (dismissed):

pwoessner (dismissed):

pwoessner (approved):

PR #155, opened 2020-06-04
Exception Dialogs

jakobmoellerdev:


Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -554,6 +554,11 @@
 
     <string name="submission_result_button_dialog_close">OK</string>
 
+    <!-- ####################################
+               Generic Error Messages
+        ###################################### -->
+    <string name="errors_storage_headline">Something went wrong</string>
```
harambasicluka, line 560: please use german as default language

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -554,6 +554,11 @@
...
+    <string name="errors_storage_button_positive">Continue</string>
```
harambasicluka, line 561: same here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DialogHelper.kt
```diff
@@ -7,14 +7,34 @@ object DialogHelper {
 
     data class DialogInstance(
         val activity: Activity,
-        val title: Int,
-        val message: Int,
-        val positiveButton: Int,
-        val negativeButton: Int? = null,
-        val cancelable: Boolean = true,
+        val title: String,
```
harambasicluka, line 10: are the other dialogs still working with this change?

harambasicluka (changes requested):

Corona-Warn-App/src/main/res/values/strings.xml
jakobmoellersap: done

Corona-Warn-App/src/main/res/values/strings.xml
jakobmoellersap: done

harambasicluka (dismissed):

harambasicluka (approved):

PR #160, opened 2020-06-04
QR & TeleTan pattern

Fabian-K:
## Checklist

* [x] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)

## Description
This PR updates the required pattern for QR and TeleTan


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/SubmissionService.kt
```diff
@@ -73,12 +73,20 @@ object SubmissionService {
...
+
+        if (scanResult.count { it == SubmissionConstants.GUID_SEPARATOR } != 1)
+            return null
+
+        val potentialGUID = scanResult.substringAfterLast(SubmissionConstants.GUID_SEPARATOR, "")
+
+        if (potentialGUID.isEmpty() || potentialGUID.length > SubmissionConstants.MAX_GUID_LENGTH)
+            return null
+
+        return potentialGUID
```
jakobmoellersap, line 89: This is not a good implementation imo. It uses null return, has too many return statements and is not described enough. I would ask to discuss this again or change the implementation here.

jakobmoellerdev (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/SubmissionService.kt
Fabian-K: @jakobmoellersap thanks for the feedback! `null` signals that the pattern is not valid. Would you prefer splitting it into something like `isValidGUID` returning Boolean and extractGUID returning String?

IMO the multiple returns make it easy to read as each if covers one constraint. Of course I can merge some of them to reduce the number of returns but I´m not sure if this adds benefit

jakobmoellersap: So I think I would create a return value that is able to only return if all conditions are not met. Furthermore you use string operations, so you could make use of lambdas like filtering on the string or the length. That way I think we will be able to reduce the return amounts :)

Regarding Null Return: Either you throw an Exception with a catch around the call or you leave it as is. But first lets try to reduce the return amounts. 👍

jakobmoellerdev (approved):

PR #200, opened 2020-06-05
Revert "Extend teleTAN to ten characters"

tklingbeil:
Reverts corona-warn-app/cwa-app-android#192


tklingbeil:
To assure compatibility, we need to hold back those changes until also the server generating the teleTANs has been modified

jakobmoellerdev (approved):

PR #264, opened 2020-06-07
Exposureapp 1079

Queenhymon:
Fix accessibility issues for
+ onboarding views
+ main views
+ submission views
+ button texts (back, share, menu)
+ silent pages (80%)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
```diff
@@ -74,6 +77,17 @@ class MainFragment : BaseFragment() {
         tracingViewModel.refreshActiveTracingDaysInRetentionPeriod()
         TimerHelper.checkManualKeyRetrievalTimer()
         submissionViewModel.refreshDeviceUIState()
+        main_scrollview.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
```
harambasicluka, line 80: please access all views via bindings
in this case it would be `binding.mainScrollView.xxx`

harambasicluka: that's relevant for all other bindings and files. I won't mark them

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
```diff
@@ -74,6 +77,17 @@ class MainFragment : BaseFragment() {
...
+    }
+
+    override fun onStart() {
+        super.onStart()
+        main_scrollview.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
+    }
+
+    private fun setContentDescription() {
+        main_header_share.setContentDescription("Teilen")
```
harambasicluka, line 89: please don't use strings directly, always put them in the strings.xml.

relevant for all strings

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingTracingFragment.kt
```diff
@@ -65,7 +67,21 @@ class OnboardingTracingFragment : BaseFragment(),
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         setButtonOnClickListener()
+<<<<<<< HEAD
```
harambasicluka, line 70: please resolve

Corona-Warn-App/src/main/res/layout/fragment_main.xml
```diff
@@ -22,6 +22,8 @@
     </data>
 
     <ScrollView
+        android:id="@+id/main_scrollview"
+        android:contentDescription="@string/heading_main"
```
harambasicluka, line 26: the string should be `main_title`

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -2786,6 +2787,9 @@ as modifying the License.
     <!-- YTXT: Body text for operating hours in contact page-->
     <string name="submission_contact_operating_hours_body">Unsere Öffnungszeiten:\nMo - Fr: 08:00 - 22:00 Uhr\nSa - So: 10:00 - 22:00 Uhr\nEs gelten die Preise Ihres\nTelefonanbieters.</string>
 
+    <!-- YTXT: Content Description for Accessibility-->
+    <string name="submission_contact_step_1_content">Im ersten Schritt Hotline anrufen und TAN erfragen, unter der Rufnummer 0800 7540002. Die Öffnungszeiten sind Montag bis Freitag von 8 bis 22 Uhr sowie Samstag und Sonntag von 10 bis 22 Uhr. Es gelten die Preise ihres Telefonanbieters.</string>
+    <string name="submission_contact_step_2_content">Im Zweiten Schritt registrieren Sie den Test per TAN-Eingabe in der App</string>
```
harambasicluka, line 2792: please add comment like in line 2790 also to this line

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -227,6 +227,7 @@
     <!-- NOTR todo insert english link in translation file https://www.bundesregierung.de/corona-warn-app-faq-englisch -->
     <string name="main_about_link">https://www.bundesregierung.de/corona-warn-app-faq</string>
 
+    <string name="hint_external_webpage">Häufige Fragen, hier finden Sie Antworten auf häufig gestellte Fragen rund um die Corona-Warn-App. Aufruf externer Infos im Web.</string>
```
harambasicluka, line 230: please add comment

Corona-Warn-App/src/main/res/layout/include_main_card.xml
```diff
@@ -42,6 +42,7 @@
     <androidx.constraintlayout.widget.ConstraintLayout
         android:id="@+id/main_card"
         style="@style/card"
+        android:contentDescription="@string/hint_external_webpage"
```
harambasicluka, line 45: that's an include, that view can be reused, so this attribute should be set where the include is used but not directly on a reusable component. If it's needed exactly in this place pass it in as an attribute.

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -2850,7 +2854,26 @@ as modifying the License.
...
+    <string name="heading_onboarding">Einführung Seite 1 von 5. Gemeinsam Corona bekämpfen</string>
+    <string name="heading_onboarding_tracing">Einführung Seite 3 von 5. Wie Sie die Risiko Ermittlung ermöglichen</string>
+    <string name="heading_onboarding_test">Einführung Seite 4 von 5. Falls Sie positiv getestet wurden</string>
+    <string name="heading_onboarding_notifications">Einführung Seite 5 von 5. Warnungen erhalten und Risiken erkennen</string>
+    <string name="heading_risk_details">Risiko Ermittlung</string>
 
+    <!-- ####################################
+          Button Tooltips
+    ###################################### -->
+    <string name="back_button">Zurück zur vorherigen Seite</string>
```
harambasicluka, line 2875: we name the stuff always in this order: `view_description` so this one should be `button_back`

harambasicluka: please also add an comment for our translators

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -2850,7 +2854,26 @@ as modifying the License.
...
+    <string name="proceed_button">Fortfahren Button</string>
```
harambasicluka, line 2876: same like here

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -2850,7 +2854,26 @@ as modifying the License.
 
     <!-- XBUT: submission result dialog close button -->
     <string name="submission_result_button_dialog_close">OK</string>
+    <!-- ####################################
+              Heading Fragments
+        ###################################### -->
+    <string name="heading_submission_contact">TAN Anfrage per Telefonanruf</string>
```
harambasicluka, line 2860: this headings should live next to the screens where they are used, like all the other strings, e.g. this one should be in the `submission_contact` block. So I would suggest a name like `submission_contact_accessibility_title`.
 Please also add the translation comments.

harambasicluka (changes requested):
Thanks for your first PR :)

PR #270, opened 2020-06-08
Extend teleTAN to 10 characters + input validation (re-submit)

tklingbeil:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TanHelper.kt
```diff
@@ -0,0 +1,33 @@
...
+import java.security.MessageDigest
+import java.util.Locale
+
+object TanHelper {
+    private const val VALID_CHARACTERS = "23456789ABCDEFGHJKMNPQRSTUVWXYZ"
+
+    fun isChecksumValid(tan: String): Boolean {
+        if (tan.trim().length != MAX_LENGTH)
+            return false
+        val subTan = tan.substring(0, MAX_LENGTH - 1).toUpperCase(Locale.getDefault())
```
uschindler, line 13: `Locale.getDefault()` is risky, as lowercasing may be largely locale dependent. Famous example is Turkish locale (quite often used in Germany). When you uppercase a lowercase "i" there you get "İ" (dotted I), which is definitely not what you expect. Gladly, you can be happy as "I" is not an allowed character for Teletans.
To be language neutral `Locale.ROOT` should be used.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TanHelper.kt
```diff
@@ -0,0 +1,33 @@
...
+        val tanDigest = MessageDigest.getInstance("SHA-256").digest(subTan.toByteArray())
```
uschindler, line 14: Converting strings to byte array without a charset is a bad idea. Please use `StandardCharsets.US_ASCII` here. Again, this is not an issue here, but this is bad code quality.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/TanInput.kt
```diff
@@ -56,9 +67,20 @@ class TanInput(context: Context, attrs: AttributeSet) : FrameLayout(context, att
...
     }
 
+    private fun limitLength(length: Int?) {
+        lengthFilter = InputFilter.LengthFilter(if (length != null) length else MAX_LENGTH)
+        tan_input_edittext.filters = arrayOf(whitespaceFilter, alphaNumericFilter, lengthFilter)
+    }
+
     private fun updateTan(text: CharSequence?) {
-        this.tan = text?.toString()
+        this.tan = text?.toString()?.toUpperCase(Locale.getDefault())
```
uschindler, line 76: same here (see below)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TanHelper.kt
```diff
@@ -0,0 +1,33 @@
...
+        var checkChar = "%02x".format(tanDigest[0])[0]
+        if (checkChar == '0') checkChar = 'G'
+        if (checkChar == '1') checkChar = 'H'
+
+        return checkChar.toUpperCase() == tan.last().toUpperCase()
```
uschindler, line 19: here it is locale independent, as it uses `char` type (not easy to see as `var` is used instead of explicit type).

uschindler (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TanHelper.kt
tklingbeil: Good to know. Thank you!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TanHelper.kt
tklingbeil: Still worth fixing it ;) Thank you!

jakobmoellerdev (approved):

PR #284, opened 2020-06-08
EN tests

ksergeevit:


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/receiver/ExposureStateUpdateReceiverTest.kt
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Test ExposureStateUpdateReceiver.
+     */
+    @Test
+    fun testExposureStateUpdateReceiver() {
+        val wm = mockk<WorkManager>()
+        every { WorkManager.getInstance(context) } returns wm
+        every { wm.enqueue(any<WorkRequest>()) } answers { mockk() }
+
+        ExposureStateUpdateReceiver().onReceive(context, intent)
```
jakobmoellersap, line 45: What is tested here?

ksergeevit: a valid received intent causes a workrequest to be enqueued. nothing more to test there actually. also, no need to schedule a job during test, so that is mocked

jakobmoellerdev (approved):

Issue #72, opened 2020-06-01
Text and Translation Contributions


marcmuschko:
Dear contributers,

first of all, we are really happy to see the amazing support we are receiving from the community, thank you very much for that! 
As we have encountered several issues and pull requests regarding translations and text improvements that we sadly had to reject, I am creating this pin thus adding it to the list of known issues as soon as possible. 

Text and translation are not handled by the development team but by a dedicated stream within the project. Once they are finalized, they will replace them alltogether which should solve all related issues.
**Therefore, to reduce efforts for everyone, we cannot accept PRs regarding texts and translations at the moment.** Once the final texts and translations are available I will close this issue.

Thank you for understanding! 
Best regards,
Marc

kedar1607:
If anyone at all wants to get english strings, they can get from https://github.com/corona-warn-app/cwa-app-android/pull/60 and put the english strings.xml file in values-en-US folder *locally* in their Android studio project *without commiting* the file. I hope that helps people understand the context.

mh-:
> Once the final texts and translations are available I will close this issue.

Will they be 'final' or will there be another chance for corrections?

marcmuschko:
@mh- 
As soon as I close this issue, we should have all texts within the app for German and English ready and approved. Then of course we certainly are looking forward for feedback because it still can be the case that we have a few typos (I don't hope so but you never know).... So yes, another chance for improvement suggestions is coming :)

mlenkeit:
Let me just take the opportunity to point out that the German texts in values-de still have inconsistencies, such as:
- COVID-19 vs. Covid-19 (upper case vs. lower case)
- Smartphone vs. Handy
- [Kassenärztlicher Notdienst](https://github.com/corona-warn-app/cwa-app-android/blob/9aada864536c2ed58b171cb41414817b562322c6/Corona-Warn-App/src/main/res/values-de/strings.xml#L240) does not seem to exist (see [Google search](https://www.ecosia.org/search?q=kassen%C3%A4rztlichen+bereitschaftsdienst&addon=chrome&addonversion=3.2.0&method=topbar)), AFAIK, it's "ärztlicher Bereitschaftsdienst" (see [116117.de](https://www.116117.de/de/index.php))

kaluju:
> Let me just take the opportunity to point out that the German texts in values-de still have inconsistencies, such as:
> 
> * ~COVID-19 vs. Covid-19 (upper case vs. lower case)~
[...]

PR #294, opened 2020-06-09
Update VersionComparator.kt

DanCuprum:
Dear Team, 

I just wish to notify you that I think to have found a (logical) bug in your above code. 
In my opinion the following code snippet below will fail, under the following constructed example condition:

        if (versionToCompareMajor > currentVersionMajor) {
            isVersionOlder = true
        } else if (versionToCompareMinor > currentVersionMinor) {
            isVersionOlder = true

currentMajor version = 3, compareMajor version = 1
currentMinor version = 1, compareMinor version = 2

In other words: If the major version on the server is two major steps older than the installed one, 
the function "isVersionOlder" will the faulty return true, which in turn would update the App to an actually *older* version. 
Thus the app will effectively be downgraded instead of upgraded, so to say.
If this behaviour is intended, then please skip my remarks. 
(It just appeared slightly odd to me, alone from the viewpoint of the logical construction of the code.)
Hope this was a bit helpful in any case.

Kind Regards
Dan(iel)

## Checklist

* [ ] Test your changes as thoroughly as possible before you commit them. Preferably, automate your test by unit/integration tests.
* [ ] If this PR comes from a fork, please [Allow edits from maintainers](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork)
* [ ] Set a speaking title. Format: {task_name} (closes #{issue_number}). For example: Use logger (closes # 41)
* [ ] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)
* [ ] Make sure that your PR is not introducing _unnecessary_ reformatting (e.g., introduced by on-save hooks in your IDE)
* [ ] Make sure that your PR does not contain changes in strings.xml (see issue #72)
* [ ] Make sure that your PR does not contain compiled sources (already set by the default .gitignore) and / or binary files

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/update/VersionComparator.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/update/VersionComparator.kt
@@ -16,6 +16,8 @@ object VersionComparator {
         val versionToCompareMajor = versionToCompareParts[0].toInt()
         val versionToCompareMinor = versionToCompareParts[1].toInt()
         val versionToComparePatch = versionToCompareParts[2].toInt()
+        
+        // ** please check my attached comments to the following code block =>
 
         if (versionToCompareMajor > currentVersionMajor) {
             isVersionOlder = true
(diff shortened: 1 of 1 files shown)


ironjan:
@DanCuprum I think this would be more fit as [an issue](https://github.com/corona-warn-app/cwa-app-android/issues/new/choose). I would pick up the issue, if it gets confirmed.

HeeTattSap:
hi @DanCuprum, I have confirmed that this is a bug. It is now fixed in https://github.com/corona-warn-app/cwa-app-android/pull/298
In the future, please create an issue if you want to report a bug. You can use pull requests if you have prepared code to fix the bug.
Thank you for finding this bug! Really appreciate it :)

PR #304, opened 2020-06-09
Extract html from strings

harambasicluka:


harambasicluka:
mistakes were made :)

PR #379, opened 2020-06-11
Adapting To TLS that is supported from Server Side

jakobmoellersap:


Levino:
And for some context I would have expected a related issue. These changes are hard for me to understand and they kinda look like you are crippling the service. Could be an attempt to make downgrading attacks easier. It probably is the contrary, but who knows without context?

PR #382, opened 2020-06-11
update camera request permission flow

Fabian-K:
## Description
This PR updates the camera permission flow:
- go back to dispatcher if denied
- directly request permission without rational dialog in between


kolyaopahle:
LGTM, lets check back if we can get texts done for the permission dialog

harambasicluka (approved):

PR #392, opened 2020-06-11
implement todos

ksergeevit:


csoltenborn:
I also had a brief look at the Date/Long serialization TODO. Might it make sense to move conversion between Long and Date to util.Converters?

pwoessner:
Hi @ksergeevit , sorry for commenting so late but we decided to close this PR because this will touch and change a very critical part of the app. We have not seen anything wrong in the implementation but changing the behaviour at the moment could cause unknown issues. I think we can revisit this change at a later point.
Thanks,
Philipp

PR #415, opened 2020-06-12
Accessibility fixes

Queenhymon:
- fix main page not announced when going back from other screens
- fix announcement of FAQ opening web browser

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/update/UpdateChecker.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/update/UpdateChecker.kt
@@ -36,7 +36,7 @@ class UpdateChecker(private val activity: LauncherActivity) {
             false
         }
 
-        if (updateNeededFromServer) {
+        if (false) {
             showUpdateNeededDialog()
         } else {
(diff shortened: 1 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/update/UpdateChecker.kt
```diff
@@ -36,7 +36,7 @@ class UpdateChecker(private val activity: LauncherActivity) {
             false
         }
 
-        if (updateNeededFromServer) {
+        if (false) {
```
Fabian-K, line 39: 🚨

Fabian-K (changes requested):

harambasicluka (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/update/UpdateChecker.kt
harambasicluka: already fixed, thx :)

harambasicluka (approved):

Fabian-K (approved):

Issue #238, opened 2020-06-07
Links are not clickable


IndianaDschones:
## Describe the bug
The URLs in `Privacy (Datenschutzerklärung)` and `Legal (Rechtliche Hinweise)` are shown as links (colored and underlined) but not clickable.

The eMail-address in `Imprint (Impressum)` is neither highlighted nor clickable.

## Expected behaviour
Highlighted URLs should be clickable and open a webbrowser.

The eMail-address should be clickable.

## Steps to reproduce the issue
Open e.g. `Legal (Rechtliche Hinweise)` and try to click on one of the shown links. Nothing happens.

marcmuschko:
Dear @IndianaDschones 

thank you for pointing this out! Clickable links, e-mails and phone numbers were in stakeholder discussion very recently.
In summary the decision was that the only clickable "link styled" elements in the app will be (at this point in time):
* navigation to technical support 
* navigation to settings 
Therefore the link styling was removed everywhere else in #256 and this issue can be closed.
If you find occurences where link styling is still applied and it might be a mistake, please let us know and reopen the issue. 
We are always happy about feedback! 

Best,
Marc

danielw93:
> Therefore the link styling was removed everywhere else in #256 and this issue can be closed.
> If you find occurences where link styling is still applied and it might be a mistake, please let us know and reopen the issue.

Dear @marcmuschko,
 
have you considered making the text area of affected pages selectable instead?

As it is right now, there is no way I can actually visit the link to the google support page explaining the exposure logging functionality, which is included in the Privacy Statement (https://github.com/corona-warn-app/cwa-app-android/blob/c3416440e644a918f4001489de85f203c7681099/Corona-Warn-App/src/main/assets/privacy_en.html#L185).

As a user I find it very frustrating to see a link and not be able to either click on it or at least copy it to paste it into my browser.

Cheers, 
Daniel

Fabian42:
Why should links not be clickable? Isn't that one of their main purposes? And yes, I agree, it should instead at least be selectable. In general, being able to select anything on the text pages would be good.

PR #443, opened 2020-06-13
Make TextView clickable in SubmissionContactFragment

athanasiosMitrentsis:
## Description
This PR adds a small change and hopefully, it will be an improvement to UX. After the first install of the application, I tested several Screens to see the current functionality. In the `SubmissionContactFragment` I tapped the `TextView` which shows the phone number and expected the dialer of my phone to open. When I saw that this is not the case, I thought that this could be a nice little improvement in this application and added an `OnClickListener` to the `TextView`


IndianaDschones:
@athanasiosMitrentsis this feature was intentionally removed, see https://github.com/corona-warn-app/cwa-app-android/issues/238#issuecomment-640231082

d4rken:
Thanks for the PR!
Due to it being removed intentionally (see https://github.com/corona-warn-app/cwa-app-android/pull/443#issuecomment-643830944), I'll close this to clean up the open PRs a bit.

Issue #448, opened 2020-06-13
Probably unwanted use of constant expression: Math.random().toInt().toString()


LostMekka:
In the closed issue #406 we saw a reference to a weird line of code regarding a HTTP header `cwa-fake`:
https://github.com/corona-warn-app/cwa-app-android/blob/970e8b859cff3c8beeca83ac83e9ed534927107a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/WebRequestBuilder.kt#L182-L183
The expression `Math.random().toInt().toString()` will always evaluate to `"0"`, since `0 <= Math.random() < 1` and `Double.toInt()` simply cuts off the fractional part. So this piece of code is quite redundant.

I do not fully understand the purpose of the `cwa-fake` header, but it seems to me that the code for it was meant to fill it with a random value, when the `faked` flag is `true`. But as it is, it will always be set to `"0"`, regardless of the value of `faked`. This might be an oversight of some kind, so I want to bring this to your attention explicitly with this issue.

And since this is a "question" type issue, here is my question:
Is it safe to always set the header to `"0"`, or is this something that needs to be fixed?

#### Additional information
The initial commit of this repository had 4 instances of the expression:
https://github.com/corona-warn-app/cwa-app-android/blob/4804888ad0d3ee4c4abfb85294bab3006b75c250/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/request/KeySubmissionRequest.kt#L45-L46
https://github.com/corona-warn-app/cwa-app-android/blob/4804888ad0d3ee4c4abfb85294bab3006b75c250/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/request/RegistrationTokenRequest.kt#L60-L61
https://github.com/corona-warn-app/cwa-app-android/blob/4804888ad0d3ee4c4abfb85294bab3006b75c250/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/request/TanRequest.kt#L50-L51
https://github.com/corona-warn-app/cwa-app-android/blob/4804888ad0d3ee4c4abfb85294bab3006b75c250/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/request/TestResultRequest.kt#L50-L51
Then in commit b195623 the usages were unified into the current code snippet referenced above. If this really is an issue, it existed before this repository was created.

---
Internal Tracking ID: EXPOSUREAPP-1966

LostMekka:
[...]

PR #463, opened 2020-06-15
fixed fake key submission not setting the cwa-fake header correctly

LostMekka:
## Description
This PR accompanies issue #448, which discusses a potential oversight in the code, that could result in a bug later on. In short, Math.random().toInt().toString() will always be evaluated to "0", which defeats the point of the faked parameter.

In the checklist above, the check mark for the tests is missing, since there currently is no code path that executes the problematic statement.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/WebRequestBuilder.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/WebRequestBuilder.kt
@@ -179,8 +179,7 @@ class WebRequestBuilder(
         val submissionPayload = KeyExportFormat.SubmissionPayload.newBuilder()
             .addAllKeys(keyList)
             .build()
-        var fakeHeader = "0"
-        if (faked) fakeHeader = Math.random().toInt().toString()
+        val fakeHeader = if (faked) "1" else "0"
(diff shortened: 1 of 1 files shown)


jakobmoellersap:
Hi, as far as I know, the CWA-Fake Header should be Random. That means that we cannot simply set Fake to 1 but should try to have an effort in actually changing the value to a random integer unequal to zero. I agree though that the current implementation does not work correctly either.

LostMekka:
Maybe I don't yet fully understand the purpose of the fake requests, but I cannot imagine, why it would be beneficial to set a random value there. If the purpose is to prevent other users to detect when I am making a real request, then I just need to make fake requests at *random intervals*.
- If the connection would be unsecured, then any other person could read the header and trivially distinguish between fake and real requests. In that case, the header would be useless.
- If the connection is secured, (HTTPS) others can only see *that* a request is sent, but the value of the fake header is encrypted, so there is no benefit from having different values there.

jakobmoellersap:
Thanks for your input. We are in the process of looking at this together with the backend team and will reach out here once we have a more clear decision on how the interface should behave.

jakobmoellersap:
Not relevant as we now have a new Approach (see Playbook)

Issue #467, opened 2020-06-15
Add support for reproducible builds


juril33t:
## Suggested Enhancement

Making the build of the app reproducible. That means providing an easy and stable process for users to rebuild the app from source and to verify that the build they downloaded from the play store actually corresponds to the published source code. 
This means: 
- That the build is made deterministic
- That a stable build environment is provided (ideally one that is easy to use, e.g. a Docker Image)
- That instructions are provided how the build environment can be used to reproduce the build
- That instructions are provided how the play-store release can be verified against a self-build .apk

An example on how this could be achieved is given by the Signal-App (https://github.com/signalapp/Signal-Android/blob/master/ReproducibleBuilds.md). They provide a Docker Image, which is used to build the App and a Script, that can compare the self-built .apk file with the one that is distributed in the play-store

## Expected Benefits

This will improve the users' trust in the app, as they can then be sure that the app they're running actually comes from the published source code and hasn't been tampered with at build- or deployment-time. 

Of course, reproducible builds have many other benefits such as better quality assurance and so on (please see https://reproducible-builds.org/docs/buy-in/ for details)

## Concrete plan

I'm no expert in this specific project, but generally these are the steps i'd take to make a build reproducible:

- Make the build deterministic: Gradle projects can often be made reproducible by adding 
```
tasks.withType(AbstractArchiveTask) {
    preserveFileTimestamps = false
    reproducibleFileOrder = true
}
```
to the build.gradle file (see https://docs.gradle.org/current/userguide/working_with_files.html#sec:reproducible_archives)
- Providing a stable build environment can be done with a Docker Image. This means that the play release version needs to be build using this image.
- Adding Instructions on how to use the Environment. This can be done, for example, in the project wiki.
- Providing a way to verify the play-store release against a self-build .apk: Signal-App does this by providing a python script to compare .apk files: https://github.com/signalapp/Signal-Android/blob/master/apkdiff/apkdiff.py Perhaps it is possible to modify this script in such a way that it also works for CWA (licence allows this).

SebastianWolf-SAP:
Dear @juril33t,

thanks for the extensive description, but we already have a corresponding issue in the documentation repository that collects all aspects of reproducible builds: https://github.com/corona-warn-app/cwa-documentation/issues/14

Therefore, we will close this issue. 

Mit freundlichen Grüßen/Best regards, 
SW
Corona Warn-App Open Source Team

PR #468, opened 2020-06-15
Added Dockerfile for Reproducible Builds #467

gutjuri:
# Description

This concerns #467 

In order to make builds reproducible, a stable build environment is needed. This Dockerfile aims to provide such an environment.

I verified that it is possible to build the app using this Image (`docker build -t cwa . && docker run --rm -v $(pwd):/project -w /project cwa ./gradlew quickBuild`)

I didn't verify that the produced build is indeed deterministic.

The Dockerfile is inspired by https://github.com/signalapp/Signal-Android/blob/master/Dockerfile but I'm not sure if a licence notice is needed, since it differs substiancially. If it is needed, i'll be glad to add one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,27 @@
+FROM ubuntu:20.04
+
+
+RUN apt-get update -y && \
+    apt-get install -y software-properties-common && \
+    apt-get update -y && \
+    apt-get install -y \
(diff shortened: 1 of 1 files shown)


d4rken:
FYI The related issue is https://github.com/corona-warn-app/cwa-backlog/issues/21

PR #475, opened 2020-06-16
Verify Gradle on each pull request

hannesa2:
Here you have the complete story https://github.com/gradle/wrapper-validation-action

image


jakobmoellersap:
I am not sure if the mentioned story is 100% fitting, as we do not accept any binary from outside our maintainer group. If I understand correctly this verifies the Checksum of the aforementioned binary.

Let me know what you think!
BR,
Jakob

hannesa2:
I've absolute no doubt that you trust yourself (or your bunch of maintainer)
It's like when a dog take care about the sausage. "Der Hund soll auf die Wurst aufpassen"
The thing is, some _independent_ should trust you,

A lot of repos do this, why not this ! Especially this ! 
At least for this case, it will prevents you from bad news

jakobmoellersap:
As far as I know what you are referencing would be the reproducible build topic. I do agree that the SHA-256 verification is useful though. Let me get it back to the team for some final decisions and then we should be good to merge this.

jakobmoellerdev (approved):

Issue #82, opened 2020-06-01
nit: Exception message is not completely right


StefanSchoof:
## Current Implementation
https://github.com/corona-warn-app/cwa-app-android/blob/90c49ae4ee75659cc96b603d0e0440e0dc0edcd3/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/tracing/TracingIntervalRepository.kt#L52

## Suggested Enhancement

```kotlin
if (to < from) throw IllegalArgumentException("to cannot be after or equal from")
```
## Expected Benefits

Invalid exception message can make it harder to analyze errors.

murksa:
A bit late to the party, but shouldn't it be "to cannot be **before** from"?

PR #608, opened 2020-06-17
Fix exception message

murksa:
## Description
Fixed an exception message (see also #82: the message was not 'not completely right' but even wrong...)


jakobmoellerdev (approved):

jakobmoellersap:
Thanks for the correction!

PR #619, opened 2020-06-18
Re-enable screenshots taking except for SubmissionTanFragment (closes  #77)

### LINKED ISSUE #77: Split "submission" activity from main activity
## Current Implementation
The current implementation has 2 activities:
- onboarding
- main activity

The onboarding activity is only shown on first start of the app. On the main class this state is checked and then either the main or the onboarding activity is started. This is perfectly good design, as the onboarding activity is mostly not used, so all fragments, images, text don't need to be loaded. 

## Suggested Enhancement
The submission of a positive COVID-19 case should be implemented as a separate activity. The transition to this should not be to hard. The fragments (in package "submission") should be removed from main activity and moved to a new activity.

## Expected Benefits
Like the onboarding, the submission code is not needed in 99% of all cases and should not be loaded when app starts. This is not necessarily an important thing with today's devices, but there are some other things to imagine:
- The submission of a positive COVID-19 result is a security/privacy sensitive task and involves scanning a QR-code or entering the teleTAN. The data involved by this should be protected. Therefore screenshots should be disabled (which also involves the preview image shown if you close app with home button - see issue #9). Pull request #41 tries to disallow screenshots by enabling/disabling them on enable/disable of the fragment. IMHO, a more clear approach would be to just disable screenshots for the submission activity. Screenshotting and app snapshot images should be allowed for all other parts (like main app showing actual status and onboarding). This seems very important to me. The enable/disable screenshotting is a setting for the activity's window and can't be enabled/disabled easily without hacks per fragment.
- The submission interface could be started without starting the app by directly triggering the activity by an URI handler: It may be possible to encode the QR code key as an URI (that's my other suggestion - should I open separate issue?). In that case users can just open the default mobile's camera app, scan the barcode and the camera app will trigger the URI encoded into it. The app can then provide an URI handler that automatically starts the submission activity in the CWA Android app. This is much more easier to understand to most people, because when they see a QR/barcode, they intuitively scan it with the camera app (recent phones have QR code reader embedded into camera app) or their favourite QR code scanning app.

Internal Tracking ID: EXPOSUREAPP-18

R4md4c:
## Description
It is not fair to block the users from taking screenshots from the entire app because of a security concern in a couple of screens.
This PR reintroduces taking screenshots from the app except for the `SubmissionTanFragment` (If you'd like to disable it for other screens please state in the comments so that I can add the interface there).

The trick here is to mark a Fragment with `ScreenshotCaptureBlockable` (not sure about the name) marker interface and once it gets attached or the user reaches that screen, the `ActivityScreenshotHelper` will set the `FLAG_SECURE` flag on the Activity Window. ~~However there is one caveat, is that once this flag is set for a Window it can't be reverted (even with using `clearFlags`) and the Activity will stay in secured mode until its destruction.~~

Tested on Emulator with API 23 and a Samsung S6.


harambasicluka:
Hi @R4md4c  ,

thank's for your detailed contribution! :) But this was implemented as a recommendation of the [BSI](https://www.bsi.bund.de/DE/Home/home_node.html) in #9 . So we won't remove it for now, but you aren't alone with your opinion and we are discussing it internally. 

That's why I'll close this PR.

Cheers,
Luka

R4md4c:
@harambasicluka But this doesn't contradict with #9, it keeps the `FLAG_SECURE` usage, however only defers it until the user reaches the screen in question. 
Anyway hopefully the internal people would agree in re-enabling the screenshots back.

corneliusroemer:
This seems to be the BSI being prohibitively restrictive in their recommendations. Rather than the BSI's recommendation being regurgitated every time this comes up: What about asking the BSI what it thinks about the new proposal?

uschindler:
Wonderful implementation. Why was this PR closed? That's really unfair to the contributor.

As said before, the iOS App allows screenshots, Android not. Why?

uschindler:
Nevertheless: the issues as described by @R4md4c could be resolved in a less Jacky way by making the submission a separate activity like the onboarding one. I opened an issue about that already more than 2 weeks ago. See #77

R4md4c:
@uschindler Actually sorry, I take back what I said earlier. I found a way to clear the flag after the Fragment gets destroyed. You can see the work done in this [commit](https://github.com/R4md4c/cwa-app-android/commit/d4ae50c4a9f448301de8ae72710767726c746dd2) (It's not visible in the PR because it got closed). The trick is to set the flag before the Fragment's View gets created.

Alestrix:
> This seems to be the BSI being prohibitively restrictive in their recommendations.

I only know what's written in #9, but there I do not see the BSI being too restrictive as their statement is limited only to "sensitive information". This PR should have been the solution to #9. not what was implemented there.

falkorichter:
closes #745 or #745 closes this one

d4rken:

R4md4c:
Closing since this PR won't add any value after #987 was marked as obsolete.

PR #620, opened 2020-06-18
Drop WindowManager.LayoutParams - FLAG_SECURE

Bengt:
This flag makes Android prohibit the app to appear in screenshots and screencasts:

https://developer.android.com/reference/android/view/WindowManager.LayoutParams#FLAG_SECURE

Using this flag drives users who want to take screenshots/casts towards circumventing this barrier. Ways of circumventing include:

- Using another Smartphone to take a photo of the app.
- Forking, applying the proposed change and compiling the app oneself.  
  (Once there is a way to run apps like this without Google Play Services.)
- Patching the installed app using XPosed.  
  <https://repo.xposed.info/module/fi.veetipaananen.android.disableflagsecure>
- Patching the Android framework Magisk.  
  <https://forum.xda-developers.com/apps/magisk/module-smali-patcher-0-7-t3680053>

As these ways to circumvent the flag exist, using this flag is essentially useless, because an app in question might always be "insecure" in the sense of this flag. All of these ways to circumvent are bad, as the inability to take screenshots compromises the ease of use, building and sideloading the app oneself requires additional effort and patching the app or the Android framework is dangerous. Even worse, the iOS version of this app does not use the equivalent iOS functionality, which demonstrates that it is not strictly necessary. In effect, using this flag compromises the rights of the user in favor of the app's rights. This is fundamentally immoral and can thus lessen the app's acceptance. Less acceptance leads to less installations and more untraced contacts, preventable infections and - in the worst case - unnecessary death. So please stop it.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
@@ -95,15 +95,6 @@ class CoronaWarnApplication : Application(), LifecycleObserver,
 
     @SuppressLint("SourceLockedOrientationActivity")
     override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
-        // prevents screenshot of the app for all activities,
-        // except for deviceForTesters build flavor, which is used for testing
-        if (BuildConfig.FLAVOR != "deviceForTesters") {
-            activity.window.setFlags(
-                WindowManager.LayoutParams.FLAG_SECURE,
-                WindowManager.LayoutParams.FLAG_SECURE
-            )
-        }
-
         // set screen orientation to portrait
         activity.requestedOrientation = ActivityInfo.SCREEN_ORIENTATION_USER_PORTRAIT
     }


tomjschwanke:
The `FLAG_SECURE` is supposed to be there, see #9

harambasicluka:
Hi @Bengt ,

thank's for your contribution! :) As @tomjschwanke already mentioned this was implemented as a recommendation of the [BSI](https://www.bsi.bund.de/DE/Home/home_node.html) in #9 . So we won't remove it for now, but you aren't alone with your opinion and we are discussing it internally. 

That's why I'll close this PR, if we are allowed to remove it we will do it ourself ;)

Cheers,
Luka

Bengt:
@tomjschwanke Thanks for the pointer. I must have overlooked that discussion when filing this pull request. I am glad I got the right answer by posting the wrong one.

@harambasicluka Thanks for the insight. After reading the discussion I better understand the problem this can solve and some value in this solution. I guess the discussion should stay focused to that one place, though.

This PR's branch in my fork is surely no longer needed as a basis for discussion, so I deleted it.

corneliusroemer:
@Bengt I think it would make sense to add your thoughts to #9 because of the valuable insights about workarounds you add.

de-graeuler:
Other apps turn the view white (blank) before going to background. If that happens before the app snapshot is taken, it would solve #9 and it would still be possible to manually take a screenshot.

PR #673, opened 2020-06-19
Random housekeeping

johnjohndoe:
# Description
Some random clean-ups.
- Move lambda out of parenthesis.
- Lift-out return.
- Let local variable be "val".


jakobmoellerdev (approved):

jakobmoellersap:
Looks good, thanks for the contribution!

Issue #332, opened 2020-06-10
[HELP WANTED][README] Text / Translations 


marcmuschko:
Dear contributers,

as we are getting into the final stage regarding all texts and translations within the app, we would appreciate your support in this area!
If you spot anything unsual within the German / English text files, **please report your findings as a comment in this issue.**

Since the texts and translations are not provided by the development team but by a dedicated stream and have to be approved by the stakeholders first (depending on the change), it is easier for the stream to check this issue for findings and provide fixes when necessary. 
Please don't create pull requests directly, the string files will be overwritten when we import changes from our translation stream.

**Edited (after we noticed several findings within legal texts and we aspire to be as transparent as possible):**
Please be aware that all legal texts including their translations (Terms of Use, Publication Details, Data Privacy Information, Legal Notices) are handled separately, 
Neither developers nor the text / translation stream have any influence on spelling or wording, as these texts require dedicated approval. 
We will certainly collect your suggestions and pass them on, but it can take time until improvements are included (if at all).

Thank you for your your help! 
Best regards,
Marc

marcmuschko:
Referencing [findings](https://github.com/corona-warn-app/cwa-app-android/issues/72#issuecomment-641032935) from issue #72

mh-:
[Privacy notice](https://www.coronawarn.app/assets/documents/cwa-privacy-notice-de.pdf):
"..., versendet Ihr Smartphone per Bluetooth Low Energy kontinuierlich zufallsgenerierte Kennnummern, auch als Zufallscodes **zu** bezeichnet" - remove the word "zu"

mh-:
Link to Privacy Notice in [Terms](https://github.com/corona-warn-app/cwa-app-android/blob/master/Corona-Warn-App/src/main/assets/terms_de.html) does not work: 
```
Die separate Datenschutzerklärung (siehe <strong>[<em>Link</em>]</strong>) ist
Bestandteil dieser Nutzungsbedingungen. 
```

bocops:
The abbreviation "CWA" is only used once in user-facing strings, in 

[...]

PR #679, opened 2020-06-19
Reverse order of Transmission Risk Levels used for Diagnosis Keys Upload (closes #678)

### LINKED ISSUE #678: Diagnosis Keys Upload: Transmission Risk Levels are applied in reverse order
## Avoid duplicates
* [x] Bug is specific for Android only - I think so, on iOS the [sort order](https://github.com/corona-warn-app/cwa-app-ios/blob/5618da6fc3faa4ecef3a6763250230501fae3774/src/xcode/ENA/ENA/Source/Services/ExposureSubmissionService.swift#L207) seems to be correct.
* [x] Bug is not already reported in another issue (at least I didn't find it)

## Describe the bug

The "Epidemiological Motivation of the Transmission Risk Level" [document](https://github.com/corona-warn-app/cwa-documentation/blob/master/transmission_risk.pdf) gives the "Transmission risk level for the Base Case" (Figure 14) as **5, 6, 8, 8, 8, 5, 3, 1, 1, 1, 1, ...** for values 0..13 of _Delay from Exposure to Consent for Upload_. This is counting backwards, 0 (today) should be assigned **5**, 1 (yesterday) should be assigned **6**, and so on.

However, `fun List<TemporaryExposureKey>.transformKeyHistoryToExternalFormat()` [(link)](https://github.com/corona-warn-app/cwa-app-android/blob/c2e5bd49093193060dded4d1545f783a498a96da/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ProtoFormatConverterExtensions.kt#L35) assigns these values the other way around: The key with the _lowest_ `rollingStartIntervalNumber`, i.e. the _oldest_ key is assigned **6** (**5** as the value for today is ignored, because the API doesn't provide today's key), and so on.

## Expected behaviour

The risk levels should be applied as explained in the "Epidemiological Motivation of the Transmission Risk Level" document, not the other way around, because then risk levels would make no sense. (This is not extremely critical in the beginning of the roll-out, when users have only 6 keys to upload, as the curve is quite symmetrical on the left end, but will cause nonsense risk estimations once users have all 14 keys saved on their device.) 

## Possible Fix

I plan to provide a Pull Request with a fix and a test case for the function.

mh-:
## Checklist

* [x] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) (if applicable)

## Description

Fixes https://github.com/corona-warn-app/cwa-app-android/issues/678


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ProtoFormatConverterExtensions.kt
```diff
@@ -33,9 +33,9 @@ object ProtoFormatConverterExtensions {
         this.sortedWith(compareBy { it.rollingStartIntervalNumber }).asReversed().take(MAXIMUM_KEYS)
 
     fun List<TemporaryExposureKey>.transformKeyHistoryToExternalFormat() =
-        this.sortedWith(compareBy { it.rollingStartIntervalNumber })
+        this.sortedWith(compareBy { it.rollingStartIntervalNumber }).asReversed()
```
ChristopherSchmitz, line 36: @mh- next to the asReversed option I see the other option to use "compareByDescending" instead of "compareBy". This option would result into better readability of code.

Do you see anything speaking against the usage of "compareByDescending"? (for sure we remove as Reversed in all the places)

mh-: @ChristopherSchmitz `compareByDescending` is probably better, if it makes the code more readable. I don't read Kotlin a lot and indeed had a hard time understanding the code. So I just copied the pattern from [.limitKeyCount()](https://github.com/corona-warn-app/cwa-app-android/blob/c2e5bd49093193060dded4d1545f783a498a96da/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ProtoFormatConverterExtensions.kt#L33) and wrote a test case to see what it's actually doing. 
The test case assumes that the keys are ordered chronologically, but if that's not required for upload to the server, we could also move the 2nd sorting to the test.
Not sure how you prefer to proceed? Do you want to edit the PR, or should I do it?

ChristopherSchmitz: @mh since we use on iOS a desc order when submitting to the server I would keep the behaviour consistent.
If you like to adjust your PR this would be wonderful. We plan to close and merge this in the next 2 hours :-)

mh-: @ChristopherSchmitz done.

ChristopherSchmitz (approved):
@mh- many thanks for your support! This looks really good.

jakobmoellerdev (approved):

PR #706, opened 2020-06-21
Theme Switcher for Dark/Light/default design (closes #450)

### LINKED ISSUE #450: Theme switcher
## Feature description

An option in the settings to choose themes of the `CWA`.

## Problem and motivation

Currently the apps theme is defined by the users system settings. A system-wide dark mode was added in `Android 10`. User who are not running the latest version of `Android` cannot use the dark theme of the `CWA`. As suggested [by the Android guide](https://developer.android.com/guide/topics/ui/look-and-feel/darktheme) the user should be offerd to switch the themes:

> When running on devices running Android 9 or earlier, the recommended theme options are:
>- Light
>- Dark
>- Set by Battery Saver (the recommended default option)
>
>When running on Android 10 (API level 29) and higher, the recommended options are different, to allow the user to override the system default:
>- Light
>- Dark
>- System default (the recommended default option)
>
>Note that if the user selects Light, then Battery Saver will not change that setting.

## Is this something you're interested in working on
No

glaukos:
## Description
Extended the settings menu to open an AlertDialog where the user can pick the UI Theme for the app (Dark, Light, or default). The preference is stored and applied on change and at application start.
This closes issue corona-warn-app/cwa-wishlist#854 .


harambasicluka:
Hi @glaukos ,

thank's for your contribution! We have to clarify the wanted pattern with our designers, but for now I think we won't implement a settings which will open an AlertDialog which is only used to inform the user.

So for now I'll leave this PR open but label it as `on hold`, so if we need it we would come back to if that's fine.

Cheers,
Luka

harambasicluka:
To keep our PRs clean I'll close this PR as it doesn't seem that this will be prioritised in the next time. 

@glaukos would you be willing to work on a solution where we provide the design spec if it comes so far? If I see an upcoming task I would contact you here.

If you want to raise more awareness of this issue you could go to our wishlist, maybe you aren't the only one: https://github.com/corona-warn-app/cwa-wishlist

Thanks for your patients,
Luka

PR #714, opened 2020-06-22
WIP: Fix tracing status, verify activity by location check (closes #491)

### LINKED ISSUE #491: Exposure Logging is in "active" state even if it is disabled
## Describe the bug

After switching off Location app still displays that Exposure Logging is active, when in fact it is not. There is a notification displayed from Play Services to turn Location back on, but it can be dismissed (even accidentally, by dismissing all notifications), and user stays in a state with exposure logging not working, false "active" status displayed in the app, and no clue how to turn in back on.

## Expected behaviour

App should always display correctly whether exposure logging is active, and guide user to turn Location back on if this is the reason for exposure logging not working. This is already handled correctly for the case when Bluetooth is off.

1. download the app
2. agree to exposure logging
3. turn location off
4. dismiss all notifications
5. check if you are in a false state of "active" exposure logging with no clue how to turn in back on

## Technical details

Android 10

## Possible Fix

I'd suggest to take a look how Immuni app from Italy handles this. If using "Location" word is not desired here app can simply display "Service inactive", and a button to activate it which brings up "Turn on COVID-19 Exposure Notifications?" prompt again and leaves it up to Exposure Notifications framework to handle turning on all required services.

cypheron:
Fix for #491

## Problem description:
Please check issue.

In short: exposure logging is shown as active even when location is disabled and hence the underlying exposure notifications is not active.

## Proposed changes
Changes are kept to an absolute minimum.
1. An additional function is added to `ConnectivityHelper` to check whether location (GPS or FINE) is turned on. Location status can be queried without requiring ANY additional permissions!
2. The location status poll is coupled together with `refreshIsTracingEnabled` in `TracingRepository`. I specifically decided to put it in that place to avoid having to maintain any objects concerning "*location*".

## Known limitations
1. No UI elements: This PR provides the minimum changes required to fix the false reporting of the exposure logging status as detailed in the issue (which I see as the priority). In the very near future, App maintainers should add UI elements to guide the user through the location activation process similar to how its done for Bluetooth. This should come with an additional help page explaining why "location" is showing up in the UI, FAQ regarding data privacy etc.. Because all of this comes with a longer review process it is not part of this PR!
2. This additional check will only function on devices running **Android 5.0** and higher. Versions below that need the location permission to ask for location status - needless to say that this is out of the equation.
3. Testing: I'm not able to test the the modified code together with Google's Exposure Notifications API, because it probably whitelists a certain signature. That's why this PR is marked as WIP. I would invite one of the SAP devs to test the change.

Thanks to @kbobrowski  and @aurisnoctis for bringing up and discussing the issue.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
@@ -1,6 +1,7 @@
 package de.rki.coronawarnapp.storage
 
 import androidx.lifecycle.MutableLiveData
+import de.rki.coronawarnapp.CoronaWarnApplication
 import de.rki.coronawarnapp.exception.ExceptionCategory
 import de.rki.coronawarnapp.exception.TransactionException
 import de.rki.coronawarnapp.exception.reporting.report
@@ -8,6 +9,7 @@ import de.rki.coronawarnapp.nearby.InternalExposureNotificationClient
 import de.rki.coronawarnapp.risk.TimeVariables.getActiveTracingDaysInRetentionPeriod
 import de.rki.coronawarnapp.transaction.RetrieveDiagnosisKeysTransaction
 import de.rki.coronawarnapp.transaction.RiskLevelTransaction
+import de.rki.coronawarnapp.util.ConnectivityHelper
 import java.util.Date
 
 /**
@@ -70,7 +72,8 @@ object TracingRepository {
     suspend fun refreshIsTracingEnabled() {
         try {
             val isEnabled = InternalExposureNotificationClient.asyncIsEnabled()
-            isTracingEnabled.value = isEnabled
+            val isActive = ConnectivityHelper.isLocationEnabled(CoronaWarnApplication.getAppContext())
+            isTracingEnabled.value = isEnabled && isActive
         } catch (e: Exception) {
             // when API is not available, ensure tracing is displayed as off
             isTracingEnabled.postValue(false)
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ConnectivityHelper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ConnectivityHelper.kt
@@ -6,11 +6,13 @@ import android.content.BroadcastReceiver
 import android.content.Context
 import android.content.Intent
 import android.content.IntentFilter
+import android.location.LocationManager
 import android.net.ConnectivityManager
 import android.net.Network
 import android.net.NetworkCapabilities
 import android.net.NetworkRequest
 import android.os.Build
+import androidx.core.location.LocationManagerCompat
 import de.rki.coronawarnapp.exception.ExceptionCategory
 import de.rki.coronawarnapp.exception.reporting.report
 import timber.log.Timber
@@ -169,6 +171,17 @@ object ConnectivityHelper {
         val activeNetwork: Network? = manager.activeNetwork
         val caps: NetworkCapabilities? = manager.getNetworkCapabilities(activeNetwork)
         return caps?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) ?: false
+    }
+
+    /**
+     * Get location enabled status.
+     *
+     * @return current location status
+     *
+     */
+    fun isLocationEnabled(context: Context): Boolean {
+        val manager = context.getSystemService(Context.LOCATION_SERVICE) as LocationManager
+        return LocationManagerCompat.isLocationEnabled(manager)
     }
 
     /**


kbobrowski:
It may be worth to have a look how Italian Immuni app handles this, grep does not show any "gps" string in the code, perhaps they are using different method

cypheron:
@kbobrowski [here it is](https://github.com/immuni-app/immuni-app-android/blob/51aeb30101f39324227c782b84615b23b49c9b55/extensions/src/main/java/it/ministerodellasalute/immuni/extensions/location/LocationStateFlow.kt)
1. to query location status they use `LocationManagerCompat` which is a helper class around the LocationManager I used - the `isLocationEnabled` function doesn't require you to specify the provider (GPS / NETWORK). Not sure if we should use this instead.
2. they register changes to location status via `PROVIDERS_CHANGED_ACTION` to a receiver. If we decide to leave the coupling of location status query with `refreshIsTracingEnabled` we don't need a callback / receiver to handle location status changes - I don't think it necessarily consumes more battery but it definitely doesn't add additional code to handle status changes and communicate them via viewmodel etc.

kbobrowski:
@cypheron I like isLocationEnabled() since this is exactly what has to be checked... But maintainers have not commented on the issue yet, not sure what is their position this issue

uschindler:
I tend to use the isLocationEnabled(), as it is more generic.

cypheron:
@kbobrowski @uschindler Found the [code](https://androidx.de/androidx/core/location/LocationManagerCompat.html) for `<LocationManagerCompat>.isLocationEnabled()`. It's safe to use the function, because it wraps `isProviderEnabled(GPS/NETWORK)` for compatibility with older Android versions. Same limitations apply to the function: minimum Android Version 5.0 because of permissions.

uschindler:
Android 5 Minimum is no issue at all, because Exposure Notification API and the CWA app ist Android 6 minimum.

cypheron:
@uschindler good to know! this means we don't need to check for android sdk version for location status.

force pushed an update with this change

jakobmoellersap:
Sorry for not replying earlier, I am looking into it and will get back to you when I collected all relevant feedback from UX, Scope Management and Dev.

uschindler:
Hi,
This PR does not yet add any extra user interface like when Bluetooth is disabled. This should also be added and like for Bluetooth it should have a button to open the settings app.

cypheron:
@uschindler Correct, this PR only fixes the issue as described in #491 with minimum changes. It fixes the false status of reporting exposure logging as active when it isn't. IMHO, UI elements should be added in a separate PR: they require additional code in view and viewmodel.

jakobmoellersap:
We are in the process of exploring a more in depth solution and I will update here what this exactly means (also for the related issue)

cypheron:
@jakobmoellersap Well, that surely sounds like a reply from management, not dev. But okay, at least this issue has got attention now and will eventually be fixed.

Keeping this PR open for updates.

Archangel1C:
A "more in depth" might be nice, but for psychologically reasons (user trust) it seems crucial to me to get the "false indication fix" released ASAP...

cypheron:
@jakobmoellersap Any updates from dev? It's been almost 2 weeks since you mentioned an in-depth solution. Don't want to sound rude but after cashing in 10 million from German citizens I'd expect at least a status update.

tkowark:
@cypheron you can always ask for an update, and we will try to give you one in a timely manner. I'll connect with the dev colleagues today to get the latest status.

As a personal side note: You can ask us without having to wave the project price tag. We (developers and community managers) did not negotiate that and our days still only have 24 hours. As visible in the Github issues, there are a lot of tasks requiring our attention and sometimes things just slip through. A quick tag and "any updates?" should generally suffice.

tkowark:
Dev sync was postponed for today. Hopefully @SebastianWolf-SAP can give an update tomorrow.

SebastianWolf-SAP:
OK, we got a small update at least. Current status is that they want to fix this issue in a major revamp of the status display which is currently work-in-progress. Reason is that there are quite some other strings attached to fixing it like it's proposed here in this PR: If we display that it's not working, we also need to take care that users are notified what they need to do in order to be able to enable it again.

However, the complete solution might take some time and we probably need a hotfix, we are currently trying to get this included as early as possible. Once again apologies for the delay and the long processing time - as @tkowark already mentioned there are a lot of tasks we need to balance in parallel right now.

Mit freundlichen Grüßen/Best regards, 
SW
Corona Warn-App Open Source Team

d4rken:

harambasicluka:
Hi, 

IMO this is solved with the following PRs:
* https://github.com/corona-warn-app/cwa-app-android/pull/1057
* https://github.com/corona-warn-app/cwa-app-android/pull/879
* https://github.com/corona-warn-app/cwa-app-android/pull/869
* https://github.com/corona-warn-app/cwa-app-android/pull/853
* https://github.com/corona-warn-app/cwa-app-android/pull/849

@cypheron do you also see it like that?

Cheers,
Luka

uschindler:
Yes it's solved since long time (version 1.2, I think)

cypheron:
@harambasicluka Yes, I agree! Issue solved, closing PR.

PR #771, opened 2020-06-27
Added 10 second delay between calls to provideDiagnosisKeys (fix #737)

### LINKED ISSUE #737: Error when communicating with Google API(10): Unable to validate key file signature.
Auf Gerät mit Android 6.0

[Scan 24.06.2020.pdf](https://github.com/corona-warn-app/cwa-app-android/files/4824611/Scan.24.06.2020.pdf)

---
Internal Tracking ID: EXPOSUREAPP-1714

kbobrowski:
## Checklist

* [x] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) - Pull Requests that are not linked to an issue with you as an assignee will be closed, except for minor fixes for typos or grammar mistakes in *documentation or code comments*.

## Description

Can no longer reproduce error 10 with 10 seconds delay between calls to provideDiagnosisKeys

fix #737 

Note that I'm only able to test that it would work by accessing live instance of the app from Play Store and reverse-engineering (calling provideDiagnosisKeys on a client extracted from heap), cannot unfortunately deploy this fork and test it with the code changes introduced here unless I get ahold of signed apk with these changes

errors_vs_delay

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RetrieveDiagnosisKeysTransaction.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RetrieveDiagnosisKeysTransaction.kt
@@ -81,8 +81,10 @@ object RetrieveDiagnosisKeysTransaction : Transaction() {
 
     override val TAG: String? = RetrieveDiagnosisKeysTransaction::class.simpleName
 
-    /** delay between provideDiagnosisKeys() calls in milliseconds,
-     *  to avoid error 10
+    /**
+     * Delay between provideDiagnosisKeys() calls in milliseconds, attempts to prevent
+     * ApiException: 10: Unable to validate key file signature: Pipe is closed.
+     * Empirically average number of exceptions decreases linearly with a log of delay.
      */
     private const val DELAY_BETWEEN_PROVIDE_DIAGNOSIS_KEYS_CALLS = 10000L
 
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RetrieveDiagnosisKeysTransaction.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RetrieveDiagnosisKeysTransaction.kt
@@ -36,6 +36,7 @@ import de.rki.coronawarnapp.transaction.RetrieveDiagnosisKeysTransaction.Retriev
 import de.rki.coronawarnapp.transaction.RetrieveDiagnosisKeysTransaction.rollback
 import de.rki.coronawarnapp.transaction.RetrieveDiagnosisKeysTransaction.start
 import de.rki.coronawarnapp.util.CachedKeyFileHolder
+import kotlinx.coroutines.delay
 import timber.log.Timber
 import java.io.File
 import java.util.Date
@@ -79,6 +80,11 @@ import java.util.concurrent.atomic.AtomicReference
 object RetrieveDiagnosisKeysTransaction : Transaction() {
 
     override val TAG: String? = RetrieveDiagnosisKeysTransaction::class.simpleName
+
+    /** delay between provideDiagnosisKeys() calls in milliseconds,
+     *  to avoid error 10
+     */
+    private const val DELAY_BETWEEN_PROVIDE_DIAGNOSIS_KEYS_CALLS = 10000L
 
     /** possible transaction states */
     private enum class RetrieveDiagnosisKeysTransactionState :
@@ -242,6 +248,7 @@ object RetrieveDiagnosisKeysTransaction : Transaction() {
         exposureConfiguration: ExposureConfiguration?
     ) = executeState(API_SUBMISSION) {
         exportFiles.forEach { batch ->
+            delay(DELAY_BETWEEN_PROVIDE_DIAGNOSIS_KEYS_CALLS)
             InternalExposureNotificationClient.asyncProvideDiagnosisKeys(
                 listOf(batch),
                 exposureConfiguration,


jakobmoellersap:
We will take care of testing this in detail on Monday.

kbobrowski:
Published scripts to reproduce it here: https://github.com/kbobrowski/en-api-exception-10-debug

Stack trace inside GMS: https://github.com/kbobrowski/en-api-exception-10-debug/blob/master/stack_trace.txt
Pipe operations (note that error only happens when PipedInputStream is closed before PipedOutputStream): https://github.com/kbobrowski/en-api-exception-10-debug/blob/master/pipes_0ms_delay.txt

This premature closing of PipedInputStream seems to never happen when delay is introduced: https://github.com/kbobrowski/en-api-exception-10-debug/blob/master/pipes_10000ms_delay.txt

kbobrowski:
@jakobmoellersap found an issue and patched it, `PipedInputStream.read` is buggy. I think we can expect fast patch by Google since it should be small change in Play Services implementation:

https://github.com/kbobrowski/en-api-exception-10-debug#bug-description

interesting part is that Google seems to be aware of this issue, they have workarounds for it GMS:

```
# bzin class

const-string v1, "#read(byte[]) returned invalid result: "

const-string v0, "\nThe InputStream implementation is buggy."
```

but in a place where `EK Export v1` header is validated it does not seem to be handled properly - which results in not enough bytes read, throwing exception, closing pipe while handling this exception and in turn `Pipe is closed` exception being passed to CWA

jakobmoellersap:
Hi again, thanks for the Update. I could reproduce the issue and also agree with your analysis. Thus we will wait 1-2 more days for a fast track patch on Google Side in the hope of a fix just like they did on the "0 keys" issue. Also all of your analysis was forwarded to Google for more visibility and we keep updating as soon as we have more information we can share publicly. Sorry that we cannot share more at this point, but my hope is that we will never need to introduce this PR workaround.

kbobrowski:
@jakobmoellersap right, hopefully this workaround won't be needed, especially since it has a potential of breaking something on the devices which have no problem right now with error 10 (transaction timeout etc). As a backup solution also shorter delay may be considered (1-2 seconds) which won't solve the problem completely but it may make it couple of times less frequent such that at least on most of the days Diagnosis Keys would be successfully provided to the framework. But hopefully we'll see quick patch from Google

kbobrowski:
It's all probably because Android 6 is still using Apache Harmony-based Java implementation, and from 7 onwards OpenJDK is used, so I guess all Android 6 devices are affected

jakobmoellersap:
Interesting thought, but we also see this on other devices not based on Android 6. Maybe they are more likely to be affected due to Harmony but I wouldn't call it out as explicit cause just yet

kbobrowski:
If it's not a bug in Java implementation then maybe there is an issue with GMS implementation which causes PipedOutputStream to send not enough bytes. Would be interesting to find out what Google has changed in the upcoming update

kbobrowski:
Everything seems to be fine with Google's implementation, looks more like an issue with Java itself, here 4096 bytes come in to the pipe and 6 bytes come out:

```
1759123: java.io.PipedOutputStream@ded319d <- PipedOutputStream()
1759132: java.io.PipedInputStream@f8a2612 <- PipedInputStream(output_stream=java.io.PipedOutputStream@ded319d)
1759140: java.io.FileInputStream@957e1e3 :: read(buffer={length=4096}) = 4096
1759146: java.io.PipedOutputStream@ded319d :: write(buffer={length=4096}, offset=0, length=4096)
1759139: java.io.PipedInputStream@f8a2612 :: read(buffer={length=16}) = 6
1759154: Exception: Invalid file header length
1759156: java.io.PipedInputStream@f8a2612 :: close()
1759159: Exception: Pipe is closed
1759166: java.io.PipedOutputStream@ded319d :: close()
1759173: Exception: Unable to validate key file signature: Pipe is closed
```

kbobrowski:
In the end it looks like it's an implementation issue on Google's side.

`PipedOutputStream.write(byte[] buffer, int offset, int length)` will block until all bytes are sent,
`PipedInputStream.read(byte[] buffer, int offset, int length)` will block only until first byte is received, but will **not** block until `length` of bytes is received, so result of this method depends on a timing, if it is called too fast then it returns less than `length` bytes.

It corresponds to the definition in documentation, although documentation does not go out of its way to explain it.

This is expected behavior which happens in most of the cases:

https://github.com/kbobrowski/en-api-exception-10-debug/blob/master/trace_all_correct.txt

but this one seems to also be valid according to the docs:

https://github.com/kbobrowski/en-api-exception-10-debug/blob/master/trace_all_error.txt

In the latter case full header is still properly received by `PipedInputStream`, but `read` method was called too early and returned just one byte (or sometimes a bit more).

Difficult to say based on this why separating calls by a timeout seems to be fixing it, probably it has to do something with timing of the start of async task which reads files

kbobrowski:
As to why separating calls to `provideDiagnosisKeys` help:

-

kbobrowski:
I think what happens is this:

If phone is under high load (calling `provideDiagnosisKeys` frequently) or simply has weaker processor then there is slightly longer delay between constructing `PipedInputStream` and calling it's `read` method. During this delay first byte is already written to `PipedInputStream` by `PipedOutputStream`, and this part of the code in `read` method passes then very fast since `in` is already more than -1:

```java
        try {
            int attempts = 3;
            while (in == -1) {
                // Are we at end of stream?
                if (isClosed) {
                    return -1;
                }
                if ((attempts-- <= 0) && lastWriter != null && !lastWriter.isAlive()) {
                    throw new IOException("Pipe broken");
                }
                // Notify callers of receive()
                notifyAll();
                wait(1000);
            }
        } catch (InterruptedException e) {
            IoUtils.throwInterruptedIoException();
        }
```

and then it flies through the rest of the method returning number less than 16. But if `read` is called before `PipedOutputStream` managed to write anything yet (time between constructing and calling method is very short), it will wait for first byte, and during this time much more than 16 bytes would already be written, so this method would return 16 as expected.

kbobrowski:
Tested it with whitelisted device and it does not look good, the error still happens even with large delay. Perhaps testing isolated `provideDiagnosisKeys` was making timing just right, which resulted in some rate of success, but testing this solution integrated into the app does not seem to solve the problem.

It's actually surprising that current implementation works at all, it looks like it is only thanks to the peculiarities of how `PipedOutputStream` and `PipedInputStream` interact in OpenJDK (former will block latter until everything is sent, this does not happen in Harmony). Closing - fix from Google should come very soon

Issue #822, opened 2020-07-05
Bug with time zones, causing exceeded quota (39508 error)


kbobrowski:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ ] Bug is not already reported in another issue

## Describe the bug

Exposure Notifications framework is resetting quota at midnight UTC, not at midnight in current time zone. This may be causing 39508.

Current design allows for initiating RetrieveDiagnosisKeysTransaction at 15:00 CEST and then next day again at 01:00 CEST. This will result in exceeded quota, as it corresponds to 13:00 UTC and 23:00 UTC on the same day.

Already reported in https://github.com/corona-warn-app/cwa-app-android/issues/774#issuecomment-653819198 https://github.com/corona-warn-app/cwa-app-android/issues/774#issuecomment-653821030 https://github.com/corona-warn-app/cwa-app-android/issues/774#issuecomment-653821218 but separating the issue here since #774 is about error 10 which is causing 39508.

eayin2:
I use UTC+7 instead of UTC+2 on Android 10 and I got following error message since today (I use this app over two weeks):
```
URSACHE 3:
Etwas ist schief gelaufen.
Fehler bei Kommunikation mit Google API(39508)
```
Removing the cache of Google Play Services didn't solve this error.

fredsleeve:
Can you post your Exposition Notification log for the last checks? At 0700 in your TZ the error should be gone.

jkrwdf:
I am in the UTC+2 (CEST) timezone and encounter this issue today for the first time (Pixel 3 XL, Android 10, GMS 20.21.17). 

My pattern matches the bug description. 

My personal action log as far as I remember:

- Opened CWA on 3.7. in the morning successfully
- Opened CWA on 4.7. in the morning successfully
- Opened CWA on 5.7. in the morning successfully
- Opened CWA on 6.7. at 00:01 with error message, reproducible multiple times 
- Opened CWA on 6.7. at 03:05 successfully

Verification Log in GMS UI (timestamps are shown obviously in CEST, number in brackets is the number of individual rows with the same timestamp):

```
3.7.: 06:53 (10)
4.7.: 08:38 (11)
[...]

Issue #321, opened 2020-06-09
Product flavor for using mock exposure notification API


pocmo:
In the code you mention a plan to add a product flavor that should use a mock exposure notification API:
https://github.com/corona-warn-app/cwa-app-android/blob/a326bb4bfc2661b3e5ff8e53a1951b5ad351a395/Corona-Warn-App/build.gradle#L67-L77

Is this something that could be done as a contribution? Looking at `InternalExposureNotificationClient` it seems like one option could be to inject a different/mock `ExposureNotificationClient` implementation that returns some dummy values. This at least would allow launching the app without any API support on the device. Did you have more elaborate plans? (e.g. controlling the mocked API via an activity or test app?).

Would be interested in contributing this if possible.

---
Internal Tracking ID: EXPOSUREAPP-1967

marcmuschko:
Dear @pocmo,

you are right, we have considered this but due to the tight time constraints and the change frequency within the app and the EN api, we were not able to ensure a properly maintained mock client and had to drop it for now.

This is definitely something that could be done as a contribution and we are interested in having this; it would make testing for collaborators without whitelisting much easier.
But we also want to be transparent about the inclusion into this repository when the mock client is developed and tested:
Mentioning time constraints and change frequency again, we cannot guarantee that we will be able to merge the feature, as we would have to maintain it.
My suggestion is a fork from master with the feature included which would then be independent from the official repository but offer a very nice addition to the project in general.

Knowing this, are you still willing to invest some effort? If so, please let me know and I'll assign you the issue! Perhaps some other colleagues are interested in collaborating together with you? 
Thanks and best,
Marc

pocmo:
Hey @marcmuschko,

thank you for your reply!

> This is definitely something that could be done as a contribution and we are interested in having this; it would make testing for collaborators without whitelisting much easier.

Is whitelisting something that is at all possible for outside contributors?

> Mentioning time constraints and change frequency again, we cannot guarantee that we will be able to merge the feature, as we would have to maintain it.

I can fully understand that you are not interested in maintaining this as part of this repository if this is not something that benefits the core team directly or in the worst case adds more work for them.

[...]

Issue #75, opened 2020-06-01
[DISCUSSION] Solution without closed source dependencys


theScrabi:
EDIT: **tl;dr**: We are working an an open implementation of the Google/Apple protocol (I call it PPCP, which may be wrong). You can find and contribute to our repository here: https://github.com/theScrabi/CoraLibre-android-sdk

Dear Corona-Warn-App developpers we have read your code and documentation. We like your effort and your open mind for the community. As there is a small but willing comunity of people who refuse to use closed source Google or Apple dependency and so can not use the Corona-Warn-App, I'd like to discuss the posibility of a Google Free FLOSS solution.

What we know so far is that it's currently hardly possible to create the corona app without the use of the closed source GMS. This comes along with the [nearby android exposure notification API](https://static.googleusercontent.com/media/www.google.com/en//covid19/exposurenotifications/pdfs/Android-Exposure-Notification-API-documentation-v1.3.2.pdf), which implements the [PPCP](https://www.apple.com/covid19/contacttracing) protocol. From what we can see in the code so far, the *contact* points between the android exposure notification API and the Corona-Warn-App are few as for the most common calls to the API [a wrapper class](https://github.com/corona-warn-app/cwa-app-android/blob/43412bdabe128dc674ec8a9ef8baa80d37292a3a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/InternalExposureNotificationClient.kt) is used. That's great! Also there are only about 14 code files that include parts from the GMS library. We therefor think it might be feasible to create a version of the app that does not require GSM services.

As statet in [this issue](https://github.com/corona-warn-app/cwa-documentation/issues/5#issuecomment-627848335) the current maintainer have no plans to implement an own exposure notificatoin API, but might be open to support an alternative if available.
What we understand so far is that the part that runs the actual BLE contact tracing as well as providing a highlevel API for exposure likelihood and key handling would need to be reimplemented. We had several thoughts about how this might be possible.

- As its offtenly staded that PPCP is close to [DP-3T](https://en.wikipedia.org/wiki/Decentralized_Privacy-Preserving_Proximity_Tracing), I question if it is possible to reuse some parts of the [prestandard DP-3T android sdk](https://github.com/DP-3T/dp3t-sdk-android/tree/prestandard).
[...]

PR #846, opened 2020-07-08
Added info about en-i13n project which enables whitelisting the app on rooted device

kbobrowski:
## Description

Information about [en-i13n project](https://github.com/kbobrowski/en-i13n) could be added in README, to increase awareness about possible way to become whitelisted on rooted device and test dev version of the app before it is released.

This has already proven beneficial in finding bugs responsible for #737 #822 and testing pull requests #771 #818 and probably others as well.

Would solve / contribute to solving:
- https://github.com/corona-warn-app/cwa-documentation/issues/214
- https://github.com/corona-warn-app/cwa-app-android/issues/321
- #75 
- https://github.com/corona-warn-app/cwa-backlog/issues/2
- #714 
- other issues that will come up in the future, as community would be able to test the application before it is rolled out via Play Store

This is working only by temporarily modifying Java Class Library used by GMS, and it all happens only in memory. There is no modification of GMS code, and any obfuscation that GMS is using is not revealed (full implementation of used hooks is [here](https://github.com/kbobrowski/en-i13n/blob/master/src/allow.ts)). Since it does not rely on obfuscated GMS classes it should robustly work across GMS updates without modifications.

It also allows affected Android 6 devices to patch a bug in GMS causing error 10 (#737) and successfully provide Diagnosis Keys to the framework.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,9 @@ The goal of this project is to develop the official Corona-Warn-App for Germany 
 
 ## Known Issues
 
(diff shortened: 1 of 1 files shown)


d4rken:

PR #876, opened 2020-07-15
Dependency Injection using Hilt

Fabian-K:
## Description
This PR starts to introduce dependency injection using hilt. So far it only covers backend connectivity. It also inlines the WebRequestBuilder into the matching Service classes.


d4rken:
A PR based on this one that adds DI via Dagger was merged with #1069

d4rken:
I understand that this PR also contains changes that add DI to various components but we need to make more granular PRs for each component to add DI.

PR #946, opened 2020-07-29
WIP: plausible deniability

Fabian-K:
🏗


Corona-Warn-App/build.gradle
```diff
@@ -244,11 +244,13 @@ dependencies {
     // TESTING
     testImplementation 'junit:junit:4.13'
     testImplementation "android.arch.core:core-testing:1.1.1"
+    testImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.3.7'
```
jakobmoellersap, line 247: this is still experimental and could cause issues. TBD

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/playbook/PlaybookImpl.kt
```diff
@@ -0,0 +1,143 @@
...
+
+    private suspend fun ignoreExceptions(body: suspend () -> Unit) {
+        try {
+            body.invoke()
+        } catch (e: Exception) {
+            Timber.d(e, "Ignoring dummy request exception")
+        }
+    }
+
+    private suspend fun <T> executeCapturingExceptions(body: suspend () -> T): Pair<T?, Exception?> {
```
jakobmoellersap, line 131: Maybe a simple throw can be enough here depending on the use case.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/service/VerificationService.kt
```diff
@@ -17,20 +17,23 @@ interface VerificationService {
...
         @Header("cwa-fake") fake: String,
+        @Header("cwa-header-padding") headerPadding: String?,
         @Body request: RegistrationRequest
     ): TestResultResponse
 
     @POST
     suspend fun getTAN(
         @Url url: String,
         @Header("cwa-fake") fake: String,
+        @Header("cwa-header-padding") headerPadding: String?,
```
jakobmoellersap, line 36: Is there a case where we do not have padding?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/playbook/PlaybookImpl.kt
Fabian-K: We need to ensure that following fake requests of the playbook are still executed even when the real request fails

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/service/VerificationService.kt
Fabian-K: I think currently you are right. The header can be omitted when the header of a request happens to be exactly as big as we need it.

Fabian-K:
Thank you for the review. Feedback is applied with 863b12d

jakobmoellerdev (dismissed):

jakobmoellerdev (approved):

Fabian42:
I'm interested: What does this change do? An emoji is not a very good description…

vaubaehn:
@Fabian42 : as far as I can interpret the code, a new function for data privacy is introduced. Randomly when opening the app, there is a fake TEK uploading to the server. It is to protect a COVID positive tested user, who uploads his keys to the server to warn the others, while his network traffic is under oberservation (e. g., by a hacker...) By this, he may deny (upon being confronted with 'his network behaviour') that he uploaded keys, because ALL CWA apps now uploading (fake) keys from time. So, network sniffing is not reliable anymore to identify positive tested users.

Dear devs, please correct me if I'm wrong.

vaubaehn:
@Fabian42 ... sorry, now realized that you most likely don't need any explanation about the code... Your original question had a different intention, I guess... ;)

Fabian42:
No, that is actually exactly what I asked about. Thanks!

PR #990, opened 2020-08-07
Fix the submission with zero keys (EXPOSUREAPP-1742)

AndroidMedaGalaxy:
## Checklist

* [x] [No submission possible : EXPOSUREAPP-1889](https://jira.itc.sap.com/browse/EXPOSUREAPP-1889) - Pull Requests that are not linked to an issue with you as an assignee will be closed, except for minor fixes for typos or grammar mistakes in *documentation or code comments*.

## Description

if there are no keys to submit, this is handled silently and submission done screen is loaded


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveOtherWarningFragment.kt
```diff
@@ -187,6 +187,7 @@ class SubmissionResultPositiveOtherWarningFragment : Fragment(),
         if (keys.isNotEmpty()) {
             submissionViewModel.submitDiagnosisKeys(keys)
         } else {
+            SubmissionService.submissionSuccessful()
```
kolyaopahle, line 190: Nitpick: From a separation of concerns perspective i would prefer it if the SubmissionService was not directly included in the Fragment but called from the ViewModel instead.

kolyaopahle (commented):
Just one small nitpick :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveOtherWarningFragment.kt
ritsam: Hi Kolya, since we do not send any network request, the current way seemed like the most efficient and straight forward.

@Fabian-K  what do you think of this?

kolyaopahle: My main concern here is not with networking but with DI, if we later add support for DI this fragment would turn into an additional injection point for the `SubmissionService`, which could be avoided if we only use the `SubmissionService` inside of the `SubmissionViewModel`

ritsam: @kolyaopahle changes have been made, please take a look.

Thank you

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/SubmissionService.kt
```diff
@@ -88,7 +88,6 @@ object SubmissionService {
 
     fun submissionSuccessful() {
         BackgroundWorkScheduler.stopWorkScheduler()
-        LocalData.numberOfSuccessfulSubmissions(1)
```
kolyaopahle, line 91: Is this not supposed to be called after a successfull submission of keys?

ritsam: that's correct, it's been rectified now.

kolyaopahle (dismissed):
LGTM 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/viewmodel/SubmissionViewModel.kt
```diff
@@ -112,6 +113,11 @@ class SubmissionViewModel : ViewModel() {
         SubmissionService.deleteTestGUID()
     }
 
+    fun submitWithNoDiagnosisKeys() {
+        BackgroundWorkScheduler.stopWorkScheduler()
+        LocalData.numberOfSuccessfulSubmissions(1)
```
Fabian-K, line 118: The SubmissionViewModel already has the SubmissionService as a dependency. Instead of duplicating the code here, we can call `SubmissionService.submissionSuccessful()`

ritsam: Had a word with Kolya yesterday, due to future changes such as DI, it was suggested that the method should be in ViewModel. I am aware of the fact that it's repeating, @kolyaopahle  what do you think? I can just call the SubmissionService.submissionSuccessful() function in Viewmodel but may not be a good idea to have inter-dependencies in ViewModel and submissionService. 

let me know what you think

Fabian-K: I think the original comment referred to not adding an additional dependency to the Fragment. The ViewModel already has a dependency to SubmissionService, therefore I see no issue with this.

Fabian-K (approved):

PR #1022, opened 2020-08-14
Plausible Deniability Config Update (EXPOSUREAPP-2214)

Fabian-K:
update the config for plausible deniability to
- not trigger a Dummy Playbook for 1% of all clients
- not repeating the playbook pattern for up to 16 days
- not trigger 1-3 repetitions for each communication to CWA/VS


kolyaopahle (approved):
LGTM 👍

Issue #912, opened 2020-07-23
[upcoming bug] exceeded quota when using with future versions of GMS


kbobrowski:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Describe the bug

I've reached out to Google to make documentation more precise as to when quota on `provideDiagnosisKeys` is set to zero (related to a bug described in https://github.com/corona-warn-app/cwa-app-android/issues/822), and was expecting description to simply state that quota is nullified at midnight UTC as it is done currently, but it was [updated to](https://developers.google.com/android/exposure-notifications/exposure-notifications-api#providediagnosiskeys):

> Days are 24-hour windows that begin at the time the method is first called.

It looks like this change will be simply triggered by GMS update, regardless of whether original or v1.5 API is used, which would result in exceeded quota in the following scenario:

- user updates Diagnosis Keys at 15:00 UTC
- user opens CWA at 02:00 UTC and update of Diagnosis Keys is triggered again
- quota is exceeded, as it was already at 14 at 15:00 UTC, and will go past 20 at 02:00 UTC update (quota is nullified only at 15:00 UTC next day)

## Possible Fix

One option is to track when Diagnosis Keys were last updated, and allow update only after 24 hours from the time this method was first called. But it would revert some progress in the efforts of reducing exposure notification delay as discussed in https://github.com/corona-warn-app/cwa-backlog/issues/2 . Following scenario would be possible:

- user updated Diagnosis Keys at 23:00
- next update is only possible next day at 23:00, despite fresh Diagnosis Keys being available already since 23 hours
- total delay in exposure notification (including delay due to consuming only daily bundles of Diagnosis Keys) is up to 48 hours, and cannot be reduced by the user in any way (currently it can be reduced to around 24 hours by opening CWA early in the morning)

Proper solution here is probably to quickly migrate to v1.5, then it would be possible to update Diagnosis Keys every 4 hours and the timing of when exactly quota is nullified would be introducing 4 hours additional delay in the worst case (even less if hourly bundles would be consumed), not 24 hours.

---
Internal Tracking ID: EXPOSUREAPP-1910

SebastianWolf-SAP:
[...]

PR #1041, opened 2020-08-19
Update UTC time to LocalTime (EXPOSUREAPP-1863)

SamuraiKek:
## Checklist

* [ ] [Link your Pull Request to an issue](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue) - Pull Requests that are not linked to an issue with you as an assignee will be closed, except for minor fixes for typos or grammar mistakes in *documentation or code comments*.

## Description

Changed the way our API quota is calculated based on the updated requirements from Google from using UTC to using the phone's local time


kbobrowski:
This will introduce exceeded quota error on devices which still use GMS version which resets quota at midnight UTC. Implementation should support both mechanisms until all devices have GMS version which resets quota at midnight local time. This pull request could be one way of implementing it: https://github.com/corona-warn-app/cwa-app-android/pull/931

kbobrowski:
by the way, it would be great if you could also reference existing github issues in pull requests (in this case https://github.com/corona-warn-app/cwa-app-android/issues/912), not only internal jira, would be much easier for the community to keep track of actions around the issues (I found this pull request just by accident), thanks!

SamuraiKek:
Hey @kbobrowski, from my understanding, it seems that the quota has always been calculated with device local time. I think there was a communication error with the guys from google and it was assumed that UTC would be used. Looking in the ENF documentation right now, it looks like it's been updated to clearly state that local time is used for both versions of the API.
CounterReset

kbobrowski:
Hi @SamuraiKek , I'm quite sure that limit on calls was set to zero at midnight UTC, the implementation of CWA actually was assuming local time in the past, which was leading to 39508 and was fixed by https://github.com/corona-warn-app/cwa-app-android/pull/818 .

Looking at the source code of GMS, version 20.30.19, it seems that currently it still defines internally "day" as midnight to midnight UTC, when it comes to setting the limit to zero. Relevant part of the code:

```smali
 120     invoke-static {}, Ljava/lang/System;->currentTimeMillis()J
 121 
 122     move-result-wide v11
 127 
 129     sget-object v14, Ljava/util/concurrent/TimeUnit;->HOURS:Ljava/util/concurrent/TimeUnit;
 133 
 134     const-wide/16 v7, 0x18
 135 
 137     invoke-virtual {v14, v7, v8}, Ljava/util/concurrent/TimeUnit;->toMillis(J)J
 138 
 139     move-result-wide v7
 140 
 141     div-long/2addr v11, v7
```

And loose translation of limit reset mechanism to java:

```java
public class Quota {

  private long day = 0;
  private int count = 0;

  public void updateCount() {
    long currentDay = System.currentTimeMillis() / TimeUnit.HOURS.toMillis(24);
    if (currentDay > day) {
      day = currentDay;
      count = 0;
    }
    if (count < 20) {
      count += 1;
    }
  }

  public boolean canCallProvideDiagnosisKeys() {
    return count < 20;
  }
}
```

I think that change from UTC to local time is upcoming, non-documented breaking change that will be introduced with future version of GMS.

If you have rooted android device you can have a look at the place where it keeps track of the number of calls, in:

`/data/data/com.google.android.gms/files/nearby/shared`

SamuraiKek:
Thanks @kbobrowski! It looks like you're right. We will try to clarify this with the team at Google before we take further action.

harambasicluka (approved):

PR #1067, opened 2020-08-23
Allow expandable long text in notification (related to #1066)

### LINKED ISSUE #1066: Enhancement to notification of risk change
## Avoid duplicates
* [x] Enhancement request is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)
* [x] If you are proposing a new feature, please do so in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)

## Current Implementation

Currently the notification does not have any title, and it cannot be expanded to view the full text

Screenshot_20200823-104942

## Suggested Enhancement

It could be displayed similar to notification which can be generated by ENF when querying `ExposureInformation`, which has a title and full text visible. Additionally it could include a message about the risk directly in the notification. Would be nice to also have the icon in the same color as original CWA logo, not grey, to make it a bit more prominent.

## Expected Benefits

Improved UX, better chance that notification is not dismissed accidentally.

kbobrowski:
## Description

It may happen that full notification text is never displayed, expanding notification allow user to read full text of the notification. I think it would be good to also include a simple title, like proposed in https://github.com/corona-warn-app/cwa-app-android/issues/1066#issuecomment-678758294 , but since this change would require strings modification I did not include it in this PR


d4rken (approved):
lgtm, thanks for your PR!

PR #1069, opened 2020-08-24
Introduce dagger 2, first risk level unit tests (EXPOSUREAPP-2207)

chris-cwa:


jakobmoellerdev (commented):
Looks good to me 👍  waiting for @d4rken to confirm.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelCalculationImpl.kt
```diff
@@ -0,0 +1,66 @@
...
+
+import com.google.android.gms.nearby.exposurenotification.ExposureSummary
+import de.rki.coronawarnapp.server.protocols.ApplicationConfigurationOuterClass
+import timber.log.Timber
+import javax.inject.Inject
+import javax.inject.Singleton
+import kotlin.math.round
+
+@Singleton
+class RiskLevelCalculationImpl @Inject constructor() : RiskLevelCalculation {
```
d4rken, line 11: With the idea that due to the interface we might replace the implementation or offer alternatives, should we maybe name this `DefaultRiskLevelCalculation`?

Corona-Warn-App/build.gradle
```diff
@@ -237,6 +237,13 @@ dependencies {
     implementation 'android.arch.lifecycle:extensions:1.1.1'
     implementation 'androidx.annotation:annotation:1.1.0'
 
+    // DAGGER
```
d4rken, line 240: Trying to build the `deviceForTesters` variant fails because it includes another file using the risk-level transaction that has not been refactored.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RiskLevelTransaction.kt
```diff
@@ -374,38 +387,59 @@ object RiskLevelTransaction : Transaction() {
...
+                highRiskScoreClass.max
+            )
+        ) {
+            Timber.v("$riskScore is above the defined min value ${highRiskScoreClass.min}")
+            return INCREASED_RISK
+        } else if (riskScore > highRiskScoreClass.max) {
+            throw RiskLevelCalculationException(
+                IllegalStateException("risk score is above the max threshold for score class")
+            )
+        }
```
d4rken, line 437: What do you think about moving the whole threshold check to `RiskScoreAnalysis`, i.e.

```kotlin
RiskScoreAnalysis.isIncreasedRisk(
    riskScore: Double,
    riskScoreClassification: ApplicationConfigurationOuterClass.RiskScoreClassification,
) : Boolean

```

d4rken (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RiskLevelTransaction.kt
chris-cwa: good idea, but not for this feature

d4rken (approved):

d4rken:
lgtm

jakobmoellerdev (approved):

jakobmoellersap:
EXPOSUREAPP-2255

PR #1073, opened 2020-08-25
Include test result retrieval in playbook for initial registration (EXPOSUREAPP-2314)

Fabian-K:
[EXPOSUREAPP-2314] Include test result retrieval in the playbook for initial registration

For plausible deniability, the test result retrieval after the initial registration (via QR or TeleTAN) needs to be part of the same playbook and not two playbooks executed sequentially. For this, this commit:
- updates the initial registration playbook to fetch the test result instead of a dummy request and returns the registration token as well as the test result
- adds an argument "skipInitialTestResultRefresh" to the SubmissionTestResultFragment to optionally skip the initial test result refresh
- includes the argument "skipInitialTestResultRefresh" when navigating from the QR Code scan or the TeleTAN input screen to the SubmissionTestResultFragment
- adds "updateTestResult" to the SubmissionRepository to be able to set the test result without having to refresh it every time from the server
- updates the SubmissionService to set the test result coming from the initial registration playbook


kolyaopahle (approved):
LGTM 👍

PR #1076, opened 2020-08-25
Storage check with optional allocation (EXPOSUREAPP-2259)

d4rken:
Preparations for  future features that require increased storage space.
Prevent features from failing unexpectedly due to missing free space on storage.
We want to make the user aware of any possible concerns regarding available storage space.

This introduces a class `DeviceStorage` that can check the free space in our app's private storage.
On newer APIs (26+), the class will also attempt to allocate additional space by asking the system delete caches if possible.

This PR adds a few test dependencies and sets up some structures to improve unit testing, i.e. see `DeviceStorageTest`.

This PR is in draft mode until #1069 is merged, so that I can directly include DI annotations.

### DIFF
--- a/Corona-Warn-App/build.gradle
+++ b/Corona-Warn-App/build.gradle
@@ -37,7 +37,7 @@ android {
         versionName "1.3.0"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
-        
+
         resConfigs "de", "en", "tr", "bg", "pl", "ro"
 
         buildConfigField "String", "DOWNLOAD_CDN_URL", "\"$DOWNLOAD_CDN_URL\""
@@ -142,6 +142,9 @@ android {
     }
 
     testOptions {
+        unitTests.all {
+            useJUnitPlatform()
+        }
         unitTests {
             includeAndroidResources = true
             returnDefaultValues = true
@@ -205,10 +208,10 @@ task jacocoTestReportDeviceRelease(type: JacocoReport, dependsOn: 'testDeviceRel
     ]
 
     def classPaths = [
-        "**/intermediates/classes/deviceRelease/**",
-        "**/intermediates/javac/deviceRelease/*/classes/**", // Android Gradle Plugin 3.2.x support.
-        "**/intermediates/javac/deviceRelease/classes/**", // Android Gradle Plugin 3.4 and 3.5 support.
-        "**/tmp/kotlin-classes/deviceRelease/**"
+            "**/intermediates/classes/deviceRelease/**",
+            "**/intermediates/javac/deviceRelease/*/classes/**", // Android Gradle Plugin 3.2.x support.
+            "**/intermediates/javac/deviceRelease/classes/**", // Android Gradle Plugin 3.4 and 3.5 support.
+            "**/tmp/kotlin-classes/deviceRelease/**"
     ]
 
     def debugTree = fileTree(dir: "$buildDir", includes: classPaths, excludes: excludes)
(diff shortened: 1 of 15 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/DeviceStorage.kt
```diff
@@ -0,0 +1,117 @@
...
+        var availableBytes = statsManager.getFreeBytes(storageUUID)
+
+        if (availableBytes < requiredBytes) {
+            val allocatableBytes = storageManager.getAllocatableBytes(storageUUID)
+            if (allocatableBytes + availableBytes >= requiredBytes) {
+                val toAllocate = requiredBytes - availableBytes
+                Timber.tag(TAG).v(
+                    "Not enough free space, allocating %d on %s.", requiredBytes, targetPath
+                )
+                storageManager.allocateBytes(storageUUID, toAllocate)
```
jakMoeller, line 46: This method may take several seconds to complete, so it should only be called from a worker thread.

Is this ensured by design?

Corona-Warn-App/build.gradle
```diff
@@ -37,7 +37,7 @@ android {
         versionName "1.3.0"
 
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
-        
+
```
jakMoeller, line 40: unintended newline?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/DeviceStorage.kt
```diff
@@ -0,0 +1,117 @@
...
+            } else {
+                requestStorageLegacy(path, requiredBytes)
+            }
+            Timber.tag(TAG).d("Requested %d from %s: %s", requiredBytes, path, result)
+            result
+        } catch (e: Exception) {
+            Timber.tag(TAG).e(e, "Failed to request %d on %s", requiredBytes, path)
+            // Android storage is a fickle beast
+            // If the space request fails, it's possible that the write attempt will still succeed
+            CheckResult(path, true, -1L, -1L)
```
jakMoeller, line 87: What if the write attempt fails, how is the use planned?

jakmoe (commented):
Looks good for me, just some clarifications before the approval.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/DeviceStorage.kt
d4rken: By design, this is not ensured by design :grin: , but must be done by the caller.
This will likely be called from the UI when making changes to options, or prior to downloading files.
We can't and don't want to make any assumptions here.

I'll clarify this in the entry method `checkSpacePrivateStorage` again. I think we can also add `@WorkerThread` to get some LINTer support with this.

Corona-Warn-App/build.gradle
d4rken: Codeformatter removed the indentation spaces on an empty line.

Someone apparently didn't run the formatter on it when making changes, I did when commiting, and apparently klint/lint doesn't check `build.gradle` as thoroughly as the `.kt` files.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/DeviceStorage.kt
d4rken: My concerns were that it has a slightly higher chance of failing via `StorageManager.getUuidForPath` in cases where the OEM did some funky things with the sdcard and the user swaps internal/external storage. I've already seen cases where they didn't give storage valid IDs.

You are right, this could be improved, instead of swallowing the error, we should fall back to the <API26 method, which has a very low chance of failing, and if that also fails, throw an exception (documented) and then the caller has to determine how to handle it.

d4rken: https://github.com/corona-warn-app/cwa-app-android/pull/1076/commits/bd244907317c1b98f4fd0685fd51667391fb283d

jakobmoellerdev (dismissed):

jakobmoellerdev (approved):

PR #1080, opened 2020-08-26
Let tests be run for all flavors

chris-cwa:
Ensure that CircleCI runs tests on all build flavors.


AlexanderAlferov (dismissed):

.circleci/config.yml
```diff
@@ -31,7 +31,7 @@ jobs:
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "Corona-Warn-App/build.gradle" }}-{{ checksum  "Server-Protocol-Buffer/build.gradle" }}
       - run:
           name: JaCoCo report
-          command: ./gradlew :Corona-Warn-App:jacocoTestReportDeviceRelease
+          command: ./gradlew :Corona-Warn-App:jacocoTestReportDeviceRelease :Corona-Warn-App:jacocoTestReportDeviceForTestersRelease
```
d4rken, line 34: This works, but I don't think it's the right approach.
With this we get unit tests on the `DeviceForTester` flavor only as side-effect of the jacocoTask.

Have you tried changing line 23? We have `quickBuild` and `quickBuildTest`, maybe we can run both there.

Corona-Warn-App/build.gradle
```diff
@@ -222,6 +222,43 @@ task jacocoTestReportDeviceRelease(type: JacocoReport, dependsOn: 'testDeviceRel
     getExecutionData().from(fileTree(dir: "$buildDir", includes: ["jacoco/testDeviceReleaseUnitTest.exec"]))
 }
 
+task jacocoTestReportDeviceForTestersRelease(type: JacocoReport, dependsOn: 'testDeviceForTestersReleaseUnitTest') {
```
d4rken, line 225: Looks good.
In the future we could setup a task that iterates over all release build flavors and generates jacoco reports.
Did you evaluate that?
I would be okay with the duplication here for now. Just curious.

d4rken:
I've removed the extra jacoco report generation as we can get the test results without now.

It's unclear how the extra coverage reports affect our results for the normal release build, and that's the important ones.

@AlexanderAlferov is checking that behavior and depending on his results we can re-add the task.

jakobmoellerdev (approved):

PR #1103, opened 2020-09-01
Initial implementation for interoperability EFGS consent (EXPOSUREAPP-2263)

SamuraiKek:
## Description

This implementation adds the new EFGS consent screen for interoperability. Navgraph is updated, only thing needed to be specified is the route to the next screen after EFGS consent


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveEfgsConsentFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+import de.rki.coronawarnapp.util.observeEvent
+
+class SubmissionResultPositiveEfgsConsentFragment : Fragment(),
+    InternalExposureNotificationPermissionHelper.Callback {
+
+    companion object {
+        private val TAG: String? = SubmissionResultPositiveEfgsConsentFragment::class.simpleName
+    }
+
+    private val efgsViewModel: EfgsConsentViewModel by activityViewModels()
```
harambasicluka, line 36: I'm personally not a big fan of using aberrations in the coding. Could you change this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveEfgsConsentFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+        submissionViewModel.submissionState.observe(viewLifecycleOwner, Observer {
+            if (it == ApiRequestState.SUCCESS) {
+                navigateToSubmissionDoneFragment()
+            }
+        })
+    }
+
+    private fun setButtonOnClickListener() {
+        val switch = binding.settingsEfgsConsentRow.settingsSwitchRowSwitch
+        binding.submissionPositiveEfgsConsentHeader.headerButtonBack.buttonIcon.setOnClickListener {
```
harambasicluka, line 145: maybe also move to variable as you did it for the switch

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveEfgsConsentFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+                ::navigateToSubmissionResultFragment
+            )
+        }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        setButtonOnClickListener()
+
+        submissionViewModel.submissionError.observeEvent(viewLifecycleOwner) {
```
harambasicluka, line 132: comments could be useful here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveEfgsConsentFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+            DialogHelper.showDialog(buildErrorDialog(it))
+        }
+
+        submissionViewModel.submissionState.observe(viewLifecycleOwner, Observer {
```
harambasicluka, line 136: also here comments could be useful

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/viewmodel/EfgsConsentViewModel.kt
```diff
@@ -0,0 +1,13 @@
+package de.rki.coronawarnapp.ui.viewmodel
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+class EfgsConsentViewModel : ViewModel() {
```
harambasicluka, line 6: maybe also remove abbreviation here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/viewmodel/EfgsConsentViewModel.kt
```diff
@@ -0,0 +1,13 @@
...
+
+    companion object {
+        val TAG: String? = TracingViewModel::class.simpleName
```
harambasicluka, line 9: I think that's not needed as we use timber.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/viewmodel/EfgsConsentViewModel.kt
```diff
@@ -0,0 +1,13 @@
...
+    }
+
+    val isEfgsConsentGranted: MutableLiveData<Boolean> = MutableLiveData<Boolean>(false)
```
harambasicluka, line 12: also here

Corona-Warn-App/src/main/res/layout/fragment_submission_positive_efgs_consent.xml
```diff
@@ -0,0 +1,147 @@
...
+            android:layout_height="wrap_content"
+            android:text="@string/submission_positive_efgs_consent_warning_button"
+            android:textAllCaps="true"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="@id/guideline_end"
+            app:layout_constraintStart_toStartOf="@id/guideline_start"
+            app:layout_constraintTop_toBottomOf="@id/guideline_action" />
+
+        <ProgressBar
+            android:id="@+id/submission_positive_other_warning_spinner"
```
harambasicluka, line 125: this should be submission_positive_efgs_consent_spinner

(maybe efgs will be replaced)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveEfgsConsentFragment.kt
```diff
@@ -0,0 +1,220 @@
...
+import de.rki.coronawarnapp.nearby.InternalExposureNotificationPermissionHelper
+import de.rki.coronawarnapp.ui.doNavigate
+import de.rki.coronawarnapp.ui.viewmodel.EfgsConsentViewModel
+import de.rki.coronawarnapp.ui.viewmodel.SubmissionViewModel
+import de.rki.coronawarnapp.ui.viewmodel.TracingViewModel
+import de.rki.coronawarnapp.util.DialogHelper
+import de.rki.coronawarnapp.util.IGNORE_CHANGE_TAG
+import de.rki.coronawarnapp.util.observeEvent
+
+class SubmissionResultPositiveEfgsConsentFragment : Fragment(),
```
harambasicluka, line 29: also add a comment what this fragment is used for

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -619,6 +619,12 @@
     <string name="settings_background_priority_card_button">"Open Device Settings"</string>
     <!-- XHED : settings(background priority) - headline on card about the current status and what to do -->
     <string name="settings_background_priority_card_headline">"Disable prioritized background activity"</string>
+    <!-- XTXT: settings(EFGS consent) - status next to switch under title -->
```
harambasicluka, line 622: Only add the strings to the german text files, even if you use placeholder texts. in this file only empty strings should be added, e.g. `<string name="settings_efgs_consent_status_active" />`

harambasicluka: relevant for all strings added in this PR

harambasicluka: And you could reference to our Lorem Ipsum string `lorem_ipsum`, that makes it easier to find this placeholders afterwards.

harambasicluka (changes requested):
I mainly looked at the code style in the layout files. Looks overall pretty good to me but I'm not sure about `efgs`, because I'm not a friend of aberrations.

And also follow the translation process like documented in our internal Wiki.

Thanks <3

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveEfgsConsentFragment.kt
SamuraiKek: Would it be descriptive enough if I changed the abbreviation to EuropeanConsentViewModel?

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -914,6 +914,18 @@
     <!-- XACT: other warning - illustration description, explanation image -->
     <string name="submission_positive_other_illustration_description">"Ein Smartphone übermittelt einen positiven Testbefund verschlüsselt ins System."</string>
 
+    <!-- Submission European Federal Gateway Server Consent -->
+    <!-- XHED: Page title for the positive result European Federal Gateway Server Consent page -->
+    <string name="submission_european_federal_gateway_server_consent_title">"Title Dummy Text"</string>
```
harambasicluka, line 919: If a text isn't defined please add `Lorem Ipsum` so that everyone sees that this is just a placeholder, you could refer to our `lorem_ipsum`.

harambasicluka: that's also relevant for the others

Corona-Warn-App/src/main/res/values/strings.xml
harambasicluka: still valide!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionEuropeanFederalGatewayServerConsentFragment.kt
```diff
@@ -0,0 +1,233 @@
...
+import de.rki.coronawarnapp.util.DialogHelper
+import de.rki.coronawarnapp.util.IGNORE_CHANGE_TAG
+import de.rki.coronawarnapp.util.observeEvent
+
+/**
+ *  This fragment is used for showing the user an option to share they're keys with other European
+ *  countries if they have been travelling outside of Germany.
+ *  If the user leaves the switch deactivated, the system dialog for submitting the keys is triggered
+ *  and the keys will only be submitted to the the German server.
+ *  Otherwise, Screen 3: Target > Germany and "Screen 4: Target specific countries" are shown, and
```
harambasicluka, line 34: maybe remove Screen 3 & Screen 4, as without context it's not very helpful :)

harambasicluka: So only the strings.

harambasicluka (changes requested):
It now looks really good to me, only some string changes have to be changed.

There is only one big question for me, why do you include this two new dependencies? I would like to see the opinion of @d4rken & @kolyaopahle here :)

Thanks!

Corona-Warn-App/build.gradle
```diff
@@ -241,6 +241,8 @@ dependencies {
     implementation('com.journeyapps:zxing-android-embedded:4.1.0') { transitive = false }
     // noinspection GradleDependency - needed for SDK 23 compatibility, in combination with com.journeyapps:zxing-android-embedded:4.1.0
     implementation 'com.google.zxing:core:3.3.0'
+    implementation 'androidx.lifecycle:lifecycle-extensions:2.1.0'
+    implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.1.0'
```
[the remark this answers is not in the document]
kolyaopahle: i was quite suprised to not already find this in our deps, especially as we make quite liberal use of `viewModelScope.launch` so i did some digging in our dependency tree and found out that our current implicit dependency is caused by our dependency on `androidx.navigation:navigation-fragment-ktx` which in its own transient dependencies asks for `lifecycle-viewmodel-ktx`. I think additionaly setting this as an explicit dep is fine and probably even good style but would also see if @d4rken shares this opinion first 👍

Corona-Warn-App/src/main/res/values-de/strings.xml
SamuraiKek: Ok, I'll will do this.

SamuraiKek:
I also realised there's no need to use two ViewModels for my fragment so I'll move the logic to SubmissionViewModel and use it as our only ViewModel for Submission fragments.

PR #1106, opened 2020-09-02
New Strings for Risikobegegnung mit niedrigem Risiko (EXPOSUREAPP-2462)

chris-cwa:


harambasicluka (approved):

janetback (approved):
UA approved.

vaubaehn:
Hi @chris-cwa , @janetback and @harambasicluka , sorry for coming in late - and I hope I don't bother too much.

I suppose the wording change of
`<item quantity="zero">"Bisher keine Begegnungen mit niedrigem Risiko"</item>`
was requested by the RKI.
This might become subject of discussion in the future, as "keine Begegnung mit niedrigem Risiko" does not exclude that there might have been an exposure with high risk... so, it's maybe unlucky/unprecise wording.
I would instead suggest to use
`<item quantity="zero">"Bisher keine Begegnungen mit einem Infektionsrisiko"</item>`
or stay with the previous wording.
If you find this hint useful, you might want to address it to RKI.

Thank you.

PR #1112, opened 2020-09-03
Improve Explanation for "Risikobegegnung mit niedrigem Risiko" (EXPOSUREAPP-1971)

chris-cwa:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/viewmodel/TracingViewModel.kt
```diff
@@ -53,6 +59,52 @@ class TracingViewModel : ViewModel() {
     val activeTracingDaysInRetentionPeriod = TracingRepository.activeTracingDaysInRetentionPeriod
     var isRefreshing: LiveData<Boolean> = TracingRepository.isRefreshing
 
+    val additionalInformationVisibility = MediatorLiveData<Int>()
```
harambasicluka, line 62: Is there a reason why you started using `MediatorLiveData`? I think we haven't used it somewhere else.

chris-cwa: because this live data generates its value by the values of 2 other livedatas - so whenever one of them is changed, this live data is also immediately updated

kolyaopahle: I've also started to use it for one of the interopt screens, i think it helps quite a lot when trying to formulate more complex view states that need to update with LiveData changes.

kolyaopahle: although im not 100% sure these specific cases could not be solved by using a formatter

kolyaopahle (approved):
LGTM 👍

PR #1139, opened 2020-09-10
Revert "Consolidate all number of retrieve diagnosis keys into one (#1088) (EXPOSUREAPP-2405, EXPOSUREAPP-2470) in DEV for 1.4 RCs

jakobmoellerdev:
This reverts commit 6e8b0535

The revert is done as we will only release the changed behaviour of the API Submission to Google in 1.5 and keep this in the interoperability feature branch

Signed-off-by: d067928 <jakob.moeller@sap.com>


vaubaehn:
Hi, could you kindly explain in a few lines, what is the reason for that decision?
I was hoping that the consolidation would decrease the chance of timeouts when submitting keyfiles to API, hence less people should have problems with (automatic) update of risk status.
Implementing it in 1.5 could prolong the time users complaining about problems for about a month or so.

@kbobrowski : FYI

harambasicluka (approved):

jakobmoellersap:
We decided to do a branching approach to cover the diff with 1.4

vaubaehn:
Thanks for your feedback!

PR #1142, opened 2020-09-10
Submission: Warning if user cancels submission (EXPOSUREAPP-2509)

AndroidMedaGalaxy:
## Implementation screenshot
submission_flow_cancelled

## Description

When a user presses the back navigation button on top or the back navigation button for Android system,
1. Users will get a warning message about cancelling the keys submission flow. They will have 2 options, Cancel and Ok respectively

2. If they choose **Yes** the user starts again from Testergebnis screen

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveOtherWarningFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveOtherWarningFragment.kt
@@ -166,7 +166,7 @@ class SubmissionResultPositiveOtherWarningFragment : Fragment(),
             R.string.submission_error_dialog_confirm_cancellation_body,
             R.string.submission_error_dialog_confirm_cancellation_button_positive,
             R.string.submission_error_dialog_confirm_cancellation_button_negative,
-            false,
+            true,
             ::navigateToSubmissionResultFragment
         ))
     }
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveOtherWarningFragment.kt
```diff
@@ -139,10 +149,28 @@ class SubmissionResultPositiveOtherWarningFragment : Fragment(),
...
+    * @see navigateToSubmissionResultFragment
+    */
+    fun handleSubmissionCancellation() {
+        DialogHelper.showDialog(DialogHelper.DialogInstance(
+            requireActivity(),
+            R.string.submission_error_dialog_confirm_cancellation_title,
+            R.string.submission_error_dialog_confirm_cancellation_body,
+            R.string.submission_error_dialog_confirm_cancellation_button_positive,
+            R.string.submission_error_dialog_confirm_cancellation_button_negative,
+            false,
```
d4rken, line 169: Why is `cancelable=false` here? It means that the user can't exit the dialog with the backbutton, which differs from other dialogs on the same screen.

d4rken (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/SubmissionResultPositiveOtherWarningFragment.kt
ritsam: I'll make that change.

d4rken (approved):
lgtm

janetback (approved):
UA approved

PR #1143, opened 2020-09-11
Fix: Line break on Imprint in german (EXPOSUREAPP-1530)

harambasicluka:
I added a hypen and a forced linebreak as the german word "Umsatzsteueridentifikationsnummer" is to long for most devices.
This word is now displayed over two lines:
image


janetback (approved):
UA Reviewed

d4rken (approved):

PR #1150, opened 2020-09-14
Version Checker with Branching for Checks on provideDiagnosisKeys below v16 of ENF (EXPOSUREAPP-2213)

jakobmoellerdev:
This PR features a small class for checking the API Version to branch off logic. It also includes a branching for provideDiagnosisKeys in the RetrieveDiagnosisKeysTransaction. It can be tested by comparing devices with different versions of GMS and seeing how many provideDiagnosisKeys calls are executed in the transaction.

Expectations:
- Versions always have exactly 8 digits, and can be compared by using their raw value
Cases:
- `<16xxxxxx` should query with 14 calls
- `<15xxxxxx` an exception is converted to 14 calls as we expect getVersion() to be unavailable due to a too old version of ENF
- `>=16xxxxxx` should query with 1 call

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
@@ -1,0 +1,34 @@
+package de.rki.coronawarnapp.util
+
+import com.google.android.gms.common.api.ApiException
+import com.google.android.gms.common.api.CommonStatusCodes
+import de.rki.coronawarnapp.nearby.InternalExposureNotificationClient
+import kotlin.math.abs
+
+class GoogleAPIVersion {
+    /**
+     * Indicates if the client runs above a certain version
+     *
+     * @return isAboveVersion, if connected to an old unsupported version, return false
+     */
+    suspend fun isAbove(compareVersion: Long): Boolean {
+        if (!compareVersion.isCorrectVersionLength) {
+            throw IllegalArgumentException("given version has incorrect length")
+        }
+        return try {
+            compareVersion > InternalExposureNotificationClient.getVersion()
+        } catch (apiException: ApiException) {
+            if (apiException.statusCode == CommonStatusCodes.API_NOT_CONNECTED) false
+            else throw apiException
+        }
+    }
+
+    // check if a raw long has the correct length to be considered an API version
+    private val Long.isCorrectVersionLength
+        get(): Boolean = abs(this).toString().length == GOOGLE_API_VERSION_FIELD_LENGTH
+
+    companion object {
+        private const val GOOGLE_API_VERSION_FIELD_LENGTH = 8
+        const val V16 = 16000000L
+    }
+}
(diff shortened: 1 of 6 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
```diff
@@ -0,0 +1,35 @@
+package de.rki.coronawarnapp.util
+
+import com.google.android.gms.common.api.ApiException
+import com.google.android.gms.common.api.CommonStatusCodes
+import de.rki.coronawarnapp.nearby.InternalExposureNotificationClient
+import kotlin.math.abs
+
+class GoogleAPIVersion {
```
d4rken, line 8: Maybe make this injectable. `@Inject constructor()`. The class is stateless, but can be recreated, so we can also tag the class `@Reusable`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/GoogleAPIVersionTest.kt
```diff
@@ -0,0 +1,79 @@
...
+
+    private lateinit var classUnderTest: GoogleAPIVersion
+
+    @BeforeEach
+    fun setUp() {
+        mockkObject(InternalExposureNotificationClient)
+        classUnderTest = GoogleAPIVersion()
+    }
+
+    @ExperimentalCoroutinesApi
```
d4rken, line 30: You could annotate the whole class with that instead of each function.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/GoogleAPIVersionTest.kt
```diff
@@ -0,0 +1,79 @@
...
+    fun `isAbove API v16 false when APIException for too low version`() {
+        coEvery { InternalExposureNotificationClient.getVersion() } throws
+                ApiException(Status(API_NOT_CONNECTED))
+
+        runBlockingTest {
+            classUnderTest.isAbove(GoogleAPIVersion.V16) shouldBe false
+        }
+    }
+
+    @AfterEach
```
d4rken, line 75: Personally I prefer AfterEach right below BeforeEach to keep a better overview of what is being setup and torn down. what do you think?

d4rken (approved):
lgtm, waiting for CircleCI after dev merge

BMItr (approved):

vaubaehn:
This is even better than #1088...
According to the label, is there any chance to find this in the release build of 1.4.0? That would be awsome!

Just one minor suggestion (hope not to bother...): The name of method `isAbove` implies a GreaterThan-logic for me, while the intended logic is GreaterOrEqual. Maybe re-naming to `isAtLeast` to avoid later confusion?

d4rken:
> According to the label, is there any chance to find this in the release build of 1.4.0? That would be awesome!

Likely :slightly_smiling_face: 

> Just one minor suggestion (hope not to bother...): The name of method isAbove implies a GreaterThan-logic for me, while the intended logic is GreaterOrEqual. Maybe re-naming to isAtLeast to avoid later confusion?

Good idea :+1:  #1159

PR #1151, opened 2020-09-14
Introduce ViewModel Injection

d4rken:
## Description
This PR adds the boilerplate necessary for combining ViewModels and dependency injection. As guide/demo the `TestRiskLevelCalculationFragment` has been refactored.

Our injected "ViewModels" are called `VDC` (View Data Controller/Component) because "ViewModel" is a confusing name and overloading the existing `viewModels` extensions would lead to confusion.

Using a `VDC` requires 3 things:
**In a dagger `Module` we need to add a `VDC.Factory` to a lookup map:**
```kotlin
@Module
abstract class TestRiskLevelCalculationFragmentModule {
    @Binds
    @IntoMap
    @VDCKey(TestRiskLevelCalculationFragmentVDC::class)
    abstract fun testRiskLevelFragment(factory: TestRiskLevelCalculationFragmentVDC.Factory): VDCFactory<out VDC>
}
```

**In the `Fragment` we need to instantiate it**
```kotlin
    @Inject lateinit var vdcSource: VDCSource.Factory
    private val vdc: TestRiskLevelCalculationFragmentVDC by vdcsAssisted(
        { vdcSource },
        { factory, handle ->
            factory as TestRiskLevelCalculationFragmentVDC.Factory
            factory.create(handle, navArgs.exampleArgument)
        }
    )
```

* `vdcSource` is a factory that has access to the lookup map we created in step one.
* `vdcsAssisted` is a little indirection that allows us to have an injected class with manual arguments, meaning we can have both injected parameters from our DI graph, and manual parameters, such as fragment/nav graph arguments.

**Finally we need the `VDC` (ViewModel) itself:**
```kotlin
class TestRiskLevelCalculationFragmentVDC @AssistedInject constructor(
    @Assisted private val handle: SavedStateHandle,
    @Assisted private val exampleArg: String?,
    private val context: Context, // App context injected
) : VDC() {

    init {
        Timber.d("VDC: %s", this)
        Timber.d("SavedStateHandle: %s", handle)
        Timber.d("Example arg: %s", exampleArg)
    }

    @AssistedInject.Factory
    interface Factory : VDCFactory<TestRiskLevelCalculationFragmentVDC> {
        fun create(
            handle: SavedStateHandle,
            exampleArg: String?
        ): TestRiskLevelCalculationFragmentVDC
    }
}
```
* Note the little differences to normal Dagger DI, i.e. `AssistedInject`,and `Assisted`. Further details: https://github.com/square/AssistedInject
* `SavedStateHandle` is a framework class that offers behavior similar to `onSavedInstance`.

**Now use the vdc/viewmodel in your fragment by moving logic into it, and keeping UI updates out of it, i.e.**
```kotlin
        binding.buttonResetRiskLevel.setOnClickListener { vdc.resetRiskLevel() }
        vdc.riskLevelResetEvent.observe2(this) {
            Toast.makeText(
                requireContext(), "Reset done, please fetch diagnosis keys from server again",
                Toast.LENGTH_SHORT
            ).show()
        }
    }
```

## How to test
* Launch the app and use the TestRiskLevelCalculationFragment
* Review the code


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/LiveDataExtensions.kt
```diff
@@ -0,0 +1,14 @@
+package de.rki.coronawarnapp.util.ui
+
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.Observer
+
+fun <T> LiveData<T>.observe2(fragment: Fragment, callback: (T) -> Unit) {
```
jakobmoellersap, line 8: function names with numbers are not great to read imo and we should try to make the naming more explicit

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModelSource.kt
```diff
@@ -0,0 +1,49 @@
...
+
+import android.os.Bundle
+import androidx.lifecycle.AbstractSavedStateViewModelFactory
+import androidx.lifecycle.SavedStateHandle
+import androidx.lifecycle.ViewModel
+import androidx.savedstate.SavedStateRegistryOwner
+import com.squareup.inject.assisted.Assisted
+import com.squareup.inject.assisted.AssistedInject
+
+class CWAViewModelSource @AssistedInject constructor(
```
jakobmoellersap, line 11: I think the implementation should be called factory as well, even though it is our source we should still stick to pattern, especially when customizing already generic code.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModelSource.kt
```diff
@@ -0,0 +1,49 @@
...
+        if (factory == null) throw IllegalStateException("Unknown ViewModel factory: $modelClass")
+
+        @Suppress("UNCHECKED_CAST")
+        var vm: T? = assistAction?.invoke(factory, handle) as? T
+
+        @Suppress("UNCHECKED_CAST")
+        if (vm == null) {
+            vm = when (factory) {
+                is SavedStateCWAViewModelFactory<*> -> factory.create(handle) as? T
+                is SimpleCWAViewModelFactory<*> -> factory.create() as? T
```
jakobmoellersap, line 33: is there not an option to just use the handle as optional parameter and then use a generic factory?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/SingleLiveEvent.kt
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * This avoids a common problem with events: on configuration change (like rotation) an update
+ * can be emitted if the observer is active. This LiveData only calls the observable if there's an
+ * explicit call to setValue() or call().
+ *
+ *
+ * Note that only one observer is going to be notified of changes.
+ * https://github.com/android/architecture-samples/blob/166ca3a93ad14c6e224a3ea9bfcbd773eb048fb0/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/SingleLiveEvent.java
+ */
+class SingleLiveEvent<T> : MutableLiveData<T>() {
```
jakobmoellersap, line 40: Shouldn't this still be LiveData? calling it event is imo not correct since an event happens exactly once. However this can be triggered multiple times, e.g. through multiple setValue() calls. Just my opinion, no high prio.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AppInjector.kt
```diff
@@ -9,4 +17,32 @@ object AppInjector {
         component = DaggerApplicationComponent.factory().create(app)
         component.inject(app)
     }
+
+    fun setup(activity: Activity) {
+        Timber.tag(TAG).d("Injecting %s", activity)
+
+        // Using lifecycle callbacks would be even more awesome,
+        // but Activity.onPreCreate isn't available for our minAPI
```
jakobmoellersap, line 25: should we branch out depending on API version?

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/risklevel/ui/TestRiskLevelCalculationFragmentCWAViewModel.kt
```diff
@@ -0,0 +1,291 @@
...
+        val exposureSummary: String = "",
+        val formula: String = "",
+        val fullConfig: String = "",
+        val exposureInfo: String = ""
+    )
+
+    fun startENFObserver() {
+        viewModelScope.launch {
+            try {
+                var workState = riskScoreState.value!!
```
jakobmoellersap, line 117: shouldn't we handle a null value more explicit?

jakobmoellerdev (commented):
Minor Remarks in the code. Overall I think we should evaluate whether the gain of using pure constructor-based DI is worth the overhead instead of using lateinit vars.

Waiting for more reviews to give approval 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/LiveDataExtensions.kt
d4rken: This is not necessary to be used, but codesugar that makes code look much nicer. I usually use this naming to denote, that a function is functionally the same, but an upgrade. Autocompletion would show this variant to someone trying to use `.observe`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModelSource.kt
d4rken: So `CWAViewModelFactory.Factory` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/SingleLiveEvent.kt
d4rken: I think `Data` would be missleading here because the data (or rather event here) is consumed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModelSource.kt
d4rken: These default factories are used when a viewmodel doesn't need extra assisted arguments, e.g. so the fragment can have

```kotlin
class Fragment .... {
    @AssistedInject.Factory
    interface Factory : SavedStateVDCFactory<GeneratorsFragmentVDC>

}
```

instead of 

```kotlin
class StrawberryViewModel .... {
    @AssistedInject.Factory
    interface Factory : SavedStateVDCFactory<GeneratorsFragmentVDC> {
         fun create(handle: SavedStateHandle): StrawberryViewModel
    }

}
```

Default parameters wouldn't work as it would then still require any user of `SimpleCWAViewModelFactory` to explicitly implement that parameter, i.e.

```kotlin
class ViewModel(private val iDontNeedTheSavedState: SavedStateHandle?)
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AppInjector.kt
d4rken: IMHO, no, as we would have different initialization behaviors depending on API version and could have bugs just on some  devices, increasing the chance for bugs that are not easily noticed/reproduced in tests. Too risky too little gain...

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/risklevel/ui/TestRiskLevelCalculationFragmentCWAViewModel.kt
d4rken: It can never be null as we initialize with a value and the data type is not nullable:
```kotlin
val riskScoreState = MutableLiveData<RiskScoreState>(RiskScoreState())
```

jakobmoellerdev (approved):

PR #1153, opened 2020-09-14
Reverted replaced wording (EXPOSUREAPP-2638)

chris-cwa:

### DIFF
--- a/Corona-Warn-App/src/main/res/values/strings.xml
+++ b/Corona-Warn-App/src/main/res/values/strings.xml
@@ -137,6 +137,8 @@
 
     <!-- XTXT: risk card - no contact yet -->
     <string name="risk_card_body_contact">"No exposure up to now"</string>
+    <!-- XTXT: risk card - no exposures with low risk -->
+    <string name="risk_card_body_contact_low_risk"/>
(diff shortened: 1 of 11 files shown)


Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -138,7 +138,7 @@
     <!-- XTXT: risk card - no contact yet -->
     <string name="risk_card_body_contact">"No exposure up to now"</string>
     <!-- XTXT: risk card - no exposures with low risk -->
-    <string name="risk_card_body_contact_low_risk">"No exposure with low risk so far"</string>
```
harambasicluka, line 141: If the string is removed in values.de it should be removed in all other languages (in my opinion), that means that line 140 and 141 have to be deleted.

harambasicluka (changes requested):

harambasicluka (changes requested):
Could you please add screenshots with low risk and in in german with the following conditions:
* zero encounters
* one or more encounters

TBH I think everything is fine, I just want to be sure :)

harambasicluka (approved):

PR #1177, opened 2020-09-17
1.3.1 into 1.4.0

d4rken:
Merge the hotfix branch into our next RC.


jakobmoellerdev (approved):

harambasicluka (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/debug/FileLoggerTree.kt
```diff
@@ -0,0 +1,86 @@
...
+            }
+            Log.i(TAG, "File logger stopped.")
+        }
+    }
+
+    override fun log(priority: Int, tag: String?, message: String, t: Throwable?) {
+        logWriter?.let {
+            try {
+                it.write("${System.currentTimeMillis()}  ${priorityToString(priority)}/$tag: $message\n")
+                it.flush()
```
BMItter, line 59: flushing is done after every logline. This happens usually in milliseconds and can penetrate storage memory. I would recommend to buffer in general but I think its ok for a temp solution.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/debug/FileLoggerTree.kt
```diff
@@ -0,0 +1,86 @@
...
+            try {
+                it.write("=== END ===\n")
+                it.close()
+            } catch (ignore: IOException) {
+            }
+            Log.i(TAG, "File logger stopped.")
+        }
+    }
+
+    override fun log(priority: Int, tag: String?, message: String, t: Throwable?) {
```
BMItter, line 55: I would recommend  to I/O async or in background thread...

BMItr (approved):

PR #1190, opened 2020-09-21
Improve low storage error message (EXPOSUREAPP-2745)

d4rken:
## Description
I've improved the low storage error message and added logic that will be used to display a formatted error message when we have smarter error tracking/display.

## How to test
* Just check the code, the actual error display is already there, I've just improved the text.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/InsufficientStorageException.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/InsufficientStorageException.kt
@@ -1,7 +1,20 @@
 package de.rki.coronawarnapp.storage
 
+import android.content.Context
+import android.text.format.Formatter
+import de.rki.coronawarnapp.util.FormattedError
 import java.io.IOException
 
 class InsufficientStorageException(
     val result: DeviceStorage.CheckResult
-) : IOException("Not enough free space (Want:${result.requiredBytes}; Have:${result.freeBytes}")
+) : IOException(
+    "Not enough free space: ${result.requiredBytes}B are required and only ${result.freeBytes}B are available."
+), FormattedError {
+
+    override fun getFormattedError(context: Context): String {
+        val formattedRequired = Formatter.formatShortFileSize(context, result.requiredBytes)
+        val formattedFree = Formatter.formatShortFileSize(context, result.freeBytes)
+        // TODO Replace with localized message when the exception is logged via new error tracking.
+        return "Not enough free space: $formattedRequired are required and only $formattedFree are available.)"
+    }
+}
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/InsufficientStorageException.kt
```diff
@@ -1,7 +1,20 @@
...
-) : IOException("Not enough free space (Want:${result.requiredBytes}; Have:${result.freeBytes}")
+) : IOException(
+    "Not enough free space: ${result.requiredBytes}B are required and only ${result.freeBytes}B are available."
+), FormattedError {
+
+    override fun getFormattedError(context: Context): String {
+        val formattedRequired = Formatter.formatShortFileSize(context, result.requiredBytes)
+        val formattedFree = Formatter.formatShortFileSize(context, result.freeBytes)
+        // TODO Replace with localized message when the exception is logged via new error tracking.
+        return "Not enough free space: $formattedRequired are required and only $formattedFree are available.)"
```
chris-cwa, line 18: why is there a ')'?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/FormattedError.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import android.content.Context
+
+interface FormattedError {
+    fun getFormattedError(context: Context): String
+}
+
+fun Throwable.tryFormattedError(context: Context): String = when {
+    this is FormattedError -> this.getFormattedError(context)
+    localizedMessage != null -> localizedMessage!!
```
chris-cwa, line 11: style: copy value of localizedMessage in local val, because localizedMessage could theoretically change between the two accesses

chris-cwa (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/FormattedError.kt
```diff
@@ -0,0 +1,13 @@
+package de.rki.coronawarnapp.util
+
+import android.content.Context
+
+interface FormattedError {
+    fun getFormattedError(context: Context): String
+}
+
+fun Throwable.tryFormattedError(context: Context): String = when {
```
chris-cwa, line 9: redundant return type specification

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/InsufficientStorageException.kt
d4rken: Brackets for everyone! :grin: fixed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/FormattedError.kt
d4rken: Good point!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/FormattedError.kt
d4rken: I think specifying a return type is okay if it makes it easier to read (and prevents accidental type changes when using `when`). Unless you feel strongly about this, I'll leave it as is.

chris-cwa (approved):

PR #1193, opened 2020-09-21
TAN Hotline no. not clickable (EXPOSUREAPP-2551)

Oliver-Zimmerman:
## Description
Made the actual TAN hotline number clickable. When clicked, the hotline number will be dialled in the same fashion as when the button is clicked. 

- Declared textview clickable within xml layout
- Reused dial logic from button and used for now clickable text  

### Test
- Navigate to hotline test submission option
- Select the blue hotline number
- Hotline number should be dialed. 

Closes JIRA ticket EXPOSUREAPP-2551


chris-cwa (dismissed):
how about moving dial() method to view model?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/fragment/SubmissionContactFragment.kt
```diff
@@ -36,6 +37,9 @@ class SubmissionContactFragment : Fragment(R.layout.fragment_submission_contact)
         binding.submissionContactButtonCall.setOnClickListener {
             dial()
         }
+        binding.submissionContactRoot.submission_contact_step_1_number.setOnClickListener {
```
d4rken, line 40: This uses `kotlinx.android.synthetic`, can we avoid that here?

d4rken (approved):

chris-cwa (approved):

PR #1201, opened 2020-09-21
QuotaCalculator v2

d4rken:
## Description
The previous one had to be removed due to unforeseen side-effects in relation to transaction rollbacks.

* Created a new `ENFClient` which currently only has a `DiagnosisKeyProvider` module, but will in the future have further modules until the singleton `InternalExposureNotificationClient` is empty.
* Added additional logging statements to better understand transaction failures via logfile.

## How to test
* Review the code
* Think about whether we have covered all quota edge cases in `SubmissionQuotaTest`
* Run the app, force key retrieval via test menu, observe logcat (filter for `SubmissionQuota`)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
```diff
@@ -0,0 +1,38 @@
...
+        token: String
+    ): Boolean {
+        Timber.tag(TAG).d(
+            "asyncProvideDiagnosisKeys(keyFiles=%s, configuration=%s, token=%s)",
+            keyFiles, configuration, token
+        )
+        return diagnosisKeyProvider.provideDiagnosisKeys(keyFiles, configuration, token)
+    }
+
+    companion object {
```
ralfgehrer, line 35: Shouldn't companion objects be on top of the class !?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
```diff
@@ -0,0 +1,38 @@
...
+    // i.e. in **[InternalExposureNotificationClient]**
+    internal val internalClient: ExposureNotificationClient
+        get() = googleENFClient
+
+    override suspend fun provideDiagnosisKeys(
+        keyFiles: Collection<File>,
+        configuration: ExposureConfiguration?,
+        token: String
+    ): Boolean {
+        Timber.tag(TAG).d(
```
ralfgehrer, line 28: The tag isn't necessary in this case IMHO. Timber logs the class name automatically.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClientLocalData.kt
```diff
@@ -0,0 +1,34 @@
...
+            putLong(PKEY_QUOTA_LAST_RESET, value.millis)
+        }
+
+    var currentQuota: Int
+        get() = prefs.getInt(PKEY_QUOTA_CURRENT, 0)
+        set(value) = prefs.edit(true) {
+            putInt(PKEY_QUOTA_CURRENT, value)
+        }
+
+    companion object {
```
ralfgehrer, line 30: at the top of the class?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
```diff
@@ -0,0 +1,110 @@
...
+    ): Void = suspendCoroutine { cont ->
+        Timber.tag(TAG).d("Performing key submission.")
+        enfClient
+            .provideDiagnosisKeys(keyFiles.toList(), configuration, token)
+            .addOnSuccessListener { cont.resume(it) }
+            .addOnFailureListener { cont.resumeWithException(it) }
+    }
+
+    companion object {
+        private val TAG: String = DefaultDiagnosisKeyProvider::class.java.simpleName
```
ralfgehrer, line 108: same as above for companion object and Timber  ... also applies to other classes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
```diff
@@ -0,0 +1,110 @@
...
+        configuration: ExposureConfiguration?,
+        token: String
+    ): Boolean {
+        return try {
+            if (keyFiles.isEmpty()) {
+                Timber.tag(TAG).d("No key files submitted, returning early.")
+                return true
+            }
+
+            val usedConfiguration = if (configuration == null) {
```
ralfgehrer, line 34: How about 
`val usedConfiguration = configuration ?:  ExposureConfiguration.ExposureConfigurationBuilder().build()`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisKeyRetrievalOneTimeWorker.kt
```diff
@@ -27,28 +27,40 @@ class DiagnosisKeyRetrievalOneTimeWorker(val context: Context, workerParams: Wor
      * @see RetrieveDiagnosisKeysTransaction
      */
     override suspend fun doWork(): Result {
-        Timber.d("Background job started. Run attempt: $runAttemptCount ")
+        Timber.tag(TAG).d("$id: doWork() started. Run attempt: $runAttemptCount")
```
ralfgehrer, line 30: Nothing major but I notice you mix how variables are logged within Timber messages:
`Timber.tag(TAG).d("$id: doWork() started. Run attempt: $runAttemptCount")` (kotlin approach)
vs 
`Timber.d("sendDebugNotification(title=%s, content=%s)", title, content)` (Timber approach)

Ideally, we would align on this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeStamper.kt
```diff
@@ -0,0 +1,12 @@
+package de.rki.coronawarnapp.util
+
+import org.joda.time.Instant
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class TimeStamper @Inject constructor() {
+
+    val nowUTC: Instant
```
ralfgehrer, line 10: not consequently used. I'd prefer Instant.now() instead of this helper.

ralfgehrer (dismissed):
Looks good overall. Commented on some minor code style things.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DiagnosisKeyProvider.kt
```diff
@@ -0,0 +1,25 @@
+package de.rki.coronawarnapp.nearby.modules.diagnosiskeyprovider
+
+import com.google.android.gms.nearby.exposurenotification.ExposureConfiguration
+import java.io.File
+
+interface DiagnosisKeyProvider {
```
chris-cwa, line 6: why is this called a provider? this class takes something / "consumes" it and does not provide something.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeStamper.kt
chris-cwa: I agree, we should probably also replace all other calls to Instant.now().

chris-cwa: I think we do need this class so that testing is easier and also we can easily fake a timestamp for the whole project

chris-cwa (changes requested):
I love it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
d4rken: While this is not always the case in current classes, new classes should put the companion object at the bottom. The goal is to follow the Kotlin styleguide: https://kotlinlang.org/docs/reference/coding-conventions.html

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
d4rken: Generally it's not a KO argument for me to explicitly log a tag, as there are some cases where not specifying an explicit tag will lead automatically generated tags that make filtering the log difficult (some coroutines and callbacks).

I've checked and it's not the case here and will remove the explicit TAG here, thanks!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
d4rken: Then we'd still need an `if` check to log a warning that the configuration was null, right? Not much gained in terms of readability.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeStamper.kt
d4rken: It's not "yet" consequently used when refactoring please make use of this class where possible. It allows us to inject mock time sources in testing and "timetravel" to test edge cases.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DiagnosisKeyProvider.kt
d4rken: Hmm, it's the interface that the class wrapping `GoogleENFClient.provideDiagnosisKeys` has to implement. So while it does both, it is more responsible for "providing keys into the ENF"  than consuming from our end. :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisKeyRetrievalOneTimeWorker.kt
d4rken: There are a few subtle differences between both approaches. In terms of readability, I think the kotlin approach is better, but for when somethingi is not logged in a `Timber.Tree` the arguments will not be evaluated. So in cases where we log stuff like `Timber.d("A list of objects: %s", listOfObjects)` the expensive operation of expanding `listOfObjects` into a string will only happen when necessary.

Though when we improve error logging, i.e. make it available to the user, there will not be many cases where this approach actually yields any performance improvements as it will be very likely logged (and thus expanded) by a planted timber-tree anyways.
Maybe we should discuss this in the techtalk.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeStamper.kt
ralfgehrer: IMHO this can be improved by switching to the ThreeTen time library entirely injecting a Clock that - which in turn can be moved in time - instead of using helper functions to solely determine the current time. Anyway, this is a bigger discussion for another time.

ralfgehrer (approved):

chris-cwa (approved):

PR #1206, opened 2020-09-22
Fix: Update data privacy (de & en) and user consent (de) (EXPOSUREAPP-2778)

harambasicluka:
This updates the data privacy in english and german, the user content is only updated in german as it's part of values-de/strings.xml, the provided has to be delivered via the translation tool, otherwise it will be overwritten every time we get a new delivery.


BMItter:
One point that I just saw by chance..
HTML Document is not valid. <!DOCTYPE html> missing in general.
Element head is also missing.
Should not have much impact as html parser are usually liberal nowadays but regardless of that both are real errors...

harambasicluka:
@BMItter I see that point, but as we are converting it to a spanned which is used for texts with markups that isn't a problem imho as we aren't rendering the HTML as it is.

BMItr (approved):

janetback (approved):
UA approved.

PR #1213, opened 2020-09-23
Handle App Config uniformly with caching (EXPOSUREAPP-2747)

d4rken:
## Description
Our appconfig handling should meet the desired specs analog to iOS.
Most of the resiliance features were already built in but we didn't fall back to the last valid config all cases, now we do.
+ Some refactoring and better test coverage.

## How to test
* Run the app (fresh install too), check log for "AppConfig"
* Use the app in adverse conditions (without internet etc.)
* Check code and unit tests, tech spec should be met (see JIRA ticket)
* The only case where you should see errors are on a fresh install without internet where no valid config has ever been loaded


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigProvider.kt
```diff
@@ -0,0 +1,106 @@
...
+
+    private val configApi: AppConfigApiV1
+        get() = appConfigAPI.get()
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal suspend fun downloadAppConfig(): ByteArray? {
+        Timber.tag(TAG).d("Fetching app config.")
+        var exportBinary: ByteArray? = null
+        var exportSignature: ByteArray? = null
+        configApi.getApplicationConfiguration(homeCountry.identifier).byteStream()
```
jakobmoellersap, line 32: do we need to call .use() on the byteStream here to close it properly?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/environment/download/DownloadCDNModule.kt
```diff
@@ -0,0 +1,50 @@
...
+    @Provides
+    fun cdnHttpClient(@HttpClientDefault defaultHttpClient: OkHttpClient): OkHttpClient =
+        defaultHttpClient.newBuilder().connectionSpecs(DOWNLOAD_CDN_CONNECTION_SPECS).build()
+
+    @Singleton
+    @DownloadCDNServerUrl
+    @Provides
+    fun provideDownloadServerUrl(): String {
+        val url = BuildConfig.DOWNLOAD_CDN_URL
+        if (!url.startsWith("https://")) throw IllegalStateException("Innvalid: $url")
```
jakobmoellersap, line 28: I think maybe we should either do a proper URL check or none at all, instead of a simple https check in the string

jakobmoellerdev (commented):
Small Questions, other than that lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/environment/download/DownloadCDNModule.kt
d4rken: What would we check though? Btw, the current check is the same we had before, I just moved the code a bit.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigProvider.kt
d4rken: Hm don't think so, the `unzip` should close the stream, it internally uses `use {}` on the bytestream.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/environment/download/DownloadCDNModule.kt
jakobmoellersap: Okay I think this is a super unimportant todo, for now should not matter.

jakobmoellerdev (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigStorage.kt
```diff
@@ -0,0 +1,39 @@
...
+        }
+        set(value) {
+            Timber.v("set(...) AppConfig: %dB", value?.size)
+
+            if (configDir.mkdirs()) Timber.v("Parent folder created.")
+
+            if (configFile.exists()) {
+                Timber.v("Overwriting %d from %s", configFile.length(), configFile.lastModified())
+            }
+            if (value != null) {
```
ralfgehrer, line 33: minor, alternativ:
value?.let {  configFile.writeBytes(value)} ?: configFile.delete()

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/AppConfigApiTest.kt
```diff
@@ -154,4 +154,9 @@ class AppConfigApiTest : BaseIOTest() {
             api.getApplicationConfiguration("DE").string() shouldBe "~appconfig"
         }
     }
+
+    @Test
+    fun `timeout and retry behavior`() {
```
ralfgehrer, line 159: Delete or implement plz

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigStorage.kt
d4rken: Will keep the current one due to better readability.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/AppConfigApiTest.kt
d4rken: :+1:

PR #1221, opened 2020-09-23
Fix various code smells and refactor our Application class

d4rken:
## Description 
Fix various code :nose: and refactor our Application class.

i.e. https://sonarcloud.io/project/issues?branch=fix%2Fflaky-downloader-test&id=corona-warn-app_cwa-app-android&resolved=false&types=CODE_SMELL

## How to test
* Review the changes 
* Launch the app


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/WatchdogService.kt
```diff
@@ -0,0 +1,76 @@
...
+import de.rki.coronawarnapp.worker.BackgroundWorkHelper
+import de.rki.coronawarnapp.worker.BackgroundWorkScheduler
+import kotlinx.coroutines.launch
+import timber.log.Timber
+import java.util.UUID
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class WatchdogService @Inject constructor(private val context: Context) {
```
jakobmoellersap, line 19: Are we sure this does not introduce additional problems based on lifecycle processing? Recommendation from Google is to leave the Logic directly in onCreate without Overhead.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CWADebug.kt
```diff
@@ -1,8 +1,24 @@
 package de.rki.coronawarnapp.util
 
+import android.app.Application
 import de.rki.coronawarnapp.BuildConfig
+import de.rki.coronawarnapp.util.debug.FileLogger
+import timber.log.Timber
 
 object CWADebug {
```
jakobmoellersap, line 8: Is this object just meant to be used in the Application Class or does it have other uses as well down the line?

d4rken: This is the only Singleton that we may keep, as debug related code is usually not injected. When the amount of "debug" init code grows smaller, the class should be split up though. For now I think this grouping is okay, what do you think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/WatchdogService.kt
d4rken: > additional problems based on lifecycle processing

Can you elaborate on that? Instantiation via Dagger does not switch threads and does not happen asynchronously.

What dagger does here would be like

```kotlin
    private lateinit var watchdogService: WatchdogService
    override fun onCreate() {
        instance = this
        super.onCreate()

        (...)
        watchdogService = Watchdog(this)

        (...)
        watchdogService.launch()
    }
```

The overhead via dagger will only cost us some nano seconds on the main thread, which I think is ok to pay to gain better abstraction. I don't think we are in an area of delay here that can negatively impact us.

jakobmoellersap: While I agree this SHOULD not impose side-effects, we still have some uncertainty when it comes to us being called by the watch-dog. Let's run this through our test-track and see if we have a behaviour change. We should definitely include some logging for the testers though to make sure the launch happens without issues.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CWADebug.kt
jakobmoellersap: Agreed, I think right now the amount of Debug-Code that would need to be in here is small in foreseeable future

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/WatchdogService.kt
d4rken: I've added additional log statements to make that visible.

mertsafter (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
```diff
@@ -23,187 +20,115 @@ import dagger.android.HasAndroidInjector
...
-                if (wakeLock.isHeld) wakeLock.release()
-            }
+        watchdogService.launch()
+    }
+
+    private val foregroundStateUpdater = object : LifecycleObserver {
+        @OnLifecycleEvent(Lifecycle.Event.ON_START)
+        fun onAppForegrounded() {
+            isAppInForeground = true
+            Timber.v("App is in the foreground")
```
ralfgehrer, line 74: Do we actually need these Timber logs? Same below.

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
d4rken: Was there before I refactored, but I also like it because it gives us more clues about what happened when the testers provide us with a log where there was an issue, we'll know whether the app was in the foreground or background etc. I'm usually a fan of extensive debug logging, huge gains when we encounter heisenbugs :grin:.

PR #1223, opened 2020-09-24
Interoperability: Submit days since onset of symptoms, Add 15th key to transmission risk vector (EXPOSUREAPP-2748, EXPOSUREAPP-2686)

chris-cwa:
This PR enables the app to submit the days since onset of symptoms to the backend and also enables it to handle both tek modes.

To test, please go to google settings and change value of 'heutigen TEK mid verkürztem Zeitraum zurückgeben'


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,54 @@
...
         keys: List<TemporaryExposureKey>,
-        transmissionRiskVector: TransmissionRiskVector
+        transmissionRiskVector: TransmissionRiskVector,
+        daysSinceOnsetOfSymptomsVector: DaysSinceOnsetOfSymptomsVector
     ) =
         keys.mapIndexed { index, key ->
             // The latest key we receive is from yesterday (i.e. 1 day ago),
             // thus we need use index+1
-            keyConverter.toExternalFormat(key, transmissionRiskVector.getRiskValue(index + 1))
+            val i = index + 1
```
jakobmoellersap, line 43: is todays key already needed?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/TransmissionRiskVectorDeterminator.kt
```diff
@@ -1,52 +1,77 @@
...
+        val SubmissionStatus.is15thKeyNeeded: Boolean
+            get() = !succeeded
+
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        fun SubmissionStatus.add15thKeyTo(currentVector: IntArray) =
+            intArrayOf(currentVector[0]) + transmissionRiskVector.raw
+
+        @Suppress("MagicNumber")
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        fun actuallyDetermine(symptoms: Symptoms) =
```
jakobmoellersap, line 26: I would argue that we call this function more explanatory. 

Maybe something like determineTransmissionRiskValuesFromSymptoms

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
jakobmoellersap: We need to be careful to not skip a value here in case we do a simple shift +1

jakobmoellersap:
We should rename this Ticket to be more self-speaking

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,70 @@
...
-            // The latest key we receive is from yesterday (i.e. 1 day ago),
-            // thus we need use index+1
-            keyConverter.toExternalFormat(key, transmissionRiskVector.getRiskValue(index + 1))
+        transmissionRiskVector: TransmissionRiskVector,
+        daysSinceOnsetOfSymptomsVector: DaysSinceOnsetOfSymptomsVector
+    ): List<KeyExportFormat.TemporaryExposureKey> {
+        val result = mutableListOf<KeyExportFormat.TemporaryExposureKey>()
+        keys.groupBy { it.daysAgo }.forEach { entry ->
+            val index = daysSinceOnsetOfSymptomsVector.indexOf(entry.key)
+            val today = entry.key == 0
```
mlenkeit, line 41: I find it quite confusing to understand what `entry.key` is referring to. At first, I was thinking it's the key of the TEK (which is actually stored in the `keyData` attribute) until I realized it's `daysAgo`.
For readability, I would suggest to add something like `val daysAgo = entry.key` as the first thing.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,70 @@
...
+    }
 
-    fun toSortedHistory(keys: List<TemporaryExposureKey>) =
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun toSortedHistory(keys: List<TemporaryExposureKey>) =
         keys.sortedWith(compareByDescending { it.rollingStartIntervalNumber })
 
-    companion object {
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    val SubmissionStatus.is15thKeyNeeded: Boolean
```
mlenkeit, line 64: Can you elaborate on the purpose of this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DefaultKeyConverter.kt
```diff
@@ -6,16 +6,20 @@ import com.google.protobuf.ByteString
...
+
+    override fun toExternalFormat(
+        key: TemporaryExposureKey,
+        riskValue: Int,
+        daysSinceOnsetOfSymptoms: Int
+    ) =
         KeyExportFormat.TemporaryExposureKey.newBuilder()
             .setKeyData(ByteString.readFrom(key.keyData.inputStream()))
             .setRollingStartIntervalNumber(key.rollingStartIntervalNumber)
             .setRollingPeriod(ROLLING_PERIOD)
```
mlenkeit, line 21: Why are we explicitly setting the rolling period here?
@jakobmoellersap do you know? After all, it was changed as part of the PR but I'm wondering whether it should be...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,70 @@
...
-        keys.mapIndexed { index, key ->
-            // The latest key we receive is from yesterday (i.e. 1 day ago),
-            // thus we need use index+1
-            keyConverter.toExternalFormat(key, transmissionRiskVector.getRiskValue(index + 1))
+        transmissionRiskVector: TransmissionRiskVector,
+        daysSinceOnsetOfSymptomsVector: DaysSinceOnsetOfSymptomsVector
+    ): List<KeyExportFormat.TemporaryExposureKey> {
+        val result = mutableListOf<KeyExportFormat.TemporaryExposureKey>()
+        keys.groupBy { it.daysAgo }.forEach { entry ->
+            val index = daysSinceOnsetOfSymptomsVector.indexOf(entry.key)
```
mlenkeit, line 40: Not sure if this is correct. If `entry.key` is semantically `daysAgo` (i.e. the age of the TEK in days), might work partially if the vector is [0;14] but even with a date, we can also have [-14;0] and with a range onset of symptoms, the values in the vector are in the 700s, 2000s, 3000s, or 4000s.

mlenkeit (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DefaultKeyConverter.kt
jakobmoellersap: Afaik, this was previously set to 144 by default. WIth 15th Key this should not be defaulted but taken from the key info.

chris-cwa: thank you

mlenkeit: @jakobmoellersap thanks, that's also what I was thinking...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DefaultKeyConverter.kt
```diff
@@ -5,17 +5,16 @@ import com.google.android.gms.nearby.exposurenotification.TemporaryExposureKey
...
+    override fun toExternalFormat(
+        key: TemporaryExposureKey,
+        riskValue: Int,
+        daysSinceOnsetOfSymptoms: Int
+    ) =
         KeyExportFormat.TemporaryExposureKey.newBuilder()
             .setKeyData(ByteString.readFrom(key.keyData.inputStream()))
             .setRollingStartIntervalNumber(key.rollingStartIntervalNumber)
-            .setRollingPeriod(ROLLING_PERIOD)
+            .setRollingPeriod(key.rollingPeriod)
```
mlenkeit, line 16: 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
+        daysAgo(
+            Instant.ofEpochMilli(rollingStartIntervalNumber * TEN_MINUTES_IN_MILLIS),
+            now
+        )
 
-        private const val MAXIMUM_KEYS = 14
-    }
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun daysAgo(instant: Instant, now: Instant = Instant()) =
+        Days.daysBetween(instant, now).days
```
mlenkeit, line 71: Clarification: how does is this affected by time zone settings and when they change?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
+            limitKeyCount(toSortedHistory(keys)),
+            transmissionRiskVectorDeterminator.determine(symptoms),
+            daysSinceOnsetOfSymptomsVectorDeterminator.determine(symptoms, VECTOR_LENGTH)
         )
 
-    fun <T> limitKeyCount(keys: List<T>): List<T> =
-        keys.take(MAXIMUM_KEYS)
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun limitKeyCount(keys: List<TemporaryExposureKey>, now: Instant = Instant()) =
+        keys.filter { it.daysAgo(now) <= MAX_DAYS_IN_PAST }
```
mlenkeit, line 33: Proposal: might be worth considering renaming `daysAgo` to `ageInDays`. I think it describes the attribute more precisely and implies that it has to be positive. It also makes it a bit clearer that we're dealing with a number. Accordingly, `MAX_DAYS_IN_PAST` should then probably be changed to `MAX_AGE_IN_DAYS`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
 
 class ExposureKeyHistoryCalculations(
     private val transmissionRiskVectorDeterminator: TransmissionRiskVectorDeterminator,
+    private val daysSinceOnsetOfSymptomsVectorDeterminator: DaysSinceOnsetOfSymptomsVectorDeterminator,
     private val keyConverter: KeyConverter
 ) {
 
+    companion object {
+        const val MAX_DAYS_IN_PAST = 14
+        const val VECTOR_LENGTH = 15
```
mlenkeit, line 17: Ask: I find this variable/constant name name too unspecific. Of what vector is this the length? It might be worth moving this into the DaysSinceOnsetOfSymptomsVectorDeterminator directly; such a parameter does not exist for the TransmissionRiskVectorDeterminator.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
-            transmissionRiskVectorDeterminator.determine(symptoms)
+            limitKeyCount(toSortedHistory(keys)),
+            transmissionRiskVectorDeterminator.determine(symptoms),
+            daysSinceOnsetOfSymptomsVectorDeterminator.determine(symptoms, VECTOR_LENGTH)
         )
 
-    fun <T> limitKeyCount(keys: List<T>): List<T> =
-        keys.take(MAXIMUM_KEYS)
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun limitKeyCount(keys: List<TemporaryExposureKey>, now: Instant = Instant()) =
```
mlenkeit, line 32: Proposal: I would suggest to rename it because its purpose has changed. It doesn't limit the key count (i..e number of keys) anymore. How about re-naming it to e.g. `removeOldKeys`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
-        keys.mapIndexed { index, key ->
-            // The latest key we receive is from yesterday (i.e. 1 day ago),
-            // thus we need use index+1
-            keyConverter.toExternalFormat(key, transmissionRiskVector.getRiskValue(index + 1))
+        transmissionRiskVector: TransmissionRiskVector,
+        daysSinceOnsetOfSymptomsVector: DaysSinceOnsetOfSymptomsVector,
+        now: Instant = Instant()
+    ): List<KeyExportFormat.TemporaryExposureKey> {
+        val result = mutableListOf<KeyExportFormat.TemporaryExposureKey>()
+        keys.groupBy { it.daysAgo(now) }.forEach { entry ->
```
mlenkeit, line 43: Remark: As mentioned elsewhere, I think here's another example where changing from `daysAgo` to `ageInDays` would help readability (i.e. _group by age in days_).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
mlenkeit, line 33: Ask: Please change the filter to `0 <= x <= MAX`, as `daysAgo`/`ageInDays` should never be negative but if it is for some reason (e.g. user manipulated the time on the device), we can eliminate potential errors. That's also how it's implemented on iOS.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
+                    )
+                )
+            }
         }
+        return result.toList()
+    }
 
-    fun toSortedHistory(keys: List<TemporaryExposureKey>) =
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun toSortedHistory(keys: List<TemporaryExposureKey>) =
```
mlenkeit, line 59: Proposal: I would prefer if we added a bit more semantics here to make it easier to understand what's the result. I think the set is sorted with the the most recent key(s) first, right? If that's the case, maybe rename it to `sortWithRecentKeyFirst`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,72 @@
...
+    ): List<KeyExportFormat.TemporaryExposureKey> {
+        val result = mutableListOf<KeyExportFormat.TemporaryExposureKey>()
+        keys.groupBy { it.daysAgo(now) }.forEach { entry ->
+            val daysAgo = entry.key
+            entry.value.forEach {
+                result.add(
+                    keyConverter.toExternalFormat(
+                        it,
+                        transmissionRiskVector.getRiskValue(daysAgo),
+                        daysSinceOnsetOfSymptomsVector[daysAgo]
```
mlenkeit, line 50: Remark: I find it confusing that a similar type of object (i.e. a vector) is accessed differently in the same context. Is there any chance to align this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/TransmissionRiskVectorDeterminator.kt
```diff
@@ -1,12 +1,19 @@
...
 import org.joda.time.Duration
 import org.joda.time.Instant
 
 class TransmissionRiskVectorDeterminator {
 
+    companion object {
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        fun numberOfDays(t0: Long, t1: Long = System.currentTimeMillis()) =
+            Duration(Instant.ofEpochMilli(t0), Instant.ofEpochMilli(t1)).standardDays.toInt()
+    }
```
mlenkeit, line 13: Clarification: how is this affected by time zone changes?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
```diff
@@ -0,0 +1,40 @@
+package de.rki.coronawarnapp.submission
+
+import de.rki.coronawarnapp.submission.TransmissionRiskVectorDeterminator.Companion.numberOfDays
+
+class DaysSinceOnsetOfSymptomsVectorDeterminator {
+
+    @Suppress("MagicNumber")
+    internal fun determine(symptoms: Symptoms, size: Int): DaysSinceOnsetOfSymptomsVector {
```
mlenkeit, line 8: Proposal: I would consider dropping the `size` parameter. I don't think we have any scenario where this is _not_ 15.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminatorTest.kt
```diff
@@ -0,0 +1,151 @@
+package de.rki.coronawarnapp.submission
+
+import org.junit.Assert
+import org.junit.Test
+
+class DaysSinceOnsetOfSymptomsVectorDeterminatorTest {
```
mlenkeit, line 6: 👍

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculationsTest.kt
```diff
@@ -16,53 +18,37 @@ class ExposureKeyHistoryCalculationsTest {
...
-                    "4",
-                    "5",
-                    "6",
-                    "7"
-                )
-            ).size
-        )
-        Assert.assertEquals(
-            14, instance.limitKeyCount(
+        val tek1 = createKey(2012, 10, 15)
```
mlenkeit, line 40: Proposal: I find it tricky to wrap your head around the different dates (not only in this test case but in all of them).

How about defining a `now` as the first variable in a test case (or maybe even in the setup) and trying to derive other dates in a declarative way, e.g. `now.minus(15, DAYS)`? I think this will massively help maintenance of the test cases.

Might require a change to `createKey` to take an Instant object or something like that.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculationsTest.kt
```diff
@@ -83,18 +69,51 @@ class ExposureKeyHistoryCalculationsTest {
 
     @Test
     fun test_toExternalFormat() {
```
mlenkeit, line 71: Ask: I think this needs more test cases. E.g. a test case with a TEK missing in the sequence, test cases with multiple TEKs on the same day, etc. Might be worth taking a look at the iOS test cases, e.g. starting https://github.com/corona-warn-app/cwa-app-ios/blob/d539979edea684be45dda527d52b9e846d618671/src/xcode/ENA/ENA/Source/Extensions/__tests__/ENTemporaryExposureKey%2BProcessingTests.swift#L211

mlenkeit (changes requested):
I added a couple of remarks in various spots. Note that this is one of the most crucial aspects of the solution and at the same time, very difficult to test in manual tests or end-to-end tests, so I'm a bit picky, e.g. on the automated tests 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
chris-cwa: @ralfgehrer

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/TransmissionRiskVectorDeterminatorTest.kt
```diff
@@ -16,13 +16,4 @@ class TransmissionRiskVectorDeterminatorTest {
             ).raw
         )
     }
-
```
mlenkeit, line 19: Ask: while I see why `test_numberOfDays` was removed, this class definitely needs more test cases, e.g. these ones here: https://github.com/corona-warn-app/cwa-app-ios/blob/8bb7273853b598a507c89acd45dad0e4157e73e7/src/xcode/ENA/ENA/Source/Models/__tests__/SymptomsOnsetTests.swift#L29-L195

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculationsTest.kt
```diff
@@ -10,66 +11,54 @@ class ExposureKeyHistoryCalculationsTest {
...
         }
-        instance = ExposureKeyHistoryCalculations(TransmissionRiskVectorDeterminator(), converter)
+
+        instance = ExposureKeyHistoryCalculations(
+            TransmissionRiskVectorDeterminator(),
+            DaysSinceOnsetOfSymptomsVectorDeterminator(),
+            converter
+        )
+
+        now = DateTime(2012, 10, 15, 0, 0)
```
mlenkeit, line 37: Ask: I would suggest to add additional test cases for `toExternalFormat` where `now` is not pointing to midnight (local time?) but different points in time (in UTC)

mlenkeit (changes requested):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculationsTest.kt
chris-cwa: this actually detected a bug! thank you!!!

ralfgehrer (dismissed):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculationsTest.kt
```diff
@@ -83,18 +77,122 @@ class ExposureKeyHistoryCalculationsTest {
...
+            result.map { it.rollingStartIntervalNumber }.toTypedArray().toIntArray()
+        )
+        Assert.assertArrayEquals(
+            intArrayOf(3998, 4005),
+            result.map { it.daysSinceOnsetOfSymptoms }.toTypedArray().toIntArray()
+        )
+        Assert.assertArrayEquals(
+            intArrayOf(0, 7),
+            result.map { it.transmissionRiskLevel }.toTypedArray().toIntArray()
+        )
```
mlenkeit, line 122: Recommendation: might be worth refactoring into three local helper functions for assertions; each function accepting `result` and an array with either the expected `rollingStartIntervalNumber`, the DSOS, or the TRL.

BMItr (dismissed):
I don't know all the vectors/intArrays by content, but code looks clean to me.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
```diff
@@ -75,4 +76,12 @@ object TimeAndDateExtensions {
...
         return TimeUnit.MILLISECONDS.toDays(millionSeconds)
     }
+
+    fun numberOfDayChanges(then: DateTime, now: DateTime = Instant().toDateTime(DateTimeZone.UTC)) =
+        Days.daysBetween(
+            then.withTime(0, 0, 0, 0),
+            now.withTime(0, 0, 0, 0)).days
+
+    fun numberOfDayChanges(millis: Long, now: DateTime = Instant().toDateTime(DateTimeZone.UTC)) =
+        numberOfDayChanges(Instant.ofEpochMilli(millis).toDateTime(DateTimeZone.UTC), now)
```
mlenkeit, line 86: Required: I find numberOfDayChanges confusing. To me, it's not clear whether changes is used as plural of change (I think it is, but I don't understand what the function is supposed to do then by reading the name) or as third-person singular of the verb (then I would expect it to return a boolean, but it doesn't).

What was wrong about ageInDays? If you find that confusing, then how about daysBetween (like the API it is wrapping) or daysBetweenIgnoreTime (inspired by Java's equalsIgnoreCase)?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
```diff
@@ -75,4 +76,12 @@ object TimeAndDateExtensions {
         val millionSeconds = secondDate - firstDate
         return TimeUnit.MILLISECONDS.toDays(millionSeconds)
     }
+
+    fun numberOfDayChanges(then: DateTime, now: DateTime = Instant().toDateTime(DateTimeZone.UTC)) =
+        Days.daysBetween(
+            then.withTime(0, 0, 0, 0),
+            now.withTime(0, 0, 0, 0)).days
```
mlenkeit, line 83: Question: do `now` and `then` need to be converted to UTC?

mlenkeit (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
ralfgehrer: I like the later.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    companion object {
+        const val VECTOR_LENGTH = 15
+    }
+
+    @Suppress("MagicNumber")
+    internal fun determine(symptoms: Symptoms): DaysSinceOnsetOfSymptomsVector {
+        return when (symptoms.symptomIndication) {
+            Symptoms.Indication.POSITIVE -> {
+                when (symptoms.startOfSymptoms) {
```
BMItter, line 15: I would put that second when Block in a seperate method for better readability. But it's just a personal preference...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,73 @@
...
+    internal fun toExternalFormat(
+        keys: List<TemporaryExposureKey>,
+        transmissionRiskVector: TransmissionRiskVector,
+        daysSinceOnsetOfSymptomsVector: DaysSinceOnsetOfSymptomsVector,
+        now: DateTime = Instant().toDateTime(DateTimeZone.UTC)
+    ): List<KeyExportFormat.TemporaryExposureKey> {
+        val result = mutableListOf<KeyExportFormat.TemporaryExposureKey>()
+        keys.groupBy { it.ageInDays(now) }.forEach { entry ->
+            val ageInDays = entry.key
+            entry.value.forEach {
```
BMItter, line 49: same her. for better readability it might be better to seperate the second forEach Block to an own method. but also just personal preference.

mlenkeit: I don't have any preference about this but if you decide to do this, I wouldn't see a need to change any of the related tests. It's just an internal refactoring of that function for improved readability.

.idea/codeStyles/Project.xml
```diff
@@ -3,7 +3,16 @@
...
+          <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
+        </value>
+      </option>
+      <option name="PACKAGES_IMPORT_LAYOUT">
+        <value>
+          <package name="" alias="false" withSubpackages="true" />
+          <package name="java" alias="false" withSubpackages="true" />
+          <package name="javax" alias="false" withSubpackages="true" />
+          <package name="kotlin" alias="false" withSubpackages="true" />
+          <package name="" alias="true" withSubpackages="true" />
```
d4rken, line 15: Bug in Android Studio, don't commit this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
```diff
@@ -0,0 +1,43 @@
+package de.rki.coronawarnapp.submission
+
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.numberOfDayChanges
+
+class DaysSinceOnsetOfSymptomsVectorDeterminator {
+
+    companion object {
```
d4rken, line 7: At the end of the class please.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
```diff
@@ -0,0 +1,43 @@
...
+        const val VECTOR_LENGTH = 15
+    }
+
+    @Suppress("MagicNumber")
+    internal fun determine(symptoms: Symptoms): DaysSinceOnsetOfSymptomsVector {
+        return when (symptoms.symptomIndication) {
```
d4rken, line 13: I think the return can be inlined, improving readability a bit.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,73 @@
...
     private val transmissionRiskVectorDeterminator: TransmissionRiskVectorDeterminator,
+    private val daysSinceOnsetOfSymptomsVectorDeterminator: DaysSinceOnsetOfSymptomsVectorDeterminator,
     private val keyConverter: KeyConverter
 ) {
 
+    companion object {
+        const val MAX_AGE_IN_DAYS = 14
+        const val TEN_MINUTES_IN_MILLIS = (10 * 60 * 1000).toLong()
+    }
+
```
d4rken, line 21: End of class please.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,73 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun sortWithRecentKeyFirst(keys: List<TemporaryExposureKey>) =
         keys.sortedWith(compareByDescending { it.rollingStartIntervalNumber })
 
-    companion object {
-
-        private const val MAXIMUM_KEYS = 14
-    }
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun TemporaryExposureKey.ageInDays(now: DateTime = Instant().toDateTime(DateTimeZone.UTC)): Int =
```
d4rken, line 67: Instant has no time zone. Instant.now() is UTC. As long as the Instant is not from some weird source, we could go without the extra conversion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ExposureKeyHistoryCalculations.kt
```diff
@@ -1,39 +1,73 @@
...
+        TimeAndDateExtensions.numberOfDayChanges(
+            Instant.ofEpochMilli(rollingStartIntervalNumber * TEN_MINUTES_IN_MILLIS).toDateTime(
```
d4rken, line 69: Maybe

```kotlin
Instant.ofEpochMilli(rollingStartIntervalNumber).plus(Duration.standardMinutes(10))
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/TransmissionRiskVector.kt
```diff
@@ -2,14 +2,13 @@ package de.rki.coronawarnapp.submission
 
 class TransmissionRiskVector(private val values: IntArray) {
 
+    companion object {
+        private const val DEFAULT_TRANSMISSION_RISK_LEVEL = 1
+    }
+
```
d4rken, line 8: End of class please.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
```diff
@@ -75,4 +76,12 @@ object TimeAndDateExtensions {
         val millionSeconds = secondDate - firstDate
         return TimeUnit.MILLISECONDS.toDays(millionSeconds)
     }
+
+    fun numberOfDayChanges(then: DateTime, now: DateTime = Instant().toDateTime(DateTimeZone.UTC)) =
+        Days.daysBetween(
+            then.withTime(0, 0, 0, 0),
```
d4rken, line 82: Could we use `.withTimeAtStartOfDay()` here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
```diff
@@ -75,4 +76,12 @@ object TimeAndDateExtensions {
         val millionSeconds = secondDate - firstDate
         return TimeUnit.MILLISECONDS.toDays(millionSeconds)
     }
+
+    fun numberOfDayChanges(then: DateTime, now: DateTime = Instant().toDateTime(DateTimeZone.UTC)) =
```
d4rken, line 80: When is the default argument for `now` used?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminatorTest.kt
```diff
@@ -0,0 +1,141 @@
...
+
+    @Test
```
d4rken, line 8: Please use jUnit5 tests.

```kotlin
import org.junit.jupiter.api.AfterEach
import org.junit.jupiter.api.BeforeEach
import org.junit.jupiter.api.Test
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminatorTest.kt
```diff
@@ -0,0 +1,141 @@
...
+    fun positive_exactDate_Yesterday() {
```
d4rken, line 9: You can use descriptive sentences in Kotlin function names for tests.

```kotlin
fun `match a positive exact date to yesterday`()
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminatorTest.kt
```diff
@@ -0,0 +1,141 @@
...
+        val daysAgo = 1
+        Assert.assertArrayEquals(
```
d4rken, line 11: Please use kotest.

```kotlin
        DaysSinceOnsetOfSymptomsVectorDeterminator().determine(
            Symptoms(
                Symptoms.StartOf.Date(System.currentTimeMillis() - 1000 * 3600 * (24 * daysAgo + 2)),
                Symptoms.Indication.POSITIVE
            )
        ) shouldBe intArrayOf(1, 0, -1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, -13)
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminatorTest.kt
```diff
@@ -0,0 +1,141 @@
...
+            intArrayOf(1, 0, -1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, -13),
+            DaysSinceOnsetOfSymptomsVectorDeterminator().determine(
+                Symptoms(
+                    Symptoms.StartOf.Date(System.currentTimeMillis() - 1000 * 3600 * (24 * daysAgo + 2)),
```
d4rken, line 15: Ideally we would not use variable time value (`currentTimeMillis`) in tests, but pass a fixed/mocked time sources.

d4rken (commented):
Generally looks good.
* The new test classes that have been created should use  jUnit5 and kotest.
* I think we don't need a lot of the DateTime conversion as Instant is already UTC. I would prefer that we make more use of the standard time classes, also for durations i.e. `Duration.standardMinutes(10)` as that then integrates well with the standard classes.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminatorTest.kt
```diff
@@ -1,141 +1,111 @@
...
-        Assert.assertArrayEquals(
-            intArrayOf(1, 0, -1, -2, -3, -4, -5, -6, -7, -8, -9, -10, -11, -12, -13),
-            DaysSinceOnsetOfSymptomsVectorDeterminator().determine(
-                Symptoms(
-                    Symptoms.StartOf.Date(System.currentTimeMillis() - 1000 * 3600 * (24 * daysAgo + 2)),
-                    Symptoms.Indication.POSITIVE
-                )
+        DaysSinceOnsetOfSymptomsVectorDeterminator().determine(
+            Symptoms(
+                Symptoms.StartOf.Date(System.currentTimeMillis() - 1000 * 3600 * (24 * daysAgo + 2)),
```
mlenkeit, line 13: Magic number: why is there a `+ 2`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
```diff
@@ -0,0 +1,54 @@
...
+    private val timeStamper: TimeStamper
+) {
+
+    @Suppress("MagicNumber")
+    internal fun determine(symptoms: Symptoms): DaysSinceOnsetOfSymptomsVector {
+        return when (symptoms.symptomIndication) {
+            Symptoms.Indication.POSITIVE ->
+                determinePositiveIndication(symptoms)
+            Symptoms.Indication.NO_INFORMATION ->
+                createDaysSinceOnsetOfSymptomsVectorWith(4000)
```
MertSafter, line 20: Would an enum or companion object be over-engineered. Currently thinking about all the "Magic numbers" to move to an enum or companion object. @chris-cwa, @d4rken, @ralfgehrer would you guys think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
```diff
@@ -0,0 +1,54 @@
...
+
+    @Suppress("MagicNumber")
+    private fun determinePositiveIndication(symptoms: Symptoms): DaysSinceOnsetOfSymptomsVector {
+        return when (symptoms.startOfSymptoms) {
+            is Symptoms.StartOf.Date ->
+                createDaysSinceOnsetOfSymptomsVectorWith(
+                    symptoms.startOfSymptoms.date.ageInDays(timeStamper.nowUTC.toLocalDate())
+                )
+            is Symptoms.StartOf.LastSevenDays ->
+                createDaysSinceOnsetOfSymptomsVectorWith(701)
```
MertSafter, line 34: Same as above

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/DaysSinceOnsetOfSymptomsVectorDeterminator.kt
ralfgehrer: IMHO that's over-engineered. They are rather magical ;)

mlenkeit: Not sure an enum is necessary (leave it up to you) but if not, a class-level (Javadoc) comment might be appropriate here to give a rough explanation what these magic numbers mean and where they come from. Let me know if you need any support with that.

ralfgehrer: I will add a link to your documentation, @mlenkeit

mlenkeit (approved):
LGTM!
@d4rken could you take another look at the usage of date-related objects such as LocalDate, Instant, DateTime, etc.?

d4rken (approved):

BMItr (approved):

PR #1227, opened 2020-09-24
Make transaction timeout settable per transaction. Increase Risk Leve…

mertsafter:
## Description
* Make transaction timeout settable for each transaction. 
* Increase transaction timeout for Risk Level Calculation needed for interoperability

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RiskLevelTransaction.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RiskLevelTransaction.kt
@@ -135,6 +135,12 @@ object RiskLevelTransaction : Transaction() {
     // Injecting here will break Test
     private val riskLevelCalculation: RiskLevelCalculation = DefaultRiskLevelCalculation()
 
+    /**
+     * The maximal runtime of the Risk Level transaction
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/RiskLevelTransaction.kt
```diff
@@ -135,6 +135,12 @@ object RiskLevelTransaction : Transaction() {
     // Injecting here will break Test
     private val riskLevelCalculation: RiskLevelCalculation = DefaultRiskLevelCalculation()
 
+    /**
+     * The maximal runtime of the Risk Level transaction
+     * In milliseconds
+     */
+    private const val RISK_LEVEL_TRANSACTION_TIMEOUT = 420000L
```
d4rken, line 142: 420s,7min, doesn't the Ticket say 8min,480s?

If you change it, make it `480 * 1000L` a bit easier to read :)

MertSafter: 😄  Done.

d4rken (approved):

ralfgehrer (approved):

PR #1231, opened 2020-09-25
Remove RKI-Hotline from terms of use (tr, en) (EXPOSUREAPP-2633)

mertsafter:
## Description
Removed RKI-Hotline number from terms of use for turkish and english translation.


BMItr (approved):

janetback (approved):
UA approved

PR #1237, opened 2020-09-25
Submission logic for interop (EXPOSUREAPP-2737)

d4rken:
## Description
This adds submission logic for interop, adding `consentToFederation` and `visistedCountries` from the app config.

To successfully test our logic, additional refactoring was necessary to further rid ourselves of the singletons. This includes the verification, submission and playbook logic which are now separated and injected.

## How to test
* Perform the submission flow, and observe logcat.
* Review refactored tests and code.

## Draft TODO

- [x] Wait for #1213 to be merged to clean country access up here (currently hardcoded)
- [x] Wait for the PR for EXPOSUREAPP-2727 (#1246) to be created and merged due to overlap that could cause conflicts


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/SubmitDiagnosisKeysTransaction.kt
```diff
@@ -54,39 +54,42 @@ object SubmitDiagnosisKeysTransaction : Transaction() {
...
+
         executeState(RETRIEVE_TAN_AND_SUBMIT_KEYS) {
-            PlaybookImpl(WebRequestBuilder.getInstance()).submission(
-                registrationToken,
-                temporaryExposureKeyList
+            val submissionData = Playbook.SubmissionData(
+                registrationToken = registrationToken,
+                temporaryExposureKeys = temporaryExposureKeyList,
+                consentToFederation = true,
+                visistedCountries = visistedCountries
```
mlenkeit, line 84: Question: Is the fallback to `DE` in case `visistedCountries` is empty built into the App Config?

.idea/codeStyles/Project.xml
```diff
@@ -3,7 +3,16 @@
     <JetCodeStyleSettings>
       <option name="PACKAGES_TO_USE_STAR_IMPORTS">
         <value>
-          <package name="kotlinx.android.synthetic" withSubpackages="true" static="false" />
+          <package name="kotlinx.android.synthetic" alias="false" withSubpackages="true" />
```
d4rken, line 6: Will fix this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/transaction/SubmitDiagnosisKeysTransaction.kt
d4rken: It is not. Good point. Will fix.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/server/SubmissionServer.kt
```diff
@@ -0,0 +1,85 @@
...
+
+        api.submitKeys(
+            authCode = authCode,
+            fake = "0",
+            headerPadding = EMPTY_HEADER,
+            requestBody = submissionPayload
+        )
+    }
+
+    suspend fun submitKeysToServerFake() = withContext(Dispatchers.IO) {
```
chris-cwa, line 59: why this method and not inject fake SubmissionServer when needed?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/PaddingTool.kt
```diff
@@ -0,0 +1,10 @@
+package de.rki.coronawarnapp.util
+
+object PaddingTool {
+    fun requestPadding(length: Int): String {
```
chris-cwa, line 4: did I miss the unit test?

Corona-Warn-App/src/test/java/testhelpers/extensions/JsonExtensions.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import com.google.gson.Gson
+import com.google.gson.JsonObject
+import okhttp3.mockwebserver.MockResponse
+
+fun String.toComparableJson() = try {
+    Gson().fromJson(this, JsonObject::class.java).toString()
+} catch (e: Exception) {
+    throw IllegalArgumentException("'$this' wasn't valid JSON")
+}
```
chris-cwa, line 11: how about
```suggestion
fun String.getValidatedJson(): String? = try {
    Gson().fromJson(this, JsonObject::class.java).toString()
} catch (e: Exception) {
    // log maybe
    null
}
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/AppConfigProviderTest.kt
```diff
@@ -185,6 +185,16 @@ class AppConfigServerTest : BaseIOTest() {
...
+    @Test
+    fun `if supportedCountryList is empty, we insert DE as fallback`() {
+        coEvery { api.getApplicationConfiguration("DE") } returns APPCONFIG_BUNDLE.toResponseBody()
+        every { verificationKeys.hasInvalidSignature(any(), any()) } returns false
+
+        runBlocking {
+            createDownloadServer().getAppConfig().supportedCountriesList shouldBe listOf("DE")
+        }
+    }
+
```
mlenkeit, line 197: @d4rken this fallback also applies then when the list of countries is displayed on the UI, isn't it? I think for the UI, there's a different kind of fallback. The `DE` fallback is only for the submission and the key download (i.e. non-UI coding).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/server/SubmissionServer.kt
d4rken: There is lot's of overlap between both methods, and they are both used in the in the playbook, we'd have to inject the real and the fake server, the fake one just uses different parameters.

I'm not opposed to injecting this somehow, just didn't see a clear way to do it when refactoring this. Would leave that for another PR, this one is already pretty huge.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/PaddingTool.kt
d4rken: You did not, this needs one definitely one. Good point, will add :+1:

Corona-Warn-App/src/test/java/testhelpers/extensions/JsonExtensions.kt
d4rken: This is an extension only used in tests, I'd prefer it to fail loudly. Will keep unless you have strong feelings about this.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/AppConfigProviderTest.kt
d4rken: I'll correct this. You are right, my approach would interfere with that behavior.

PR #1254, opened 2020-09-29
Texts for error recovery notice (EXPOSUREAPP-1851)

d4rken:
We need to show a notice if we perform error recovery for existing users affected by #642 .
This only adds the texts so we can start on translations, they will be required for a different PR with the actual logic.

### DIFF
--- a/Corona-Warn-App/src/main/res/values-de/strings.xml
+++ b/Corona-Warn-App/src/main/res/values-de/strings.xml
@@ -1164,6 +1164,10 @@
     <string name="errors_generic_button_negative">"Details"</string>
     <!-- XTXT: error dialog - text when no error description is available  -->
     <string name="errors_generic_text_unknown_error_cause">"Ein unbekannter Fehler ist aufgetreten."</string>
+    <!-- XTXT: error dialog - text when a catastrophic error occured from which the app recovered automatically via data reset -->
+    <string name="errors_generic_text_catastrophic_error_recovery_via_reset">"Aufgrund eines technischen Problems wurde Ihre App zurückgesetzt. Das hat jedoch keine Auswirkungen auf die Verwendung der App. Sie werden weiterhin über Risiko-Begegnungen benachrichtigt und können andere warnen, falls Sie positiv auf COVID-19 getestet wurden."</string>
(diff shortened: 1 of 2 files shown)


janetback (approved):
UA Reviewed

harambasicluka (approved):

vaubaehn:
Hi @d4rken and @harambasicluka ,

while I really like the automatic recovery, I see an issue with the current version of the text presented to users in the dialog box:
For users who had previously registered a Covid test with the app, there is the inconvenient consequence that their token for retrieval of the test result is lost after crash of EPS/automatic recovery. So they need to wait for a notification by doctor or health authorities by phone/mail, or if in doubt, need to proactively call related institutions to get their result in timely manner. For some users, getting their test result online is the only reason to install CWA (according to some Google Play Store reviews).
If these users read current information after automatic recovery that app is still fully functional, they may feel fooled in best case (similar to people with API39508, who need to read that app is functional but infection risk cannot be determined...)
So, I would suggest to change the text in that way, that exposure logging and warning is not affected by app reset, but a potentially registered Covid test would be lost. To have a more detailed explanation of what happened to app data and what to do in case of loss of Covid test, a new FAQ section could be created, and the intended link out from dialog box (details) could link to the new section instead of linking to cause9002, as this is not really helpful in that situation.
Cheers, V.

d4rken:
Valid concerns @vaubaehn , I've forwarded this internally.

svengabr:
@vaubaehn the new Popup text will be introduced in 1.5 can be found here: https://github.com/corona-warn-app/cwa-website/issues/429

We still need to write a good entry for the FAQ website.

vaubaehn:
Hi @svengabr ,
Thanks, I found it.
If you like, I could write a proposal for FAQ entry as a starting point. Just ping me, if you welcome support here.

svengabr:
@vaubaehn I would really appreciate it! You could write a text and add it as a comment here: https://github.com/corona-warn-app/cwa-website/issues/429

vaubaehn:
@svengabr 
Cool, I'll try till tomorrow lunch time.

PR #1235, opened 2020-09-25
Wrap EncryptedSharedPreferences initialization in a retry mechanism (EXPOSUREAPP-1851)

d4rken:
_New PR to merge into 1.4.x instead of 1.5.x_

## Description
Wrapped the `EncryptedSharedPreferences` initialization in an exponential backoff mechanism with randomized delay, to mitigate issues caused by the system's keystore being busy. The idea is that after a few retries the congestion has cleared and we can get access to the system's keystore.

We'll retry for a maximum of 15s, delay at least 25ms and at most 3s between tries.

Should mitigate #642, possibly also  #1214, #1121.

Bug description:
* https://github.com/corona-warn-app/cwa-app-android/issues/642#issuecomment-697603175
* https://issuetracker.google.com/issues/158234058
* https://github.com/google/tink/issues/321

## How to test
* Difficult to reproduce :frowning_face: 
* Understand the bug as described above (see links), review the code.
* Add a test exception in the preference creation code, observe behavior.

<details>
<summary>Example log output</summary>

```kotlin
2020-09-23 00:34:12.839 4424-4424/de.rki.coronawarnapp.dev V/RetryMechanism: Executing attempt: Attempt(count=1, totalDelay=0, lastDelay=0, exception=null)
2020-09-23 00:34:12.839 4424-4424/de.rki.coronawarnapp.dev D/EncryptedPreferencesFactory$create: Creating EncryptedSharedPreferences instance.
2020-09-23 00:34:13.138 4424-4424/de.rki.coronawarnapp.dev V/RetryMechanism: Executing attempt: Attempt(count=2, totalDelay=25, lastDelay=25, exception=java.security.KeyException)
2020-09-23 00:34:13.139 4424-4424/de.rki.coronawarnapp.dev D/EncryptedPreferencesFactory$create: Creating EncryptedSharedPreferences instance.
2020-09-23 00:34:13.255 4424-4424/de.rki.coronawarnapp.dev V/RetryMechanism: Executing attempt: Attempt(count=3, totalDelay=50, lastDelay=25, exception=java.security.KeyException)
2020-09-23 00:34:13.255 4424-4424/de.rki.coronawarnapp.dev D/EncryptedPreferencesFactory$create: Creating EncryptedSharedPreferences instance.
2020-09-23 00:34:13.368 4424-4424/de.rki.coronawarnapp.dev V/RetryMechanism: Executing attempt: Attempt(count=4, totalDelay=75, lastDelay=25, exception=java.security.KeyException)
[...]

PR #1256, opened 2020-09-29
Perform data reset for users already affected by the encryption issues (EXPOSUREAPP-1851/EXPOSUREAPP-2958)

d4rken:
~~**DO NOT MERGE** but review :wink:~~
~~This still needs the final translations for the info dialog after a reset.~~

## Description
The mitigation from #1235 only helps users not already affected by the issue (see #642).

Existing users will still keep crashing. There is graceful recovery because the encryption key is gone.
They have to reset their app data. We can do this for them. If the user has automatic updates enabled in Google Play, there may be users who have already been affected, don't know it yet, and where it will be fixed automatically too. They will only see a dialog when they open their app the next time, informing them about what happened (see mock screenshot).

The result should be that the app automatically starts working again if an affected user updates it. We will loose the encrypted data, but the automatic tasks should be able to work again and thus the app  perform it's duty.

**When do we do it?**
* It is a `GeneralSecurityException("decryption failed")`
* The encrypted shared preferences actually exist (otherwise it's a different error cause)
* This is the first error we encountered when creating the encrypted shared preferences, which means that there was no other error before we encountered this one. This allows us to only perform the reset for users upgrading from <1.4.0 to 1.4.0+, as we don't want to reset for any other case potentially hiding other issues.

**What do we do?**
* If the specific error case is matched
* Deleted the encrypted shared preferences
* Delete the encrypted database
* Retry creation of the encrypted shared preferences

## How to test
* Open the app
* Review the code
* Review the tests, are all edge cases covered?
* Install the app, complete onboarding, insert a `throw GeneralSecurityException("decryption failed")` ~~into the `EncryptedPreferencesFactory`~~ actually, insert it into the [SecurityHelper](https://github.com/corona-warn-app/cwa-app-android/pull/1256/files#diff-d05886a3ee6f8f08529a36f8e9fb936bR49), otherwise we keep throwing. Install the new version without data reset. You should see the app reset and then display the dialog, once.

### DIFF
--- a/Corona-Warn-App/config/detekt.yml
+++ b/Corona-Warn-App/config/detekt.yml
@@ -568,7 +568,7 @@ style:
     active: false
   ReturnCount:
     active: true
-    max: 2
+    max: 6
     excludedFunctions: 'equals'
     excludeLabeled: false
     excludeReturnFromLambda: true
--- a/Corona-Warn-App/src/device/java/de.rki.coronawarnapp/ui/main/MainFragment.kt
+++ b/Corona-Warn-App/src/device/java/de.rki.coronawarnapp/ui/main/MainFragment.kt
@@ -22,6 +22,8 @@ import de.rki.coronawarnapp.ui.viewmodel.SubmissionViewModel
 import de.rki.coronawarnapp.ui.viewmodel.TracingViewModel
 import de.rki.coronawarnapp.util.DialogHelper
 import de.rki.coronawarnapp.util.ExternalActionHelper
+import de.rki.coronawarnapp.util.di.AppInjector
+import de.rki.coronawarnapp.util.errors.RecoveryByResetDialogFactory
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -47,6 +49,10 @@ class MainFragment : Fragment() {
     private val submissionViewModel: SubmissionViewModel by activityViewModels()
     private var binding: FragmentMainBinding by viewLifecycle()
 
+    private val errorResetTool by lazy {
+        AppInjector.component.errorResetTool
+    }
+
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
@@ -66,6 +72,19 @@ class MainFragment : Fragment() {
         setContentDescription()
 
         showOneTimeTracingExplanationDialog()
+    }
+
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
+        if (errorResetTool.isResetNoticeToBeShown) {
+            RecoveryByResetDialogFactory(this).showDialog(
+                detailsLink = R.string.errors_generic_text_catastrophic_error_encryption_failure,
+                onDismiss = {
+                    errorResetTool.isResetNoticeToBeShown = false
+                }
+            )
+        }
     }
 
     override fun onResume() {
@@ -148,7 +167,7 @@ class MainFragment : Fragment() {
     private fun toSubmissionIntro() {
         findNavController().doNavigate(
             MainFragmentDirections.actionMainFragmentToSubmissionIntroFragment()
-            )
+        )
     }
 
     private fun showPopup(view: View) {
--- a/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
+++ b/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
@@ -77,9 +77,8 @@ class MainFragment : Fragment() {
         super.onActivityCreated(savedInstanceState)
 
         if (errorResetTool.isResetNoticeToBeShown) {
-            // TODO replace with correct string
             RecoveryByResetDialogFactory(this).showDialog(
-                detailsLink = R.string.main_about_link,
+                detailsLink = R.string.errors_generic_text_catastrophic_error_encryption_failure,
                 onDismiss = {
                     errorResetTool.isResetNoticeToBeShown = false
                 }
--- a/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
+++ b/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
@@ -22,6 +22,8 @@ import de.rki.coronawarnapp.ui.viewmodel.SubmissionViewModel
 import de.rki.coronawarnapp.ui.viewmodel.TracingViewModel
 import de.rki.coronawarnapp.util.DialogHelper
 import de.rki.coronawarnapp.util.ExternalActionHelper
+import de.rki.coronawarnapp.util.di.AppInjector
+import de.rki.coronawarnapp.util.errors.RecoveryByResetDialogFactory
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -47,6 +49,10 @@ class MainFragment : Fragment() {
     private val submissionViewModel: SubmissionViewModel by activityViewModels()
     private var binding: FragmentMainBinding by viewLifecycle()
 
+    private val errorResetTool by lazy {
+        AppInjector.component.errorResetTool
+    }
+
     override fun onCreateView(
         inflater: LayoutInflater,
         container: ViewGroup?,
@@ -65,6 +71,20 @@ class MainFragment : Fragment() {
         setButtonOnClickListener()
 
         showOneTimeTracingExplanationDialog()
+    }
+
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
+        if (errorResetTool.isResetNoticeToBeShown) {
+            // TODO replace with correct string
+            RecoveryByResetDialogFactory(this).showDialog(
+                detailsLink = R.string.main_about_link,
+                onDismiss = {
+                    errorResetTool.isResetNoticeToBeShown = false
+                }
+            )
+        }
     }
 
     override fun onResume() {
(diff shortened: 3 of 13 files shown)


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
```diff
@@ -22,6 +22,8 @@ import de.rki.coronawarnapp.ui.viewmodel.SubmissionViewModel
 import de.rki.coronawarnapp.ui.viewmodel.TracingViewModel
 import de.rki.coronawarnapp.util.DialogHelper
 import de.rki.coronawarnapp.util.ExternalActionHelper
+import de.rki.coronawarnapp.util.di.AppInjector
```
harambasicluka, line 25: Why is this the Fragment in `.../deviceForTesters/...`?

d4rken: :rage: that's the split code issue again, good spot. Need to add it for both fragments `device` and `deviceForTesters`.

harambasicluka: It's about the file not about the line.

Corona-Warn-App/src/device/java/de.rki.coronawarnapp/ui/main/MainFragment.kt
```diff
@@ -68,6 +74,19 @@ class MainFragment : Fragment() {
         showOneTimeTracingExplanationDialog()
     }
 
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
+        if (errorResetTool.isResetNoticeToBeShown) {
```
ralfgehrer, line 80: How about moving this block to a dedicated function "showResetNoticeIfNeeded"

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
```diff
@@ -67,6 +73,19 @@ class MainFragment : Fragment() {
         showOneTimeTracingExplanationDialog()
     }
 
+    override fun onActivityCreated(savedInstanceState: Bundle?) {
+        super.onActivityCreated(savedInstanceState)
+
+        if (errorResetTool.isResetNoticeToBeShown) {
```
ralfgehrer, line 79: How about moving this block to a dedicated function "showResetNoticeIfNeeded"

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
```diff
@@ -0,0 +1,137 @@
...
+        } catch (e: Exception) {
+            // If anything goes wrong, we return false, so that our caller can rethrow the original error.
+            Timber.e(e, "Error while performing the reset.")
+            false
+        }
+    }
+
+    private fun performReset(): Boolean {
+        // Delete encrypted shared preferences file
+
```
ralfgehrer, line 110: no empty line plz

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
```diff
@@ -0,0 +1,137 @@
...
+            return false
+        }
+
+        resetPerformedAt = Instant.now()
+        isResetNoticeToBeShown = true
+
+        return true
+    }
+
+    companion object {
```
ralfgehrer, line 132: to top of class

ralfgehrer (dismissed):

Corona-Warn-App/src/device/java/de.rki.coronawarnapp/ui/main/MainFragment.kt
chris-cwa: <personal opinion>
I don't like methods that are only called from one location. Use doc to make more readable.
</personal opinion>

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
```diff
@@ -0,0 +1,137 @@
...
+    private var isResetWindowConsumed: Boolean
+        get() = prefs.getBoolean(PKEY_EA1851_WAS_WINDOW_CONSUMED, false)
+        set(value) = prefs.edit {
+            putBoolean(PKEY_EA1851_WAS_WINDOW_CONSUMED, value)
+        }
+
+    private var resetPerformedAt: Instant?
+        get() {
+            val performedAt = prefs.getLong(PKEY_EA1851_RESET_PERFORMED_AT, -1)
+            return if (performedAt == -1L) null else Instant.ofEpochMilli(performedAt)
```
chris-cwa, line 49: use Kotlin style
```suggestion
            prefs.getLong(PKEY_EA1851_RESET_PERFORMED_AT, -1).also{
                return if (it == -1L) null else Instant.ofEpochMilli(performedAt)
            }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/SecurityHelper.kt
```diff
@@ -23,26 +23,44 @@ import android.annotation.SuppressLint
...
 import de.rki.coronawarnapp.util.security.SecurityConstants.DB_PASSWORD_MAX_LENGTH
 import de.rki.coronawarnapp.util.security.SecurityConstants.DB_PASSWORD_MIN_LENGTH
 import de.rki.coronawarnapp.util.security.SecurityConstants.ENCRYPTED_SHARED_PREFERENCES_FILE
+import timber.log.Timber
 import java.security.SecureRandom
 
 /**
  * Key Store and Password Access
  */
 object SecurityHelper {
```
chris-cwa, line 40: this should be an injectable class, ideally with interface (maybe out of scope for this PR -> 1 more for refactoring epic)

chris-cwa (dismissed):

Corona-Warn-App/src/device/java/de.rki.coronawarnapp/ui/main/MainFragment.kt
chris-cwa: > How about moving this block to a dedicated function "showResetNoticeIfNeeded"

I don't like methods that are only called from one location. Use doc to make more readable.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/ui/main/MainFragment.kt
chris-cwa: > How about moving this block to a dedicated function "showResetNoticeIfNeeded"

I don't like methods that are only called from one location. Use doc to make more readable.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
chris-cwa, line 49: define const INVALID = -1L to make it more clear?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
d4rken: We attempt to follow the official Kotlin conventions, which say `companion object` last: https://kotlinlang.org/docs/reference/coding-conventions.html

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
d4rken: Agree, that's nicer, will do! Btw we need to use `let` not `also`. Only `let` modifies the returned result.

Corona-Warn-App/src/device/java/de.rki.coronawarnapp/ui/main/MainFragment.kt
d4rken: I concur with @chris-cwa here. I also think `errorResetTool.isResetNoticeToBeShown` is descriptive enough that no comment is warranted. But let's revisit the MainFragment for refactoring in one of the next sprints in any case!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/SecurityHelper.kt
d4rken: Yep out of scope here, would require too many changes, but should be refactored as part of our no-more-singletons tickets.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
d4rken: Not here, it's so close to each other. I would have also used `null` for the check but `getLong` doesn't support `Long?` as default value.

Corona-Warn-App/config/detekt.yml
```diff
@@ -568,7 +568,7 @@ style:
     active: false
   ReturnCount:
     active: true
-    max: 2
+    max: 6
```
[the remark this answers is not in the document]
d4rken: I agree but I also think we need to raise this as it is  improving code readability.

Changing the `tryResetIfNecessary` method to use less returns would reduce readability imho.

While we could remove this change, and then add `Supress(ReturnCount)` to the `tryResetIfNecessary`, then we would have to remove that again when we change the `ReturnCount` value in an extra PR.

So unless someone is opposed to increasing the `ReturnCount` value itself, I would keep it as is.

AlexanderAlferov (approved):

ralfgehrer (approved):

chris-cwa (approved):

PR #1266, opened 2020-09-30
Solve unreliable tests

d4rken:
## Description
Few tests are unreliable.
Related to singletons and build flavors.

## How to test
* Look through the code changes

### DIFF
--- a/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/util/security/VerificationKeysTest.kt
+++ b/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/util/security/VerificationKeysTest.kt
@@ -1,17 +1,29 @@
 package de.rki.coronawarnapp.util.security
 
+import de.rki.coronawarnapp.environment.EnvironmentSetup
 import de.rki.coronawarnapp.exception.CwaSecurityException
 import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.matchers.shouldBe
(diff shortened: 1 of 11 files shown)


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/util/security/VerificationKeysTest.kt
```diff
@@ -54,18 +66,21 @@ class VerificationKeysTest {
     }
 
     companion object {
+        private const val PUB_KEY =
```
MertSafter, line 69: What do you think about moving the companion object to the beginning of the class? I personally like to see all consts/companions at first of the class so I know what consts are used in the code

mertsafter (approved):
LGTM. Noted one small code change but it's not a must.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/util/security/VerificationKeysTest.kt
d4rken: I like the opposite, I think the constants distract from the global variables and constructor arguments. 
Our goal is to follow the Kotlin code conventions which recommend companion objects at the end: 
https://kotlinlang.org/docs/reference/coding-conventions.html

SamuraiKek (approved):

chris-cwa (approved):

PR #1269, opened 2020-10-01
Improve API test fragment and environment switching (DEV)

d4rken:
## Description
* Align environment switching with iOS (type selection instead of a toggle)
* Simplify build config and remove the ALTERNATIVE environment type
* Refactored the API test fragment a bit to move some code over to the viewmodel

## How to test
* Review code+tests
* Open the API test fragment, switch the environment, test a few other options too.


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/api/ui/TestForAPIFragment.kt
```diff
@@ -118,19 +118,11 @@ class TestForAPIFragment : Fragment(R.layout.fragment_test_for_a_p_i),
 
     private var lastSetCountries: List<String>? = null
 
+    @SuppressLint("SetTextI18n")
```
MertSafter, line 121: After your Refactoring this seems to be obsolete.

mertsafter (commented):
LGMT. One commit will be pushed by my side regarding some code cleanup of the fragment layout xml

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/api/ui/TestForAPIFragment.kt
d4rken: Still needed to stop the warnings regarding hardcoded strings (which are okay because we won't add test texts to the translation process).

mertsafter (approved):
LGTM. Approve from me if pipeline run is successful

chris-cwa (approved):

PR #1286, opened 2020-10-01
Reduce code duplication (EXPOSUREAPP-2844)

d4rken:
## Description
Previously we had two `MainFragment` where the only different were the menu entries to the 2 test fragments we have. This also duplicated the `nav_graph.xml` in turn it caused by subtle issues and cost us time when switching between variants caused build errors until cache was cleared and where features were added to only one of the fragments.

This PR removes the extra `MainFragment`, introduces a `TestMenuFragment` as root for all test related screens, and hooks that into the main `nav_graph` via nested nav-graph inclusion. In `device` flavor builds, the nested nav-graph is a stub (empty) and the menu entry is hidden.

Future PRs can refactor the existing test fragments and move the mixed test code into their own Fragments to offer improved modularity and testing.

## How to test
* Review code etc.
* Deploy both `device` and `deviceForTesters` variant, the former should not have a test menu, while the later should have one.
* Navigate the test menu


d4rken:
@ UA I removed the test related texts from the strings.xml here. I don't see much sense in including test related strings in the translation process, using up valuable translation resources. IMHO test/debugging related strings should be in English and can be hardcoded.

BMItr (dismissed):
nice testmenu

janetback (changes requested):
Hi, if this is appearing on UI, could you please add a ' ' (space) between Test an (Manual Test), i.e.
"Android API Test (Manual Test)"
Sorry for being pedantic.

d4rken:
@janetback These are strings only visible in the menus for testers, I can definitely fix the spacing, maybe we should even consider just removing the "test only" strings from the `strings.xml` translation files.

janetback (dismissed):
UA reviewed

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/menu/ui/TestMenuFragmentViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
+import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
+
+class TestMenuFragmentViewModel @AssistedInject constructor() : CWAViewModel() {
+
+    val testMenuData = MutableLiveData<List<TestMenuItem>>()
+    val showTestScreenEvent = SingleLiveEvent<TestMenuItem>()
+
+    init {
+        loadMenu()
```
chris-cwa, line 17: please not in constructor - maybe create load method to call from <fragment>.onViewCreated()?

Corona-Warn-App/src/deviceForTesters/res/drawable/ic_bug.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="24dp"
+    android:width="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:fillColor="#000"
+        android:pathData="M14,12H10V10H14M14,16H10V14H14M20,8H17.19C16.74,7.22 16.12,6.55 15.37,6.04L17,4.41L15.59,3L13.42,5.17C12.96,5.06 12.5,5 12,5C11.5,5 11.04,5.06 10.59,5.17L8.41,3L7,4.41L8.62,6.04C7.88,6.55 7.26,7.22 6.81,8H4V10H6.09C6.04,10.33 6,10.66 6,11V12H4V14H6V15C6,15.34 6.04,15.67 6.09,16H4V18H6.81C7.85,19.79 9.78,21 12,21C14.22,21 16.15,19.79 17.19,18H20V16H17.91C17.96,15.67 18,15.34 18,15V14H20V12H18V11C18,10.66 17.96,10.33 17.91,10H20V8Z" />
+</vector>
```
chris-cwa, line 10: \n @ eof missing

Corona-Warn-App/src/deviceForTesters/res/drawable/ic_coffee.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    android:height="24dp"
+    android:width="24dp"
+    android:viewportWidth="24"
+    android:viewportHeight="24">
+    <path
+        android:fillColor="#000"
+        android:pathData="M2,21H20V19H2M20,8H18V5H20M20,3H4V13A4,4 0 0,0 8,17H14A4,4 0 0,0 18,13V10H20A2,2 0 0,0 22,8V5C22,3.89 21.1,3 20,3Z" />
+</vector>
```
chris-cwa, line 10: \n @ eof missing

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_menu_adapter_item.xml
```diff
@@ -0,0 +1,51 @@
...
+        android:layout_marginEnd="16dp"
+        style="@style/body2"
+        android:layout_marginBottom="8dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toEndOf="@+id/imageView"
+        app:layout_constraintTop_toBottomOf="@+id/title"
+        tools:text="Description" />
+
+</androidx.constraintlayout.widget.ConstraintLayout>
```
chris-cwa, line 51: \n @ eof missing

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_menu.xml
```diff
@@ -0,0 +1,27 @@
...
+
+    <androidx.recyclerview.widget.RecyclerView
+        android:id="@+id/test_menu_list"
+        android:layout_width="match_parent"
+        android:layout_height="0dp"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@+id/toolbar" />
+</androidx.constraintlayout.widget.ConstraintLayout>
```
chris-cwa, line 27: \n @ eof missing

harambasicluka:
> @janetback These are strings only visible in the menus for testers, I can definitely fix the spacing, maybe we should even consider just removing the "test only" strings from the `strings.xml` translation files.

I like the idea, if we do so and only remove the localdata strings we would have a clean string file!

ralfgehrer (dismissed):

PR #1290, opened 2020-10-02
Fail builds fast if an expected file is missing (DEV)

d4rken:
If `test_environments.json` is missing, fail deployment on device for flavor `deviceForTesters` instead of silently falling back to `prod_environments.json`.

## How to test
Delete/Rename the existing `test_environments.json` and try to deploy each build flavor


chris-cwa (dismissed):

BMItr (dismissed):

ralfgehrer (dismissed):

AlexanderAlferov:
I don't think we need this right now: pipelines (CircleCI and Internal) are configured.

PR #1291, opened 2020-10-02
Submission result request handling fix (EXPOSUREAPP-3001)

BMItr:
## Relates to
https://jira.itc.sap.com/browse/EXPOSUREAPP-3001

## Description
If you provide a testresult e.g. by QR Code the result is usually pending until the final result is there.
In case of a final Positiv Result the background requests (fetchTestResults) are stopped (this is expected behaviour).
In case of a final Negativ Result there was an ongoing resultFetching by App-Resuming in Background even the final result there. 
TestResult will only be fetched in correct states now.


kolyaopahle (approved):
passed local testing, LGTM 👍

chris-cwa (approved):

PR #1295, opened 2020-10-02
Federation/EUR key-package file support (EXPOSUREAPP-3025)

d4rken:
## Description
This adds build time config support for switching the keyfile download behavior from countries to *EUR*.

## How to test
* Update your `test_environments.json` with the latest variant.
* Switch to an environment that supports EUR and observe logcat download behavior.


ralfgehrer (approved):

BMItr (approved):
tested with new test_environments.json
LGTM

PR #1296, opened 2020-10-02
Translation Delivery (DEV)

service-tip-git:
Translation Delivery


service-tip-git:
This pull request is for translation delivery. Please review and approve them,  @D067796

Corona-Warn-App/src/main/res/values-bg/strings.xml
```diff
@@ -373,7 +373,7 @@
     <!-- YTXT: one time risk explanation dialog - pointing to the faq page for more information-->
     <string name="risk_details_explanation_dialog_faq_body">"За повече информация вижте страницата „ЧЗВ“."</string>
     <!-- XLNK: risk explanations and informations - pointing to the faq page for more information and contains hyperlink-->
-    <string name="risk_details_explanation_faq_body_with_link"><a href="https://www.coronawarn.app/en/faq/#encounter_but_green">"За повече информация вижте страницата „ЧЗВ“."</a></string>
+    <string name="risk_details_explanation_faq_body_with_link"><a href="https://www.coronawarn.app/de/faq/#encounter_but_green">"За повече информация вижте страницата „ЧЗВ“."</a></string>
```
ralfgehrer, line 376: note: we need to fix that Country label for all but German languages manually again while updating the default translations.

ralfgehrer (approved):

Corona-Warn-App/src/main/res/values-bg/strings.xml
harambasicluka: Could you create an bug for the current sprint to split text and url? So that we don't have to do this any longer?

harambasicluka (approved):

PR #1299, opened 2020-10-02
Add coroutine dispatcher injection to fix tests (DEV)

d4rken:
## Desc
Some of our tests are flaky.
The initial idea was to just retry those, but that failed because it caused issues with verifying calls on mocks (exact call count).

This PR adds a `DispatcherProvider` which can (and should) be injected wherever we need to use `withContext` to switch coroutine contexts, the previous workarounds were removed.

Under test the provided dispatchers can then be replaced, i.e. with `Dispatchers.Unconfined` for testing.

## How to test
* Review code, deploy the app, fetch keys.


d4rken:
And it's still flaky, wasn't the dispatchers...

d4rken:
Might still be the dispatchers, I forgot to replace all instances.

BMItr (dismissed):
LGTM

ralfgehrer (dismissed):

PR #1313, opened 2020-10-05
Translation of legal strings to EN/TR (EXPOSUREAPP-2919, EXPOSUREAPP-2931, EXPOSUREAPP-2938, EXPOSUREAPP-2942)

kolyaopahle:
Update for some german texts with en & tr.


Corona-Warn-App/src/main/res/values-en/legal_strings.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
```
harambasicluka, line 1: As this texts are under our control and english is our fallback language values/legal_strings.xml has to be updated, if english is selected and no dedicated translated strings are present the fallback strings will be used. This means one file we don't have to keep update. So please move this over to values/legal_strings.xml

Corona-Warn-App/src/main/res/values-tr/legal_strings.xml
```diff
@@ -1,24 +1,15 @@
 <?xml version="1.0" encoding="utf-8"?>
-<resources>
-    <string name="interop_consent_body" />
```
harambasicluka, line 3: Is this string not needed any longer?

harambasicluka (changes requested):

ralfgehrer (dismissed):

janetback (changes requested):
Hi Luka, 
at the very end of "submission_positive_other_warning_privacy_body" please replace 'Data Privacy Information' with 'Data Privacy'.
Thanks!

janetback (approved):
UA approved

PR #1320, opened 2020-10-06
Introduce VM injection for home screen and enable screen testing (EXPOSUREAPP-2946, EXPOSUREAPP-2948)

d4rken:
## Description
This PR adds VM injection to the `MainFragment` (now called `HomeFragment`) and provides a first example for testing screens via instrumentation tests.

* I've added a name tag to the `Context` injection to make it absolutely clear what type of context it is, so if you now want `Context` constructor injected you need to label it ` @AppContext val context: Context`
* `TracingViewModel`, `SettingsViewModel` and `SubmissionViewModel` are now injected, i.e. see `HomeFragmentViewModel`, so we are using composition to modularise the logic for UI elements that are reused. The viewmodels are still to complicated and need further refactoring but that's for another PR.
* This PR adds a first example on how to test screens via instrumentation tests (see `HomeFragmentTest`), currently nothing is being tested though because `TracingViewModel` etc. are still too complex to be efficiently mocked. What we have now though is a mechanism to launch a fragment on it's own and provide a mocked viewmodel.
* Check out `HomeFragmentViewModel` and note that we inject a `DispatcherProvider` to make coroutine testing easier, and that the nested viewmodels are added as "child viewmodels" via our `CWAViewModel`

## How to test
* Run the app, click around on the home screen, risk details screen and tracing settings screen.
* Run the unit tests (will work)
* Run the instrumentation test  `HomeFragmentTest`, it will launch the fragment but fail when trying to do data-binding


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -0,0 +1,69 @@
...
+    val settingsViewModel: SettingsViewModel,
+    val submissionViewModel: SubmissionViewModel
+) : CWAViewModel(
+    dispatcherProvider = dispatcherProvider,
+    childViewModels = listOf(tracingViewModel, settingsViewModel, submissionViewModel)
+) {
+
+    val showInteropDeltaOnboarding = SingleLiveEvent<Unit>()
+    val showTracingExplanation = SingleLiveEvent<Long>()
+    val showErrorResetDialog = SingleLiveEvent<Boolean>()
```
chris-cwa, line 29: sealed event class?

Corona-Warn-App/build.gradle
```diff
@@ -166,6 +166,12 @@ android {
                 srcDirs = ['src/device']
             }
         }
+        test {
+            java.srcDirs += "$projectDir/src/testShared/java"
+        }
+        androidTest {
+            java.srcDirs += "$projectDir/src/testShared/java"
```
ralfgehrer, line 173: same folder?

d4rken: Yes, shared folder for test classes used in both unit and instrumentation tests, needs to be added to `test` and `androidTest` to be available.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
d4rken: Note sure what you mean. Something like `data class Event<T>(val value: T)`? Don't think we need that here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -0,0 +1,152 @@
...
+import de.rki.coronawarnapp.util.di.AutoInject
+import de.rki.coronawarnapp.util.errors.RecoveryByResetDialogFactory
+import de.rki.coronawarnapp.util.ui.observe2
+import de.rki.coronawarnapp.util.ui.viewBindingLazy
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactoryProvider
+import de.rki.coronawarnapp.util.viewmodel.cwaViewModels
+import javax.inject.Inject
+
+/**
+ * After the user has finished the onboarding this fragment will be the heart of the application.
```
ralfgehrer, line 21: Kind of obvious IMHO

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -0,0 +1,152 @@
...
+            RecoveryByResetDialogFactory(this).showDialog(
+                detailsLink = R.string.errors_generic_text_catastrophic_error_encryption_failure,
+                onDismiss = { vm.errorResetDialogDismissed() }
+            )
+        }
+    }
+
+    override fun onResume() {
+        super.onResume()
+        // refresh required data
```
ralfgehrer, line 79: remove comment plz

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -0,0 +1,152 @@
...
+                findNavController().doNavigate(
+                    HomeFragmentDirections.actionMainFragmentToSettingsTracingFragment()
+                )
+            }
+        }
+    }
+
+    private fun setupTestResultCard() {
+        binding.apply {
+            val toSubmissionResult = {
```
ralfgehrer, line 105: val not fun ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
ralfgehrer, line 29: do we ever emit "false" as a value for showErrorResetDialog? Unit would suffice otherwise.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/TracingExplanationDialog.kt
```diff
@@ -0,0 +1,39 @@
...
+        val infoPeriodLogged =
+            context.getString(R.string.risk_details_information_body_period_logged)
+        val infoPeriodLoggedAssessment =
+            context.getString(
+                R.string.risk_details_information_body_period_logged_assessment,
+                activeTracingDaysInRetentionPeriod.toString()
+            )
+        val infoFAQ = context.getString(R.string.risk_details_explanation_dialog_faq_body)
+
+        // display the dialog
```
ralfgehrer, line 25: I would prefer function calls over section comments

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
```diff
@@ -4,69 +4,58 @@ import android.os.Bundle
...
-            settingsViewModel.updateManualKeyRetrievalEnabled(false)
+            vm.updateRiskDetails()
         }
         binding.riskDetailsButtonEnableTracing.setOnClickListener {
             findNavController().doNavigate(
                 RiskDetailsFragmentDirections.actionRiskDetailsFragmentToSettingsTracingFragment()
             )
         }
+
+        // Make the links clickable and convert to hyperlink
```
ralfgehrer, line 51: That's what the function says IMHO

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragmentViewModel.kt
```diff
@@ -0,0 +1,36 @@
...
+    dispatcherProvider: DispatcherProvider,
+    val tracingViewModel: TracingViewModel,
+    val settingsViewModel: SettingsViewModel
+) : CWAViewModel(
+    dispatcherProvider = dispatcherProvider,
+    childViewModels = listOf(tracingViewModel, settingsViewModel)
+) {
+
+    fun refreshData() {
+        // refresh required data
```
ralfgehrer, line 21: same same

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/SmartLiveData.kt
```diff
@@ -38,7 +48,42 @@ class SmartLiveDataProperty<T : Any>(
     }
 }
 
-class SmartLiveData<T : Any>(
+class SmartSingleLiveEvent<T : Any>(
```
ralfgehrer, line 51: In that case, the file should be renamed too.
Seems like a strange workaround overall, though.

ralfgehrer, line 51: Hint hint .. are we using LiveData for things it is not designed for ;)

ralfgehrer (dismissed):
Nice PR. Great work.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
d4rken: The `val` value here is the anonymous function we are declaring. I find this more readable and light weight than using a `fun` function type, which requires more to keep the linter from complaining.

I'd only use the later part for more complex scenarios with nested returns.

Screenshot from 2020-10-06 16-31-09

https://stackoverflow.com/questions/48112081/why-does-kotlin-have-two-syntaxes-for-lambdas-anonymous-functions

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
d4rken: Think it's an okay description about the class, I don't think it's super obvious. Let's keep it for now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
d4rken: ofc :smile:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
d4rken: We do not, good point :+1:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/TracingExplanationDialog.kt
d4rken: Agree, neither seems necessary here though, I'll remove the comment.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/SmartLiveData.kt
d4rken: We kind of are, I removed this class, it was not used.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
chris-cwa: like SymptomCalendarEvent - so that we need only 1 life data for the event - as agreed on at https://teams.microsoft.com/l/entity/com.microsoft.teamspace.tab.wiki/tab::32332847-46cb-4995-a3eb-e183b8c68f28?context=%7B%22subEntityId%22%3A%22%7B%5C%22pageId%5C%22%3A21%2C%5C%22sectionId%5C%22%3A22%2C%5C%22origin%5C%22%3A2%7D%22%2C%22channelId%22%3A%2219%3Ac51bf302889a44e28119b0efc59b8f7d%40thread.tacv2%22%7D&tenantId=42f7676c-f455-423c-82f6-dc2d99791af7

d4rken: :sweat_smile:  you are absolutely right, changed. Looks much better now :+1:

chris-cwa (approved):
now it's perfect!

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentTest.kt
```diff
@@ -0,0 +1,46 @@
...
+import testhelpers.BaseUITest
+
+@RunWith(AndroidJUnit4::class)
+class HomeFragmentTest : BaseUITest() {
+
+    @MockK lateinit var viewModel: HomeFragmentViewModel
+
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this, relaxed = true)
```
BMItter, line 22: you can get rid of Annotation.init by using @RunWith(MockKJUnitRunner::class) if you want

PS: if it wouldn't be you I would have to fight for my techLead advising not to use Junit4 ;-)

BMItter:
Looks very good to me. Left one comment & approved

BMItr (approved):

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentTest.kt
d4rken: I'd leave it as it is for now, as it's the same pattern then that we use for unit-tests. Also helps if we need to change to different runners in the future.

PR #1323, opened 2020-10-06
Introduce translations for the button to skip the symptoms screen (EXPOSUREAPP-3088)

ralfgehrer:


Oliver-Zimmerman (dismissed):

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -851,8 +851,10 @@
     <string name="submission_test_result_positive_steps_warning_others_heading">"Andere warnen"</string>
     <!-- YTXT: Body text for for other warnings screen-->
     <string name="submission_test_result_positive_steps_warning_others_body">"Teilen Sie Ihre Zufalls-IDs und warnen Sie andere.\nHelfen Sie, das Infektionsrisiko für andere genauer zu bestimmen, indem Sie auch angeben, wann eventuelle Corona-Symptome zum ersten Mal aufgetreten sind."</string>
-    <!-- XBUT: positive test result : continue button -->
-    <string name="submission_test_result_positive_continue_button">"Weiter"</string>
+    <!-- XBUT: positive test result : continue button with symptoms -->
+    <string name="submission_test_result_positive_continue_button">"Weiter mit Symptom-Erfassung"</string>
```
harambasicluka, line 855: Have you checked that this string isn't used anywhere else? We have this sometimes...

ralfgehrer: Thanks for pointing that out. Indeed it was used in multiple places. Introduced new ids which can be used be the updated implementation.

harambasicluka (approved):

janetback (approved):
UA approved

PR #1324, opened 2020-10-06
Information on how to scan QR Code (EXPOSUREAPP-2581)

Oliver-Zimmerman:
## Description
Added a screen before the QR Code Scan screen in submission flow that explains how the QR Code scanning works.

- Created layout XML file
- Created Fragment 
- Created ViewModel
- Added required string files
- Updated navigation. 
- Added required illustrations

## Test
Begin the submission process, select QR option - you should be presented with the new screen. When next is clicked, you will be taken to the original QR scan screen. Back functionality should work appropriately.


harambasicluka:
Was built based on release/1.5.x should be based on release/1.6.x for that I put it into draft until it's fixed.

PR #1329, opened 2020-10-06
Fix/seperate link from text translation (EXPOSUREAPP-3056)

AndroidMedaGalaxy:
## Description

texts and links / urls are seperated. 

## Screenshot
ezgif-7-1aa4633a2fdc

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
@@ -53,7 +53,7 @@ class RiskDetailsFragment : Fragment(R.layout.fragment_risk_details) {
     private fun setUpWebLinks() {
         binding.riskDetailsInformationLowriskBodyUrl.setUrl(
             R.string.risk_details_explanation_dialog_faq_body,
-            "FAQ",
+            getString(R.string.risk_details_explanation_dialog_faq_body_hyperlink_text),
             getString(R.string.risk_details_explanation_faq_body_with_link)
         )
     }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
@@ -5,17 +5,19 @@ import android.text.method.LinkMovementMethod
 import android.view.View
 import android.view.accessibility.AccessibilityEvent
 import androidx.fragment.app.Fragment
-import androidx.fragment.app.activityViewModels
 import androidx.navigation.fragment.findNavController
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.databinding.FragmentRiskDetailsBinding
-import de.rki.coronawarnapp.timer.TimerHelper
 import de.rki.coronawarnapp.ui.doNavigate
 import de.rki.coronawarnapp.ui.main.MainActivity
 import de.rki.coronawarnapp.ui.viewmodel.SettingsViewModel
 import de.rki.coronawarnapp.ui.viewmodel.TracingViewModel
 import de.rki.coronawarnapp.util.convertToHyperlink
+import de.rki.coronawarnapp.util.di.AutoInject
 import de.rki.coronawarnapp.util.ui.viewBindingLazy
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactoryProvider
+import de.rki.coronawarnapp.util.viewmodel.cwaViewModels
+import javax.inject.Inject
 
 /**
  * This is the detail view of the risk card if additional information for the user.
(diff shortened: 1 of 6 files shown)


Corona-Warn-App/src/main/res/layout/fragment_risk_details.xml
```diff
@@ -269,11 +269,7 @@
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:layout_marginTop="@dimen/spacing_normal"
-                        android:text="@string/risk_details_explanation_faq_body_with_link"
                         android:visibility="@{FormatterRiskHelper.formatVisibilityBehaviorLowLevelRisk(tracingViewModel.riskLevel)}"
-                        android:focusable="true"
```
ralfgehrer, line 274: Can this be safely removed? Did you check on older Android versions?

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -373,8 +373,8 @@
     <string name="risk_details_explanation_dialog_title">"Information zur Funktionsweise der Risiko-Ermittlung"</string>
     <!-- YTXT: one time risk explanation dialog - pointing to the faq page for more information-->
     <string name="risk_details_explanation_dialog_faq_body">"Weitere Informationen finden Sie in den FAQ."</string>
-    <!-- XLNK: risk explanations and informations - pointing to the faq page for more information and contains hyperlink-->
-    <string name="risk_details_explanation_faq_body_with_link"><a href="https://www.coronawarn.app/de/faq/#encounter_but_green">Weitere Informationen finden Sie in den FAQ.</a></string>
+    <!-- XLNK: FAQ URL pointing to the faq page in german. Need to use the URL for english for all other languages-->
+    <string name="risk_details_explanation_faq_body_with_link">"https://www.coronawarn.app/de/faq/#encounter_but_green"</string>
```
ralfgehrer, line 377: Not sure we can reuse the same key. Probably better to introduce a new one. How was this handled in the past @harambasicluka ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
```diff
@@ -51,8 +51,11 @@ class RiskDetailsFragment : Fragment(R.layout.fragment_risk_details) {
      * Make the links clickable and convert to hyperlink
      */
     private fun setUpWebLinks() {
-        binding.riskDetailsInformationLowriskBodyUrl
-            .convertToHyperlink(getString(R.string.risk_details_explanation_faq_body_with_link))
+        binding.riskDetailsInformationLowriskBodyUrl.setUrl(
+            R.string.risk_details_explanation_dialog_faq_body,
+            "FAQ",
```
ralfgehrer, line 56: Why is "FAQ" not a string ressource?

ralfgehrer (changes requested):
Please retarget this onto release/1.5.x so we don't have to fix translation deliveries manuall anymore

Corona-Warn-App/src/main/res/layout/fragment_risk_details.xml
ritsam: I'll have to check on the older android version but have no device for that.

Best is to keep that parameter then.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
ritsam: I can add it as a String source. In fact this will help us when the translated Strings to not contain the word "FAQ"

Also refer to LocationTracingStatusCardBodyTextView.kt
I used this as a reference. Perhaps the string resource can be used there as well

BMItter:
One Note because I just saw this PR by random: The whole line "Weitere Informationen finden Sie in den FAQ" is intended to be linked. Not only FAQ..

Corona-Warn-App/src/main/res/values-de/strings.xml
harambasicluka: I think it's fine if @ritsam checked that it's not used somewhere else.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
harambasicluka: If it's like @BMItter mentioned that the whole string is clickable you should use: `convertToHyperlink` instead of `setUrl`. Please provide screenshot for before and after. To check if it still looks the same as this PR should only separate text and url.

harambasicluka (changes requested):

harambasicluka:
> Please retarget this onto release/1.5.x so we don't have to fix translation deliveries manuall anymore

Not sure if that's possible as I think we have to do a roundtrip with translation to get this right in all files.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/riskdetails/RiskDetailsFragment.kt
ritsam: I think I will have to revert back to the original implementation with _convertToHyperlink_  then. 

because current implementation only shows part of the String ....FAQ... as link and not the complete text as link.

@ralfgehrer  what do you think?

ritsam:
closing pull request to create a cleaner implementation of the branch

PR #1330, opened 2020-10-06
code smell fixes (EXPOSURAPP-3058)

AndroidMedaGalaxy:
## Description

Code smell fixes 
image


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
```diff
@@ -22,8 +22,8 @@ class GoogleAPIVersion @Inject constructor() {
             val currentVersion = InternalExposureNotificationClient.getVersion()
             currentVersion >= compareVersion
         } catch (apiException: ApiException) {
-            if (apiException.statusCode == CommonStatusCodes.API_NOT_CONNECTED) false
-            else throw apiException
+            if (apiException.statusCode != CommonStatusCodes.API_NOT_CONNECTED) throw apiException
+            else return false
```
ralfgehrer, line 26: no "else" needed

ritsam: removing else resulted in Type mismatch error. But will check for any alternate implementation to remove that

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
```diff
@@ -22,11 +22,15 @@ class GoogleAPIVersion @Inject constructor() {
             val currentVersion = InternalExposureNotificationClient.getVersion()
             currentVersion >= compareVersion
         } catch (apiException: ApiException) {
-            if (apiException.statusCode == CommonStatusCodes.API_NOT_CONNECTED) false
-            else throw apiException
+            isGoogleAPIConnected(apiException)
```
ritsam, line 25: @ralfgehrer  the catch block expects a boolean in return hence we need to return 'false'  here.

feel free to comment on it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
ralfgehrer: ok then

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
d4rken, line 25: Extracting the function here makes it less readable, because it's unexpected that it's a function that throws an exception.

Something like this would be nicer:

```kotlin
        if (apiException.statusCode == CommonStatusCodes.API_NOT_CONNECTED) {
            throw apiException
        }
        return false
```

d4rken (changes requested):
One codestyle thing, otherwise lgtm.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/GoogleAPIVersion.kt
ritsam: implemented

d4rken (approved):
lgtm, let's wait for CircleCI to give it's blessing.

PR #1331, opened 2020-10-06
Fix: Pop-up dialog for cancelation of the key upload appears on the wrong screen (EXPOSUREAPP-2915)

kolyaopahle:
## Description
This pr fixes an issue with the test result submission flow where the whole submission would be canceled on back press on the final screen instead of allowing for correction of user input in previous steps.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/fragment/SubmissionResultPositiveOtherWarningFragment.kt
```diff
@@ -120,30 +110,10 @@ class SubmissionResultPositiveOtherWarningFragment :
             initiateWarningOthers()
         }
         binding.submissionPositiveOtherWarningHeader.headerButtonBack.buttonIcon.setOnClickListener {
-            handleSubmissionCancellation()
+            activity?.onBackPressed()
```
kolyaopahle, line 113: Not sure if this is the proper way to handle backwards navigation here, usage between this implementation and an additional navigation path differs in the app.
Any input on which way to choose is appreciated :+1:

harambasicluka:
I'm not sure about the version.

kolyaopahle:
> I'm not sure about the version, maybe it has to go in 1.6.0.

I'll keep a lookout on the jira ticket, will retarget if necessary

kolyaopahle:
1.5 was confirmed as the target release for this PR

harambasicluka (changes requested):
You removed the popup, but shouldn't it be moved to another fragment?

harambasicluka (dismissed):
Works on my machine.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/fragment/SubmissionResultPositiveOtherWarningFragment.kt
ralfgehrer: does "findNavController().popBackStack()" do the trick? Otherwise this seems fine

ralfgehrer (approved):

harambasicluka (approved):

PR #1332, opened 2020-10-06
Correct Protobuf format (EXPOSUREAPP-3071, EXPOSUREAPP-3049)

d4rken:
Correct a missmatch between app and server protobuf format.

App:
https://github.com/corona-warn-app/cwa-app-android/blob/release/1.5.x/Server-Protocol-Buffer/src/main/proto/applicationConfiguration.proto

Server:
https://github.com/corona-warn-app/cwa-server/blob/master/common/protocols/src/main/proto/app/coronawarn/server/common/protocols/internal/app_config.proto

## How to test
* Build the version, switch environments, check UI and log for supportedCountries data.

### DIFF
--- a/Server-Protocol-Buffer/src/main/proto/applicationConfiguration.proto
+++ b/Server-Protocol-Buffer/src/main/proto/applicationConfiguration.proto
@@ -13,7 +13,9 @@ message ApplicationConfiguration {
 
     ApplicationVersionConfiguration appVersion = 5;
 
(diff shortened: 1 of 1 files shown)


ralfgehrer (approved):

BMItr (approved):

harambasicluka (approved):
Works on my machine, I get Belgium and Germany for WRU-XA.

PR #1334, opened 2020-10-06
Add option to skip Symptoms Screens from Test Result (EXPOSUREAPP-3075)

Oliver-Zimmerman:
## Description
Added the option to, on receiving a positive result, continue without sharing any symptom details before even beginning the symptom flow via a second button. 

- Adjusted test result screen layout to include new button
- Adjusted navigation
- Changed logic to set symptom indication to no indication when the new button is pressed

## Test
Scan QR positive QR code. You will see two buttons, one will start symptom flow and one will skip symptom flow. Once at the Other Warning screen, you should be able to submit from both flows. 

Screen-Recording-20201006-161217


harambasicluka (dismissed):
Works on my device! 👍 

Please name the branch the next time like we agreed on and it's written down in our internal wiki: `feature/3075-skip-symptom-submission`
And please also provide screenshots for a UI implementation. As in this case the flow (back and forth navigation) is relevant I would be happy to have a gif :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/fragment/SubmissionTestResultFragment.kt
```diff
@@ -32,6 +32,7 @@ class SubmissionTestResultFragment : Fragment(R.layout.fragment_submission_test_
     private val binding: FragmentSubmissionTestResultBinding by viewBindingLazy()
 
     private var skipInitialTestResultRefresh = false
+    private var skipSymptomSubmission = false
```
ralfgehrer, line 35: IMHO this should rather be a paramter for continueIfTracingEnabled() --> continueIfTracingEnabled(skipSymptomSubmission = false/true) instead of a predefined value

Oliver-Zimmerman: I agree this is more elegant. I will adapt this now.

ralfgehrer (approved):
Works nicely on my device

BMItr (approved):
LGTM 
​​​​​​​
​​​​​​​

PR #1348, opened 2020-10-07
Add r8/proguard configuration to keep GSON related classes.

d4rken:
Specifically keep data class members used in retrofit responses.


ralfgehrer (approved):
Thanks

d4rken:
We specifically needed

```
-keepclassmembers,allowobfuscation class * {
  @com.google.gson.annotations.SerializedName <fields>;
}
```

I also added the rest of the recommended rules though.

harambasicluka (approved):

PR #1352, opened 2020-10-08
Taskcontroller logic (EXPOSUREAPP-2743)

d4rken:
This works towards refactoring our transactions, which will be replaced by our new `Task` concept.
`TaskController` will control these, meaning:

* Handle creation
* Collect results
* Provide access to progress feedback
* Manage queue and skip behavior
* Allow creation and access via injection
* There is a test menu entry to allow testers to observe what is currently happening

This PR introduces Kotlin `Flow<T>` as reactive means of observing changes to current active tasks, and observe progress from active tasks.

## Testing
* Review the code changes
* Check that our unit tests cover all edge cases you can think of
* There is a new test menu entry just for observing and testing the task controller, play around with that.

device-2020-10-12-191541


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/ExampleArguments.kt
```diff
@@ -0,0 +1,7 @@
+package de.rki.coronawarnapp.task.example
```
mlenkeit, line 1: Wouldn't the examples better fit into `src/test` rather than `src/main`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskConfig.kt
```diff
@@ -0,0 +1,11 @@
+package de.rki.coronawarnapp.task
+
+interface TaskConfig {
+
+    val executionMode: ExecutionMode
+
+    enum class ExecutionMode {
```
mlenkeit, line 7: Does an execution mode make sense here? When I have two tasks and I want pass them to the controller(?) to execute them, isn't that when I want to specify whether these two tasks should be executed in parallel or sequentially? The task itself shouldn't know/care about this, IMO.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/ExampleArguments.kt
d4rken: Correct, will do later.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskConfig.kt
d4rken: So the question would be whether the execution mode is something inherent to the task type?
Should the caller have control over the execution mode, or does the task specify which execution mode it requires?

Maybe the task doesn't support concurrency, then someone calling it with `Mode.Parallel` could break things :thinking: 

If the caller doesn't know about other tasks being started by other callers (possibly of the same type), can it even make the correct decision to pick an execution mode?

mlenkeit: > So the question would be whether the execution mode is something inherent to the task type?

Exactly.

How about a combination of `supportedExecutionMode` which is inherent to the task and the actual `executionMode`  that's set by the caller?

But that already seems to complex and might seem like we are designing something though we don't know if we'll ever need it.

Anyway, I think having it on task-level only is not correct. What happens if you schedule two tasks, one with ENQUEUE and one with PARALLEL? How would they be executed?

d4rken: > Anyway, I think having it on task-level only is not correct. What happens if you schedule two tasks, one with ENQUEUE and one with PARALLEL? How would they be executed?

Execution mode would be specific to the task type. "TaskType=A (ENQUEUE)" does not intefere with "TaskType=B (PARALLEL)".

mlenkeit: Do you already have specific scenarios in mind where we _need_ to distinguish between parallel and enqueue? As mentioned above, I strongly recommend not design (for) something where we don't even know if we need it, just because we can 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskData.kt
```diff
@@ -0,0 +1,22 @@
...
+    val request: TaskRequest<*>
+    val createdAt: Instant
+    val startedAt: Instant?
+    val completedAt: Instant?
+    val error: Throwable?
+    val result: Task.Result?
+    val state: State
+    val type: TaskType
+
+    enum class State {
```
mlenkeit, line 17: How would you distinguish a successfully finished task from a failed finished task?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -0,0 +1,160 @@
...
+                    "Sibling are:\n%s", siblingTasks.joinToString("\n")
+                )
+
+                // Handle **[ExecutionMode]** here
+                when {
+                    siblingTasks.isEmpty() || data.config.executionMode == ExecutionMode.PARALLEL -> {
+                        this[data.id] = data.toRunningState()
+                    }
+                    data.config.executionMode == ExecutionMode.ENQUEUE -> {
+                        Timber.tag(TAG).d("Postponing task %s", data)
```
mlenkeit, line 130: Logically this is kind-of if-else, isn't it? Can we make this more obvious?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskConfig.kt
d4rken: I will remove the mode for PARALLEL, the current interfaces allow us to easily readd if necessary.

We need to add a new one though right? `SKIP` The `RetrieveDiagnosisKeysTransaction` transaction currently skips if already running :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskData.kt
d4rken: A finished task always has a result, a task that failed, or was canceled will have an exception.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskConfig.kt
mlenkeit: We would we even submit a task if it's going to be skipped anyway? Is it the task itself that would decide if it shall be skipped?
Additionally, if we decide to add this to the TaskConfig, I'm wondering whether executionMode is the right place or whether this should be something separate.

vaubaehn:
Hi, looks great.
I have a question: How would future debugging look like, if a task is buggy or runs into a crash? With transactions it was more or less easy, if a stack trace contained the clear name of a crashed transaction (from non-minified code). In the current implementation there is the random UUID only to identify a buggy task. I wonder if there is a benefit if a task could additionally be registered with a clear name, especially when many tasks are in the pipe or are running in parallel (if supported later). Or did I miss something?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/ExampleArguments.kt
d4rken: Will be moved once we have another task, otherwise we will have compilation errors because dagger can't construct the map for injection.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskCancelationException.kt
```diff
@@ -0,0 +1,5 @@
+package de.rki.coronawarnapp.task
+
+import java.util.concurrent.CancellationException
+
+class TaskCancelationException : CancellationException()
```
ralfgehrer, line 5: typo: TaskCance**l**lationException

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskFactory.kt
```diff
@@ -0,0 +1,20 @@
...
+
+interface TaskFactory<
+    ProgressType : Task.Progress,
+    ResultType : Task.Result
+    > {
+
+    interface Config {
+        val collisionBehavior: CollisionBehavior
+
+        enum class CollisionBehavior {
```
ralfgehrer, line 11: I'd prefer this not to be inlined in the Config.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskState.kt
```diff
@@ -0,0 +1,47 @@
...
+import kotlin.reflect.KClass
+
+interface TaskState {
+    val request: TaskRequest
+    val createdAt: Instant
+    val startedAt: Instant?
+    val completedAt: Instant?
+    val error: Throwable?
+    val result: Task.Result?
+    val state: State
```
ralfgehrer, line 13: state within a state might be confusing. Maybe  "progress" or "executionState" to be moe precise?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskFactory.kt
d4rken: It's specific to the task factories configuration. It's a value that classes that implement `interface TaskFactory.Config` need to provide, i think it's clean that the interface would also be the location that defines the `enum`.

If it would be more globally used, I might agree here, but this should only ever be declared by the task factory, and read by the `TaskController`.

I don't think extracting this increases readability or code understanding. If you don't feel strongly about this, I'd keep it here. Let's see how the others feel, I don't feel too strongly about it either.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskState.kt
d4rken: Will change it to `executionState`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/InternalTaskState.kt
```diff
@@ -0,0 +1,31 @@
...
+internal data class InternalTaskState(
+    internal val id: UUID = UUID.randomUUID(),
+    override val request: TaskRequest,
+    override val createdAt: Instant,
+    override val startedAt: Instant? = null,
+    override val completedAt: Instant? = null,
+    override val error: Throwable? = null,
+    override val result: Task.Result? = null,
+    internal val config: TaskFactory.Config,
+    internal val deferred: Deferred<Task.Result>,
```
ralfgehrer, line 18: result vs deferred? What's the difference as it is of the same result type. Maybe just a naming "issue"

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/ExampleTaskModule.kt
```diff
@@ -0,0 +1,19 @@
...
+
+import dagger.Binds
+import dagger.Module
+import dagger.multibindings.IntoMap
+import de.rki.coronawarnapp.task.Task
+import de.rki.coronawarnapp.task.TaskFactory
+import de.rki.coronawarnapp.task.TaskTypeKey
+
+@Module
+abstract class ExampleTaskModule {
```
ralfgehrer, line 11: Shouldn't this be renamed to QueingTaskModule. Same for the package. Example hints to be sth. that rather belongs in the test

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskFactory.kt
```diff
@@ -0,0 +1,20 @@
+package de.rki.coronawarnapp.task
+
+interface TaskFactory<
```
ralfgehrer, line 3: There are a lot of Task_Sth_ classes in the task package - some of which are interfaces, others are implementation, models, ... 
How about introducing an "... .task.internal" package to host the implementation details and all public interfaces remain in task

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
```diff
@@ -0,0 +1,80 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
+import timber.log.Timber
+import java.io.File
+import java.util.UUID
+import javax.inject.Inject
+import javax.inject.Provider
+
+open class QueueingTask @Inject constructor() : Task<DefaultProgress, QueueingTask.Result> {
```
ralfgehrer, line 17: Should be removed in the upcomig PRs or moved to the test package and tested

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -0,0 +1,215 @@
...
+        }
+        taskScope.launch { processMap() }
+    }
+
+    private suspend fun processMap() = internalTaskData.updateSafely {
+        Timber.tag(TAG).d("Processing task data (count=%d)", size)
+        Timber.tag(TAG).v("Tasks before processing: %s", this.values)
+
+        // Procress all unprocessed finished tasks
+        values.toList()
```
ralfgehrer, line 122: Could be moved to a dedicated function

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -0,0 +1,215 @@
...
+                this[data.id] = data.copy(
+                    completedAt = timeStamper.nowUTC,
+                    result = result,
+                    error = error
+                )
+                Timber.tag(TAG).i("Task is now FINISHED: %s", data)
+            }
+
+        // Start new tasks
+        values.toList()
```
ralfgehrer, line 142: Could be moved to a dedicated function

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -0,0 +1,219 @@
...
+
+    /**
+     * Don't re-use taskrequests, create new requests for each submission.
+     * They contain unique IDs.
+     */
+    fun submit(request: TaskRequest) {
+        if (!taskFactories.containsKey(request.type.java)) {
+            throw MissingTaskFactoryException(request::class)
+        }
+        Timber.tag(TAG).i("Task submitted: %s", request)
```
ralfgehrer, line 68: This log only holds true after the ".offer" completed with true. It could also fail

ralfgehrer (dismissed):
Nice work

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -0,0 +1,219 @@
...
+        taskQueue.offer(request)
```
ralfgehrer, line 69: IMHO the result of offer should be return, or - to stick to the exceptions - false should be mapped to an Exception.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    private fun InternalTaskState.toSkippedState(): InternalTaskState {
+        Timber.tag(TAG).d("Task was skipped (type=%s, id=%s)", type, id)
+        return copy(
+            startedAt = timeStamper.nowUTC,
+            completedAt = timeStamper.nowUTC
+        ).also {
+            Timber.tag(TAG).i("Starting new task: %s", it)
```
mlenkeit, line 195: Copy-paste mistake, should be "Skipping task", right?

mlenkeit, line 195: How about consistently logging both task type and task id?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskFactory.kt
```diff
@@ -0,0 +1,20 @@
...
+    ProgressType : Task.Progress,
+    ResultType : Task.Result
+    > {
+
+    interface Config {
+        val collisionBehavior: CollisionBehavior
+
+        enum class CollisionBehavior {
+            ENQUEUE,
+            SKIP_IF_ALREADY_RUNNING
```
mlenkeit, line 13: I know I can be a PITA if it comes to naming... I started thinking about what would be the difference between `SKIP_IF_ALREADY_RUNNING` and `SKIP_IF_STILL_RUNNING`. IMO, it depends on the perspective and on the understanding that this Config is for tasks of the same type.
To be very explicit, how about `SKIP_IF_SIBLING_RUNNING`? That's pretty much exactly describing what the TaskController is doing.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskState.kt
```diff
@@ -0,0 +1,47 @@
...
+    val createdAt: Instant
+    val startedAt: Instant?
+    val completedAt: Instant?
+    val error: Throwable?
+    val result: Task.Result?
+    val executionState: ExecutionState
+    val type: KClass<out Task<*, *>>
+
+    val isFinished: Boolean
+        get() = executionState == ExecutionState.FINISHED
```
mlenkeit, line 17: I see `completedAt` but `isFinished`. How about aligning this? So either, `finishedAt` (which I would prefer; `completedAt` has a notion of "success") or `isCompleted`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskState.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    val isFailed: Boolean
+        get() = isFinished && error != null
+
+    val isSkipped: Boolean
+        get() = isFinished && result == null && error == null
```
mlenkeit, line 23: I find this check too implicit. It's making assumptions about how `toSkippedState` works (i.e. that it's not setting any result or error). Is there any existing enum we can use to explicitly check if the task was skipped?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/InternalTaskState.kt
d4rken: `Deferred value is a non-blocking cancellable future &mdash; it is a [Job] with a result.`

Renaming to `job`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/ExampleTaskModule.kt
d4rken: Will be moved into the test package once we have one of the real transactions refactored. Currently here to keep Dagger happy.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
d4rken: Will be moved in an upcoming PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
chris-cwa: I don't like the idea of extracting methods that are called only from a single location. This method is not too complex and sufficiently documented in my opinion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
chris-cwa: I don't like the idea of extracting methods that are called only from a single location. This method is not too complex and sufficiently documented in my opinion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
chris-cwa: I don't like the idea of extracting methods that are called only from a single location. This method is not too complex and sufficiently documented in my opinion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
chris-cwa: I don't like the idea of extracting methods that are called only from a single location. This method is not too complex and sufficiently documented in my opinion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
d4rken: offer will always return `true` here.
We are using a `Channel(Channel.UNLIMITED)` as queue.

```
When capacity is Channel.UNLIMITED — it creates a LinkedListChannel. This channel has a linked-list buffer of unlimited capacity (limited only by available memory). Sending to this channel never suspends, and offer always returns true.
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
d4rken: Can never return false here, see:
https://github.com/corona-warn-app/cwa-app-android/pull/1352#discussion_r503728263

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
d4rken: Yep :sweat_smile:

d4rken: Will make sure to consistently log the whole task state, logging `it` here logs the whole task state. I'll clean up the logging a bit: :+1:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskState.kt
d4rken: Not more or less implicit than failed tasks have an exception and successful tasks have a result, so if skipped tasks would have neither. In some way we have to make a contract about that, I think doing that in the interface of `TaskState` fits.
More explicitly defining with a boolean value seems more error prone, logicwise.

After a task is finished, it's "finished" state is defined by result+error, the `isSkipped` is just a helper here.

chris-cwa (approved):

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskState.kt
d4rken: I'm merging the PR now, but if this is an issue we can fix this in an extra PR.

PR #1356, opened 2020-10-08
Unreliable background priority (EXPOSUREAPP-3094)

kolyaopahle:
## Description
This PR removes the switch for background prioritisation from the settings fragment as they are not stable on all devices.
Im not sure if these changes are complete or if more work on ui flow is needed. Will open in draft state.

image


Corona-Warn-App/src/main/res/values-en/strings.xml
```diff
@@ -633,11 +633,11 @@
     <!-- XACT: settings(background priority) - illustraction description -->
     <string name="settings_background_priority_illustration_description"/>
     <!-- XTXT: settings(background priority) - explains user what to do on card if background priority is enabled -->
-    <string name="settings_background_priority_card_body">"To disable prioritized background activity, please deactivate it in the device settings."</string>
+    <string name="settings_background_priority_card_body">"To enable / disable prioritized background activity, please change it in the device settings"</string>
```
harambasicluka, line 636: We don't change the english strings, that has to go over the translation process.

harambasicluka: please revert all changes in this file.

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -642,11 +642,11 @@
     <!-- XACT: settings(background priority) - illustraction description -->
     <string name="settings_background_priority_illustration_description" />
     <!-- XTXT: settings(background priority) - explains user what to do on card if background priority is enabled -->
-    <string name="settings_background_priority_card_body">"To disable prioritized background activity, please deactivate it in the device settings."</string>
+    <string name="settings_background_priority_card_body">"To enable / disable prioritized background activity, please change it in the device settings"</string>
```
harambasicluka, line 645: same as for the english file, we don't change this.

harambasicluka (dismissed):
Only checked the strings.

Corona-Warn-App/src/main/res/layout/fragment_settings_background_priority.xml
```diff
@@ -57,34 +57,84 @@
...
+                    <androidx.constraintlayout.widget.ConstraintLayout
+                        style="@style/row"
+                        android:layout_width="0dp"
+                        android:layout_height="wrap_content"
+                        android:focusable="true"
+                        app:layout_constraintEnd_toEndOf="parent"
+                        app:layout_constraintStart_toStartOf="parent"
+                        app:layout_constraintTop_toTopOf="parent">
+
+                        <androidx.constraintlayout.widget.ConstraintLayout
```
ralfgehrer, line 77: ConstraintLayout within a ConstraintLayout should not be necessary.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/settings/SettingsBackgroundPriorityFragment.kt
```diff
@@ -40,19 +39,12 @@ class SettingsBackgroundPriorityFragment :
...
-            it.setOnClickListener {
-                val isPriorityEnabled = settingsViewModel.isBackgroundPriorityEnabled.value == true
-
-                if (!isPriorityEnabled) {
-                    (requireActivity() as MainActivity).apply {
-                        startActivitySafely(powerManagement.disableBatteryOptimizationsIntent)
-                    }
-                }
+        settingsRow.setOnClickListener {
+            (requireActivity() as MainActivity).apply {
```
ralfgehrer, line 46: Could (requireActivity() as MainActivity) to be null?  Add a"?." before the apply.

SabineLoss (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/settings/SettingsBackgroundPriorityFragment.kt
kolyaopahle: it cant be null but could raise an exception if no main activity is found, i could replace it with a null check that does nothing instead. Which behaviour would you prefer?

Corona-Warn-App/src/main/res/layout/fragment_settings_background_priority.xml
kolyaopahle: this is taken directly from the `include_settings_switch_row` include and is (i think) used to apply the row style to some of the children of the parent element

ralfgehrer: And now you can remove the nested ConstraintLayout, too @kolyaopahle

AlexanderAlferov (approved):
Tested. Looks good!

chris-cwa (approved):

PR #1359, opened 2020-10-08
Improving Sonarcloud Scans (EXPOSUREAPP-2772)

d4rken:
* Moved duplicate command calls into custom commands to reduce duplicate code
* Optimized call order of saving and restoring build and gradle caches
* Moved sonarcloud into a sepperate job that depends deviceReleaseUnitTests
* Reduce parallelism to 3 (from 4), currently the overhead from 4 offers no performance gain
* Adjusted function names and descriptions
* Removed jetitifier requirement to speed up compile times


AlexanderAlferov (dismissed):

harambasicluka (changes requested):
Please put the dependency changed in 1.6.0.

AlexanderAlferov (approved):

harambasicluka (approved):

PR #1372, opened 2020-10-09
Country list on Risiko-Ermittlung AFTER WLAN ON (EXPOSUREAPP-3152)

Oliver-Zimmerman:
## Description
Small fix for issue 3152 where the country list would not appear after switching Wifi off and then back on again because the network callbacks defined were not working properly. 

Adjusted the network connectivity check and is now working as expected.

## Test

- Install the App, switch of WLAN, 
- Do the onboarding > Contry list missing 
- on the Homescreen choose Exposure Logging > Country List - Country List Missing check device settings 
- go to homescreen
- go to smartphone settings, turn WLAN on
- go back to the app, check WLAN via "Häufige Fragen" > Browser starts, external website appears
- go back to the app, choose Exposure Logging > Country List - Country List Missing check device settings ---> **List should now appear where previously it did not.**

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/interoperability/InteroperabilityConfigurationFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/interoperability/InteroperabilityConfigurationFragment.kt
@@ -26,15 +26,27 @@ class InteroperabilityConfigurationFragment :
 
     private val binding: FragmentInteroperabilityConfigurationBinding by viewBindingLazy()
 
+    private var isNetworkCallbackRegistered = false
+    private val networkCallback = object : ConnectivityHelper.NetworkCallback() {
+        override fun onNetworkAvailable() {
+            vm.getAllCountries()
+            unregisterNetworkCallback()
+        }
+
+        override fun onNetworkUnavailable() {
+            // NOOP
+        }
+    }
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        if (ConnectivityHelper.isNetworkEnabled(CoronaWarnApplication.getAppContext())) {
-            vm.getAllCountries()
-        }
-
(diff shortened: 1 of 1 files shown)


ralfgehrer (dismissed):

BMItter:
@Oliver-Zimmerman Hi Oliver, I just did a quick look. I think without somekind of callback or observer this would indicate another problem: wifi usually needs some seconds to establish a connection. If you return to the App or Onboarding before having a connection, it could result in country list not showing...

Oliver-Zimmerman:
@BMItter interesting, I didn't consider this. I will test it again to see if this issue occurs and adjust based on my findings. Did you test to see if this issue was occurring? I imagine it would only appear if you enable WLAN right before going to the country list screen.

BMItter:
Yes, I did a quick test.. maybe it would be a solution to leave the callback there and combine it with you enhencement (ConnectivityHelper)..

Oliver-Zimmerman:
@BMItter I put the registerNetworkCallback within the ConnectivityHelper check and this now seems to be working as expected.

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/interoperability/InteroperabilityConfigurationFragment.kt
```diff
@@ -88,4 +91,11 @@ class InteroperabilityConfigurationFragment :
         super.onDestroy()
         unregisterNetworkCallback()
     }
+
+    override fun onResume() {
+        super.onResume()
+        if (ConnectivityHelper.isNetworkEnabled(CoronaWarnApplication.getAppContext())) {
```
ralfgehrer, line 97: minor:  if could be moved into the registerNetworkCallback()

ralfgehrer (approved):

BMItter:
lgtm. quicktest on Android 10,11 passed.

BMItr (approved):

PR #1375, opened 2020-10-09
Fix behavior of last 3 hours test mode switch (EXPOSUREAPP-3179)

d4rken:
Previously the days for which the last 3 hours were fetched, was the latest day in the servers index for a country, but if a day is not complete, the day will not be visible in the index. This PR fixes this behavior by artificially adding the one additional date (`LATEST_DAY_IN_INDEX+1DAY`).

## Testing
* Try it without this PR, check in the logs what the `KeyFileDownloader` downloads.
* Try with this PR, check that the actual last 3 hours are downloaded.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyFileDownloader.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyFileDownloader.kt
@@ -92,6 +92,7 @@ class KeyFileDownloader @Inject constructor(
                         Timber.tag(TAG).w("Missing keyfile for : %s", keyInfo)
                         null
                     } else {
+                        Timber.tag(TAG).v("Providing available key: %s", keyInfo)
                         path
                     }
                 }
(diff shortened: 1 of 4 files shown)


mlenkeit (commented):
What if there are less than 3 hour packages available for the given date?

mlenkeit:
> What if there are less than 3 hour packages available for the given date?

Just for the record: I think it's not required to fix this, as the 3 Hour mode is only used for testing, and it wouldn't work as the name suggests only between midnight and 2 am or so. I can live with that 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyFileDownloader.kt
```diff
@@ -171,11 +172,21 @@ class KeyFileDownloader @Inject constructor(
...
+                    it.plus(lastDayInIndex.plusDays(1))
+                } else {
+                    it
+                }
+            }
+            Timber.tag(TAG).v("Day index with (fake) today entry: %s", indexWithToday)
+
+            indexWithToday.sortedDescending().mapNotNull { day ->
                 // Limit reached, return null (filtered out) instead of new CountryHours object
                 if (remainingItems <= 0) return@mapNotNull null
```
d4rken, line 191: @mlenkeit then we rollover to the previous day, see `remainingItems`.

d4rken: I'll push a PR for 1.6.x that changes to 24 hour mode and includes a unit-test for the rolloverbehavior.

BMItr (approved):

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyFileDownloader.kt
ralfgehrer: Max is fine with this change, too.

PR #1385, opened 2020-10-12
Disabled dates which exceed 21 days in the past (EXPOSUREAPP-3191)

SamuraiKek:
## Description

This PR fixes a bug where you could select days in the symptom submission calendar which were beyond 21 days in the past.
## How to test
- Get a positive result by scanning a QR code.
- Start the submission of keys process and on the Symptoms screen, select "Yes" and click next
- On the calendar view, the days that are beyond 21 days in the past are greyed out and can't be selected.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/calendar/CalendarDayViewHolder.kt
```diff
@@ -39,7 +39,7 @@ class CalendarDayViewHolder(v: View) : RecyclerView.ViewHolder(v) {
         textView.contentDescription = day.date.toString(talkBackDateFormat)
 
         // If date is after today - then disable click listener
```
harambasicluka, line 41: Maybe also adjust the comment to reflect the new logic.

harambasicluka (dismissed):
lgtm, it just would be nice if you could adjust the comment.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/calendar/CalendarDayViewHolder.kt
SamuraiKek: I just saw your comments, but I have adjusted the comments to reflect the new logic in the latest commit.

harambasicluka (dismissed):

ralfgehrer (changes requested):
Screenshot_20201013-142201


Although September 21st cannot be selected, I suppose it should be greyed out as well !?

d4rken:
There is one codesmell according to Sonar, can this be fixed before the merge?

SamuraiKek:
> There is one codesmell according to Sonar, can this be fixed before the merge?
Yep, smelling fresh now.

SamuraiKek:
> Although September 21st cannot be selected, I suppose it should be greyed out as well !?
Past dates are greyed out now. It used to work yesterday but a commit I did today somehow broke it.

kolyaopahle (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/calendar/CalendarDayViewHolder.kt
```diff
@@ -55,8 +55,8 @@ class CalendarDayViewHolder(v: View) : RecyclerView.ViewHolder(v) {
                 textView.setBackgroundResource(R.drawable.calendar_today_back)
                 textView.setTextColor(ContextCompat.getColor(context, R.color.colorCalendarTodayText))
             }
-            // Future
-            day.date.isAfter(today) -> {
+            // Future & past exceeding 21 days
+            day.date.isAfter(today) || day.date.isBefore(today.minusDays(ONSET_PERIOD)) -> {
```
kolyaopahle, line 59: as a notice for further refactoring, since this is used in multiple places it might make sense to move these checks into a function. Especially if more conditions are introduced in the future.

kolyaopahle (approved):

ralfgehrer (approved):
works as expected

PR #1388, opened 2020-10-13
1.4.x into 1.5.x (DEV)

d4rken:


d4rken:
I previously did "master into 1.4.x" and resolved a lot of conflicts caused by prior stuff. We are pulling these resolved conflicts and our commit history straight here to not get caught by conflicts when needing to do hotfixes in the future.

PR #1390, opened 2020-10-13
Hide app content in app launcher (EXPOSUREAPP-2612)

ralfgehrer:
**Description**
- The app preview in the app launcher is blank (no content)
- The ability to create screenshots withing the app is not affected

**How to test**
Run the app and validate the above-mentioned criteria on any screen you can navigate to:

- You should be able to take screenshots of any screen
- The app launcher shall not show the app's content (on none-rooted devices)

**Screenshots**
Screenshot_20201013-113702
Screenshot_20201013-113712

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
@@ -5,6 +5,7 @@ import android.app.Application
 import android.content.Context
 import android.content.IntentFilter
 import android.os.Bundle
+import android.view.WindowManager
 import androidx.localbroadcastmanager.content.LocalBroadcastManager
 import androidx.work.Configuration
 import androidx.work.WorkManager
@@ -82,10 +83,11 @@ class CoronaWarnApplication : Application(), HasAndroidInjector {
                 localBM.unregisterReceiver(it)
                 errorReceiver = null
             }
+            disableAppLauncherPreviewAndScreenshots(activity)
         }
 
         override fun onActivityStarted(activity: Activity) {
-            // NOOP
+            enableAppLauncherPreviewAndScreenshots(activity)
         }
 
         override fun onActivityDestroyed(activity: Activity) {
@@ -97,7 +99,7 @@ class CoronaWarnApplication : Application(), HasAndroidInjector {
         }
 
         override fun onActivityStopped(activity: Activity) {
-            // NOOP
+            disableAppLauncherPreviewAndScreenshots(activity)
         }
 
         override fun onActivityCreated(activity: Activity, savedInstanceState: Bundle?) {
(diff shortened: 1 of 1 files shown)


d4rken:
Works on a Huawei CLT-L09 (Android 8.1.0): 
[device-2020-10-13-140333](https://user-images.githubusercontent.com/1439229/95858323-35c2a380-0d5d-11eb-93d9-2b0df13ccf2b.png)

Works on a Pixel 3a Emulator (Android 11):
[device-2020-10-13-141526](https://user-images.githubusercontent.com/1439229/95859262-9c948c80-0d5e-11eb-95ad-0ee0277485e3.png)


Doesn't work on a rooted Pixel 3a (Android 11):
[device-2020-10-13-140120](https://user-images.githubusercontent.com/1439229/95858334-3b1fee00-0d5d-11eb-9255-b0f03faed653.png)

So is the flag here automatically ignored if the device is rooted? Seems weird :thinking: Does it check the unlocked bootloader?

ralfgehrer:
> Works on a Huawei CLT-L09 (Android 8.1.0):
> [device-2020-10-13-140333](https://user-images.githubusercontent.com/1439229/95858323-35c2a380-0d5d-11eb-93d9-2b0df13ccf2b.png)
> 
> Works on a Pixel 3a Emulator (Android 11):
> [device-2020-10-13-141526](https://user-images.githubusercontent.com/1439229/95859262-9c948c80-0d5e-11eb-95ad-0ee0277485e3.png)
> 
> Doesn't work on a rooted Pixel 3a (Android 11):
> [device-2020-10-13-140120](https://user-images.githubusercontent.com/1439229/95858334-3b1fee00-0d5d-11eb-9255-b0f03faed653.png)
> 
> So is the flag here automatically ignored if the device is rooted? Seems weird 🤔 Does it check the unlocked bootloader?

see https://developer.android.com/reference/android/view/Display#FLAG_SECURE

Seems to me, that on rooted devices, the app launcher can choose itself what to show, and as the app sandboxing is circumvented the encrypted data can be accessed nevertheless. Hence several posts mentioned that the flag is not reliable on rooted devices.

d4rken (dismissed):
lgtm, if possible confirm behavior on an unrooted Pixel 3a (API30), but not a blocker, current behavior is already better than before.

harambasicluka (dismissed):

ralfgehrer:
@d4rken tested in a non-rooted Pixel 3A. Same behavior as on the rooted one.
With @BMItter's changes, screenshots are disabled. We can only have one of the two options: 
a) reliably hide app content in the launcher but disable screenshots within the app
OR
b) hide app content in the launcher for some devices but allow screenshots within the app.

d4rken:
I think screenshots should not be disabled. I don't think an app should tell the user what he/she can do with their device. If they want to take a screenshot, they can drop their phone on a copier :shrug:, disabling it just makes it cumbersome. If someone wants to take screenshots, let them.

I also don't see a huge risk with the app content showing in the launcher, if the device is locked, no one can see it, if the device is unlocked, someone can just open the app and view it. While I agree that you may show someone else your device and could inadvertently reveal the corona app's screen content, I don't think it's an important enough "feature" to warrant disabling screenshots.

Final decision probably lies with the POs here then.

BMItr (approved):
already tested, with known result. approved

d4rken (approved):

PR #1393, opened 2020-10-13
Calculate the apps versionName and versionCode (EXPOSUREAPP-2772)

d4rken:
Instead of manually increasing values everywhere, we define variables for `MAJOR`,`MINOR`,`PATCH` and `BUILD` in `gradle.properties`.

This attempts to solve two issues:
* Due to development on multiple release candidates, and the need for `versionCode` to increase, it can happen that due to a hotfix (version bump) on i.e. `release/1.4.x`, we need to update an APK build from `release/1.5.x` too, just to increase the `versionCode`. The current calculation scheme would space minor versions `10000` iterations apart. We also get `99` RCs, `99` patches, `99` minor updates, `99` :balloon:  and unlimited major versions of course.
* Pipelines can't easily read the `versionCode` configured (file format/formatting) in `build.gradle`, but can easily read the `bash` like syntax in `gradle.properties`. As long as the pipeline code calculated the `versionCode` the same way, it can easily get the version information.

Example output:
```bash
# In properties
VERSION_MAJOR=1
VERSION_MINOR=6
VERSION_PATCH=0
VERSION_BUILD=1
# Calculated
Current VERSION_MAJOR: 1
Current VERSION_MINOR: 6
Current VERSION_PATCH: 0
Current VERSION_BUILD: 1
Used versionCode: 1060001
Used versionName: 1.6.0-RC1
```

Version 2, update 22, on it's 14th hotfix  :dizzy_face: would look like this:
```bash
# In properties
VERSION_MAJOR=2
VERSION_MINOR=22
VERSION_PATCH=14
VERSION_BUILD=1
# Calculated
Current VERSION_MAJOR: 2
Current VERSION_MINOR: 22
Current VERSION_PATCH: 14
Current VERSION_BUILD: 1
Used versionCode: 2221401
Used versionName: 2.22.14-RC1
```

### DIFF
--- a/Corona-Warn-App/build.gradle
+++ b/Corona-Warn-App/build.gradle
@@ -32,6 +32,11 @@ def environmentExtractor = { File path ->
 }
 
 android {
+    println("Current VERSION_MAJOR: ${VERSION_MAJOR}")
+    println("Current VERSION_MINOR: ${VERSION_MINOR}")
+    println("Current VERSION_PATCH: ${VERSION_PATCH}")
+    println("Current VERSION_BUILD: ${VERSION_BUILD}")
+
     compileSdkVersion 29
     buildToolsVersion "29.0.3"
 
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/build.gradle
```diff
@@ -32,15 +32,29 @@ def environmentExtractor = { File path ->
 }
 
 android {
+    println("Current VERSION_MAJOR: ${VERSION_MAJOR}")
```
ralfgehrer, line 35: we don't really need  the println,  do we

ralfgehrer (approved):

kaddaSz (approved):

Corona-Warn-App/build.gradle
d4rken: Not for us, but I think it's an easy addition which will allow pipeline development to check in the gradle output/logs what values were read. Can remove it in a future PR when the new pipeline is :rocket:.

AlexanderAlferov (approved):

PR #1406, opened 2020-10-14
Increase transaction timeout (EXPOSUREAPP-3238)

d4rken:
Raise the transaction timeout to 8 minutes (from 3 minute) to prevent slower devices from unnecessarily canceling transactions. 
There is no apparent downside to increasing it until we would hit the 10 minute mark.
At 10 minutes it becomes problematic because that's the maximum time we get for background execution in worker tasks, and we want to "timeout" on our own terms not because the worker manager kills us.


harambasicluka (dismissed):

ralfgehrer (dismissed):

AlexanderAlferov:
getTransactionTimeout (de.rki.coronawarnapp.risk.TimeVariablesTest) should be updated as well

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/TimeVariables.kt
```diff
@@ -34,7 +34,7 @@ object TimeVariables {
      * The maximal runtime of a transaction
      * In milliseconds
      */
-    private const val TRANSACTION_TIMEOUT = 180000L
+    private const val TRANSACTION_TIMEOUT = 8 * 60 * 1000L
```
kolyaopahle, line 37: Maybe also add the background task system timeout as a comment here so in future it wont be set higher than 10 minutes by accident?

AlexanderAlferov (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/TimeVariables.kt
```diff
@@ -33,8 +33,10 @@ object TimeVariables {
     /**
      * The maximal runtime of a transaction
      * In milliseconds
+     * Stay below 10min with this timeout!
+     * We only 10min background execution time via WorkManager.
      */
-    private const val TRANSACTION_TIMEOUT = 180000L
+    private const val TRANSACTION_TIMEOUT = 8 * 60 * 1000L
```
chris-cwa, line 39: why does the docs say "stay below 10 minutes" but the implementation is 3 minutes? what am I missing?

harambasicluka (approved):

PR #1407, opened 2020-10-14
TaskController: Support timeout behavior (EXPOSUREAPP-2842)

d4rken:
This PR adds timeout behavior to our `TaskController`.
Each tasks **has to** specificy a timeout value via their `TaskFactory.Config` class.
If the timeout is reached, the task will be killed by the controller and processed.
The final TaskState will be `result = null` and `error = TimeoutCancellationException`.

The unit tests check timeout, timeout on queued tasks, and timeout for tasks running in parallel.


chris-cwa (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskFactory.kt
```diff
@@ -1,11 +1,18 @@
...
 interface TaskFactory<
     ProgressType : Task.Progress,
     ResultType : Task.Result
     > {
 
     interface Config {
+        /**
+         * The maximal runtime of the task, before it is canceled by the controller
+         */
+        val executionTimeout: Duration
```
ralfgehrer, line 14: I'm wondering if made  sense  to 
a) set a default
b) enforce a maximum timeout

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/task/testtasks/timeout/TimeoutTask2.kt
```diff
@@ -0,0 +1,9 @@
+package de.rki.coronawarnapp.task.testtasks.timeout
+
+import javax.inject.Provider
+
+class TimeoutTask2 : BaseTimeoutTask() {
```
ralfgehrer, line 5: There is no difference between  TimeoutTask and TimeoutTask2, is there? In that case, he could remove this one !?

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskFactory.kt
d4rken: > set a default
Because we (the TaskController itself) doesn't know how the tasks will be used, I don't think we can set a reasonable timeout, and I like that when someone creates a new task, they are forced to think about and specify a timeout.

> enforce a maximum timeout
The maximum timeout is basically Long.MAX_VALUE in milliseconds. Similar to my previous point, the TaskController should not be aware or take part in specifying too much default behavior for Tasks.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/task/testtasks/timeout/TimeoutTask2.kt
d4rken: They are the same, but the TaskController differentiates between two task types (to decide on parallel execution) by comparing the classes, we need a second task class here to test timeouts in parallel execution.

PR #1410, opened 2020-10-14
Refactor submission flow fragments (EXPOSUREAPP-2746) 

Oliver-Zimmerman:
## Description
Initial refactor of the submission fragments to introduce the MVVM pattern and dependency injection mostly for navigation-related tasks. 

Also migrated SymptomIntroduction and SymptomCalendar fragments into the submission/fragment folder for continuity. 

## Test
Run the app and test the submission flow, including all options on the dispatcher screen as well as submitting a positive result with and without providing onset of symptoms. There should be no difference in functionality.


d4rken (commented):
lgtm

~~What do you think about moving the files in this PR? i.e. subpackages for each screen that then contain fragment+viewmodel+module?~~

Let's not do it in this PR. Will create conflicts with mine.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivityModule.kt
```diff
@@ -6,15 +6,15 @@ import de.rki.coronawarnapp.ui.interoperability.InteroperabilityConfigurationFra
...
 import de.rki.coronawarnapp.ui.riskdetails.RiskDetailsFragmentModule
+import de.rki.coronawarnapp.ui.submission.SubmissionActivityModule
 
 @Module(
     includes = [
         OnboardingDeltaInteroperabilityModule::class,
         HomeFragmentModule::class,
-        RiskDetailsFragmentModule::class
+        RiskDetailsFragmentModule::class,
+        SubmissionActivityModule::class
```
d4rken, line 17: Small nitpick, let's call it `SubmissionFragment*s*Module`, as has nothing to do with activities.

d4rken (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivityModule.kt
Oliver-Zimmerman: Of course! I'll change that now

d4rken (approved):

BMItr (approved):
Code LGTM, flow tested on Pixel 2 & works fine

​​​

PR #1412, opened 2020-10-15
Converted risk level transaction to task (EXPOSUREAPP-2842)

chris-cwa:
Using CWAs new task controller for risk level calculation.

Change to user experience of the app is not intended and should keep this PR from being approved.

- [x]  unit tests for risk level task and RiskLevels class
- [x] check changes in https://github.com/corona-warn-app/cwa-app-android/pull/1418
- [x] wait for https://github.com/corona-warn-app/cwa-app-android/pull/1418 to be approved


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+import javax.inject.Singleton
+import kotlin.math.round
+
+@Singleton
+class DefaultRiskLevels @Inject constructor() : RiskLevels {
+
+    companion object {
+
+        private var TAG = DefaultRiskLevels::class.simpleName
+
```
ralfgehrer, line 30: remove empty lines plz

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+        private const val DECIMAL_MULTIPLIER = 100
+    }
+
+    private fun calculateRiskScore(
+        attenuationParameters: ApplicationConfigurationOuterClass.AttenuationDuration,
+        exposureSummary: ExposureSummary
+    ): Double {
+
```
ralfgehrer, line 38: remove empty line plz

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+                .plus(defaultBucketOffset)
+
+        Timber.v("Formula used: ($maximumRiskScore / $normalizationDivisor) * $weightedAttenuationDuration")
+
+        val riskScore = (maximumRiskScore / normalizationDivisor) * weightedAttenuationDuration
+
+        return round(riskScore.times(DECIMAL_MULTIPLIER)).div(DECIMAL_MULTIPLIER)
+    }
+
+    private fun Int.capped() =
```
ralfgehrer, line 73: personally, I prefer to have override functions first, then public, then private

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+            Timber.tag(TAG).v("update the risk level with $riskLevel")
+            val lastCalculatedRiskLevelScoreForRollback =
+                RiskLevelRepository.getLastCalculatedScore()
+            updateRiskLevelScore(riskLevel)
+            rollbackItems.add {
+                updateRiskLevelScore(lastCalculatedRiskLevelScoreForRollback)
+            }
+
+            // risk level calculation date update
+            val lastCalculatedRiskLevelDate = LocalData.lastTimeRiskLevelCalculation()
```
ralfgehrer, line 92: LocalData should be injected IMHO

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+            }
+        }
+
+    /**
+     * Updates the Risk Level Score in the repository with the calculated Risk Level
+     *
+     * @param riskLevel
+     */
+    private fun updateRiskLevelScore(riskLevel: RiskLevel) {
+        val lastCalculatedScore = RiskLevelRepository.getLastCalculatedScore()
```
ralfgehrer, line 218: RiskLevelRepository should be injected

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+        } else {
+            this
+        }
+
+    override fun updateRepository(riskLevel: RiskLevel, time: Long) {
+        val rollbackItems = mutableListOf<RollbackItem>()
+        try {
+            Timber.tag(TAG).v("update the risk level with $riskLevel")
+            val lastCalculatedRiskLevelScoreForRollback =
+                RiskLevelRepository.getLastCalculatedScore()
```
ralfgehrer, line 85: RiskLevelRepository should be injected

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Updates the Risk Level Score in the repository with the calculated Risk Level
+     *
+     * @param riskLevel
+     */
+    private fun updateRiskLevelScore(riskLevel: RiskLevel) {
+        val lastCalculatedScore = RiskLevelRepository.getLastCalculatedScore()
+        if (RiskLevel.riskLevelChangedBetweenLowAndHigh(lastCalculatedScore, riskLevel)) {
+            NotificationHelper.sendNotification(
+                CoronaWarnApplication.getAppContext().getString(R.string.notification_body),
```
ralfgehrer, line 221: rather use LazyString

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,243 @@
...
+
+    fun withinDefinedLevelThreshold(riskScore: Double, min: Int, max: Int) =
+        riskScore >= min && riskScore <= max
+
+    /**
+     * Make a call to the backend to retrieve the current application configuration values
+     *
+     * @return the [ApplicationConfigurationOuterClass.ApplicationConfiguration] from the backend
+     */
+    private suspend fun getApplicationConfiguration(): ApplicationConfigurationOuterClass.ApplicationConfiguration =
```
d4rken, line 203: Inject `AppConfigProvider`  and access the config from there please

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,135 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.asFlow
+import org.joda.time.Duration
+import timber.log.Timber
+import javax.inject.Inject
+import javax.inject.Provider
+
+class RiskLevelTask @Inject constructor(
+    private val riskLevels: RiskLevels,
+    private val context: Context
```
d4rken, line 29: I think you need to annotate this with `@AppContext'

d4rken (changes requested):
Looking good

d4rken:
Didn't do an indepth review yet, just short lookthrough.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,135 @@
...
+) : Task<DefaultProgress, RiskLevelTask.Result> {
+
+    private val internalProgress = ConflatedBroadcastChannel<DefaultProgress>()
+    override val progress: Flow<DefaultProgress> = internalProgress.asFlow()
+
+    private var isCanceled = false
+
+    override suspend fun run(arguments: Task.Arguments) = try {
+        Timber.d("Running with arguments=%s", arguments)
+        Result(
```
d4rken, line 39: Don't nest a whole block like this please

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,135 @@
...
+                        } -> UNKNOWN_RISK_INITIAL
+
+                        else -> LOW_LEVEL_RISK
+                    }.also {
+                        checkCancel()
+                        updateRepository(it, System.currentTimeMillis())
+                    }
+                }
+            } else {
+                RiskLevelRepository.setLastCalculatedRiskLevelAsCurrent()
```
d4rken, line 75: Maybe early abort instead of the big if-else?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,135 @@
...
+    ) : TaskFactory<DefaultProgress, Result> {
+
+        override val config: TaskFactory.Config = Config()
+        override val taskProvider: () -> Task<DefaultProgress, Result> = {
+            taskByDagger.get()
+        }
+    }
+
+    companion object {
+        private val RISK_LEVEL_TRANSACTION_TIMEOUT = Duration.standardMinutes(8)
```
d4rken, line 132: Improved readability if we inline this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
chris-cwa: OK.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
chris-cwa: OK.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
chris-cwa: OK.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/RiskLevelAndKeyRetrievalBenchmark.kt
```diff
@@ -29,7 +33,7 @@ class RiskLevelAndKeyRetrievalBenchmark(
      */
     suspend fun start(
         callCount: Int,
-        callback: (resultInfo: String) -> Unit
+        callback: Callback
```
ralfgehrer, line 36: Can you please give "Callback" a more descriptive name

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,234 @@
...
+                )
+        /** we only return outdated risk level if the threshold is reached AND the active tracing time is above the
+        defined threshold because [UNKNOWN_RISK_INITIAL] overrules [UNKNOWN_RISK_OUTDATED_RESULTS] */
+        return timeSinceLastDiagnosisKeyFetchFromServer.millisecondsToHours() >
+            TimeVariables.getMaxStaleExposureRiskRange() && isActiveTracingTimeAboveThreshold()
+    }
+
+    override fun calculationNotPossibleBecauseNoKeys() =
+        (TimeVariables.getLastTimeDiagnosisKeysFromServerFetch() == null).also {
+            if (it)
```
ralfgehrer, line 80: .also.if(true) ? Why not simplify

`if(TimeVariables.getLastTimeDiagnosisKeysFromServerFetch() == null) { Timber.... }`

Is the same, isn't it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,234 @@
...
+                .plus(defaultBucketOffset)
+
+        Timber.v("Formula used: ($maximumRiskScore / $normalizationDivisor) * $weightedAttenuationDuration")
+
+        val riskScore = (maximumRiskScore / normalizationDivisor) * weightedAttenuationDuration
+
+        return round(riskScore.times(DECIMAL_MULTIPLIER)).div(DECIMAL_MULTIPLIER)
+    }
+
+    fun Int.capped() =
```
ralfgehrer, line 190: should be private (or internal with a @VisibleForTesting annotation)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,234 @@
...
+        if (this > TimeVariables.getMaxAttenuationDuration()) {
+            TimeVariables.getMaxAttenuationDuration()
+        } else {
+            this
+        }
+
+    fun withinDefinedLevelThreshold(riskScore: Double, min: Int, max: Int) =
```
ralfgehrer, line 197: should be private (or internal with a @VisibleForTesting annotation)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,234 @@
...
+        riskScore >= min && riskScore <= max
+
+    /**
+     * Updates the Risk Level Score in the repository with the calculated Risk Level
+     *
+     * @param riskLevel
+     */
+    fun updateRiskLevelScore(riskLevel: RiskLevel) {
```
ralfgehrer, line 205: should be private (or internal with a @VisibleForTesting annotation)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,234 @@
...
+                "active tracing time ($activeTracingDurationInHours h) is above threshold " +
+                    "($durationTracingIsActiveThreshold h): $it"
+            )
+            if (it) {
+                Timber.tag(TAG).v("active tracing time is not enough")
+            }
+        }
+    }
+
+    fun calculateRiskScore(
```
ralfgehrer, line 152: should be private (or internal with a @VisibleForTesting annotation)

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/risk/RiskLevelsTest.kt
```diff
@@ -1,16 +1,43 @@
...
+
+    @Before
+    fun setUp() {
+        MockKAnnotations.init(this)
+        riskLevels = DefaultRiskLevels(appConfigProvider)
+    }
+
+    @Test
+    fun `is within defined level threshold`() {
+        // positive
```
ralfgehrer, line 26: I'd prefer separate test cases, so it's obvious by the test case name which scenario failed, instead of having to look at the line number.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
chris-cwa: will be done in another PR

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
chris-cwa: will be done in another PR

kolyaopahle:
i made some small changes to the risk transaction in https://github.com/corona-warn-app/cwa-app-android/pull/1418 please keep those in mind when updating this pr 👍

chris-cwa:
> i made some small changes to the risk transaction in #1418 please keep those in mind when updating this pr 👍

thank you!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,238 @@
...
+
+    companion object {
+        private var TAG = DefaultRiskLevels::class.simpleName
+        private const val DECIMAL_MULTIPLIER = 100
+    }
+
+    override fun updateRepository(riskLevel: RiskLevel, time: Long) {
+        val rollbackItems = mutableListOf<RollbackItem>()
+        try {
+            Timber.tag(TAG).v("update the risk level with $riskLevel")
```
d4rken, line 36: As a general thing: I'd really like it if log statements start with a capital letter, their are sentences after all.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,238 @@
...
+            }
+
+            // risk level calculation date update
+            val lastCalculatedRiskLevelDate = LocalData.lastTimeRiskLevelCalculation()
+            LocalData.lastTimeRiskLevelCalculation(time)
+            rollbackItems.add {
+                LocalData.lastTimeRiskLevelCalculation(lastCalculatedRiskLevelDate)
+            }
+        } catch (error: Exception) {
+            Timber.tag(TAG).e(error)
```
d4rken, line 51: Please always log errors with an error message, e.g.  "Updating the RiskLevelRepository failed."

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,238 @@
...
+
+            try {
+                Timber.tag(TAG).d("Initiate Rollback")
+                for (rollbackItem: RollbackItem in rollbackItems) rollbackItem.invoke()
+            } catch (rollbackException: Exception) {
+                Timber.tag(TAG).e(rollbackException)
```
d4rken, line 57: Same here, e.g. "RiskLevelRepository rollback failed."

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,238 @@
...
+                )
+        /** we only return outdated risk level if the threshold is reached AND the active tracing time is above the
+        defined threshold because [UNKNOWN_RISK_INITIAL] overrules [UNKNOWN_RISK_OUTDATED_RESULTS] */
+        return timeSinceLastDiagnosisKeyFetchFromServer.millisecondsToHours() >
+            TimeVariables.getMaxStaleExposureRiskRange() && isActiveTracingTimeAboveThreshold()
+    }
+
+    override fun calculationNotPossibleBecauseNoKeys() =
+        (TimeVariables.getLastTimeDiagnosisKeysFromServerFetch() == null).also {
+            if (it)
```
d4rken, line 81: When formatting forces us to go multi-line, please add brackets to `if` statements.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,238 @@
...
+                    .v("no last time diagnosis keys from server fetch timestamp was found")
+        }
+
+    override suspend fun calculationNotPossibleBecauseTracingIsOff() =
+        // this applies if tracing is not activated
+        !InternalExposureNotificationClient.asyncIsEnabled()
+
+    override suspend fun isIncreasedRisk(): Boolean {
+        val lastExposureSummary = getNewExposureSummary()
+        // retrieve application configuration
```
d4rken, line 92: We only need comments for things that are not obvious.

We don't need to explain what `val appConfiguration = appConfigProvider.getAppConfig()` is doing, right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,238 @@
...
+    override fun calculationNotPossibleBecauseNoKeys() =
+        (TimeVariables.getLastTimeDiagnosisKeysFromServerFetch() == null).also {
+            if (it)
+                Timber.tag(TAG)
+                    .v("no last time diagnosis keys from server fetch timestamp was found")
+        }
+
+    override suspend fun calculationNotPossibleBecauseTracingIsOff() =
+        // this applies if tracing is not activated
+        !InternalExposureNotificationClient.asyncIsEnabled()
```
d4rken, line 88: We can get this without static access now, try `ENFClient.isTracingEnabled.first()`.

If possible use the injected ENFClient with `ENFClient.internalClient` so we can remove more singleton access. `InternalExposureNotificationClient` will be gone in the future.

chris-cwa:
> i made some small changes to the risk transaction in #1418 please keep those in mind when updating this pr 👍

#1418 has been merged to this branch.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
d4rken: The `context` needs to be annotated, not the class.
Because there are multiple contexts through out an Android app we want to tell dagger which specific one we want.
It's just a custom "tag" we created with the name "AppContext", which we apply in `AndroidModule.kt`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -56,15 +59,22 @@ class TracingRepository @Inject constructor(
             LocalData.lastTimeDiagnosisKeysFromServerFetch()
     }
 
-    // TODO shouldn't access this directly
-    val internalIsRefreshing = MutableStateFlow(false)
+    private val retrievingDiagnosisKeys = MutableStateFlow(false)
+    private val internalIsRefreshing =
+        retrievingDiagnosisKeys.combine(AppInjector.component.taskController.tasks) { retrievingDiagnosisKeys, tasks ->
```
d4rken, line 64: Please don't use static access if the the whole class is already injected, you can get the taskController via constructor injection here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -77,17 +87,15 @@ class TracingRepository @Inject constructor(
...
     fun refreshDiagnosisKeys() {
         scope.launch {
-            internalIsRefreshing.value = true
+            retrievingDiagnosisKeys.value = true
             try {
                 RetrieveDiagnosisKeysTransaction.start()
-                RiskLevelTransaction.start()
-            } catch (e: TransactionException) {
-                e.cause?.report(ExceptionCategory.EXPOSURENOTIFICATION)
+                AppInjector.component.taskController.submit(DefaultTaskRequest(RiskLevelTask::class))
```
d4rken, line 93: Constructor injection.

d4rken (changes requested):

d4rken (changes requested):
Got this Exception on launch:

```kotlin
E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-3 @coroutine#80
    Process: de.rki.coronawarnapp.test, PID: 21703
    de.rki.coronawarnapp.task.MissingTaskFactoryException: No task factory mapped for class de.rki.coronawarnapp.task.common.DefaultTaskRequest (Kotlin reflection is not available)
        at de.rki.coronawarnapp.task.TaskController.submit(TaskController.kt:68)
        at de.rki.coronawarnapp.storage.TracingRepository$refreshRiskLevel$1.invokeSuspend(TracingRepository.kt:168)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ExposureStateUpdateWorker.kt
```diff
@@ -7,39 +7,33 @@ import com.google.android.gms.common.api.ApiException
...
-
-            RiskLevelTransaction.start()
+            InternalExposureNotificationClient
+                .asyncGetExposureSummary(token).also {
+                    ExposureSummaryRepository.getExposureSummaryRepository()
+                        .insertExposureSummaryEntity(it)
+                    Timber.v("exposure summary state updated: $it")
+                }
+
+            AppInjector.component.taskController.submit(DefaultTaskRequest(RiskLevelTask::class))
```
kolyaopahle, line 33: Couldn't we constructor inject this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+            attenuationParameters,
+            lastExposureSummary
+        ).also {
+            Timber.tag(TAG).v("Calculated risk with the given config: $it")
+        }
+
+        // get the high risk score class
+        val highRiskScoreClass =
+            riskScoreClassification.riskClassesList.find { it.label == "HIGH" }
+                ?: throw RiskLevelCalculationException(IllegalStateException("No high risk score class found"))
```
kolyaopahle, line 112: is there a more robust way to handle this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+
+        return false
+    }
+
+    override fun isActiveTracingTimeAboveThreshold(): Boolean {
+        val durationTracingIsActive = TimeVariables.getTimeActiveTracingDuration()
+        val durationTracingIsActiveThreshold =
+            TimeVariables.getMinActivatedTracingTime().toLong()
+
+        val activeTracingDurationInHours = durationTracingIsActive.millisecondsToHours()
```
kolyaopahle, line 138: i would maybe reorder these so that both variables are worked on in order before being compared. This is only personal preference though.

```suggestion
        val durationTracingIsActive = TimeVariables.getTimeActiveTracingDuration()
        val activeTracingDurationInHours = durationTracingIsActive.millisecondsToHours()
        
        val durationTracingIsActiveThreshold =
            TimeVariables.getMinActivatedTracingTime().toLong()
```

harambasicluka (commented):
Looks really nice and is easy to read.
I just would like to get clarification on the two comments :)

I tested it on an Samsung Galaxy S8 and on on OnePlus 8.
Look very good, but I just tested the manual mode and the riskscore is calculated when you navigate from one screen to another. In my opinion this shouldn't happen, I will retest it with the latest RC to see if that's caused by this PR or it's already in.
-> ok it's at least in 1.6.0/RC5 I'll create a bug ticket for this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+                        "Time since last exposure calculation is null"
+                    )
+                )
+        /** we only return outdated risk level if the threshold is reached AND the active tracing time is above the
+        defined threshold because [UNKNOWN_RISK_INITIAL] overrules [UNKNOWN_RISK_OUTDATED_RESULTS] */
+        return timeSinceLastDiagnosisKeyFetchFromServer.millisecondsToHours() >
+            TimeVariables.getMaxStaleExposureRiskRange() && isActiveTracingTimeAboveThreshold()
+    }
+
+    override fun calculationNotPossibleBecauseNoKeys() =
```
SamuraiKek, line 77: I would rename this function to calculationNotPossibleBecauseOfNoKeys, to keep the naming consistent with the function above.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevels.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    fun calculationNotPossibleBecauseOfOutdatedResults(): Boolean
+
+    /**
+     * true if threshold is reached / if the duration of the activated tracing time is above the
+     * defined value
+     */
+    fun isActiveTracingTimeAboveThreshold(): Boolean
+
+    suspend fun isIncreasedRisk(): Boolean
```
d4rken, line 18: What do you think about refactoring this such that the `DefaultRiskLevels` implementation is not responsible for both retrieving the data and working on it?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+
+    override suspend fun calculationNotPossibleBecauseTracingIsOff() =
+        // this applies if tracing is not activated
+        !enfClient.isTracingEnabled.first()
+
+    override suspend fun isIncreasedRisk(): Boolean {
+        val lastExposureSummary = getNewExposureSummary()
+        val appConfiguration = appConfigProvider.getAppConfig()
+        Timber.tag(TAG).v("Retrieved configuration from backend")
+        // custom attenuation parameters to weight the attenuation
```
SamuraiKek, line 93: Weigh, not weight.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevels.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    suspend fun calculationNotPossibleBecauseTracingIsOff(): Boolean
```
d4rken, line 20: Moving things like early abort checks (e.g. `calculationNotPossibleBecauseTracingIsOff`) out of this interface? Into the task?

I think it's already A LOT better than before, but I have still trouble following the code path through the different steps. In my head I always try to imagine how the data is passed from one step to the next :thinking:.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -70,6 +78,10 @@ class TracingRepository @Inject constructor(
         }
     }
 
+    private fun List<TaskInfo>.isRiskLevelTaskRunning() = find {
```
d4rken, line 81: I think there is a method called `List<T>.any { }` which returns true if an of the elements match the condition.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/risk/RiskLevelsTest.kt
```diff
@@ -1,16 +1,48 @@
...
 import de.rki.coronawarnapp.server.protocols.internal.AttenuationDurationOuterClass
+import io.kotest.matchers.shouldBe
+import io.mockk.MockKAnnotations
+import io.mockk.impl.annotations.MockK
 import junit.framework.TestCase.assertEquals
+import org.junit.Before
 import org.junit.Test
 
-class RiskLevelCalculationTest {
+class RiskLevelsTest {
```
d4rken, line 14: Please extend `BaseTest` which automatically set's up logging so that we can see log output if the classes produce any.

d4rken (commented):
Love it so far, thinking about whether it worth it to put some things into their own class. Maybe we can have a short to discuss some ideas, brainstorm a bit.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ExposureStateUpdateWorker.kt
d4rken: Can't yet, will make this possible soon.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+            InternalExposureNotificationClient.asyncGetExposureSummary(googleToken)
+
+        return exposureSummary.also {
+            Timber.tag(TAG)
+                .v("Generated new exposure summary with $googleToken")
+        }
+    }
+
+    companion object {
+        private var TAG = DefaultRiskLevels::class.simpleName
```
BMItter, line 246: TAG should be final (val) I think..
I also like private val TAG = DefaultRiskLevels::class.java.simpleName very much, because you get a String instead of String?, but please, don't change it if you don't want (just a personal note)

BMItter:
It's really a pleasure not to see the old Transactions anymore! Great Work! :heart:

chris-cwa:
> It's really a pleasure not to see the old Transactions anymore! Great Work! ❤️

Thanks! 😃

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ExposureStateUpdateWorker.kt
d4rken: Will be made possible by #1503

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
chris-cwa: What do you mean by 'robust'? An exception has to be thrown in order to abort the task. All exceptions are caught by the task controller.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+        if (withinDefinedLevelThreshold(
+                riskScore,
+                highRiskScoreClass.min,
+                highRiskScoreClass.max
+            )
+        ) {
+            Timber.tag(TAG)
+                .v("$riskScore is above the defined min value ${highRiskScoreClass.min}")
+            return true
+        } else if (riskScore > highRiskScoreClass.max) {
```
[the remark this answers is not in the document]
chris-cwa: it was that day before, I tried to no change logic. I guess invalid values should not lead to a result

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -0,0 +1,249 @@
...
+     */
+    @VisibleForTesting
+    internal fun updateRiskLevelScore(riskLevel: RiskLevel) {
+        val lastCalculatedScore = RiskLevelRepository.getLastCalculatedScore()
+        if (RiskLevel.riskLevelChangedBetweenLowAndHigh(
+                lastCalculatedScore,
+                riskLevel
+            ) && !LocalData.submissionWasSuccessful()
+        ) {
+            NotificationHelper.sendNotification(
```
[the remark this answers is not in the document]
chris-cwa: This also was not changed to the way it behaved before. I you want I can change it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
chris-cwa: Should we then also rename the field `attenuationParameters.weights`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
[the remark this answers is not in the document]
harambasicluka: It was just an idea, I wouldn't insist on that :)

harambasicluka (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: the exception is fine, i was just thinking about the whole process of resolving the highest risk class by comparing a list of instance properties to a string.

kolyaopahle: kind of seems like this might fail if the classes are changed and we forget to update here, there is no real hard coupling that enforces compile time correctness

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
SamuraiKek: I don't think so. Weight is used for defining the "heaviness" of something, and the verb "to weigh" is used to find the weight of something. The attenuationParameters indeed have weights. Sorry I don't mean to sound like the grammar police :D

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,166 @@
...
+            // If there is no connectivity the transaction will set the last calculated
+            // risk level
+            if (!isNetworkEnabled(context)) {
+                RiskLevelRepository.setLastCalculatedRiskLevelAsCurrent()
+                return Result(UNDETERMINED)
+            }
+            with(riskLevels) {
+                return Result(
+                    when {
+                        !enfClient.isTracingEnabled.first().also {
```
d4rken, line 59: Can we turn this into an early abort condition like the above pattern used with `isNetworkEnabled`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,166 @@
...
+                        } -> NO_CALCULATION_POSSIBLE_TRACING_OFF
+
+                        calculationNotPossibleBecauseOfNoKeys().also {
+                            checkCancel()
+                        } -> UNKNOWN_RISK_INITIAL
+
+                        calculationNotPossibleBecauseOfOutdatedResults().also {
+                            checkCancel()
+                        } -> if (backgroundJobsEnabled)
+                            UNKNOWN_RISK_OUTDATED_RESULTS
```
d4rken, line 70: No multiline `if` without braces please.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,166 @@
...
+    override suspend fun run(arguments: Task.Arguments): Result {
+        try {
+            Timber.d("Running with arguments=%s", arguments)
+            // If there is no connectivity the transaction will set the last calculated
+            // risk level
+            if (!isNetworkEnabled(context)) {
+                RiskLevelRepository.setLastCalculatedRiskLevelAsCurrent()
+                return Result(UNDETERMINED)
+            }
+            with(riskLevels) {
```
d4rken, line 56: What about removing one level of indentation here?

```kotlin

return when {
....
}
.let { Result(it) }
.also { ... }


```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,166 @@
...
+                        } -> INCREASED_RISK
+
+                        !isActiveTracingTimeAboveThreshold().also {
+                            checkCancel()
+                        } -> UNKNOWN_RISK_INITIAL
+
+                        else -> LOW_LEVEL_RISK
+                    }.also {
+                        checkCancel()
+                        updateRepository(it, System.currentTimeMillis())
```
d4rken, line 85: Please inject and use `TimeStamper` to improve testability.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -0,0 +1,166 @@
...
+                .v("Generated new exposure summary with $googleToken")
+        }
+    }
+
+    private fun checkCancel() {
+        if (isCanceled) throw TaskCancellationException()
+    }
+
+    private val backgroundJobsEnabled: Boolean
+        get() = ConnectivityHelper.autoModeEnabled(CoronaWarnApplication.getAppContext()).also {
```
d4rken, line 123: We have a non static variant available for that, see `BackgroundModeStatus`, Could we use that here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: i'll resolve this as we discussed a fix for it internally

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
chris-cwa: I didn't know that! Thank you so much!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
chris-cwa: good idea

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
chris-cwa: I believe it reduces the good overview that you quickly get in the current state by just looking at the code.

harambasicluka: Agree with @chris-cwa currently it's just nice to read.

AlexanderAlferov:
Tested. Looks good.
But I confirm Luka's comment regarding manual mode and risk auto-update when navigating from one screen to another.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ExposureStateUpdateWorker.kt
```diff
@@ -9,42 +9,36 @@ import com.squareup.inject.assisted.Assisted
...
```
d4rken, line 38: Inject the taskController

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -162,16 +172,9 @@ class TracingRepository @Inject constructor(
...
 
-            // refresh the risk level
-            try {
-                RiskLevelTransaction.start()
-            } catch (e: TransactionException) {
-                e.cause?.report(ExceptionCategory.INTERNAL)
-            } catch (e: Exception) {
-                e.report(ExceptionCategory.INTERNAL)
-            }
+            AppInjector.component.taskController.submit(DefaultTaskRequest(RiskLevelTask::class))
```
d4rken, line 175: Injecet the taskController.

d4rken (dismissed):
Two cases where we can inject instead of static access, but otherwise looks good.

Successfully tested on a Pixel 3 @ Android 11

d4rken (approved):
lgtm

BMItr (approved):

ralfgehrer:
Successfully tested on Pixel (Android 10) and Pixel 5 (Android 11)

PR #1418, opened 2020-10-16
Homescreen Notifications, even though keys were submitted (EXPOSUREAPP-3181, EXPOSUREAPP-3182)

kolyaopahle:
## Description
This PR adds a few additional checks to the test result fetching and risk level calculation tasks to check for prior test submission and stops the notification when tests have been submitted already.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -615,6 +615,10 @@ object LocalData {
         )
     }
 
+    fun submissionWasSuccessful(): Boolean {
```
ralfgehrer, line 618: val submissionWasSuccessful = submissionWasSuccessful

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
```diff
@@ -89,13 +89,18 @@ class DiagnosisTestResultRetrievalPeriodicWorker(
      * @see TestResult
      */
     private fun initiateNotification(testResult: TestResult) {
+        if (LocalData.isTestResultNotificationSent() || LocalData.submissionWasSuccessful()) {
```
ralfgehrer, line 92: return if ...

kolyaopahle: afaik this is not possible as return if always requires an else branch

d4rken (approved):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
kolyaopahle: fixed :)

harambasicluka (approved):

PR #1429, opened 2020-10-19
Converted submission transaction to task (EXPOSUREAPP-2482)

chris-cwa:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionTask.kt
```diff
@@ -0,0 +1,114 @@
...
+    ) : TaskFactory<DefaultProgress, Task.Result> {
+
+        override val config: TaskFactory.Config = Config()
+        override val taskProvider: () -> Task<DefaultProgress, Task.Result> = {
+            taskByDagger.get()
+        }
+    }
+
+    companion object {
+        private val SUBMISSION_TASK_TIMEOUT = Duration.standardMinutes(8)
```
d4rken, line 110: Can we inline this?

d4rken (commented):
Few readability improvements.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionTask.kt
```diff
@@ -0,0 +1,114 @@
...
+                    registrationToken,
+                    getHistory(),
+                    true,
+                    supportedCountries()
+                ).also {
+                    checkCancel()
+                }
+            )
+        }
+        SubmissionService.submissionSuccessful()
```
kolyaopahle, line 46: As a note for further refactoring, i really don't like the `SubmissionService` in its current state. Its just a agglomeration of grown over time functions with no real central goal. Id really like to spend some time here but obviously not in this PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionTask.kt
kolyaopahle: I agree, its only used in one place and the variable that its assigned to is already descriptive enough in my opinion

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionTask.kt
```diff
@@ -0,0 +1,114 @@
...
+                toNewConfig {
+                    addSupportedCountries(FALLBACK_COUNTRY)
+                }.also {
+                    Timber.w("Country list was empty, corrected: %s", it.supportedCountriesList)
+                }
+            } else {
+                this
+            }.supportedCountriesList
+        }
+    }
```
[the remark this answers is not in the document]
chris-cwa: simplified this part

SamuraiKek (dismissed):
Lgtm. Code is readable and easy to follow.

BMItter:
I just tested the submission flow. Negativ/Positive tests were all right, but after putting in symtoms (Keine Angaben) I got a missingTaskFactoryException (No Task factory mapped for class). I will send you the report internally.

BMItr (approved):
lgtm :muscle:, above mentioned exception is gone, passed submission tests.

SamuraiKek (approved):

PR #1460, opened 2020-10-23
Delta onboarding: legal strings back to normal strings (EXPOSUREAPP-3078)

harambasicluka:
Just moved legal strings back to default strings, that they can go through the normal translation process.

delta-onboarding-strings


ralfgehrer (approved):

SabineLoss (approved):
UA reviewed

Issue #948, opened 2020-07-30
CWA risk update timestamp different to Google timestamp (by one or two minutes)


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Describe the bug

The last update time shown by CWA is sometimes one minute or two minutes earlier or later than the update time shown in the Google ENS UI.

## Expected behaviour

This is a minor issue.
The timestamp shown by CWA for the last update should be the same as the timestamp shown in the Google ENS UI in the overview screen and in the exposure checks log.

## Steps to reproduce the issue

1. View CWA risk status
Note text similar to "Updated: Today, 07:54"
2. Settings > Google > COVID-19 Exposure Notifications > three-dot icon > Exposure checks
Note latest timestamp in exposure checks list.

## Technical details

- Mobile device: Samsung Galaxy A5 (2017) SM-A520F
- Android version: 8.0.0
- CWA Version: ~~1.1.1~~ 2.4.2
- Google Exposure Notifications System Version: ~~15202902003~~ v1.8 (212116000)

Edit: Updated steps to reproduce for newer ENS UI, updated software versions and verified that the issue still occasionally occurs in current version of CWA and ENS.
 
---
Internal Tracking ID: [EXPOSUREAPP-1991](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-1991)

GPclips:
Hello @MikeMcC399,

thanks for pointing this out. I will take this into the discussion with our development team.

Best regards,
LMM

Corona-Warn-App Open Source Team

GPclips:
Hello @MikeMcC399 ,

thanks again for your investigation. Our developers have been informed. This bug probably occurs due to different rounding strategies and will be inspected as soon as possible.

Best regards,
LMM

Corona-Warn-App Open Source Team

GPclips:
https://jira.itc.sap.com/browse/EXPOSUREAPP-1991

MikeMcC399:
I noticed the Google Exposure Notification Framework (Version 15202902003) showing a time one minute into the future.
The text was "Last checked for potential exposure today at 19:49" whilst the system clock was showing the time as 19:48.
[...]

PR #1473, opened 2020-10-25
Monitor ENF calculation more closely for better progress feedback (EXPOSUREAPP-2743)

d4rken:
## Why
We want to display better progress feedback, we want to know when the download is happening and when we are waiting for the ENF to check the keys.

The goal is to show two different progress types, this PR does not yet add that, but adds the necessary base classes. For now the `isRefreshing` state will be displayed on the risk card if either the transaction is on-going, or the ENF is still calculating.

## How
This PR adds the `CalculationTracker`, which is a set of classes that can track the ENF calcluations that happen outside of our app process.
The idea is:
* Every time we submit submit key files to the ENF, we store an identifier (the token we also pass to the ENF).
* Register for both `ACTION_EXPOSURE_STATE_UPDATED` and `ACTION_EXPOSURE_NOT_FOUND`
* When we receive a broadcast for either action, the token is included returned to us by the ENF and we mark the respective entry as finished.

To the rest of the app we expose a `CalculationTracker.calculations: Flow<Map<String, Calculation>>`:
* We store the last 5 started calculations.
* If any of those does not have a `finishedAt` timestamp set by us, it's still calculating.
* We consider the calculation still in progress as long as there is at least one unfinished calculation.
* The data should be accessed via `ENFClient` which offers easier access with
    * `isCurrentlyCalculating(): Flow<Boolean>`
    * `latestFinishedCalculation(): Flow<Calculation?>`

To notice any issues within the ENF that don't throw us an exception, there is a timeout checker running within the calculation tracker. 
Every 5 minutes it will evaluate the `startedAt` timestamps of all known calculations. If it exceeds the timeout (currently 60 minutes,  @mlenkeit thoughts on that?, might be a bit high), the calculation is considered "timed out", get the result type "TIMEOUT", and given the current time as `finishedAt` timestamp (which marks it as no longer "in calculation").

The calculation data is persisted everytime there is an update to the data map. The class `CalculationTrackerStorage` loads and stores the map as serialized JSON file to our private app storage.

## Testing
* Observe the log for `CalculationTracker`
* Start risk level calculations and observe the log and the progress state of the risk cards.


d4rken:
In another PR we should consider replacing the timestamp on the risk card, which currently reflects the last time the download was finished, but it makes more sense if it's the timestamp of the newest tracked calculation (which would IMHO be the fix for #948 .

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
```diff
@@ -0,0 +1,68 @@
...
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.mapNotNull
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.flow.onStart
+import kotlinx.coroutines.flow.shareIn
+import kotlinx.coroutines.plus
+import timber.log.Timber
+import kotlin.coroutines.CoroutineContext
+
+class HotDataFlow<T : Any>(
```
d4rken, line 20: This PR also introduces a nice helper class dubbed `HotDataFlow`.

It's a flowable that
* Does lazy init. The start value is provided via a suspending function that is only executed if someone subscribes
* Shares all updates between all subscribers as long as there is at least one subscriber.
* Allows threadsafe updates to it's contents as it only allows its data to be changed by giving it an "update function" that takes the current value and can return a new one.

d4rken:
I've changed the `ENF.isCalculating` to only check the state of the newest submission. With batched submission there is only a small chance for overlapping calculations and it reduces the chance that we actually run into a timeout and the user sees a prolonged "progress" state. As we currently submit all key files each time, each new calculation would superseed the previous anyways.

vaubaehn:
You [invited for feedback](https://github.com/corona-warn-app/cwa-app-android/issues/948#issuecomment-715325151). First class. Nothing else to say. 👍

mlenkeit:
@d4rken as discussed, let's set the timeout to 15 minutes initially, should already be quite conservative.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/calculationtracker/CalculationTrackerStorage.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val mutex = Mutex()
+    private val storageDir = File(context.filesDir, "calcuation_tracker")
+    private val storageFile = File(storageDir, "calculations.json")
+    private var lastCalcuationData: Map<String, Calculation>? = null
+
+    private val gson by lazy {
+        if (storageDir.mkdirs()) {
+            Timber.v("Created %s", storageDir)
+        }
+        GsonBuilder().create()
```
ralfgehrer, line 28: I'm surprised, not injected :) ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
ralfgehrer: Turned out very nicely.

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/calculationtracker/CalculationTrackerStorage.kt
d4rken: :fearful:

ralfgehrer (approved):

BMItter:
passed deep look & test. lgtm

BMItr (approved):

PR #1480, opened 2020-10-26
Added logic for showing notification on risk level change (EXPOSUREAPP-2569)

SamuraiKek:
### Description

This PR introduces a new dialog which informs the user what to do when the risk level lowers from red to green.
### Steps to reproduce

Due to the nature of how the risk is calculated, there is currently no easy way of testing this.
Prerequisites include:
- Having two android phones, staying close to each other, both having ENF active.
- One of which needs to get a positive result.
- From the phone with the positive result, submit the keys to the server.
- Once the keys from the first phone have been downloaded and the risk calculated, wait for the risk status on the second phone to change to high, 
- Reset the risk level on the second phone by going into the testers menu.
- With the risk level reset, the notification should pop up


Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -72,6 +72,8 @@
     <string name="preference_interoperability_selected_country_codes"><xliff:g id="preference">"preference_interoperability_selected_country_codes"</xliff:g></string>
     <!-- NOTR -->
     <string name="preference_interoperability_all_countries_selected">preference_interoperability_all_countries_selected</string>
+    <!-- NOTR -->
+    <string name="preference_has_risk_status_lowered">preference_has_risk_status_lowered</string>
```
d4rken, line 76: Don't use strings.xml for preference keys add a static string variable to `LocalData`

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -371,6 +371,19 @@
...
     <string name="risk_details_deadman_notification_body">"Bitte öffnen Sie die Corona-Warn-App, um die Risiko-Ermittlung zu aktualisieren."</string>
 
+    <!-- ####################################
+               Risk Lowered Dialog
+    ###################################### -->
+
+    <!-- XHED: risk lowered - dialog headline -->
+    <string name="risk_lowered_dialog_headline">"Änderung Ihres Risikostatus"</string>
+    <!-- XTXT: risk lowered - dialog content -->
+    <string name="risk_lowered_dialog_body">"Die Risiko-Begegnung, die zu einem erhöhten Risiko für Sie geführt hat, liegt mehr als 14 Tage zurück. Daher wird das Infektionsrisiko wieder als niedrig für Sie eingestuft. \n So verhalten Sie sich richtig:\n 1. Wenn Sie keine Symptome von COVID-19 aufweisen, halten Sie sich an die allgemein geltenden Verhaltensregeln zu Abstand und Hygiene. \n 2f. Wenn Sie Symptome einer Infektion mit COVID-19 aufweisen, empfehlen wir, dass Sie Ihren Arzt aufsuchen und sich testen lassen."</string>
```
d4rken, line 381: ` \n ` Spaces after and before newlines?

`2f`?

Sure that the string is correct?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -177,4 +180,25 @@ class HomeFragment : Fragment(R.layout.fragment_home), AutoInject {
             setOnClickListener { homeMenu.showMenuFor(it) }
         }
     }
+
+    private fun showRiskLevelLoweredDialogIfNeeded() {
```
d4rken, line 184: The viewmodel should decide whether the dialog needs to be shown, preferably via `SingleLiveEvent`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -357,6 +357,34 @@ object LocalData {
             )
         }
 
+    /**
+     * Gets a boolean that tells weather the risk level
```
d4rken, line 361: "whether", not :partly_sunny:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -357,6 +357,34 @@ object LocalData {
...
+     * decreased or not.
+     *
+     * @return boolean if risk status lowers from red to green
+     */
+    fun hasRiskStatusLowered(): Boolean {
```
d4rken, line 366: Since this is only used for the dialog, let name this `isUserToBeNotifiedOfLoweredRiskLevel` so no one accidentally uses this data for something else.

d4rken (changes requested):

Corona-Warn-App/src/main/res/values-de/strings.xml
janetback, line 381: Just checking - is '\n 2f.' ok here?

janetback (changes requested):

Corona-Warn-App/src/main/res/values-de/strings.xml
SamuraiKek: No, it's not. It's a typo. I'm getting it fixed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
SamuraiKek: I always mix these two up.  It's fixed now

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
SamuraiKek: Yeah, you're right,  I've changed it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
SamuraiKek: Ok, the showing of the notification is now handled by the viewmodel.

Corona-Warn-App/src/main/res/values-de/strings.xml
SamuraiKek: I've removed the spaces and the typo.

Corona-Warn-App/src/main/res/values/strings.xml
SamuraiKek: Ok, I've implemented a static string in LocalData and removed the string from strings.xml

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
d4rken: The logic and decisions are still being made in the Fragment and not the ViewModel, it would also be good to not require a manual refresh call. Maybe we can have call.

janetback (dismissed):
UA approved

d4rken (dismissed):
lgtm

harambasicluka (dismissed):

janetback (dismissed):
UA approved

d4rken (dismissed):

d4rken (approved):
lgtm if tests pass

harambasicluka (approved):

PR #1489, opened 2020-10-27
Improve logging and data emission (DEV)

d4rken:
Small change that cleans up the code for timeout code in the CalculationTracker, and a change to `HotDataFlow` to only emit new values if they actually changed.


harambasicluka (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
```diff
@@ -51,6 +52,7 @@ class HotDataFlow<T : Any>(
     }
 
     val data: Flow<T> = internalFlow
+        .distinctUntilChanged()
```
ralfgehrer, line 55: Shouldn't this class be more general-purpose? Adding this limits the use cases.

d4rken: Which case would no longer work though?

I think this "restrictions" fits very well here as the only way to threadsafe get data and update it, is via the `updateSafely` method which requires you to return a result. If you don't want to change anything you would return the same object you got, and without `distinctUntilChanged` trigger an emission.

New subscribers get the latest value, and I don't see a usecase where current subscribers would want an emission if nothing changes. (At least when I implemented this pattern in other apps, i always want no emissions if nothing changed, and never required the reverse case.

ralfgehrer (approved):

PR #1492, opened 2020-10-28
Grant one more encryption reset attempt for users on v1.6.0 (EXPOSUREAPP-3313)

d4rken:
The mitigation for the encryption issue is already in place but we are still trying to help those users that are just updating the app in the background automatically hoping for a solution to their error when opening the app.

Either the mitigation works or it doesn't so resetting the data should only happens once.
If the mitigation does not work and the issue reappears we don't want to keep resetting the data.

@vaubaehn noticed that users who upgraded to 1.5.0 before 1.5.1 was available, and who were affected by the encryption issue, the data reset attempt was already consumed without helping those users, due to a bug in the previous implementation, and it should be retried.

This PR adds a new stored [boolean value](https://github.com/corona-warn-app/cwa-app-android/compare/release/1.6.x...fix/encryption-reset-per-version?expand=1#diff-069e5febb8fd6d0280bfacf45d686c5799dd786629f330fd8126e61532e6ba4bR138), ignoring the previous one, such that for everyone affected by the issue, the recovery through data reset will be attempted once again if the error case matches.

## Testing
Sadly this can't be easily tested again and manually editing the code to throw exceptions, and reviewing code + unit-tests is required.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/security/EncryptionErrorResetTool.kt
@@ -4,6 +4,7 @@ import android.content.Context
 import android.content.SharedPreferences
 import androidx.core.content.edit
 import de.rki.coronawarnapp.storage.DATABASE_NAME
+import de.rki.coronawarnapp.util.TimeStamper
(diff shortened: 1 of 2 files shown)


ralfgehrer (approved):

chris-cwa (approved):
ok

PR #1495, opened 2020-10-28
1.5.x into 1.6.x

d4rken:


AlexanderAlferov (dismissed):

BMItr (approved):
Tested & lgtm.
PS: we should really get rid of the AlertDialog.Builder used in this way. But I ll adress this later, as it is unrelated here.

ralfgehrer (approved):

PR #1503, opened 2020-10-29
Dependency injection for workers (DEV)

d4rken:
This adds the necessary boilerplate code to use dependency injection in `WorkManager` workers.

The principles are very similar to how viewmodels are injected.
* For each worker we have a factory
* All of our worker factories are mapped by their class key into a map (`Map<WorkerClass, WorkerClass.Factory>`).
* We have a custom "WorkerFactoryFactory" that we give to `WorkManager` when our app is instantiated.
* Whenever a new worker needs to be created, `WorkManager` will look up the class in the app and use the provided factory to create it.
* The `WorkManager` always provides the same arguments to each worker (`Context` and `WorkerParameters`), and we use `AssistedInject` to combine them with what ever we need injected from our side.

Internally `AssistedInjected` generates something like

```kotlin
// The class is generated by dagger, so from the DI graph we get the Playbook
class BackgroundNoiseOneTimeWorker_AssistedFactory @Inject constructor(
    private val playbook: Provider<Playbook>
) : BackgroundNoiseOneTimeWorker.Factory {

   // The create method is called by the worker factory to provide the manual parameters that are not part of our DI graph
    override fun create(
        context: Context,
        workerParams: WorkerParameters
    ): BackgroundNoiseOneTimeWorker {
        return BackgroundNoiseOneTimeWorker(
            context,
            workerParams,
            playbook.get()
        )
    }
}

```

## Testing
* Check that I didn't forget a worker (see [here](https://github.com/corona-warn-app/cwa-app-android/compare/release/1.7.x...feature/dependency-injection-for-workers?expand=1#diff-ec171db19783c398709b85ce3ac5979c793c89e7556b47aad5cfe6857e03cb65))
* Do a little smoke test
* Perform a QR Code or TAN submission


Corona-Warn-App/build.gradle
```diff
@@ -329,6 +329,7 @@ dependencies {
     implementation 'com.google.dagger:dagger-android-support:2.28.1'
     kapt 'com.google.dagger:dagger-compiler:2.28.1'
     kapt 'com.google.dagger:dagger-android-processor:2.28.1'
+    kaptTest 'com.google.dagger:dagger-compiler:2.28.1'
```
d4rken, line 332: So dagger generates code for classes that only exist in test packages.

Corona-Warn-App/build.gradle
```diff
@@ -366,6 +367,8 @@ dependencies {
     androidTestImplementation "io.kotest:kotest-assertions-core-jvm:4.3.0"
     androidTestImplementation "io.kotest:kotest-property-jvm:4.3.0"
 
+    testImplementation "io.github.classgraph:classgraph:4.8.90"
```
d4rken, line 370: Allows us to scan our own code in tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/worker/WorkerBinder.kt
```diff
@@ -0,0 +1,58 @@
...
+import dagger.multibindings.IntoMap
+import de.rki.coronawarnapp.nearby.ExposureStateUpdateWorker
+import de.rki.coronawarnapp.worker.BackgroundNoiseOneTimeWorker
+import de.rki.coronawarnapp.worker.BackgroundNoisePeriodicWorker
+import de.rki.coronawarnapp.worker.DiagnosisKeyRetrievalOneTimeWorker
+import de.rki.coronawarnapp.worker.DiagnosisKeyRetrievalPeriodicWorker
+import de.rki.coronawarnapp.worker.DiagnosisTestResultRetrievalPeriodicWorker
+
+@Module
+abstract class WorkerBinder {
```
d4rken, line 15: Similar to our viewmodels we add a map of all worker-workerfactory combinations to our DI graph

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/BackgroundNoiseOneTimeWorker.kt
```diff
@@ -41,4 +40,7 @@ class BackgroundNoiseOneTimeWorker(
 
         return result
     }
+
+    @AssistedInject.Factory
+    interface Factory : InjectedWorkerFactory<BackgroundNoiseOneTimeWorker>
```
d4rken, line 45: Every worker needs a factory. The `InjectedWorkerFactory` just hides the boilerplate code:

```
interface Factory {
    fun create(context: Context, workerParams: WorkerParameters)
}
```

d4rken: Note: There is one peculiarity with `AssistedInject`, the parameter names of the `@AssistedInject.Factory` need to be written exactly the same as the constructor arguments annotated with `@Assisted`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/worker/WorkerBinderTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        val ourWorkerClasses = scanResult
+            .getSubclasses("androidx.work.ListenableWorker")
+            .filterNot { it.name.startsWith("androidx.work") }
+
+        Timber.v("Our project contains %d worker classes.", ourWorkerClasses.size)
+        ourWorkerClasses.forEach { Timber.v("Existing: ${it.name}") }
+
+        val boundFactories = factories.keys.map { it.name }
+        val existingFactories = ourWorkerClasses.map { it.name }
+        boundFactories shouldContainAll existingFactories
```
d4rken, line 55: Now isn't this some voodoo @chris-cwa  :grin: 

While it's not super awesome, or is it? :smirk: 

It will prevent anyone from accidentally not a worker factory to the DI graph and not noticing it because it's only rarely launched.

chris-cwa: This is so beautiful! What's it doing later?

chris-cwa (approved):
Another urhahn symphony.

Corona-Warn-App/build.gradle
chris-cwa: very helpful hints in this PR - why not put them as comments to the code for future generations to also be able to see it?

AlexanderAlferov (approved):
Looks great

PR #1513, opened 2020-11-01
 Deadman Notification if exposure check was not successful for 36h (EXPOSUREAPP-2332)

AlexanderAlferov:
**Description**
Schedule a notification in 36 hours after successful risk calculation.
This is "last try" notification: if after this notification application would not be able to calculate risk successfully, then it would not show this notification anymore.

**Steps to reproduce**
Risk calculation should fail and never succeed in 36 hours (either in foreground or background).
1) Wait until risk calculation finished
2) Go to settings -> Date & time
3) Add two days to current date
4) Deadman notification should appear

Remember to get back to your time line! 🕓  🔙   🚙💨


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationOneTimeWorker.kt
```diff
@@ -0,0 +1,49 @@
...
+ */
+class DeadmanNotificationOneTimeWorker @AssistedInject constructor(
+    @Assisted val context: Context,
+    @Assisted workerParams: WorkerParameters,
+    private val sender: DeadmanNotificationSender
+) :
+    CoroutineWorker(context, workerParams) {
+
+    /**
+     * Work execution
```
d4rken, line 25: Comments are good, but should help the developer better understand the code, e.g. explain why certain decisions were made. Comments that say the same as a function name just add noise to the code when reading it.

I'd say this is a superfluous comment, what do you think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationPeriodicWorker.kt
```diff
@@ -0,0 +1,51 @@
...
+ */
+class DeadmanNotificationPeriodicWorker @AssistedInject constructor(
+    @Assisted val context: Context,
+    @Assisted workerParams: WorkerParameters,
+    private val scheduler: DeadmanNotificationScheduler
+) :
+    CoroutineWorker(context, workerParams) {
+
+    /**
+     * Work execution
```
d4rken, line 25: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,85 @@
...
+import javax.inject.Inject
+import kotlin.random.Random
+
+@Reusable
+class DeadmanNotificationSender @Inject constructor(
+    @AppContext private val context: Context
+) {
+
+    /**
+     * Notification channel id
```
d4rken, line 24: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @see NotificationConstants.NOTIFICATION_CHANNEL_ID
+     */
+    private val channelId =
+        context.getString(NotificationConstants.NOTIFICATION_CHANNEL_ID)
+
+    /**
+     * Create pending intent to main activity
```
d4rken, line 32: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,85 @@
...
+        PendingIntent.getActivity(
+            context,
+            0,
+            Intent(context, MainActivity::class.java),
+            0
+        )
+
+    /**
+     * Build notification
+     * Create notification with defined title, content text and visibility.
```
d4rken, line 44: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,85 @@
...
+            .setContentIntent(createPendingIntentToMainActivity())
+            .setAutoCancel(true)
+            .setContentTitle(title)
+            .setContentText(content)
+
+        return builder.build()
+    }
+
+    /**
+     * Build and send notification with predefined title and content.
```
d4rken, line 71: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationWorkBuilder.kt
```diff
@@ -0,0 +1,45 @@
...
+import de.rki.coronawarnapp.worker.BackgroundConstants
+import org.joda.time.DateTimeConstants
+import java.util.concurrent.TimeUnit
+import javax.inject.Inject
+
+@Reusable
+class DeadmanNotificationWorkBuilder @Inject constructor() {
+
+    /**
+     * Build one time work
```
d4rken, line 18: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationWorkBuilder.kt
```diff
@@ -0,0 +1,45 @@
...
+            )
+            .setBackoffCriteria(
+                BackoffPolicy.EXPONENTIAL,
+                BackgroundConstants.BACKOFF_INITIAL_DELAY,
+                TimeUnit.MINUTES
+            )
+            .build()
+
+    /**
+     * Build periodic work
```
d4rken, line 34: Superfluous comment?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -88,6 +89,7 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
         scheduleWork()
         checkShouldDisplayBackgroundWarning()
         doBackgroundNoiseCheck()
+        scheduleDeadmanNotification()
```
d4rken, line 92: We don't really need an extra method if it just calls another one line method.

Do we need to do this in `onResume` though?

Could we use something like `ForegroundState.kt` to just execute it any time the app is moved from background to foreground?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/deadman/DeadmanNotificationOneTimeWorkerTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        sender = sender
+    )
+
+    @Test
+    fun `create worker`() {
+        createWorker()
+    }
+
+    @Test
+    fun `run worker success`() = runBlockingTest2(permanentJobs = true) {
```
d4rken, line 47: If there was no specific issue that required this variant `runBlockingTest2(permanentJobs = true)`, use `runBlockingTest` or just add `suspend` to the test method.

`runBlockingTest2` is a workaround for coroutine scopes where a job never suspends (and never should).

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSchedulerTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        workBuilder = workBuilder
+    )
+
+    @Test
+    fun `one time work was scheduled`() = runBlockingTest2(permanentJobs = true) {
+        coEvery { timeCalculation.getDelay() } returns 10L
+
+        createScheduler().scheduleOneTime()
+
+        verify(exactly = 1) { workManager.enqueueUniqueWork(
```
d4rken, line 70: FYI, there is a verify method called `verifySequence`, you can add multiple methods to it and it will check that those and only those were called.

d4rken (changes requested):
Need to think about whether we need to trigger this via `MainActivity.onResume` but otherwise looking good :+1:

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/deadman/DeadmanNotificationOneTimeWorkerTest.kt
d4rken: Edit to my previous comment, don't add `suspend` to test functions. It will not pick them up correctly.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
AlexanderAlferov: Removed extra method.
Right now it would require additional checks if we want to place this call in ForegroundState (onboarding is finished).
I would say it should be streamlined with other background workers if we want to switch to ForegroundState, mb as part of BackgroundWorkScheduler refactoring in future.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSchedulerTest.kt
AlexanderAlferov: Thank you for this note, using it now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationOneTimeWorker.kt
AlexanderAlferov: Agree, removed superfluous comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -105,6 +106,7 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
         scheduleWork()
         checkShouldDisplayBackgroundWarning()
         doBackgroundNoiseCheck()
+        deadmanScheduler.schedulePeriodic()
```
d4rken, line 109: Could this be done in the `CoronaWarnApplication`?

d4rken (changes requested):
Looking good, nice tests! Will test on device now.

harambasicluka (changes requested):
Tested it on day one, worked like expected, but right now it's not working on my device.

harambasicluka:
It worked on a OnePlus, had to wait a few minutes and prio background was turned on (on OnePlus it's called "don't optimise"). So it's unreliable as expected...

harambasicluka (dismissed):
Fine from my side. Just one tiny tiny remark :)

And as mentioned I was able to get the notification on a device where background prio was turned on.

harambasicluka:
> Tested it on day one, worked like expected, but right now it's not working on my device.

I think that's to the fact that we removed the popup to activate prio background. And as I tested it on day one it was likely activated.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,64 @@
...
+            .setContentIntent(createPendingIntentToMainActivity())
+            .setAutoCancel(true)
+            .setContentTitle(title)
+            .setContentText(content)
+
+        return builder.build()
+    }
+
+    fun sendNotification() {
+        if (CoronaWarnApplication.isAppInForeground) {
```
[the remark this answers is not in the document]
AlexanderAlferov: Tried - was not working from background.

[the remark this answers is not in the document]
d4rken: In what way? It should work, if it's not working please check why, maybe there is a bug that needs to be fixed.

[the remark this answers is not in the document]
AlexanderAlferov: Found the issue. Implemented.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,64 @@
...
+            return
+        }
+        val title = context.getString(R.string.risk_details_deadman_notification_title)
+        val content = context.getString(R.string.risk_details_deadman_notification_body)
+        val notification =
+            buildNotification(title, content) ?: return
+        with(NotificationManagerCompat.from(context)) {
+            notify(Random.nextInt(), notification)
```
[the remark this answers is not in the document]
AlexanderAlferov: We need random id for notification.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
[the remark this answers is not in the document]
d4rken: Please always explain what the issue was :smiley: .

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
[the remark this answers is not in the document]
d4rken: Why? We do not need a random notification id. The ID is specific to a notification, we should not share IDs between different types of notification, but for the same notification, the ID usually stay the same, otherwise we wouldn't be able to update, or cancel it. :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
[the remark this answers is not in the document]
d4rken: Ok a missunderstanding.

harambasicluka (commented):
Code LGTM, still doing some device testing.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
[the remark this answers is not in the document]
AlexanderAlferov: Nothing from code perspective - was confused with naming :)

harambasicluka (dismissed):
Successful tested on a device, still not reliable, but I think that's not a problem from your implementation.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationScheduler.kt
```diff
@@ -0,0 +1,60 @@
...
+import androidx.work.ExistingWorkPolicy
+import androidx.work.WorkManager
+import dagger.Lazy
+import dagger.Reusable
+import javax.inject.Inject
+
+@Reusable
+class DeadmanNotificationScheduler @Inject constructor(
+    val timeCalculation: DeadmanNotificationTimeCalculation,
+    val workManager: Lazy<WorkManager>,
```
d4rken, line 13: Why is this `Lazy`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    suspend fun sendNotification() {
+        if (foregroundState.isInForeground.first()) {
+            return
+        }
+        val title = context.getString(R.string.risk_details_deadman_notification_title)
+        val content = context.getString(R.string.risk_details_deadman_notification_body)
+        val notification =
+            buildNotification(title, content) ?: return
+        with(NotificationManagerCompat.from(context)) {
```
d4rken, line 61: You can inject this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationScheduler.kt
AlexanderAlferov: Otherwise it wouldn't be injected in main class (CoronaWarnApplication).

d4rken: I think it should work, what error do you get?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationSender.kt
AlexanderAlferov: Done

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationScheduler.kt
AlexanderAlferov: `java.lang.RuntimeException: Unable to create application de.rki.coronawarnapp.CoronaWarnApplication: java.lang.IllegalStateException: WorkManager is not initialized properly.  You have explicitly disabled WorkManagerInitializer in your manifest, have not manually called WorkManager#initialize at this point, and your Application does not implement Configuration.Provider.`
DeadmanScheduler is being injected along with WorkManager, but it also needs an initialised WorkManager. WorkManager is being initialised right in OnCreate in CoronaWarnApplication (line 61), but after scheduler is being injected.

d4rken (approved):

harambasicluka (approved):

PR #1514, opened 2020-11-01
Accept invalidly handed out qr-codes (COMMUNITY)

thepiwo:
Incorrectly generated QR-Codes have been handed out apparently, these contain an extra `/` before the GUID, e.g. `https://localhost//?GUID`. This fix will allow people with such QR-Codes to also use the CWA to register their test.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/QRScanResult.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/QRScanResult.kt
@@ -22,7 +22,7 @@ data class QRScanResult(val rawResult: String) {
     companion object {
         // regex pattern for scanned QR code URL
         val QR_CODE_REGEX: Pattern = Pattern.compile(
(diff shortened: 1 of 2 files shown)


thomasaugsten:
@thepiwo can you provide via e-mail a photo of the QR Code form and the doctor/test lab where you receive this QR Code?
thomas.augsten@sap.com

thepiwo:
Thanks @thomasaugsten I did email you

harambasicluka:
@thepiwo we are currently discussing some other changes to the config as well, I think we have to provide a solution ourselves. So I'll leave this PR open but I think we won't use it to speed up the process as otherwise you would have to include the changes or we would have spread one fix over multiple PRs. 
Sorry for that, but thanks for initiative and providing a solution!

Cheers,
Luka

thepiwo:
Thanks @harambasicluka will this change itself be integrated, if so it would be cool if my commit gets cherry-picked. Otherwise, thanks anyways :)

harambasicluka:
As already announced, we had to implement our own solution in #1523 , but we won't exclude `//` from the regex. New qr codes shouldn't contain this, we are currently communicating this.

PR #1515, opened 2020-11-02
Show current env when launching app (DEV)

d4rken:
When testing or setting something up, we spend a lot of time (too much time) going back and forth to make sure we are on the right environment.

This PR adds a toast message on `deviceForTester` builds that shows us the current environment.

## How to test
* Check that the environments are displayed only for `deviceForTesters`.
* Check that switching the environments causes the toast message to change.


AndroidMedaGalaxy (dismissed):
Tested on the actual device
LGTM    \\(◦'⌣'◦)/

BMItr (dismissed):

PR #1520, opened 2020-11-03
Improved AppConfig logic for 1.7.x and 1.8.x (EXPOSUREAPP-3455)

d4rken:
This PR refactores the `AppConfigProvider` to prepare for the changes to the app config in 1.7.x and 1.8.x

* The app config (`ConfigData`) consists now of "per module" sub interfaces, e.g. there is a `RiskCalculationConfig` that is created by the `RiskCalculationConfig.Mapper` such that changes to individual parts of the original protobuf config are hidden and can be worked on in parallel.
* The `AppConfigProvider` now offers a `val appConfig: Flow<ConfigData>` to allow other app components to react to config changes (e.g. reset RiskCalculuation, EXPOSUREAPP-2718 @chris-cwa )
* `ConfigData` exposes time variables (server time and the offset to our local time) to allow other app components to abort e.g. Riskcalclulation if the time is incorrect (EXPOSUREAPP-2998)
* There is a new test menu solely for viewing the current app config, seeing when it was last refreshed and to force a refresh.

## Test
* Smoke test the app
* Review that the tests in the `appconfig` package cover all edge cases you can think off
* Play a bit with the app config test menu


d4rken:
Can be reviewed, still fixing the last unit tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/download/AppConfigServer.kt
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+            if (exportBinary == null || exportSignature == null) {
+                throw ApplicationConfigurationInvalidException()
+            }
+
+            if (verificationKeys.hasInvalidSignature(exportBinary, exportSignature)) {
+                throw ApplicationConfigurationCorruptException()
+            }
+            exportBinary!!
```
chris-cwa, line 60: !! shouldn't be used, change it so that the compiler regards it as non-null

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -0,0 +1,43 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun AppConfig.ApplicationConfiguration.getMappedSupportedCountries(): List<String> {
+        if (supportedCountriesList == null) return emptyList()
+
+        if (supportedCountriesList.size == 1 && !VALID_CC.matches(supportedCountriesList.single())) {
+            Timber.w("Invalid country data, clearing. (%s)", supportedCountriesList)
+            return emptyList()
+        }
+
+        return supportedCountriesList
```
chris-cwa, line 31: convert to when?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigData.kt
```diff
@@ -0,0 +1,23 @@
...
+import de.rki.coronawarnapp.appconfig.mapping.ConfigMapping
+import org.joda.time.Duration
+import org.joda.time.Instant
+
+interface ConfigData : ConfigMapping {
+
+    val updatedAt: Instant
+
+    /**
+     * serverTime + localOffset = updatedAt
```
ralfgehrer, line 12: minor: shouldn't this comment rather go to updatedAt

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/DefaultConfigData.kt
```diff
@@ -0,0 +1,14 @@
...
+import org.joda.time.Duration
+import org.joda.time.Instant
+
+data class DefaultConfigData(
+    val serverTime: Instant,
+    val mappedConfig: ConfigMapping,
+    override val localOffset: Duration,
+    override val isFallback: Boolean
+) : ConfigData, ConfigMapping by mappedConfig {
+    override val updatedAt: Instant = serverTime.plus(localOffset)
```
ralfgehrer, line 13: what about 
"* If **[isFallback]** returns true,
     * you should probably ignore the time offset."

ralfgehrer (commented):
Great work.

ralfgehrer (dismissed):

d4rken:
Offline mode behavior is not working correctly yet...

d4rken:
Make sure to check offline (airplane mode) behavior too.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigProvider.kt
```diff
@@ -1,149 +1,45 @@
...
-            }
-        }
-
-        val countryCheck = sanityChecked.supportedCountriesList
-        if (countryCheck.size == 1 && !VALID_CC.matches(countryCheck.single())) {
-            Timber.w("Invalid country data, clearing. (%s)", this.supportedCountriesList)
-            sanityChecked = sanityChecked.toNewConfig {
-                clearSupportedCountries()
+    suspend fun getAppConfig(): ConfigData = mutex.withLock {
+        Timber.tag(TAG).v("getAppConfig()")
```
ralfgehrer, line 34: Could be removed IMHO

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigSource.kt
```diff
@@ -0,0 +1,82 @@
...
+    private val parser: ConfigParser,
+    private val dispatcherProvider: DispatcherProvider
+) {
+
+    suspend fun retrieveConfig(): ConfigData = withContext(dispatcherProvider.IO) {
+        Timber.v("retrieveConfig()")
+        val (serverBytes, serverError) = try {
+            server.downloadAppConfig() to null
+        } catch (e: Exception) {
+            Timber.tag(TAG).w(e, "Failed to download AppConfig from server .")
```
ralfgehrer, line 26: rather an error (.e.) log?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/download/AppConfigServer.kt
```diff
@@ -0,0 +1,94 @@
...
+            )
+            Instant.parse(rawDate, DATE_FORMAT)
+        } catch (e: Exception) {
+            Timber.e("Failed to get server time.")
+            localTime
+        }
+        val offset = Duration(serverTime, localTime)
+        Timber.tag(TAG).v("Time offset was %dms", offset.millis)
+        return ConfigDownload(
+            rawData = rawConfig,
```
ralfgehrer, line 75: above "calculations" could be split into dedicated functions for easier readability

ralfgehrer (dismissed):
Successfully tested with and without internet connection on a Pixel (Android 10).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigProvider.kt
d4rken: I like a bit more verbose logging :blush:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigSource.kt
d4rken: As this would also trigger when no internet is available, it's more of a warning, isn't it? An error will be thrown and logged if then also no fallback config is available.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/download/AppConfigServer.kt
d4rken: Moved the extraction into local extension functions to improve readability.

ralfgehrer (dismissed):

harambasicluka (commented):
tested on device, works like expected 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/serialization/adapter/ByteArrayAdapter.kt
```diff
@@ -0,0 +1,20 @@
...
+
+class ByteArrayAdapter : TypeAdapter<ByteArray>() {
+    override fun write(out: JsonWriter, value: ByteArray?) {
+        if (value == null) out.nullValue()
+        else value.toByteString().base64().let { out.value(it) }
+    }
+
+    override fun read(reader: JsonReader): ByteArray? = when (reader.peek()) {
+        NULL -> reader.nextNull().let { null }
+        else -> reader.nextString().decodeBase64()!!.toByteArray()
```
chris-cwa, line 18:
```suggestion
        else -> reader.nextString().decodeBase64()?.toByteArray() ?: byteArrayOf()
```

chris-cwa (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/serialization/adapter/ByteArrayAdapter.kt
chris-cwa: alternativ:
```suggestion
        else -> reader.nextString().decodeBase64()?.toByteArray()
```

d4rken: This would change the returned data though, if we are not reading a NULL token, then we expect there to be valid data, we want to crash then, not return "null" as if there would be no data.

AFAICT (see [here](https://github.com/square/okio/blob/master/okio/src/commonMain/kotlin/okio/-Base64.kt#L31)), the only way for `decodeBase64` to return null is on invalid data. So if we then return null in the not `NULL` token case, then we treat not existing data and invalid data the same. Would that be desired behavior?

chris-cwa: ok, if a crash is ok with everyone - let's do it!

d4rken: I think the crash needs to happen, otherwise the classes above, using gson (with this adapter) to deserialize something, wouldn't know that there is something wrong with the data if there was no exception. What do you think?

chris-cwa: I believe this app should never crash in production. :( but my opinion is not the only one that counts.

d4rken: That's not what I meant.

The code we are talking about is part of a custom type adapter for data serialization via Gson.

The TypeAdapter does not have enough information to determine whether returning `null` or throwing an exception is the correct approach, so it should not make that decision, but throw an exception so that the "higher level classes" know about this issue and can make the decision as they have more information available.

```kotlin
    fun getLastTimeStampData() : Instant? {
        return try {
            gson.fromJson(someFile)
        } catch (e: Exception) {
            Timber.e(e, "Our data was corrupted, returning EPOCH")
            Instant.EPOCH
        }
    }
```

The `getLastTimeStampData` knows about the data's purpose and can make a decision whether we should return null in an error case, or a fallback value. If we catch the decoding issue in the `ByteArrayAdapter` and just return `null`, then `getLastTimeStampData` could no longer distinguish between "no value exists" and "value exists but data corrupted/invalid".

d4rken: Thinking about it, it would be good to throw a more specific Exception than a Kotlin NPE. Will add that.

d4rken: Let me add a unit test for before it's merged.

chris-cwa (dismissed):

chris-cwa (approved):

ralfgehrer (approved):

kolyaopahle (approved):

AlexanderAlferov (approved):

PR #1524, opened 2020-11-03
UPPER Case QR Code can be accepted (EXPOSUREAPP-3546)

AndroidMedaGalaxy:
## Description

QR Codes in Upper cases can now be accpted

### Implmentation info
REGEX adjusted and test cases have been added

Please click [here](https://regex101.com/r/RIDKZC/2/) if you want to test more QR code scnearios

## Steps to reproduce
step 1. Try the [QR Code generator](https://www.qr-code-generator.com/) and use the following examples.
HTTPS://LOCALHOST/?123456-12345678-1234-4dA7-b166-B86d85475064
https://localhost/?123456-12345678-1234-4DA7-B166-B86D85475ABC

step 2. modify them however you like (Upper case or lowercase letter) 
step 3. Scan them with the phone.
Step 4. You should not see the error for invalid QR


ritsam:
@harambasicluka @mlenkeit  Created this PR for UPPER case QR codes and the other PR had Rebase issues

harambasicluka:
Small remark, nothing to change now: the ticket is a bug, according to our internal wiki that means the branch has to start with `fix/` and also the ticket number should be included: `fix/3546-upper-case-qr-codes`.

harambasicluka:
Fixes the issue on the iOS repo: https://github.com/corona-warn-app/cwa-app-ios/issues/1438

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/service/submission/ScanResultTest.kt
```diff
@@ -27,6 +29,29 @@ class ScanResultTest {
...
+
+        scanResult = QRScanResult("HTTPS://LOCALHOST//?$guid")
+        scanResult.isValid shouldBe false
+
+        scanResult = QRScanResult("HTTPS://LOCALHOST///?$guid")
+        scanResult.isValid shouldBe false
+
         // more invalid tests checks
         scanResult = QRScanResult("http://localhost/?$guid")
         scanResult.isValid shouldBe false
```
chris-cwa, line 57: if you want, you could put every test into its own method - this way the result of every test could be seen quickly in circle ci / GitHub

chris-cwa (dismissed):

mlenkeit (dismissed):
LGTM 👍

Supporting Chris' comment above regarding the test cases.

d4rken (changes requested):
Can we write a better regex :sweat_smile: ?

harambasicluka:
> Can we write a better regex 😅 ?

What do you mean by "better" a correct one or a beautifuler?

d4rken (commented):
:+1: 

https://regex101.com/r/gIMezQ/1

d4rken (approved):

mlenkeit (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/QRScanResult.kt
```diff
@@ -9,24 +9,19 @@ data class QRScanResult(val rawResult: String) {
...
-            "^((^https:\\/{2}localhost)(\\/\\?)[A-Fa-f0-9]{6}" +
-                    "[-][A-Fa-f0-9]{8}[-][A-Fa-f0-9]{4}[-][A-Fa-f0-9]{4}[-][A-Fa-f0-9]{4}[-][A-Fa-f0-9]{12})\$"
-        )
-        const val GUID_SEPARATOR = '?'
-        const val MAX_QR_CODE_LENGTH = 150
-        const val MAX_GUID_LENGTH = 80
+        val QR_CODE_REGEX: Pattern = ("^" + // Match start of string
+            "(?:https:\\/{2}localhost)" + // Match `https://localhost`
+            "(?:\\/{1}\\?)" + // Match the query param `/?`
+            "([a-f\\d]{6}[-][a-f\\d]{8}[-](?:[a-f\\d]{4}[-]){3}[a-f\\d]{12})" + // Match the UUID
```
harambasicluka, line 23: I think the UUID has to be upper case, but please clarify with @thomasaugsten 
```suggestion
            "([A-F\\d]{6}[-][A-F\\d]{8}[-](?:[A-F\\d]{4}[-]){3}[A-F\\d]{12})" + // Match the UUID
```

harambasicluka: only `https` and `localhost` are allowed to be case insensitive.

harambasicluka (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/service/submission/QRScanResult.kt
harambasicluka: Clarified internally, the whole string is case insensitive.

harambasicluka (approved):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/service/submission/ScanResultTest.kt
ritsam: @chris-cwa  that's been doen. tried to segregate valid and invalid tests, along with upper lower and mixed case letters

ritsam:
> > Can we write a better regex 😅 ?
> 
> @d4rken what do you mean by "better" a correct one or a beautifuller?

perhaps he meant more readable regex. Beautidication = ✅  

Done now!  (▀̿Ĺ̯▀̿ ̿)

ascherp:
> ## Description
> 
> QR Codes in Upper cases can now be accpted
> ### Implmentation info
> 
> REGEX adjusted and test cases have been added
> 
> Please click [here](https://regex101.com/r/RIDKZC/2/) if you want to test more QR code scnearios
> ## Steps to reproduce
> 
> step 1. Try the [QR Code generator](https://www.qr-code-generator.com/) and use the following examples.
> [HTTPS://LOCALHOST/?123456-12345678-1234-4dA7-b166-B86d85475064](HTTPS://LOCALHOST/?123456-12345678-1234-4dA7-b166-B86d85475064)
> https://localhost/?123456-12345678-1234-4DA7-B166-B86D85475ABC
> 
> step 2. modify them however you like (Upper case or lowercase letter)
> step 3. Scan them with the phone.
> Step 4. You should not see the error for invalid QR

A good offline-tool to quick-fix the localhost// problem for yourself is to use https://wiki.ubuntuusers.de/QtQR/.
In version 1.7.1 of the app, upper/lower case is not an issue anymore, but the two // after localhost are still a problem. Example, if you have https://localhost//?AD1ED0-... a tiny modification to https://localhost/?AD1ED0-... will do the trick.

The bug is very annoying as it unnecessarily delays the entire process. One character makes the difference ;-)

A

PR #1534, opened 2020-11-04
Re-add common codestyle (DEV)

d4rken:
The previous code style was removed from the repo by https://github.com/corona-warn-app/cwa-app-android/commit/be4c773dc0480faac9230ef12e7da2d738e5beee

Accidentally? @BMItter sus :eyes: 

This PR re-adds the codestyle that the project has been using up to now.
Running the formatter on the whole project shows about 35 files that don't use this code style now, likely added during the period the file was missing.

Codestyle configurations should be part of the source code such that the auto-formatter behaves the same for everyone, otherwise random changes will be included in commits, when the formatter is run before commits (which it should be).

We are not using the default codestyle, but an adjusted version that fits the majority of the files and conforms to what klint+detekt expects.


BMItter:
was removed by accident in github :baby_chick:
thx for readding it

BMItr (approved):

ralfgehrer (approved):

PR #1543, opened 2020-11-05
Risk calculation failed card - ui (EXPOSUREAPP-3051)

SamuraiKek:
### Description

This PR introduces a new risk card for when the comparison of the random id's with the server fails. 

We are also trying to separate the risk cards in different views based on the risk status, starting with this one. Up until now, we've only had one view with a data class responsible for all the states of the risk card which is growing quite clunky and hard to follow.  (TracingCardState)

This PR is open as draft for now as there are other views that need to be implemented for the rest of the risk card states and with my availability reducing, other people could contribute to it as well.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/riskcards/RiskCardNoInternetStateProvider.kt
```diff
@@ -0,0 +1,37 @@
...
+import javax.inject.Inject
+
+@Reusable
+class RiskCardNoInternetStateProvider @Inject constructor(
+    tracingRepository: TracingRepository
+) {
+
+    val state: Flow<RiskCardNoInternet> = combine(
+        RiskLevelRepository.riskLevelScoreLastSuccessfulCalculated.onEach {
+            Timber.v("riskLevelScoreLastSuccessfulCalculated: $it")
```
ritsam, line 21: do you think we can use natural words here instead ofusing camel case variables?
More of a question than a suggestion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -91,7 +102,9 @@ class HomeFragment : Fragment(R.layout.fragment_home), AutoInject {
         }
 
         vm.showLoweredRiskLevelDialog.observe2(this) {
-            if (it) { showRiskLevelLoweredDialog() }
+            if (it) {
+                showRiskLevelLoweredDialog()
+            }
```
ritsam, line 107:
```suggestion
 if (it) { showRiskLevelLoweredDialog() }
```

Probably your lint settings changed it. but makes sense to have it in single line

ritsam (commented):
onViewCreated() method can have less lines. detekt issue might crop up I feel.

You think moving the _observe2_ declarations to another funtion make sense?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
SamuraiKek: You're right, it was the linter doing it's stuff. I agree, it should be a one liner.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/riskcards/RiskCardNoInternetStateProvider.kt
SamuraiKek: It might be a good idea. I just followed the other state providers as an example for this (SubmissionCardsStateProvider) and they all have the Timber messages like this. Right now it looks like the Timber logs differ from person to person and  I don't really think we have some guidelines for them.

SamuraiKek:
Closing this draft PR as it's no longer relevant.

PR #1545, opened 2020-11-05
Converted RetrieveDiagnosisKeysTransaction to Task (Exposureapp-2908)

chris-cwa:

PR #1550, opened 2020-11-06
Update Risk Score more than once a day (EXPOSUREAPP-3455)

d4rken:
Use a hybrid approach of day and hour key packages to provide risk score updates as often as the quota allows.

* Change the download mechanisms to the new download & sync behavior.
* Smarter timeout behavior that times out individual downloads to allow proceeding with subsets before the global timeout hits.
* Allow cache invalidation via protobuf config.
* Allow custom timeouts via protobuf config.
* Download day packages on any type of connection, hourly packages only on metered.
* New test fragment

## TODO
- [x] Finish tests
- [x] Finish test menu
- [x] Merge #1545 into release/1.7.x first

## Out of scope
* New error card for the home screen that shows (EXPOSUREAPP-3411)

## Testing
* Smoke test & Check logs
* Check unit tests for edge cases (package `de.rki.coronawarnapp.diagnosiskeys`)
* Compare against [requirements](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/55e830687219ab9bdb810c2b8e01b9b477610cc5/docs/spec/key-download.md)


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragmentViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+import de.rki.coronawarnapp.util.coroutine.DispatcherProvider
+import de.rki.coronawarnapp.util.network.NetworkStateProvider
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
+import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
+import kotlinx.coroutines.flow.map
+
+class KeyDownloadTestFragmentViewModel @AssistedInject constructor(
+    dispatcherProvider: DispatcherProvider,
+    private val syncSettings: KeyPackageSyncSettings,
+    private val appConfigProvider: AppConfigProvider,
```
ralfgehrer, line 19: appConfigProvider is not used

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_keydownload.xml
```diff
@@ -0,0 +1,74 @@
...
+                android:layout_height="wrap_content"
+                android:text="Is metered network: ??" />
+
+            <Switch
+                android:id="@+id/allow_metered_toggle"
+                android:layout_width="match_parent"
+                android:layout_height="wrap_content"
+                android:text="Hourly pkgs on metered networks" />
+
+            <LinearLayout
```
ralfgehrer, line 37: Three layers of nested linear layouts!? I can only accept that for test fragment ;)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/DownloadConfigMapper.kt
```diff
@@ -2,20 +2,93 @@ package de.rki.coronawarnapp.appconfig.mapping
...
+                    day = LocalDate.parse(it.date, DAY_FORMATTER),
+                    hour = LocalTime.parse("${it.hour}", HOUR_FORMATTER)
+                )
+            } catch (e: Exception) {
+                Timber.e(e, "Failed to parse invalidated hour metadata: %s", it)
+                null
+            }
+        }
+
+    sealed class InvalidatedKeyFile : KeyDownloadConfig.InvalidatedKeyFile {
```
ralfgehrer, line 67: How about moving this to its own file?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ExposureDetectionConfig.kt
```diff
@@ -3,9 +3,13 @@ package de.rki.coronawarnapp.appconfig
...
 import de.rki.coronawarnapp.appconfig.mapping.ConfigMapper
 import de.rki.coronawarnapp.server.protocols.internal.ExposureDetectionParameters
+import org.joda.time.Duration
 
 interface ExposureDetectionConfig {
 
+    val maxExposureDetectionsPerDay: Int
+    val minTimeBetweenDetections: Duration
+
     val exposureDetectionConfiguration: ExposureConfiguration
```
mlenkeit, line 13: What is this doing here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ExposureDetectionConfig.kt
```diff
@@ -3,9 +3,13 @@ package de.rki.coronawarnapp.appconfig
 import com.google.android.gms.nearby.exposurenotification.ExposureConfiguration
 import de.rki.coronawarnapp.appconfig.mapping.ConfigMapper
 import de.rki.coronawarnapp.server.protocols.internal.ExposureDetectionParameters
+import org.joda.time.Duration
 
 interface ExposureDetectionConfig {
 
+    val maxExposureDetectionsPerDay: Int
```
mlenkeit, line 10: Might be worth renaming to `maxExposureDetectionsPerUTCDay` for future generations 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/DownloadConfigMapper.kt
```diff
@@ -2,20 +2,93 @@ package de.rki.coronawarnapp.appconfig.mapping
...
-            keyDownloadParameters = rawConfig.androidKeyDownloadParameters
+            individualDownloadTimeout = rawParameters.individualTimeout(),
+            overallDownloadTimeout = rawParameters.overAllTimeout(),
+            invalidDayETags = rawParameters.mapDayEtags(),
+            invalidHourEtags = rawParameters.mapHourEtags()
         )
     }
 
+    private fun KeyDownloadParametersAndroid.individualTimeout(): Duration {
+        return if (downloadTimeoutInSeconds == 0) Duration.standardSeconds(60)
```
mlenkeit, line 29: Do we need these default/fallback values here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/DownloadConfigMapper.kt
```diff
@@ -2,20 +2,93 @@ package de.rki.coronawarnapp.appconfig.mapping
...
+import de.rki.coronawarnapp.server.protocols.internal.KeyDownloadParameters.KeyDownloadParametersAndroid
+import org.joda.time.Duration
+import org.joda.time.LocalDate
+import org.joda.time.LocalTime
+import org.joda.time.format.DateTimeFormat
+import timber.log.Timber
 import javax.inject.Inject
 
 @Reusable
 class DownloadConfigMapper @Inject constructor() : KeyDownloadConfig.Mapper {
```
mlenkeit, line 16: I'm not so fond of unnecessary shortcuts in naming... why not calling this `KeyDownloadParametersMapper` according to the ProtoBuf structure?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/ExposureDetectionConfigMapper.kt
```diff
@@ -6,22 +6,48 @@ import dagger.Reusable
...
-        ExposureDetectionConfigContainer(
+    override fun map(rawConfig: AppConfig.ApplicationConfiguration): ExposureDetectionConfig {
+        val exposureParams = rawConfig.androidExposureDetectionParameters
+        return ExposureDetectionConfigContainer(
             exposureDetectionConfiguration = rawConfig.mapRiskScoreToExposureConfiguration(),
-            exposureDetectionParameters = rawConfig.androidExposureDetectionParameters
+            exposureDetectionParameters = exposureParams,
+            maxExposureDetectionsPerDay = exposureParams.maxExposureDetectionsPerDay(),
+            minTimeBetweenDetections = exposureParams.minTimeBetweenExposureDetections(),
+            overAllDetectionTimeout = exposureParams.overAllDetectionTimeout()
```
mlenkeit, line 21: overAll -> overall 
https://www.dict.cc/?s=overall

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+
+    internal suspend fun invalidateCachedKeys(invalidatedKeyFiles: Collection<KeyDownloadConfig.InvalidatedKeyFile>) {
+        if (invalidatedKeyFiles.isEmpty()) {
+            Timber.tag(tag).d("No invalid files to delete.")
+            return
+        }
+
+        val badEtags = invalidatedKeyFiles.map { it.etag }
+        val toDelete = keyCache.getAllCachedKeys()
+            .filter { badEtags.contains(it.info.etag) }
```
mlenkeit, line 25: That's a bit of a shortcut, isn't it? It's ignoring whether it's a day or hour package...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        Timber.tag(tag).w("Deleting invalidated cached keys: %s", toDelete.joinToString("\n"))
+        keyCache.delete(toDelete.map { it.info })
```
mlenkeit, line 28: These files should not be deleted yet. They should be downloaded again and if the download was successful, overwrite the cached file.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+
+        when {
+            cachedKey.type == CachedKeyInfo.Type.LOCATION_DAY -> {
+                // If there is no matching day on the server, our cached key is stale
+                return@filter !serverHasMatchingDay
+            }
+            cachedKey.type == CachedKeyInfo.Type.LOCATION_HOUR && serverHasMatchingDay -> {
+                // A cached hour for which a server day exists, means we don't need the hour anymore
+                // If there is no match, then we can't decide yet, and need to check the server for hours
+                return@filter true // Stale
```
mlenkeit, line 58: Hour packages should only be discarded once the Day Package has been downloaded.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
d4rken: How likely is it to have a collision here though? What gain would the extra comparison offer?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+        if (serverHasMatchingHour) {
+            // Our hour is still on the server
+            return@filter false // Not stale
+        }
+
+        // If we couldn't find match against the server data, our cache entry is probably stale
+        return@filter true
+    }
+
+    internal suspend fun getCompletedCachedKeys(
```
mlenkeit, line 80: Naming: sounds as if there could also be _incomplete cached keys_. How about `getDownloadedCachedKeys`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
```diff
@@ -80,5 +80,7 @@ object TimeAndDateExtensions {
 
     fun LocalDate.ageInDays(now: LocalDate) = Days.daysBetween(this, now).days
 
-    fun Instant.toLocalDate() = this.toDateTime(DateTimeZone.UTC).toLocalDate()
+    fun Instant.toLocalDate(timezone: DateTimeZone = DateTimeZone.UTC) = this.toDateTime(timezone).toLocalDate()
```
chiljamgossow, line 83: This is for sure an improvement.

I'm confused though, wouldn't one use the local timezone as default for a local date?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/DownloadConfigMapper.kt
d4rken: A protobuf config value that is missing defaults to `0` for numbers, right? Unless we want different behavior for "missing value", I'd like to always map it to sane defaults.

d4rken: In other cases I'm especially mapping it to fallbacks to prevent "division by 0 exceptions".

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyDownloadTool.kt
```diff
@@ -0,0 +1,69 @@
...
+                /**
+                 * To try legacy migration, we attempt to the etag as checksum.
+                 * Removing the quotes, the etag can represent the file's MD5 checksum.
+                 */
+                val etagAsChecksum = downloadInfo.etagWithoutQuotes
+                val continueDownload = !legacyKeyCache.tryMigration(etagAsChecksum, saveTo)
+                continueDownload // Continue download if no migration happened
+            }
+
+        val dlInfo = withTimeout(downloadConfig.individualDownloadTimeout.millis) {
```
harambasicluka, line 39: I assume `dl` stands for download? Everywhere else it's written out, so why not also do it here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyDownloadTool.kt
```diff
@@ -0,0 +1,69 @@
...
+            keyServer.downloadKeyFile(
+                locationCode = keyInfo.location,
+                day = keyInfo.day,
+                hour = keyInfo.hour,
+                saveTo = saveTo,
+                precondition = preconditionHook
+            )
+        }
+        Timber.tag(TAG).v("Dowwnload finished: %s -> %s", cachedKey, saveTo)
```
harambasicluka, line 48: typo :D

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.withContext
+import org.joda.time.LocalDate
+import timber.log.Timber
+import javax.inject.Inject
+
+@Reusable
+class DaySyncTool @Inject constructor(
```
mlenkeit, line 26: Naming: how about `DayPackageSyncTool`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+
+import de.rki.coronawarnapp.appconfig.KeyDownloadConfig
+import de.rki.coronawarnapp.diagnosiskeys.server.LocationCode
+import de.rki.coronawarnapp.diagnosiskeys.storage.CachedKey
+import de.rki.coronawarnapp.diagnosiskeys.storage.CachedKeyInfo
+import de.rki.coronawarnapp.diagnosiskeys.storage.KeyCacheRepository
+import de.rki.coronawarnapp.storage.DeviceStorage
+import timber.log.Timber
+
+open class BaseSyncTool(
```
mlenkeit, line 11: naming: How about `BaseKeyPackageSyncTool`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ExposureDetectionConfig.kt
d4rken: Direct exposure of the protobuf class as a workaround until specific mapping is available, will check whether we can remove it already. But I think it's used by the ENF client, and there is no specific mapping available yet, and because it will be changed with 1.8.x, not introducing an abstraction here now, saves us some conflicts.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+    tag = TAG
+) {
+
+    /**
+     * returns true if the sync was successful
+     * and false if not all files have been synced
+     */
+    internal suspend fun syncMissingDays(
+        availableLocations: List<LocationCode>,
+        forceSync: Boolean
```
mlenkeit, line 46: when is this ever set?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/preferences/FlowPreference.kt
```diff
@@ -0,0 +1,77 @@
+package de.rki.coronawarnapp.util.preferences
+
+import android.content.SharedPreferences
+import androidx.core.content.edit
+import com.google.gson.Gson
+import de.rki.coronawarnapp.util.serialization.fromJson
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+
+class FlowPreference<T> constructor(
```
chiljamgossow, line 10: This is nice!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+        Timber.tag(TAG).v("syncMissingDays(availableCountries=%s)", availableLocations)
+
+        val downloadConfig: KeyDownloadConfig = configProvider.getAppConfig()
+        invalidateCachedKeys(downloadConfig.invalidDayETags)
+
+        val missingDays = availableLocations.mapNotNull {
+            determineMissingDays(it, forceSync)
+        }
+        if (missingDays.isEmpty()) {
+            Timber.tag(TAG).i("There were no missing days.")
```
mlenkeit, line 57: log: missing _Day Packages_

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+            return SyncResult(successful = true, newPackages = emptyList())
+        }
+
+        Timber.tag(TAG).d("Downloading missing days: %s", missingDays)
```
mlenkeit, line 61: log: Day Packages

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
d4rken: I agree that it's a bit confusing, the actual time classes are called `LocalTime` and `LocalDate`, internally we always work with `UTC` time. We need `LocalTime` and `LocalDate` for the calculations though (especially formatting). I'm open for suggestions if you have an idea for a better name.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/DownloadConfigMapper.kt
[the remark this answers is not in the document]
d4rken: Good point, I'll add a comment.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+        return SyncResult(
+            successful = downloads.size == downloadedDays.size,
+            newPackages = downloadedDays
+        )
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun expectNewDayPackages(cachedDays: List<CachedKey>): Boolean {
+        val yesterday = timeStamper.nowUTC.toLocalDate().minusDays(1)
+        val newestDay = cachedDays.map { it.info.createdAt }.maxOrNull()?.toLocalDate()
```
mlenkeit, line 81: Are we sure we can use createdAt to find the most recent Day Package? What if the user clears the cache, files are downloaded in bulk, and an older package created (stored) before a more recent package.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+    keyCache = keyCache,
+    deviceStorage = deviceStorage,
+    tag = TAG
+) {
+
+    /**
+     * returns true if the sync was successful
+     * and false if not all files have been synced
+     */
+    internal suspend fun syncMissingDays(
```
mlenkeit, line 44: naming: syncMissingDayPackages

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun expectNewDayPackages(cachedDays: List<CachedKey>): Boolean {
+        val yesterday = timeStamper.nowUTC.toLocalDate().minusDays(1)
+        val newestDay = cachedDays.map { it.info.createdAt }.maxOrNull()?.toLocalDate()
+
+        return yesterday != newestDay
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal suspend fun determineMissingDays(location: LocationCode, forceSync: Boolean): LocationDays? {
```
mlenkeit, line 87: naming: determineMissingDayPackages

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragmentViewModel.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    fun download() {
+        isSyncRunning.postValue(true)
+        launch {
+            keyPackageSyncTool.syncKeyFiles()
+            isSyncRunning.postValue(false)
+        }
+    }
+
+    fun clearDownloads() {
```
BMItter, line 45: DRY - funs are nearly the same...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -168,19 +211,14 @@ class DownloadDiagnosisKeysTask @Inject constructor(
...
+        val wantedLocations = if (environmentSetup.useEuropeKeyPackageFiles) {
+            listOf("EUR")
+        } else {
+            requestedCountries ?: appConfigProvider.getAppConfig().supportedCountries
+        }.map { LocationCode(it) }
+
+        return keyPackageSyncTool.syncKeyFiles(wantedLocations)
     }
 
     private fun checkCancel() {
```
mlenkeit, line 224: naming: throwIfCancelled?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -70,18 +73,30 @@ class DownloadDiagnosisKeysTask @Inject constructor(
...
 
             val requestedCountries = arguments.requestedCountries
-            val availableKeyFiles = getAvailableKeyFiles(requestedCountries)
+            val keySyncResult = getAvailableKeyFiles(requestedCountries)
             checkCancel()
 
-            val totalFileSize = availableKeyFiles.fold(0L, { acc, file ->
-                file.length() + acc
-            })
+            val trackedDetections = enfClient.latestCalculations().first()
```
mlenkeit, line 85: why can there be multiple "latest calculations"? is this really "calculations" or "exposure detections"?

harambasicluka (commented):
some small remarks, will now do device testing.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -70,18 +73,30 @@ class DownloadDiagnosisKeysTask @Inject constructor(
...
+            val now = timeStamper.nowUTC
+
+            if (wasLastDetectionPerformedRecently(now, exposureConfig, trackedDetections)) {
+                // At most one detection every 6h
+                return object : Task.Result {}
+            }
+
+            if (hasRecentDetectionAndNoNewFiles(now, keySyncResult, trackedDetections)) {
+                //  Last check was within 24h, and there are no new files.
+                return object : Task.Result {}
```
mlenkeit, line 95: the return statement could be wrapped in a function such as `return abortWithEmptyResult()` for readability.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
mlenkeit, line 85: Let's not be lazy and call this `trackedExposureDetections` 😉

harambasicluka:
Spotted an app crash on start with a clean install:
````
2020-11-12 19:00:38.739 9323-9379/de.rki.coronawarnapp.test V/OkHttp: <-- HTTP FAILED: de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404
2020-11-12 19:00:38.741 9323-9366/de.rki.coronawarnapp.test W/AppConfigRetriever: Failed to download AppConfig from server .
    de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404
        at de.rki.coronawarnapp.http.HttpErrorParser.intercept(HttpErrorParser.kt:44)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at de.rki.coronawarnapp.http.interceptor.RetryInterceptor.intercept(RetryInterceptor.kt:15)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.kt:221)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at de.rki.coronawarnapp.http.interceptor.WebSecurityVerificationInterceptor.intercept(WebSecurityVerificationInterceptor.kt:11)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
2020-11-12 19:00:38.741 9323-9366/de.rki.coronawarnapp.test V/AppConfigStorage: get() AppConfig
2020-11-12 19:00:38.742 9323-9366/de.rki.coronawarnapp.test E/AppConfigStorage: Couldn't load config.
    java.io.FileNotFoundException: /data/user/0/de.rki.coronawarnapp.test/files/appconfig_storage/appconfig.json (No such file or directory)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:231)
        at java.io.FileInputStream.<init>(FileInputStream.java:165)
        at de.rki.coronawarnapp.appconfig.download.AppConfigStorage.getStoredConfig(AppConfigStorage.kt:97)
        at de.rki.coronawarnapp.appconfig.AppConfigSource$retrieveConfig$2.invokeSuspend(AppConfigSource.kt:49)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
```

As aligned with @d4rken this will be fixed with https://github.com/corona-warn-app/cwa-app-android/pull/1553 !

harambasicluka:
Went on WRU-XD to the new test menu, enabled `Hourly okg on metered netowrks`, clicked on `Download` and the app crashed

```
2020-11-12 19:32:07.561 14442-14592/de.rki.coronawarnapp.test V/AppConfigStorage: Overwriting 426 from 1605205927000
2020-11-12 19:32:07.564 14442-14481/de.rki.coronawarnapp.test D/KeySync:HourSync: No invalid files to delete.
2020-11-12 19:32:07.571 14442-14488/de.rki.coronawarnapp.test V/OkHttp: --> GET <ENDPOINT>/version/v1/diagnosis-keys/country/EUR/date/2020-11-12/hour
2020-11-12 19:32:07.572 14442-14488/de.rki.coronawarnapp.test V/OkHttp: --> END GET
2020-11-12 19:32:07.722 14442-14488/de.rki.coronawarnapp.test V/OkHttp: <-- HTTP FAILED: de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404
2020-11-12 19:32:07.742 1687-3390/system_process W/ActivityManager:   Force finishing activity de.rki.coronawarnapp.test/de.rki.coronawarnapp.ui.main.MainActivity
2020-11-12 19:32:07.758 14442-14466/de.rki.coronawarnapp.test I/Process: Sending signal. PID: 14442 SIG: 9
2020-11-12 19:32:07.907 1687-1916/system_process W/InputDispatcher: channel '2f9a4cb de.rki.coronawarnapp.test/de.rki.coronawarnapp.ui.main.MainActivity (server)' ~ Consumer closed input channel or an error occurred.  events=0x9
2020-11-12 19:32:07.907 1687-1916/system_process E/InputDispatcher: channel '2f9a4cb de.rki.coronawarnapp.test/de.rki.coronawarnapp.ui.main.MainActivity (server)' ~ Channel is unrecoverably broken and will be disposed!
```

harambasicluka:
Also happens when the `Hourly pkg on metered networks` is disabled:

```
2020-11-12 19:35:01.199 15349-15402/de.rki.coronawarnapp.test D/KeySync:HourSync: No invalid files to delete.
2020-11-12 19:35:01.202 15349-15434/de.rki.coronawarnapp.test V/OkHttp: --> GET <ENDPOINT>/version/v1/diagnosis-keys/country/EUR/date/2020-11-12/hour
2020-11-12 19:35:01.202 15349-15434/de.rki.coronawarnapp.test V/OkHttp: --> END GET
2020-11-12 19:35:01.325 15349-15434/de.rki.coronawarnapp.test V/OkHttp: <-- HTTP FAILED: de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404
2020-11-12 19:35:01.327 15349-15402/de.rki.coronawarnapp.test E/DownloadDiagnosisKeysTask: de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404
        at de.rki.coronawarnapp.http.HttpErrorParser.intercept(HttpErrorParser.kt:44)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at de.rki.coronawarnapp.http.interceptor.RetryInterceptor.intercept(RetryInterceptor.kt:15)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.kt:221)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at de.rki.coronawarnapp.http.interceptor.WebSecurityVerificationInterceptor.intercept(WebSecurityVerificationInterceptor.kt:11)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
2020-11-12 19:35:01.328 15349-15402/de.rki.coronawarnapp.test D/DownloadDiagnosisKeysTask: Initiate Rollback
2020-11-12 19:35:01.339 15349-15402/de.rki.coronawarnapp.test I/DownloadDiagnosisKeysTask: Finished (isCanceled=false).
2020-11-12 19:35:01.340 15349-15402/de.rki.coronawarnapp.test D/TaskController: Task ended: InternalTaskState(id=d04652bf-ccb3-4012-a2d6-3a1b4d003acd, request=DefaultTaskRequest(type=class de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask (Kotlin reflection is not available), arguments=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask$Arguments@4e4e04c, id=b5e8bd36-4b43-4fab-989c-6200e745ec7b), createdAt=2020-11-12T18:35:00.399Z, startedAt=null, finishedAt=null, error=null, result=null, config=Config(executionTimeout=PT480S, collisionBehavior=ENQUEUE), job="coroutine#99":LazyDeferredCoroutine{Cancelled}@428b1ad, task=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask@4448ce2)
2020-11-12 19:35:01.342 15349-15402/de.rki.coronawarnapp.test D/TaskController: Processing task data (count=2)
2020-11-12 19:35:01.343 15349-15402/de.rki.coronawarnapp.test V/TaskController: Tasks before processing: [InternalTaskState(id=26167883-4eac-4273-b4c0-8905fa340c1d, request=DefaultTaskRequest(type=class de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask (Kotlin reflection is not available), arguments=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask$Arguments@78d516f, id=b03a7fb5-6d9d-4cfd-b7ac-05dcf2357be5), createdAt=2020-11-12T18:34:59.802Z, startedAt=2020-11-12T18:34:59.811Z, finishedAt=2020-11-12T18:35:00.929Z, error=de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404, result=null, config=Config(executionTimeout=PT480S, collisionBehavior=ENQUEUE), job="coroutine#11":LazyDeferredCoroutine{Cancelled}@f314a73, task=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask@49ded30), InternalTaskState(id=d04652bf-ccb3-4012-a2d6-3a1b4d003acd, request=DefaultTaskRequest(type=class de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask (Kotlin reflection is not available), arguments=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask$Arguments@4e4e04c, id=b5e8bd36-4b43-4fab-989c-6200e745ec7b), createdAt=2020-11-12T18:35:00.399Z, startedAt=2020-11-12T18:35:00.932Z, finishedAt=null, error=null, result=null, config=Config(executionTimeout=PT480S, collisionBehavior=ENQUEUE), job="coroutine#99":LazyDeferredCoroutine{Cancelled}@428b1ad, task=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask@4448ce2)]
2020-11-12 19:35:01.345 15349-15402/de.rki.coronawarnapp.test E/TaskController: Task failed: InternalTaskState(id=d04652bf-ccb3-4012-a2d6-3a1b4d003acd, request=DefaultTaskRequest(type=class de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask (Kotlin reflection is not available), arguments=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask$Arguments@4e4e04c, id=b5e8bd36-4b43-4fab-989c-6200e745ec7b), createdAt=2020-11-12T18:35:00.399Z, startedAt=2020-11-12T18:35:00.932Z, finishedAt=null, error=null, result=null, config=Config(executionTimeout=PT480S, collisionBehavior=ENQUEUE), job="coroutine#99":LazyDeferredCoroutine{Cancelled}@428b1ad, task=de.rki.coronawarnapp.diagnosiskeys.download.DownloadDiagnosisKeysTask@4448ce2)
    de.rki.coronawarnapp.exception.http.NotFoundException: error during web request, http status 404
        at de.rki.coronawarnapp.http.HttpErrorParser.intercept(HttpErrorParser.kt:44)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at de.rki.coronawarnapp.http.interceptor.RetryInterceptor.intercept(RetryInterceptor.kt:15)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.logging.HttpLoggingInterceptor.intercept(HttpLoggingInterceptor.kt:221)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at de.rki.coronawarnapp.http.interceptor.WebSecurityVerificationInterceptor.intercept(WebSecurityVerificationInterceptor.kt:11)
        at okhttp3.internal.http.RealInterceptorChain.proceed(RealInterceptorChain.kt:109)
        at okhttp3.internal.connection.RealCall.getResponseWithInterceptorChain$okhttp(RealCall.kt:201)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:517)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
        at java.lang.Thread.run(Thread.java:764)
```

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragmentViewModel.kt
d4rken: Not sure what you mean here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+    private val tag: String
+) {
+
+    internal suspend fun invalidateCachedKeys(invalidatedKeyFiles: Collection<KeyDownloadConfig.InvalidatedKeyFile>) {
+        if (invalidatedKeyFiles.isEmpty()) {
+            Timber.tag(tag).d("No invalid files to delete.")
+            return
+        }
+
+        val badEtags = invalidatedKeyFiles.map { it.etag }
```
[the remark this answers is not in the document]
d4rken: :roll_eyes:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
d4rken: In `KeyPackageSyncTool`

```kotlin
        val successfulSync = dayPackageSyncTool.syncMissingDays(
            availableLocations = locations,
            forceSync = lastDownload == null || !lastDownload.successful
        )
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
d4rken: Not sure what you mean.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
d4rken: Not sure what you mean.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
d4rken: > why can there be multiple "latest calculations"?

It's a `Flow`, it will emit multiple values over time,  but always the latest.

>  is this really "calculations" or "exposure detections"?

Calculation by the ENF.

d4rken: It's not about "being lazy".
Horizontal space is limited, identiation and wrap around makes things less readable, readers attention span is limited.
It's about providing the reader with the right amount of information to understand the code, and not overburden them with unnecessary noise.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
d4rken: Good point :+1:.

What do you think about this:
https://github.com/corona-warn-app/cwa-app-android/pull/1550/commits/700b3511da0db3e7c690cd28b994790c7060cefb#diff-e8b0c086fb3151818210311734b3595f9bf8fa8b304d188da76982fe3ea4ee52R53

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
d4rken: Why not?
Either they are invalid or they are not, why would we want to resubmit files that we know are invalid? I think it's better to leave out a file than to submit an invalid one. For all we know, the invalid file could break the ENF. If the download that is supposed to overwrite the file fails for some reasons, then the invalid file is still there.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
d4rken: > Hour packages should only be discarded once the Day Package has been downloaded.

What's the relation to the line of code you commented on?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
mlenkeit: This is not only for invalid files that break ENF. Let's me try to explain...

If there's a rogue file on CDN that breaks ENF and that has already been downloaded by the client, the client might shows unexpected behavior (whatever that means in detail).
1. with the current implementation (delete before download)
    1. if the download fails, the client has one file less on the device and would or would not run Exposure Detection and Risk Calculation (depending on whether it's an hour our day package). If it runs, ENF should not fail anymore.
    1. if the download succeeds, everything is fine.
1. with the expected implementation (replace with new download instead of delete before download)
    1. if the download fails, the client still has the old file. Exposure Detection and Risk Calculation might _still_ fail 
    1. if the download succeeds, everything is fine.

But there's also the scenario where there's just a (maybe just partially) bad file that doesn't break ENF but leads to wrong results:
1. with the current implementation (delete before download)
    1. if the download fails, the client has one file less on the device and would or would not run Exposure Detection and Risk Calculation (depending on whether it's an hour our day package). If it runs, ENF might return a different number of matches. This might be lower for the user if a file is missing. This might falsly hide a red card because we removed the data too early.
    1. if the download succeeds, everything is fine.
1. with the expected implementation (replace with new download instead of delete before download)
    1. if the download fails, the client still has the old file. Exposure Detection and Risk Calculation would could still show the incorrect result but IMO that's better than no result (if the missing file leads to result)
    1. if the download succeeds, everything is fine.

So in the end, I'd rather have clients continue to fail than not showing a risk although there is one.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
```diff
@@ -0,0 +1,99 @@
...
+            Timber.tag(tag).d("No invalid files to delete.")
+            return
+        }
+
+        val badEtags = invalidatedKeyFiles.map { it.etag }
+        val toDelete = keyCache.getAllCachedKeys()
+            .filter { badEtags.contains(it.info.etag) }
+
+        Timber.tag(tag).w("Deleting invalidated cached keys: %s", toDelete.joinToString("\n"))
+        keyCache.delete(toDelete.map { it.info })
```
mlenkeit, line 28: Just tagging this again to not loose track, see https://github.com/corona-warn-app/cwa-app-android/pull/1550/#discussion_r519977834

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
mlenkeit: _cached hour for which a server day exists, means we don't need the hour anymore_ suggests that the Hour Package is deleted _before_ the Day Package is downloaded. It should be deleted after the Day Package has been downloaded successfully.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DaySyncTool.kt
mlenkeit: Yes, I think something like this is more reliable!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DayPackageSyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+        return SyncResult(
+            successful = downloads.size == downloadedDays.size,
+            newPackages = downloadedDays
+        )
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun expectNewDayPackages(cachedDays: List<CachedKey>): Boolean {
+        val yesterday = timeStamper.nowUTC.toLocalDate().minusDays(1)
+        val newestDay = cachedDays.map { it.info.toDateTime() }.maxOrNull()?.toLocalDate()
```
mlenkeit, line 81: Re-commenting, see https://github.com/corona-warn-app/cwa-app-android/pull/1550/#discussion_r520307066

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -118,6 +134,38 @@ class DownloadDiagnosisKeysTask @Inject constructor(
...
+        now: Instant,
+        exposureConfig: ExposureDetectionConfig,
+        trackedDetections: Collection<Calculation>
+    ): Boolean {
+        val lastDetection = trackedDetections.maxByOrNull { it.startedAt }
+        val nextDetectionAt = lastDetection?.startedAt?.plus(exposureConfig.minTimeBetweenDetections)
+        if (exposureConfig.maxExposureDetectionsPerUTCDay == 0) {
+            Timber.tag(TAG).w("Exposure detections are disabled! maxExposureDetectionsPerUTCDay=0")
+            return true
+        }
```
mlenkeit, line 147: IMO, this doesn't belong here. This should be a separate check.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/HourPackageSyncTool.kt
```diff
@@ -0,0 +1,163 @@
...
+        return downloads.map { (location, day, missingHour) ->
+            withContext(context = dispatcherProvider.IO) {
+                launcher(location, day, missingHour)
+            }
+        }
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun expectNewHourPackages(cachedHours: List<CachedKey>, now: Instant): Boolean {
+        val previousHour = now.toLocalTime().minusHours(1)
```
mlenkeit, line 123: Is this UTC?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/DownloadConfigMapper.kt
ralfgehrer: I was wondering the same thing when reviewing and I agree with the overwrite. Maybe the function name could hint that 0 is not a valid value (will be replaced)

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/CachedKeyListItem.kt
```diff
@@ -0,0 +1,12 @@
+package de.rki.coronawarnapp.test.keydownload.ui
+
+import de.rki.coronawarnapp.diagnosiskeys.storage.CachedKeyInfo
+import de.rki.coronawarnapp.util.lists.HasStableId
+
+data class CachedKeyListItem(
+    val info: CachedKeyInfo,
+    val fileSize: Long
+) : HasStableId {
+    override val stableId: Long
```
ralfgehrer, line 10: Minor: how about  introducing a typealias for stableId so we don't pass Long params araound

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragmentViewModel.kt
```diff
@@ -0,0 +1,75 @@
...
+class KeyDownloadTestFragmentViewModel @AssistedInject constructor(
+    dispatcherProvider: DispatcherProvider,
+    private val syncSettings: KeyPackageSyncSettings,
+    networkStateProvider: NetworkStateProvider,
+    private val keyPackageSyncTool: KeyPackageSyncTool,
+    private val keyCacheRepository: KeyCacheRepository
+) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
+
+    val currentCache = runBlocking {
+        // TODO runBlocking is not nice, how can we solve this better?
```
ralfgehrer, line 26: It's only consumed by the fragment as far as I can see. Why not make it a suspending function and call it lifecycle-aware from the fragment?

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyFileDownloadAdapter.kt
```diff
@@ -0,0 +1,70 @@
...
+    override fun onBindBaseVH(holder: VH, position: Int) {
+        val item = data[position]
+        holder.itemView.setOnLongClickListener {
+            deleteAction(item)
+            true
+        }
+        holder.bind(item)
+    }
+
+    class VH(
```
ralfgehrer, line 40: minor: rather spell it out, although it should be kinda obvious

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyFileDownloadAdapter.kt
```diff
@@ -0,0 +1,70 @@
...
+        R.layout.fragment_test_keydownload_adapter_line, parent
+    ), BindableVH<CachedKeyListItem, FragmentTestKeydownloadAdapterLineBinding> {
+
+        override val viewBinding = lazy { FragmentTestKeydownloadAdapterLineBinding.bind(itemView) }
+
+        override val onBindData: FragmentTestKeydownloadAdapterLineBinding.(key: CachedKeyListItem) -> Unit = { item ->
+            locationInfo.text = item.info.location.identifier
+
+            val shortSize = Formatter.formatShortFileSize(context, item.fileSize)
+            typeInfo.text = when (item.info.type) {
```
ralfgehrer, line 52: minor: the when cases could be moved to dedicated "formatter" functions

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/HourPackageSyncTool.kt
```diff
@@ -0,0 +1,163 @@
...
+    private val configProvider: AppConfigProvider,
+    private val dispatcherProvider: DispatcherProvider
+) : BaseKeyPackageSyncTool(
+    keyCache = keyCache,
+    deviceStorage = deviceStorage,
+    tag = TAG
+) {
+
+    /**
+     * returns true if the sync was successful
```
ralfgehrer, line 44: Comment is outdated, isn't it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/HasStableId.kt
```diff
@@ -0,0 +1,5 @@
+package de.rki.coronawarnapp.util.lists
+
+interface HasStableId {
```
ralfgehrer, line 3: Isn't that somewhat misused? E.g. the Comparable interface is implemented if a class fulfills that criteria, otherwise it does not implement this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
ralfgehrer: toLocaleDateFormat ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/KeyPackageSyncTool.kt
```diff
@@ -0,0 +1,138 @@
...
+        val lastDownload = syncSettings.lastDownloadDays.value
+        Timber.tag(TAG).d("Synchronizing available days (lastDownload=%s).", lastDownload)
+
+        syncSettings.lastDownloadDays.update {
+            KeyPackageSyncSettings.LastDownload(startedAt = timeStamper.nowUTC)
+        }
+
+        val successfulSync = dayPackageSyncTool.syncMissingDayPackages(
+            availableLocations = locations,
+            forceSync = lastDownload == null || !lastDownload.successful
```
mlenkeit, line 83: I still don't quite understand this. What does forceSync refer to from the tech spec?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DayPackageSyncTool.kt
```diff
@@ -0,0 +1,142 @@
...
+    tag = TAG
+) {
+
+    /**
+     * returns true if the sync was successful
+     * and false if not all files have been synced
+     */
+    internal suspend fun syncMissingDayPackages(
+        availableLocations: List<LocationCode>,
+        forceSync: Boolean
```
mlenkeit, line 46: Took me quite some time to fully understand what forceSync is doing and when it's set (logically).
Just from the name, I was expecting this to force the sync with CDN regardless of any caching, but that's not the case.

How about renaming this to something like `forceIndexLookup` or `forceIndexComparison` (or DayPackageIndex instead of Index for extra clarity/verbosity 😄)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/CachedKeyInfo.kt
```diff
@@ -18,7 +20,7 @@ data class CachedKeyInfo(
     @ColumnInfo(name = "day") val day: LocalDate, // i.e. 2020-08-23
     @ColumnInfo(name = "hour") val hour: LocalTime?, // i.e. 23
     @ColumnInfo(name = "createdAt") val createdAt: Instant,
-    @ColumnInfo(name = "checksumMD5") val checksumMD5: String?,
+    @ColumnInfo(name = "checksumMD5") val etag: String?, // ETag
```
mlenkeit, line 23: Does it make sense to add some kind of getter for `etagWithQuotes` and `etagWithoutQuotes`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/calculationtracker/DefaultCalculationTracker.kt
```diff
@@ -25,7 +26,8 @@ class DefaultCalculationTracker @Inject constructor(
     @AppScope private val scope: CoroutineScope,
```
mlenkeit, line 26: naming: I don't think this is tracking the calculation. This is tracking the Exposure Detection.

The sequence is:
1. Key Download
2. Exposure Detection (aka provideDiagnosisKeys)
3. Risk Calculation (via Exposure Summary/Exposure Windows)

mlenkeit:
c31937b3d63d6ec189dc0bc512e783f7e1ff75b9 doesn't compile for me:

```
failed
:Corona-Warn-App:compileDeviceForTestersDebugKotlin
Corona-Warn-App/build/generated/source/navigation-args/deviceForTesters/debug/de/rki/coronawarnapp/test/crash/ui/SettingsCrashReportFragmentDirections.kt
Unresolved reference: action_crashReportFragment_to_settingsCrashReportDetailsFragment
Corona-Warn-App/build/generated/source/navigation-args/deviceForTesters/debug/de/rki/coronawarnapp/test/menu/ui/TestMenuFragmentDirections.kt
Unresolved reference: action_testMenuFragment_to_settingsCrashReportFragment
Unresolved reference: action_test_menu_fragment_to_appConfigTestFragment
Unresolved reference: action_test_menu_fragment_to_debugOptionsFragment
Unresolved reference: action_test_menu_fragment_to_keyDownloadTestFragment
Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/appconfig/ui/AppConfigTestFragment.kt
Unresolved reference: test_appconfig_fragment
Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/crash.ui/SettingsCrashReportFragment.kt
Unresolved reference: action_testMenuFragment_to_settingsCrashReportFragment
Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/debugoptions/ui/DebugOptionsFragment.kt
Unresolved reference: test_debugoptions_fragment
Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragment.kt
Unresolved reference: test_keydownload_fragment
Compilation error
```

mlenkeit:
When I run c31937b3d63d6ec189dc0bc512e783f7e1ff75b9 against my local CWA server, after an initial download of all available data, moving the app to background, and opening again, I see the following requests immediately:
1. /version/v1/diagnosis-keys/country/EUR/date
1. /version/v1/diagnosis-keys/country/EUR/date/2020-11-10/hour
1. /version/v1/diagnosis-keys/country/EUR/date
1. /version/v1/diagnosis-keys/country/EUR/date/2020-11-10/hour

Looks like something is scheduled twice.

mlenkeit:
/version/v1/diagnosis-keys/country/EUR/date/2020-11-10/hour is requested despite (real) metered network (and Is metered network? true) in the Developer Menu.

mlenkeit:
Everytime I open the the client, it sends a request to /version/v1/diagnosis-keys/country/EUR/date although the the most recent Day Package has already been downloaded.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
chiljamgossow: I see why you use the UTC date as it is independent from timezone changes. Maybe toTimeZoneAdjustedLocalDate. However it is only minor thing, especially as it has been there before.

d4rken:
> Everytime I open the the client, it sends a request to /version/v1/diagnosis-keys/country/EUR/date although the the most recent Day Package has already been downloaded.

Fixed with https://github.com/corona-warn-app/cwa-app-android/pull/1550/commits/f3f548f718acf5e952dda96bdcaccf05d8b2273d

> Looks like something is scheduled twice.

Fixed with https://github.com/corona-warn-app/cwa-app-android/pull/1550/commits/42807e55b4f8c984b01ec6755dec98eb0e8c8343

chiljamgossow (commented):
It looks fine to me (with my limited knowledge of the project).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/CachedKeyInfo.kt
d4rken: I've switched all code to treat etags as they are and not remove any quotes, special handling should no longer be required.

d4rken: Edge case would be invalidation during the first 14 days of 1.7.x beacuse it wouldn't match the etags we have already in db (etag as checksum without quotes).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/calculationtracker/DefaultCalculationTracker.kt
d4rken: Renamed to `TrackedExposureDetections`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/HasStableId.kt
d4rken: >  E.g. the Comparable interface is implemented if a class fulfills that criteria, otherwise it is not.

Isn't that what we are doing here? An adapter class can implement this interface to signal that it provides a `stableId`. So for all classes that offer this "feature" by implementing this interface, we can use classes on the adapter that don't need to know anything but the stable ID. i.e. `AsyncDiffUtilAdapter`.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyFileDownloadAdapter.kt
d4rken: Yes, but for test only code this is clean enough :relaxed:

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/CachedKeyListItem.kt
d4rken: Don't see the benefit here though :thinking: ? The stableID is a long, hiding that from the developer, and somehow obscuring the relation to recyclerview adapter's "stableID" seems to be disadvantageous.

This is not a custom property we invented. https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.Adapter#hasStableIds()

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/HourPackageSyncTool.kt
d4rken: > Is this UTC? It should be...

It is, I've renamed the method to make it more clear.

> And why is this line so different from expectNewDayPackages?

One calls `nowUTC` and the other get's it passed as parameter as to not call it two times and work on two different timestamps.

d4rken: > It is, I've renamed the method to make it more clear.

Kept the name for now, to prevent conflicts with other changes. Will change in 1.8.x.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragmentViewModel.kt
BMItter: Dry means, Don't repeat yourself (codestyle).. as download and clearDownloads is nearly the same the variable part could be passed as parameter. But its just style, not an issue. So feel free

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
d4rken: I agree that it should probably be named something like "toUTCLocalDate" or something similar, but closing this for now as I noticed we also use this line of code for the risk calculation which is currently being changed on `release/1.8.x` I want to avoid unnecessary conflicts there.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseSyncTool.kt
mlenkeit: Rather unlikely, yes, but... 😄

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/CachedKeyInfo.kt
mlenkeit: I don't think that should be a problem...

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyDownloadTestFragmentViewModel.kt
d4rken: https://github.com/corona-warn-app/cwa-app-android/pull/1550/commits/982ad98b82cf056d4afc30597de53750584e4e68

mlenkeit (approved):
I think we can leave it as-is for now to build the first RC and to the necessary minor changes in a follow-up PR for the next RC

harambasicluka (approved):

ralfgehrer (approved):

harambasicluka:
Would appreciate more than 2 approvals on such a large PR.

BMItr (approved):

PR #1553, opened 2020-11-07
Improved app config handling (Identifier, local fallback) (DEV)

d4rken:
* The app config now offers an `identifier` field that is unique, and represents the current app config. If the identifier changes, the app config should be considered changed (needed for EXPOSUREAPP-2718, risk calc reset after config change).
* There is now a local default app config (@mlenkeit :heart: ) that we can fall back to if all else fails which allows us to guarantee that the app config is never null, which makes our app behavior much more consistent.
* The app config update call will now switch scopes, previously the callers scope was used, now the app scope is used. We want to prevent that a canceled caller's scope, will cancel the config retrieval, we'll want that new config in any case, it should never be aborted. I think this was the cause for the error @harambasicluka produced [here](https://github.com/corona-warn-app/cwa-app-android/pull/1545#pullrequestreview-525158395). I can't reproduce it locally though, can you confirm whether this fixes it Luka?

## Testing
* Force the app into edge cases, fresh launch, no internet etc.


harambasicluka:
Getting the same error as in #1550 , but the app isn't crashing any longer as promised by @d4rken ! <3

```
2020-11-12 19:10:03.846 10448-10686/de.rki.coronawarnapp.test V/AppConfigStorage: get() AppConfig
2020-11-12 19:10:03.849 10448-10686/de.rki.coronawarnapp.test E/AppConfigStorage: Couldn't load config.
    java.io.FileNotFoundException: /data/user/0/de.rki.coronawarnapp.test/files/appconfig_storage/appconfig.json (No such file or directory)
        at java.io.FileInputStream.open0(Native Method)
        at java.io.FileInputStream.open(FileInputStream.java:231)
        at java.io.FileInputStream.<init>(FileInputStream.java:165)
        at de.rki.coronawarnapp.appconfig.download.AppConfigStorage.getStoredConfig(AppConfigStorage.kt:98)
        at de.rki.coronawarnapp.appconfig.AppConfigSource$retrieveConfig$2.invokeSuspend(AppConfigSource.kt:53)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:104)
2020-11-12 19:10:03.850 10448-10686/de.rki.coronawarnapp.test W/AppConfigRetriever: Current or fallback config was unavailable, using default.
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/download/AppConfigServer.kt
```diff
@@ -55,12 +53,20 @@ class AppConfigServer @Inject constructor(
             exportBinary
         }
 
+        // If this is a cached response, we need the original timestamp to calculate the time offset
+        val localTime = response.getCacheTimestamp() ?: timeStamper.nowUTC
+
+        // Shouldn't happen, but hey ¯\_(ツ)_/¯
```
harambasicluka, line 59: 🥰

harambasicluka (commented):
LGTM, also works on my device which crashed before.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/download/DefaultAppConfigSource.kt
```diff
@@ -0,0 +1,30 @@
+package de.rki.coronawarnapp.appconfig.download
+
+import android.content.Context
+import dagger.Reusable
+import de.rki.coronawarnapp.util.HashExtensions.toSHA256
+import de.rki.coronawarnapp.util.di.AppContext
+import javax.inject.Inject
+
+@Reusable
+class DefaultAppConfigSource @Inject constructor(
```
mlenkeit, line 10: For now, this should only be required in the tests, so I would suggest to move it there.

d4rken: Why would this only be required in tests?

mlenkeit: The SHA256 thingy is only required for tests

harambasicluka (approved):

ralfgehrer (approved):
Works like a charm. Could almost be considered offline-mode ready ;)

PR #1556, opened 2020-11-09
Reminder to share a positive test result (EXPOSUREAPP-3551)

ralfgehrer:
- Notification is shown 2 hours after the user has seen the positive test result (after he opened the screen with the positive test result) to remind the user in case the test result was not shared
- After 4 hours an additional reminder is shown if the test result is still not shared-
- No further reminder is shown
- On Android, the scheduling might not be exactly 2h
- A click on the notification directly leads to the test result screen from where others can be warned
- After a reset of the App the notification shall be canceled

**How to test:**
To shorten the time span, developers can adopt the duration in the NotificationConstants class (see POSITIVE_RESULT_NOTIFICATION_INITIAL_OFFSET and POSITIVE_RESULT_NOTIFICATION_INTERVAL).

1.  Scan a positive QR Code or TeleTan.
2. The notification will be triggered either in (a) the positive test result screen or (b) on the home screen that shows a positive test result for the first time.

positivetestresultnotification


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationReceiver.kt
```diff
@@ -0,0 +1,51 @@
...
+            POSITIVE_RESULT_NOTIFICATION_ID ->
+                showPositiveResultNotification(context, requestCode)
+            else ->
+                Timber.tag(TAG).v("NotificationReceiver received an undefined request code: %s", requestCode)
+        }
+    }
+
+    private fun showPositiveResultNotification(context: Context, notificationId: Int) {
+        if (LocalData.numberOfRemainingPositiveTestResultReminders > 0) {
+            LocalData.numberOfRemainingPositiveTestResultReminders -= 1
```
chris-cwa, line 30: As this is the bc receiver class it is only supposed to be the access point for triggers (application layer). I would really love all the reminder logic to be extracted to a class within the domain layer.
An injectable class that could maybe have those members for:
- resetReminding() -> when reminding is needed, because submission was done
- isRemindingNeeded() -> whether or not to display reminders
- onReminderShownOnce() -> when the reminder Is shown

This way we have a single point of responsibility for the reminder logic. I guess testing would also be simplified.

chris-cwa (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -74,6 +78,11 @@ class HomeFragmentViewModel @AssistedInject constructor(
 
     private var isLoweredRiskLevelDialogBeingShown = false
 
+    suspend fun observeTestResultToSchedulePositiveTestResultReminder() =
+        submissionCardsStateProvider.state
+            .first { it.isPositiveSubmissionCardVisible() }
```
ralfgehrer, line 83: After an app reset via the settings menu, the deviceUiState is still "PAIRED_POSITIVE" which will falsely trigger the notification. This needs to be fixed.

chiljamgossow (commented):
Ok

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
ralfgehrer: Resolved. The SubmissionRepository state has not been reset on app reset.

chris-cwa (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationReceiver.kt
```diff
@@ -0,0 +1,31 @@
...
+
+class NotificationReceiver : BroadcastReceiver() {
+
+    @Inject lateinit var testResultNotificationService: TestResultNotificationService
+
+    override fun onReceive(context: Context, intent: Intent) {
+        AndroidInjection.inject(this, context)
+        when (val notificationId = intent.getIntExtra(NOTIFICATION_ID, Int.MIN_VALUE)) {
+            POSITIVE_RESULT_NOTIFICATION_ID ->
+                testResultNotificationService.showPositiveTestResultNotification(notificationId)
```
harambasicluka, line 22: Maybe also add a log here, to be sure that it was executed, otherwise we only see the `else` path.

harambasicluka (commented):
LGTM, will do device testing now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationReceiver.kt
ralfgehrer: The NotificationHelper logs all built notifications but will add it anyway in case it gets lost somewhere in between

harambasicluka (approved):

PR #1560, opened 2020-11-09
"Invalid Tan" message is displayed when the TAN input screen is opened (EXPOSUREAPP-3646)

Oliver-Zimmerman:
### Description

As soon as the user opens the screen for teletan input, a message appears stating that the TAN is invalid. This message remains until the complete TAN has been entered.

**Expected**: Only if the TAN is invalid the message should appear

**Fix**: Added check to only do valid tan check if the length of the TAN is at the max length. Changed unit tests to reflect this. 

### Steps to reproduce

Go to TAN input screen, enter an invalid TAN - the error will now only appear once the user has filled in all the spaces.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
@@ -10,7 +10,7 @@ data class Tan(
 
     val areCharactersValid = allCharactersValid(value)
     val isTanValidFormat = value.length == MAX_LENGTH && isChecksumValid(value)
-    val isTanValid = areCharactersValid && isTanValidFormat
+    val isTanValid = if (value.length != MAX_LENGTH) true else areCharactersValid && isTanValidFormat
 
     companion object {
         const val MAX_LENGTH = 10
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
@@ -8,10 +8,10 @@ data class Tan(
     val value: String
 ) {
 
-    val areCharactersValid = allCharactersValid(value)
-    val isTanValidFormat = value.length == MAX_LENGTH && isChecksumValid(value)
-    val isTanValid = areCharactersValid && isTanValidFormat
     val isCorrectLength = value.length == MAX_LENGTH
+    val areCharactersValid = allCharactersValid(value)
+    val isTanValidFormat = isCorrectLength && isChecksumValid(value)
+    val isTanValid = areCharactersValid && isTanValidFormat
 
     companion object {
         const val MAX_LENGTH = 10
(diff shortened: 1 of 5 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
```diff
@@ -10,7 +10,7 @@ data class Tan(
 
     val areCharactersValid = allCharactersValid(value)
     val isTanValidFormat = value.length == MAX_LENGTH && isChecksumValid(value)
-    val isTanValid = areCharactersValid && isTanValidFormat
+    val isTanValid = if (value.length != MAX_LENGTH) true else areCharactersValid && isTanValidFormat
```
ralfgehrer, line 13: please add some tests

BMItr (dismissed):
code lgtm. tested on Pixel 2 device

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
ralfgehrer: Also, shouldn't it rather be (value.length < MAX_LENGTH)

ralfgehrer (commented):
Works well on Pixel (Android 10)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/tan/Tan.kt
Kolpa: just from a readabilty point of view id prefer if the check for no input is performed in a different place, isTanValid isn't reall true if there is no tan at all.

ralfgehrer (dismissed):

Kolpa (changes requested):
just from a readabilty point of view id prefer if the check for no input is performed in a different place, isTanValid isn't reall true if there is no tan at all.

Oliver-Zimmerman:
> just from a readabilty point of view id prefer if the check for no input is performed in a different place, isTanValid isn't reall true if there is no tan at all.

@Kolpa 
It's not necessarily a check for no input, the issue was that the tan was always invalid when it wasn't at full length so the error message was always there. This addition was just to make sure that the tan validation check is only done if the tan is at full length (complete)

Kolpa:
Just to make my point clear :D i totally agree with the solution and i like it! But the property is called isTanValid, which is quite simply, not true for a tan that is shorter than max length.

So i propose that you do the length check in a different spot and use it in the formatter to decide to not show the error even if the tan is technically "incorrect".

I think doing that way would help a lot with readability for further people looking at this code. Since a check for is tan valid should not return true when a tan is actually invalid but we don't want to display the error.

Oliver-Zimmerman:
@Kolpa 
Noted and implemented. Sorry I missed the initial point of your comment. I have separated the length and validity logic so that it reads better.

kolyaopahle (approved):
Perfect, thats exactly what i wanted LGTM 👍

BMItr (approved):

PR #1563, opened 2020-11-09
Implement specific logs to analyse (EXPOSUREAPP-3440)

AndroidMedaGalaxy:
### Description
Log improvement where the PR tries to introduce more point of logs and information about Background Workers and their operations

### Steps to reproduce
Test Menu > Debug Options > Logfiled enabled togled to true > Share Log button > Observe the saved file location

How to get to the file.  
1.  You can use Device Explorer from Android Studio
2.  Can use File explorer in the device to get to the file


harambasicluka (changes requested):
IMHO the ticket was about adding a log when the high risk notification is send and also log "the notification permissions for the app to ensure everything is configured right". So please add logs in this locations.

ritsam:
> IMHO the ticket was about adding a log when the high risk notification is send and also log "the notification permissions for the app to ensure everything is configured right". So please add logs in this locations.

Hi Luka,

The changes have been pushed. Please check them out

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/BackgroundNoiseOneTimeWorker.kt
```diff
@@ -38,6 +40,7 @@ class BackgroundNoiseOneTimeWorker @AssistedInject constructor(
             }
         }
 
+        Timber.d("$id: doWork() finished with %s", result)
```
chiljamgossow, line 43: I there a reason you do not write Timber.d("$id: doWork() finished with **$result**")?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
```diff
@@ -43,19 +43,20 @@ class DiagnosisTestResultRetrievalPeriodicWorker @AssistedInject constructor(
...
 
-        Timber.d("Background job started. Run attempt: $runAttemptCount")
+        Timber.d("$id: doWork() started. Run attempt: $runAttemptCount")
         BackgroundWorkHelper.sendDebugNotification(
             "TestResult Executing: Start", "TestResult started. Run attempt: $runAttemptCount "
         )
 
         if (runAttemptCount > BackgroundConstants.WORKER_RETRY_COUNT_THRESHOLD) {
-            Timber.d("Background job failed after $runAttemptCount attempts. Rescheduling")
+            Timber.d("$id Background job failed after $runAttemptCount attempts. Rescheduling")
```
chiljamgossow, line 52: Minor: Maybe also use doWork() instead of Background job in the log here for consistency?

chiljamgossow (dismissed):
Some minor suggestions. Since it is only for debugging, it's also fine if you leave it as it is.

harambasicluka (dismissed):
It's fine from my point of view but please discuss the open points with @chiljamgossow :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
ritsam: Makes sense.

harambasicluka (approved):

chiljamgossow (approved):
Looks fine to me

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/BackgroundNoiseOneTimeWorker.kt
d4rken: Doesn't matter here I think, but the difference would be:

The solution with `%s` using string formatting would defer the `toString` evaluation until it's actually needed. So in the scenario where `result` is a complex object (i.e. a list of complex objects), we would not evaluate it `toString()` if logging is disabled (internal optimization in Timber).

PR #1567, opened 2020-11-10
Verify right risk state: Exposure Logging stopped vs. Unknown Risk (EXPOSUREAPP-3594)

Oliver-Zimmerman:
### Description

Bug where the risk card stays at Unknown Risk when Exposure Logging is no longer active, rather than Exposure Logging Stopped. 

### Fix
Added various checks to TracingCard state that make sure TracingStatus isn't set to Inactive. if it is, then it defaults to the tracing off option. 

_NOTE: I feel like this is fixing a symptom rather than the root cause. Perhaps a better method than these additional checks is checking why the risk score isn't being set to NO_CALCULATION_POSSIBLE_TRACING_OFF. This is because risk calculation is done daily I believe_

### Steps to reproduce
Check out from release/1.7.x
Clean install the app
Proceed with onboarding (turn on exposure notification)
After onboarding go to settings and turn off exposure notification
Go back to main screen
Risk Card should now be set to Exposure Logging Stopped


chris-cwa (dismissed):

AlexanderAlferov:
Tested. Code looks good.
Waiting for additional tests 🧪 🔬

Oliver-Zimmerman:
> Tested. Code looks good.
> Waiting for additional tests 🧪 🔬

@AlexanderAlferov tests implemented! Please have a look

AndroidMedaGalaxy (approved):
New Unit tests reviewed.

LGTM ヽ(o⌣oヾ)

ralfgehrer (approved):
thx for adding the tests

BMItr (approved):

chris-cwa (approved):

AlexanderAlferov (approved):

PR #1569, opened 2020-11-10
New quota error (EXPOSUREAPP-3511)

d4rken:
A new quota calculation limit error dialog.

* Catch the error and wrap the `ApiException` in a custom one that implements `HasHumanReadableError` which provides localized title and string
* Forward that to the error reporting routines and make sure that it displays the title, if available, and otherwise use a fallback.
* Removed the `"Tägliche Aktualisierung"` label from the risk card.
* I also made the StackTrace text selectable. Not great, but not as terrible as having to screenshot the error dialog multiple times. 

Screenshot from 2020-11-10 20-15-32

## Testing

Screenshot from 2020-11-10 20-14-43


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/exception/reporting/ErrorReportReceiver.kt
```diff
@@ -52,31 +47,44 @@ class ErrorReportReceiver(private val activity: Activity) : BroadcastReceiver()
...
-                            DialogHelper.DialogInstance(
-                                activity,
-                                title,
-                                "$detailsTitle:\n$stack",
-                                confirm
-                            )
-                        ).run {}
-                    }
-                ))
+        val dialogTitle = if (intent.hasExtra(ReportingConstants.ERROR_REPORT_TITLE_EXTRA)) {
```
ralfgehrer, line 50: Minor: could be moved to a dedicated function in order to get rid of the @Suppress("LongMethod")

ralfgehrer:
Popup did not show on my Pixel (Android 10)

Screenshot 2020-11-11 at 10 50 25

d4rken:
@ralfgehrer Was the app in the foreground? Did the error reporter trigger? Check further down in the log please.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/exception/reporting/ErrorReportReceiver.kt
d4rken: I currently don't see an easy way to restructure the code due to overlapping use of properties (e.g. `title`). I'd prefer to not refactor it in this PR, the whole error reporting via intents could need an overhaul too :frowning: 

Would it be okay to leave it as is?

ralfgehrer: Sure, I don't expect minors to be fixed.

ralfgehrer:
> @ralfgehrer Was the app in the foreground? Did the error reporter trigger? Check further down in the log please.

Yes, was in foreground. Reporter did not trigger. In fact, not to be found in the log anywhere.

harambasicluka:
I get the expected popup.

ralfgehrer (dismissed):
Works as expected (local copy of the branch was outdated. Sorry for the false alarm).

ralfgehrer (approved):

harambasicluka (approved):

janetback (approved):
UA reviewed

PR #1572, opened 2020-11-11
Override EXPECT_NEW_PACKAGES if we revoked key packages (EXPOSUREAPP-3455)

d4rken:
We need to set `force index lookup` to `true` if we revoked any local key packages, as we might still have the newest package, so the check (i.e. `expectNewDayPackages`) could return `false` because we revoked a key package that wasn't the newest one (which is what the check mostly looks at).


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
```diff
@@ -14,18 +14,28 @@ open class BaseKeyPackageSyncTool(
     private val tag: String
 ) {
 
-    internal suspend fun revokeCachedKeys(revokedKeyPackages: Collection<KeyDownloadConfig.RevokedKeyPackage>) {
+    /**
+     * Returns true if any of our cached keys were revoked
+     */
+    internal suspend fun revokeCachedKeys(
```
ralfgehrer, line 20: Optional: add some tests for this

ralfgehrer (approved):
Looks good.

kolyaopahle (approved):
LGTM 👍

harambasicluka (commented):
Tested on device

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
d4rken: https://github.com/corona-warn-app/cwa-app-android/pull/1580

PR #1573, opened 2020-11-11
Reduce appconfig calls for exposure detection tracker (DEV)

d4rken:
`getAppConfig()` may force a config refresh, we don't need that though for the time tracker on each check interval, the trackers timeout check is fine just working with the latest available config.


ralfgehrer (approved):

kolyaopahle (approved):
LGTM 👍

PR #1575, opened 2020-11-11
Parse test JSON files and init unit tests with respective parameters (EXPOSUREAPP-3456)

AlexanderAlferov:
Created ExposureWindowsCalculationTest in de.rki.coronawarnapp.nearby.windows

Added multiple logs level for better debugging: right now only configuration and comparison table is shown:
`private val logLevel = LogLevel.ONLY_COMPARISON`

Test class contains only one test that should cover everything.


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/windows/ExposureWindowsCalculationTest.kt
```diff
@@ -0,0 +1,400 @@
...
+        // 1 - Load and parse json file
+        val jsonFile = Paths.get("src", "test", "resources", fileName).toFile()
+        jsonFile shouldNotBe null
+        val jsonString = FileReader(jsonFile).readText()
+        jsonString.length shouldBeGreaterThan 0
+        val json = Gson().fromJson<ExposureWindowsJsonInput>(jsonString)
+        json shouldNotBe null
+
+        // 2 - Check test cases
+        json.testCases.map { case -> checkTestCase(case) }
```
BMItter, line 95: map fun.., someone usually expects some data is mapped by this fun (not a check).
using 'forEach' might do it... :wink:

BMItr (approved):
test cases are fine and behaving with expected result! good work :+1:
left one comment..

PR #1583, opened 2020-11-12
Detect changes to app config (EXPOSUREAPP-2718)

chris-cwa:
You can test by locally setting LocalData.lastConfigId("dummy") on app creation. RiskLevelTask request must be submitted to task controller.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
@@ -1,0 +1,24 @@
+package de.rki.coronawarnapp.appconfig
+
+import de.rki.coronawarnapp.risk.RiskLevel
+import de.rki.coronawarnapp.risk.RiskLevelTask
+import de.rki.coronawarnapp.storage.LocalData
+import de.rki.coronawarnapp.storage.RiskLevelRepository
+import de.rki.coronawarnapp.task.TaskController
+import de.rki.coronawarnapp.task.common.DefaultTaskRequest
+import kotlinx.coroutines.flow.onEach
+import javax.inject.Inject
+
+class ConfigChangeDetector @Inject constructor(
+    private val appConfigProvider: AppConfigProvider,
+    private val taskController: TaskController
+) {
+
+    fun start() = appConfigProvider.currentConfig.onEach {
+        if (it.identifier != LocalData.lastConfigId()) {
+            RiskLevelRepository.setRiskLevelScore(RiskLevel.UNDETERMINED)
+            taskController.submit(DefaultTaskRequest(RiskLevelTask::class))
+            LocalData.lastConfigId(it.identifier)
+        }
+    }
+}
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
@@ -2,27 +2,51 @@ package de.rki.coronawarnapp.appconfig
 
 import androidx.annotation.VisibleForTesting
 import de.rki.coronawarnapp.risk.RiskLevel
+import de.rki.coronawarnapp.risk.RiskLevelData
 import de.rki.coronawarnapp.risk.RiskLevelTask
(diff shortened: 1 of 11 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -0,0 +1,36 @@
...
+import de.rki.coronawarnapp.task.common.DefaultTaskRequest
+import kotlinx.coroutines.flow.onEach
+import javax.inject.Inject
+
+class ConfigChangeDetector @Inject constructor(
+    private val appConfigProvider: AppConfigProvider,
+    private val taskController: TaskController
+) {
+
+    fun start() = appConfigProvider.currentConfig.onEach { check(it.identifier) }
```
chris-cwa, line 18: must collect here

harambasicluka:
Why is this PR a draft, I would like to have it in 1.7.0/RC3 tomorrow. @chris-cwa & @d4rken do you think that's feasible?

d4rken:
I've moved "saving the last config identifier" into the risk calculation task. Otherwise we run into the racecondition that there may already be a task running, with the old config, and we don't start a new calculation because the task uses `SKIP_IF_SIBLING_RUNNING`, but still save the new config ID as the last one used.

Calculating  a new risk level is cheap and we can do it rather one time too much than one time less than we needed.

To prevent a task submission loop like:

`new config -> new task (retrieves new config) -> new config -> new task`

we filter config emissions with [`distinctUntilChangedBy`](https://github.com/corona-warn-app/cwa-app-android/pull/1583/files#diff-9a68f259a436ecaa2f4a4191a234f5b2e3147fd4d26b107598ac5b60cb487b79R29).

"Wort case edge case" would be an overlap in tasks such that our task submission collides with an existing one that still works with the old app config. In that case the old config ID would be saved by the task, and on next app launch, the config with the new ID will trigger a risk level reset + task submission again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelData.kt
```diff
@@ -0,0 +1,30 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class RiskLevelData @Inject constructor(
+    @AppContext private val context: Context
+) {
+
+    private val prefs by lazy {
+        context.getSharedPreferences("risklevel_localdata", Context.MODE_PRIVATE)
```
ralfgehrer, line 15: You might want to use a const val for the key

ralfgehrer (approved):

BMItr (approved):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/risk/RiskLevelDataTest.kt
```diff
@@ -0,0 +1,40 @@
...
+        every { context.getSharedPreferences("risklevel_localdata", Context.MODE_PRIVATE) } returns preferences
+    }
+
+    fun createInstance() = RiskLevelData(context = context)
+
+    @Test
+    fun `update last used config identifier`() {
+        createInstance().apply {
+            lastUsedConfigIdentifier shouldBe null
+            lastUsedConfigIdentifier = "Banana"
```
chiljamgossow, line 31: A val for Banana?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -0,0 +1,63 @@
...
+            Timber.i("New config id ($newIdentifier) differs from last one ($oldConfigId), resetting.")
+            RiskLevelRepositoryDeferrer.resetRiskLevel()
+            taskController.submit(DefaultTaskRequest(RiskLevelTask::class))
+        } else {
+            Timber.v("Config identifier ($oldConfigId) didn't change, NOOP.")
+        }
+    }
+
+    @VisibleForTesting
+    internal object RiskLevelRepositoryDeferrer {
```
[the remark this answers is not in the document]
d4rken: A workaround for mocking. The simple object here can mocked, while there were issues mocking the `RiskLevelRepository`.

This indirection should be removed once `RiskLevelRepository` is no longer a singleton.

PR #1589, opened 2020-11-12
Never report CancellationException (closes #1588)

### LINKED ISSUE #1588: Application reports that something went wrong with JobCancellationException
## Describe the bug

See this tweet: https://twitter.com/filippovdmitry/status/1326603774199291904?s=20

## Expected behavior

Canceling an internal operation is normal behavior and should not have been reported as "Something went wrong".

---
Internal Tracking ID:  EXPOSUREAPP-3974

elizarov:
There is a number of places in the app with the following code pattern:

```
try {
    // do some suspending operation
} catch(e: Exception) {
    e.report(ExceptionCategory.EXPOSURENOTIFICATION) // or similar
}
```

In all such places if the corresponding suspending gets cancelled for any reason, the resulting CancellationException will get reported as an application crash. This fix prevents it from happening, since a cancellation exception is a part of a normal operation, not a failure that should be reported.

Fixes #1588


d4rken:
Can you change the base branch to `release/1.7.x` or `release/1.8.x` ?

elizarov:
Rebased onto `release/1.8.x`

PR #1591, opened 2020-11-12
Improve cancelation exception handling (DEV)

d4rken:
Roman raised a good point in #1588  

We have a few operations launched in viewmodels (on the viewmodelscope).
While most are quite quick, each could raise a cancellation exception (if the stars align right).

None of these coroutines actually wrrant a crash on cancellation though.
Most refactored code already uses the `CWAViewModel.launch` method to launch suspending functions, so I think it would be good coding to catch cancellation exceptions there and just log them.

This PR catches the cancellation exception for the `CWAViewModel.launch` method, and refactores few calls that didn't use it, to use it.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModel.kt
```diff
@@ -29,7 +29,13 @@ abstract class CWAViewModel constructor(
     fun launch(
         context: CoroutineContext = dispatcherProvider.Default,
         block: suspend CoroutineScope.() -> Unit
-    ): Job = viewModelScope.launch(context = context, block = block)
+    ) {
+        try {
+            viewModelScope.launch(context = context, block = block)
+        } catch (e: CancellationException) {
+            Timber.w(e, "launch()ed coroutine was canceled.")
```
chris-cwa, line 36: what's with the brackets? severity should be e in my opinion

chris-cwa (approved):

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModel.kt
d4rken: > what's with the brackets?

Wordplay on the functions name :grin: 

> severity should be e in my opinion

As argued by Roman in #1588 a cancelation is not an error but part of normal behavior. Reporting this as error would distract us from actual errors. It's still a warning and not a `.d` because we might want to look into what operations regularly run into this.

PR #1597, opened 2020-11-13
Risk Card has "Unknown Risk" after 1.7.0 (RC2) Update (EXPOSUREAPP-3723)

Oliver-Zimmerman:
### Description

Risk card is set to unknown after an update, regardless of state stored in the previous version.  

### Fix
Initial risk score in RiskLevelRepository was set to UNKNOWN_RISK_INITIAL, it is now changed to the last successfully calculated score which is stored in LocalData. If there isn't one (ie. Fresh install), it is set to the UNDETERMINED state.

### Steps to reproduce
1. App is installed 1.7.0(1) or 1.6 RC7
2. Risk Card  has a state (green or red) - timetravel forward a day if necessary to get LOW risk level
3. do the update to 1.7.0 (2)
4. Previously, the risk level would initially be unknown. Now it should stay the same.


kolyaopahle (approved):
looks kind of weird without error handlingt at first but since LocalData.getLastSuccessfullRisk always returns Undetermined if nothing has been written yet i guess this is fine 👍

kolyaopahle:
actually, on a second note, is this fine? as before it would be UNKNOWN_RISK_INITIAL and now, if no risk is set it is UNDETERMINED. Could this cause any issues?

Oliver-Zimmerman:
> actually, on a second note, is this fine? as before it would be UNKNOWN_RISK_INITIAL and now, if no risk is set it is UNDETERMINED. Could this cause any issues?

I believe this will only ever happen in the case of a fresh install, as soon as the appropriate amount of time has passed, it will change to the correct risk level. In the case of a fresh install and having it set to Unknown, the same thing happens.

kolyaopahle:
> @kolyaopahle
I believe this will only ever happen in the case of a fresh install, as soon as the appropriate amount of time has passed, it will change to the correct risk level. In the case of a fresh install and having it set to Unknown, the same thing happens.
Also, if you check the function setLastCalculatedRiskLevelAsCurrent() in RiskLevelRepistory, you can see that if it is set to UNDETERMINED, it gets set to UNKNOWN_RISK_INITIAL

sounds good :)

ralfgehrer (approved):

PR #1601, opened 2020-11-13
Sonar cloud scan for external pr (DEV)

AlexanderAlferov:
Enable Sonar Cloud Scan for external PRs  📡 ☁️ 🔍


d4rken (commented):
I'm wondering whether this is has the same concerns als CircleCI on external PRs, could we expose secrets? :thinking: 

I'm also a bit confused because when searching for security related topics to SonarCloud and external pull requests, what I read sounds like SonarCloud on external PRs is a feature that is not yet implemented?

e.g.
https://jira.sonarsource.com/browse/MMF-1371
https://github.com/viadee/camunda-kafka-polling-client/issues/15

On second look it looks like this is only an issue if TravisCI launches SonarCube as TravisCI does not have the option to hide secrets from external PRs :eyes: 

Let's talk about this shortly on monday to make sure it behaves as expected.

ralfgehrer (approved):

d4rken (approved):

PR #1603, opened 2020-11-13
Diagnosis keys task uses window mode (EXPOSUREAPP-3538)

chris-cwa:


BMItter:
ExposureSummary is deprected and should no be used anymore.
Use ExposureWindows now -> ENFClient.kt :ok_hand:

chris-cwa:
waiting for other end v2 branches to merge to 3456

PR #1584, opened 2020-11-12
Enhance enf client (EXPOSUREAPP-3540)

BMItr:
This PR implements one additional part of the new ExposureWindowMode (Enhance ENFClient).

- new ExposureWindowProvider with Interface
- new provideDiagnosisKeys with Interface
- adjusted ENFclientTest to work properly
- deprecated outdated funs (LegacyMode)

This PR does not address the new calculation tracking (later PR).

PR #1605, opened 2020-11-13
ENFClient v2 - ExposureDetectionTracker without identifier (EXPOSUREAPP-3540)

BMItr:
This PR implements one additional part of the new ExposureWindowMode (#1584 ).

We don't have a token anymore and can't provide one to the ExposureDetectionTracker.
This PR implements the opportunity to use ExposureDetectionTracker with & without identifier.
Note: ExposureDetectionTracker was CalculationTracker some days ago (just to prevent confusion).

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
@@ -40,7 +40,7 @@ class ENFClient @Inject constructor(
         configuration: ExposureConfiguration?,
         token: String
     ): Boolean {
-        //NO-OP
+        // NO-OP
         return false
     }
 
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
@@ -39,30 +39,17 @@ class ENFClient @Inject constructor(
         keyFiles: Collection<File>,
         configuration: ExposureConfiguration?,
         token: String
-    ): Boolean = submitDiagnosisKeys(keyFiles, configuration, token)
-
-    override suspend fun provideDiagnosisKeys(keyFiles: Collection<File>): Boolean = submitDiagnosisKeys(keyFiles)
-
-    private suspend fun submitDiagnosisKeys(
-        keyFiles: Collection<File>,
-        configuration: ExposureConfiguration? = null,
-        token: String? = null
     ): Boolean {
-        Timber.d(
-            "asyncProvideDiagnosisKeys(keyFiles=%s, configuration=%s, token=%s)",
-            keyFiles, configuration, token
-        )
+        //NO-OP
+        return false
+    }
 
-        if (keyFiles.isEmpty()) {
+    override suspend fun provideDiagnosisKeys(keyFiles: Collection<File>): Boolean {
+        Timber.d("asyncProvideDiagnosisKeys(keyFiles=$keyFiles)")
+
+        return if (keyFiles.isEmpty()) {
             Timber.d("No key files submitted, returning early.")
-            return true
-        }
-
-        Timber.d("Forwarding %d key files to our DiagnosisKeyProvider.", keyFiles.size)
-
-        return if (token != null) {
-            exposureDetectionTracker.trackNewExposureDetection(token)
-            diagnosisKeyProvider.provideDiagnosisKeys(keyFiles, configuration, token)
+            true
         } else {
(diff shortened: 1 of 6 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
```diff
@@ -40,19 +40,8 @@ class ENFClient @Inject constructor(
...
-
-        return if (keyFiles.isEmpty()) {
-            Timber.d("No key files submitted, returning early.")
-            true
-        } else {
-            Timber.d("Forwarding %d key files to our DiagnosisKeyProvider.", keyFiles.size)
-            exposureDetectionTracker.trackNewExposureDetection(token)
-            diagnosisKeyProvider.provideDiagnosisKeys(keyFiles, configuration, token)
-        }
+        // NO-OP
```
d4rken, line 43: Maybe `throw UnsupportedOperationException()` until this is removed?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/DefaultExposureDetectionTracker.kt
```diff
@@ -93,34 +94,16 @@ class DefaultExposureDetectionTracker @Inject constructor(
         }
     }
 
-    override fun finishExposureDetection(identifier: String, result: Result) {
-        Timber.i("finishExposureDetection(token=%s, result=%s)", identifier, result)
```
d4rken, line 97: Please re-add a log statement directly at function start, the following methods using `updateSafely` are async, if there is any issue there we'd still like to know that the function was called, but failed somewhere later.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/DefaultExposureDetectionTracker.kt
```diff
@@ -134,6 +117,54 @@ class DefaultExposureDetectionTracker @Inject constructor(
         }
     }
 
+    private fun finishExposureDetectionWithoutIdentifier(
```
d4rken, line 120: Would an extension function improve readability here? 

Something like `fun Map<String, TrackedExposureDetection>.findOrCreateByIdentifier(identifier: String?)`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/DefaultExposureDetectionTracker.kt
```diff
@@ -134,6 +117,54 @@ class DefaultExposureDetectionTracker @Inject constructor(
...
+    ) {
+        Timber.d(
+            "finishExposureDetectionWithoutIdentifier(): " +
+                "Get identifier of newest unfinished detection or create new identifier"
+        )
+        val identifier = map
+            .map { it.value }
+            .filter { it.finishedAt == null }
+            .maxByOrNull { it.startedAt.millis }
+            ?.identifier ?: kotlin.run {
```
d4rken, line 132: A bit difficult to read. How could we make this easier on the :eyes:

BMItter: good point. 
_?.identifier ?: kotlin.run_ is out.
readability improved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/DefaultExposureDetectionTracker.kt
BMItter: I changed it to:
private fun Map<String, TrackedExposureDetection>.findUnfinishedOrCreateIdentifier(): String {

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/DefaultExposureDetectionTracker.kt
BMItter: added.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ENFClient.kt
BMItter: throwing UnsupportedOperationException() now

kolyaopahle (approved):
LGTM 👍 all handleable changes seem solved, lets get this merged to get the main branch ready, tests can be fixed later before we merge into release

PR #1608, opened 2020-11-16
Fix passive config requests triggering a refresh (EXPOSUREAPP-3774)

d4rken:
Make config sharing behavior `SharingStarted.Lazily`. (start once, never release)

Otherwise classes that are just interested in the current config, if there is no other subscriber active,
would launch the `startValueProvider` and pull a config each time.

The ExposureDetectionTracker is such a candidate that would like to use the latest config every ~3min.

While this already no longer happens since "Track config changes" (4d1ecb5),
because there is now a class that is permanently subscribed to `currentConfig`, we should not rely on side-effects,
so let's make the sharing behavior explicit.


kolyaopahle (approved):
Tests are passing and i can confirm that this also fixes fetching behaviour that I've seen while building the ENFv2 calculations 👍

chiljamgossow (approved):

PR #1612, opened 2020-11-16
Improve HotDataFlow behavior (EXPOSUREAPP-3777)

d4rken:
* Prevent re-execution of past submissions when the observable goes cold.
* Guard internal value updates with a mutex in case the value update is not as thread-safe as we think it is.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
@@ -15,6 +15,8 @@ import kotlinx.coroutines.flow.onCompletion
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.shareIn
 import kotlinx.coroutines.plus
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import timber.log.Timber
 import kotlin.coroutines.CoroutineContext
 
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
```diff
@@ -36,19 +38,24 @@ class HotDataFlow<T : Any>(
...
-        updateActions.collect { updateAction ->
-            currentValue = updateAction(currentValue).also {
-                currentValue = it
-                send(it)
+        updateActions
+            .onCompletion {
+                Timber.tag(tag).v("updateActions resetReplayCache().")
+                updateActions.resetReplayCache()
+            }
+            .collect { updateAction ->
```
d4rken, line 54: I'm suspicious that our access of `currentValue` in `.collect { }` was not thread-safe, which is why I added the mutex.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
```diff
@@ -36,19 +38,24 @@ class HotDataFlow<T : Any>(
...
+        Timber.tag(tag).v("startValue=%s", currentValue)
 
-        updateActions.collect { updateAction ->
-            currentValue = updateAction(currentValue).also {
-                currentValue = it
-                send(it)
+        updateActions
+            .onCompletion {
+                Timber.tag(tag).v("updateActions resetReplayCache().")
+                updateActions.resetReplayCache()
```
d4rken, line 52: When we unsubscribe, and the `HotDataFlow` goes cold, and then resubscribe, we would replay all previous update actions on the new value.

I don't think this affected the current issue, but it's a bug nonetheless.

ralfgehrer (dismissed):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/flow/HotDataFlowTest.kt
```diff
@@ -122,6 +123,48 @@ class HotDataFlowTest : BaseTest() {
         coVerify(exactly = 1) { valueProvider.invoke(any()) }
     }
 
+    data class TestData(
+        val number: Long = 1
+    )
+
+    @Test
+    fun `value updates 2`() {
```
chiljamgossow, line 131: You could be more verbose here describing what the test is checking

chiljamgossow (dismissed):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/flow/HotDataFlowTest.kt
d4rken: Done

BMItr (approved):
Code lgtm. Test are running fine now.

ralfgehrer (approved):

PR #1616, opened 2020-11-16
Don't limit the DownloadDiagnosisKeysTask to once a day (EXPOSUREAPP-3771)

ralfgehrer:
Expected result:
- Regular exposure checks are visible in the exposure log around every 4 hours (increased frequency for exposure checks)

Note for testing:
- only time travel into the future, please.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -135,9 +135,9 @@ class TracingRepository @Inject constructor(
         )
 
         // check if the keys were not already retrieved today
-        val keysWereNotRetrievedToday =
+        val keysWereNotAlreadyRetrieved =
             LocalData.lastTimeDiagnosisKeysFromServerFetch() == null ||
-                currentDate.withTimeAtStartOfDay() != lastFetch.withTimeAtStartOfDay()
```
chiljamgossow, line 140: Yes, that was awkward as future days and past days would result in true

chiljamgossow (dismissed):

chiljamgossow:
I couldn't test on a device (do not have one yet)

harambasicluka:
Will do the device testing for @chiljamgossow .

Oliver-Zimmerman:
LGTM :+1:

Tested on Samsung S9 Plus

Oliver-Zimmerman (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -198,13 +198,13 @@ class DownloadDiagnosisKeysTask @Inject constructor(
             DateTimeZone.UTC
         )
         return (LocalData.lastTimeDiagnosisKeysFromServerFetch() == null ||
-            currentDate.withTimeAtStartOfDay() != lastFetch.withTimeAtStartOfDay()).also {
+            currentDate.isAfter(lastFetch)).also {
```
d4rken, line 201: What effect will this have? Won't this always be try, because unless we have a sub-milliseconds racecondition, `timeStamper.nowUTC` will always be after `LocalData.lastTimeDiagnosisKeysFromServerFetch()`.

d4rken (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -135,9 +135,9 @@ class TracingRepository @Inject constructor(
...
+                currentDate.isAfter(lastFetch)
```
d4rken, line 140: Same here, against what case of unwated execution does the check of `currentDate.isAfter(lastFetch)` protect us?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
chris-cwa: i agree. what harm would it do if we just delete the complete condition?

d4rken (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
d4rken: I think it's already effectively removed, because `currentDate.isAfter(lastFetch)`  will almost always return true.

d4rken:
It should be tested how often the task is now executed, especially when opening/closing the UI.

We should observe the webrequests in the log. Even if we run the task more often:
* `KeyPackageSyncTool` should not make a network request unless new day or hour packages are expected (based on last  data and current timestamp)
* AppConfig will get called each time. This may still be an issue, because despite our current caching we will still load the app-config headers from the server to compare the ETag against the one we have cached.

After that, before submission we would evaluate these conditions and abort if called too often: https://github.com/corona-warn-app/cwa-app-android/blob/996121d8c06a2dff4e3456dc909d94d698c945af/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt#L90

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
ralfgehrer: Yes, my thinking too. The check solely protects us from someone time traveling backward.
IMHO all these checks should not be implemented by the Task but rather by the KeyPackageSyncTool.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
ralfgehrer: see above. Same check.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
Oliver-Zimmerman: Yes, I tested again and it seems it is always returning true.

harambasicluka (commented):
Could you please add some tests. Will let it run over the night and check the logs tomorrow.

harambasicluka:
clean install, that's what I see directly after the onboarding
image

I only got the "Unknown Risk" after a force close.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
ralfgehrer: Implemented this differently now. Making sure the task if only run once every hour.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
ralfgehrer: Removed it entirely while ensuring that the task is only being run once every hour.

ralfgehrer:
> clean install, that's what I see directly after the onboarding
> 
> I only got the "Unknown Risk" after a force close.

Fixed.

mlenkeit:
Download looks better, I'll keep it running for some time.
But after I installed the client from this branch (over 1.7.0-RC3), the risk card changed from red to green and from 4 encounters to zero without updating the timestamp. Is that just another fix that's missing in this branch here?

harambasicluka (dismissed):
Coding looks fine, ticket is solved, tested it over night, logs & exposure logs in os settings looks as expected. Also the initial wrong state is resolved.

d4rken:
> But after I installed the client from this branch (over 1.7.0-RC3), the risk card changed from red to green and from 4 encounters to zero without updating the timestamp. Is that just another fix that's missing in this branch here?

And it never changed back to the red card?

mlenkeit:
> > But after I installed the client from this branch (over 1.7.0-RC3), the risk card changed from red to green and from 4 encounters to zero without updating the timestamp. Is that just another fix that's missing in this branch here?
> 
> And it never changed back to the red card?

I'll tell you in 30 minutes when the next Risk Calculation runs...

mlenkeit:
> > But after I installed the client from this branch (over 1.7.0-RC3), the risk card changed from red to green and from 4 encounters to zero without updating the timestamp. Is that just another fix that's missing in this branch here?
> 
> And it never changed back to the red card?

It's a bit chaotic.
- previous risk calculation was at 7:56, red with 4 encounters
- next one should run no earlier than 11:56
- at 11:05, I installed this branch and the card turned green with 0 encounters
- at 11:58, I opened the app, nothing happened
- at 12:01, I terminated the app, opened it again
- client downloaded new hour package as expected, runs the risk calculation (which should have run at 11:58 already), returns to a red card with 4 encounters; timestamp did not update (still shows 07:56)

mlenkeit:
And another navigation inside the app leads to:
image

ralfgehrer:
> > > But after I installed the client from this branch (over 1.7.0-RC3), the risk card changed from red to green and from 4 encounters to zero without updating the timestamp. Is that just another fix that's missing in this branch here?
> > 
> > 
> > And it never changed back to the red card?
> 
> It's a bit chaotic.
> 
> * previous risk calculation was at 7:56, red with 4 encounters
> * next one should run no earlier than 11:56
> * at 11:05, I installed this branch and the card turned green with 0 encounters
> * at 11:58, I opened the app, nothing happened
> * at 12:01, I terminated the app, opened it again
> * client downloaded new hour package as expected, runs the risk calculation (which should have run at 11:58 already), returns to a red card with 4 encounters; timestamp did not update (still shows 07:56)
> 
> I have logs, videos, etc.

At 11:05 - the download task run - hence it won't run for another hour (with the current implementation) IF the app is not terminated (next "regular" download would be at 12:05)

At 12:01 - app was killed, download triggered by opening the app - next download should be expected at 13:01

To my understanding, the download task runs as expected BUT there seems to be some issue regarding the UI states and their update.

d4rken (approved):

harambasicluka (approved):

PR #1623, opened 2020-11-16
Refactor app config to source, check validity and improve edgecase handling (EXPOSUREAPP-3774,EXPOSUREAPP-3771)

d4rken:
We noticed earlier that despite the app-configs cache header being for our caching mechanism, we would still  check the server's headers. As we are now removing more interval checks (e.g. #1616), we want to make sure that hit the CDN only if really required, even if it's just headers.

* This PR splits up the config source into: remote (CDN), local (storage) and default (assets), because the tests were becoming difficult to understand.
* The `ConfigData` now has method `isValid(nowUTC: Instant): Boolean`.
* Edgecases related to storage we noticed in 1.6.0 are addressed.

### Test
Smoke test the app and play with the remote config test menu.
Most changed files come from refactoring the packages. The key classes are:

* [`AppConfigSource` which makes the decision which config source to use](https://github.com/corona-warn-app/cwa-app-android/pull/1623/files#diff-d00988a4dcf1d5e07b764b6758bfdc6c0691a2247efdebd071e0f3ac51d79182)
* [`RemoteAppConfigSource` which gets the config from the CDN](https://github.com/corona-warn-app/cwa-app-android/pull/1623/files#diff-94f47d27bb3296c50e47e58113a346852cf80aab77bbf84214fa45493c97aee3)
* [`LocalAppConfigSource` which loads it from disk](https://github.com/corona-warn-app/cwa-app-android/pull/1623/files#diff-5476554bde030b16b4a63bf5c5e50dc11c102897b6d7c220ed4573e4de2d7b7c)
* [`DefaultAppConfigSource` which loads from assets](https://github.com/corona-warn-app/cwa-app-android/pull/1623/files#diff-78355d503b02143bce640ef23ff2a1b4bb53f462a5c3189914ec189b30cced4f)
* [`AppConfigStorage` which should not allow invalid parsed GSON to be returned ](https://github.com/corona-warn-app/cwa-app-android/pull/1623/files#diff-f44a321d8af888969de3e495f0c08cf999fe8257b508bbd1118a1537397d3fa0)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/sources/fallback/DefaultAppConfigSource.kt
```diff
@@ -0,0 +1,31 @@
...
+) {
+
+    fun getRawDefaultConfig(): ByteArray {
+        return context.assets.open("default_app_config.bin").readBytes()
+    }
+
+    fun getConfigData(): ConfigData = ConfigDataContainer(
+        mappedConfig = configParser.parse(getRawDefaultConfig()),
+        serverTime = Instant.EPOCH,
+        localOffset = Duration.standardHours(1),
```
ralfgehrer, line 26: Why is the default +1h? If we deal with time zones properly we would not need an offset, do we?

ralfgehrer (dismissed):
Works well on Android 10 with the Pixel. Also tested edge cases such as no Wifi/Airplane mode.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/sources/fallback/DefaultAppConfigSource.kt
d4rken: Should we set it to 0?
I set it to something that is not "great" (0), but also not terrible (2+hours, which would once we validate the time offset, prevent key submission).

We could discuss whether using the default config for key submission would be acceptable? @mlenkeit  If it is the offset should be below the allowed time diff otherwise above.

d4rken: Offset from server to local time.

https://github.com/corona-warn-app/cwa-app-android/blob/a1d8465e2d988e83d56b1979ddbe5fabd85527de/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/download/AppConfigServer.kt#L64

mlenkeit: but then I would set it to zero for the default.

d4rken: Done.

ralfgehrer (approved):

mlenkeit (approved):
LGTM as far as I can tell.

PR #1624, opened 2020-11-16
Fix appconfig display for testers (DEV)

d4rken:
Testers are not able to read the app config in the test menu due to the `toString()` method requiring the config getters and the shrinker strips them. The little bit of extra space due to not shrinking the protobuf generated classes is well worth making it easier for testers to confirm correct behavior.

### How test
* Compare the app config test menu in debug and release mode with/without the change

Before in release mode:

Now in release mode:


ralfgehrer (approved):

Oliver-Zimmerman (approved):
Tested on Samsung S9 plus.

PR #1626, opened 2020-11-17
custom view Horizontal list of flags (EXPOSUREAPP-3736)

AndroidMedaGalaxy:
### Description
Custom Control for Horizontal List of Flags

### Steps to reproduce
Can be obeserved in interoprability while submission flow or in the settings

### Screenshot
horizontal list


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,98 @@ package de.rki.coronawarnapp.ui.view
...
+            country_list_entry_label.text = countryNames
+        }
+    }
+}
+
+/**
+ * Adapter for the country flags grid
+ */
+class CountryAdapterFlags internal constructor(
+    context: Context?,
```
d4rken, line 71: Why is the context nullable?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,98 @@ package de.rki.coronawarnapp.ui.view
...
+    countryList: List<Country>?
```
d4rken, line 72: Why is the list nullable? I'd prefer it to be not nullable, if there are no countries, the list is empty.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,98 @@ package de.rki.coronawarnapp.ui.view
...
+        holder.flagImage.setImageResource(country.iconRes)
+    }
+
+    // total number of cells
+    override fun getItemCount(): Int {
+        return countryList!!.size
+    }
+
+    // stores and recycles views as they are scrolled off screen
+    inner class ViewHolder internal constructor(itemView: View) : RecyclerView.ViewHolder(itemView) {
```
d4rken, line 96: Check out `KeyFileDownloadAdapter.kt` as example implementation, it does not have to be as elaborate (but would be nice), the gist is to delegate the view specific details (inflation & binding) to the `ViewHolder` and to take as few dependencies via constructor (i.e. no `Context'), if we can get it from the RecyclerView (i.e. getting a context from the `parent: ViewGroup`).

https://github.com/corona-warn-app/cwa-app-android/blob/ec910ab5c8c156d069c2b7c384b1d1c40dfd21ce/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/keydownload/ui/KeyFileDownloadAdapter.kt#L15

Pass the parent to the ViewHolder, let the ViewHolder inflate it's own layout (modularization), and only pass the data class (here `Country`) to a bind method in the ViewHolder.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,98 @@ package de.rki.coronawarnapp.ui.view
...
+import kotlinx.android.synthetic.main.view_country_list_entry_flag_container.view.*
 import java.text.Collator
 
 class CountryList(context: Context, attrs: AttributeSet) :
     LinearLayout(context, attrs) {
-
+    private var attributeSet = attrs
+    private var view: View? = null
     private var _list: List<Country>? = null
     var list: List<Country>?
```
d4rken, line 23: Can we make this non-nullable?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,98 @@ package de.rki.coronawarnapp.ui.view
...
                 Collator.getInstance().compare(a.first, b.first)
             }
-            ?.forEachIndexed { index, (label, iconRes) ->
-                inflate(context, R.layout.view_country_list_entry, this)
-                val child = this.getChildAt(index)
-                child.apply {
-                    findViewById<ImageView>(R.id.country_list_entry_image).setImageResource(iconRes)
-                    findViewById<TextView>(R.id.country_list_entry_label).text = label
+
+        view = inflate(context, R.layout.view_country_list_entry_flag_container, this)
```
d4rken, line 46: This is a linear layout, and `view_country_list_entry_flag_container` contains two more `LinearLayout`. Do we need 3 nested `LinearLayouts`?

Corona-Warn-App/src/main/res/layout/view_country_list_entry_flag_item.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
d4rken, line 2: `LinearLayout` with one child, could maybe be replaced with a `FrameLayout`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
ritsam: resolved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
ritsam: it is resolved now

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
ritsam: Please check the new implmentation out when you can

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
ritsam: the new changes didn't require the variable view. So should be resolved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
ritsam: as we're setting orientation in _init_ as HORIZONTAL, I feared that it might break the orientation and I kept the inner Linear layout.

But I removed the nested Linear Layouts and noticed no issues.  So yeah. this has been implmented

Corona-Warn-App/src/main/res/layout/view_country_list_entry_flag_item.xml
ritsam: I concur!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,130 @@ package de.rki.coronawarnapp.ui.view
...
+        context.withStyledAttributes(attributeSet, R.styleable.SimpleStepEntry) {
+            flagGrid.layoutManager = GridLayoutManager(context, FLAG_COLUMNS)
+            flagGrid.adapter = adapterCountryFlags
+            val iterator: Iterator<Country> = list.iterator()
+            while (iterator.hasNext()) {
+                countryNames += resources.getString(iterator.next().labelRes)
+                if (iterator.hasNext()) countryNames += ", "
             }
+            country_list_entry_label.text = countryNames
+        }
```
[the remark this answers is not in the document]
d4rken: The top level viewmodel (i.e. Fragment's) should compute the list of data classes (`Country`) and pass it to the the view, the view should pass it to the internal adapter, the adapter binds the `Country' objects to the RecyclerView.ViewHolder's

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,130 @@ package de.rki.coronawarnapp.ui.view
...
 import de.rki.coronawarnapp.ui.Country
+import de.rki.coronawarnapp.util.lists.BindableVH
+import kotlinx.android.synthetic.main.view_country_list_entry_flag_container.view.*
 import java.text.Collator
 
 class CountryList(context: Context, attrs: AttributeSet) :
     LinearLayout(context, attrs) {
-
+    private var attributeSet = attrs
     private var _list: List<Country>? = null
```
d4rken, line 23: Can't this just be a non-nullable empty list ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,130 @@ package de.rki.coronawarnapp.ui.view
...
-                inflate(context, R.layout.view_country_list_entry, this)
-                val child = this.getChildAt(index)
-                child.apply {
-                    findViewById<ImageView>(R.id.country_list_entry_image).setImageResource(iconRes)
-                    findViewById<TextView>(R.id.country_list_entry_label).text = label
-                }
+        list?.let { prepareFlags(it) }
+    }
+
+    private fun prepareFlags(list: List<Country>) {
```
d4rken, line 50: Why are we passing `list` and then using `_list`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,130 @@ package de.rki.coronawarnapp.ui.view
...
+/**
+ * Adapter for the country flags grid
+ */
+class CountryAdapterFlags internal constructor(
+    countryList: List<Country>
+) :
+    BaseCountryFlagAdapter<CountryAdapterFlags.CountryFlagViewHolder>() {
+    // total number of cells
+
+    private val countryList: List<Country>? = countryList
```
d4rken, line 79: I think the country list does not change for the adapter.

Try if we can just use a constructor property.

```kotlin
constructor(
    private val countryList: List<Country>
)
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,130 @@ package de.rki.coronawarnapp.ui.view
...
+    override fun getItemCount(): Int = countryList!!.size
+
+    override fun getItemId(position: Int): Long = countryList?.get(position)!!.iconRes.hashCode().toLong()
```
d4rken, line 82: `getItemId` only has an effect if we have `setStableIds`, either set that, or remove the `getItemId`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CountryList.kt
```diff
@@ -2,47 +2,130 @@ package de.rki.coronawarnapp.ui.view
...
+            ViewCountryListEntryFlagItemBinding.bind(itemView)
+        }
+
+        override val onBindData: ViewCountryListEntryFlagItemBinding.(key: Country) -> Unit = { item ->
+            countryListEntryImage.setImageResource(item.iconRes)
+        }
+    }
+}
+
+abstract class BaseCountryFlagAdapter<T : BaseCountryFlagAdapter.VH> : RecyclerView.Adapter<T>() {
```
d4rken, line 111: Why do we need the abstract class?

d4rken:
Looks like this now:

Screenshot from 2020-11-20 17-19-51

d4rken (approved):

chiljamgossow (approved):

PR #1632, opened 2020-11-17
Remove debug notifications (DEV)

d4rken:
They are not actively maintained, not really used and just creating techdebt.

### Testing
* Smoke test the app
* Check that the debug menu still works okay


SamuraiKek (dismissed):
Tested the debug menu and everything looks good. I was able to switch environments properly, the logs were working and, as expected, the background notifications were no longer there.

harambasicluka:
Testers are using this. Need to discuss how we go ahead.

harambasicluka (dismissed):
Clarified with testers, they are also only going into the logs. If a tester isn't sure how to work with logs we are happy to help!

PR #1636, opened 2020-11-17
Fix blank card and tracing state (EXPOSUREAPP-3823)

d4rken:
This fixes blank UI elements because the `GeneralTracingStatus` did not emit a value, because `ENFClient.isTracingEnabled` did not emit a value.

Two potentional issues could have caused this which are fixed by this PR:
* The polling loop used for `isTracingEnabled` could exit prematurely or not run at all if the scope on which it was running was canceled. `isActive` would return false, and the `while (...) {...}` would never execute.
* Additionally we were also using `sendBlocking` in combination with `Dispatchers.Default` in two other polling loops (`BackgroundModeStatus`), and on devices with low CPU core count, the `sendBlocking` could starve the coroutines of threads to run on.

The PR ensures that we first emit a value and then check `isActive` before repeating the loop, and makes it use `suspend send`
 instead of `sendBlocking`. Additionally we also use `shareLatest` for `isTracingEnabled` such that there is only one polling loop running at a time for it.

## Testing
Test the home screen UI on low end devices, close/open the app.
Toggle stuff that effects the tracing state (i.e. Bluetooth ON/OFF), to make sure the values are propagated correctly and stale values are not cached.


SamuraiKek:
I've tested it on my Nokia 3 device which is hella slow, over the span of roughly an hour. I've played around with the bluetooth settings and the location settings and didn't find any unexpected behaviour. Everything seems to be working fine and no blank card was encountered.

SamuraiKek (approved):

chris-cwa (approved):

PR #1615, opened 2020-11-16
Failed risk calculation card (EXPOSUREAPP-4088)

SamuraiKek:
### Description
This PR is aiming to refactor the home fragment cards as well as the risk details screen while introducing a new risk card for when the risk calculation fails.
We are planning to add a RecyclerView to the HomeFragment and RiskDetailsFragment that will hold all of the different cards.
As this is quite a big undertaking, this PR is going to be open as DRAFT until everything is implemented. 

This checklist is meant to keep track of the progress done so far:
- [x] Create new risk card for failed risk calculation state.
- [x] Create new risk card for risk calculation stopped.
- [x] Create new risk card for low level risk.
- [x] Create new risk card for high level risk.
- [x] Create data classes for risk cards.
- [ ] Remove grey risk card along with old implementation of risk cards.
- [ ] Create new covid test cards for all test states (untested, fetching, failed, positive, negative, done...).
- [ ] Create a new FAQ card.
- [ ] Create a new state provider that provides the right risk card to the HomeFragmentViewModel.
- [ ] Introduce RecyclerView approach for the home fragment.
- [ ] Introduce RecyclerView approach for the risk details fragment.
- [ ] Create class that handles the state of the risk calculation.
- [ ] Add necessary unit tests.

PR #1643, opened 2020-11-17
Sneak Preview: ENF Window Mode

chris-cwa:
- not ready to test on devices
- UI not changed


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
```diff
@@ -22,95 +19,37 @@ class DefaultDiagnosisKeyProvider @Inject constructor(
...
-    ): Boolean {
-        Timber.d("Using non-legacy key provision.")
-
-        if (!submissionQuota.consumeQuota(1)) {
-            Timber.w("Not enough quota available.")
-            // TODO Currently only logging, we'll be more strict in a future release
-            // return false
+        if (!googleAPIVersion.isAtLeast(GoogleAPIVersion.V15)) {
+            // Actually this shouldn't happen
+            Timber.d("No key files submitted because client uses an old unsupported version")
```
d4rken, line 30: Let's make that a `Timber.e`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
```diff
@@ -22,95 +19,37 @@ class DefaultDiagnosisKeyProvider @Inject constructor(
...
+            return false
```
d4rken, line 31: I think we should throw an exception here, this is worthy of making it user visible as the app will not be able to function. 

We could create `googleAPIVersion.requireV15()` that throws a specific exception if the requirement is not satisfied. :thinking: 

What do you think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
```diff
@@ -22,95 +19,37 @@ class DefaultDiagnosisKeyProvider @Inject constructor(
...
-        Timber.d("Using LEGACY key provision.")
-
-        if (!submissionQuota.consumeQuota(keyFiles.size)) {
-            Timber.w("Not enough quota available.")
-            // TODO What about proceeding with partial submission?
-            // TODO Currently only logging, we'll be more strict in a future release
-            // return false
+        if (!submissionQuota.consumeQuota(1)) {
+            Timber.w("No key files submitted because not enough quota available.")
+            return false
```
d4rken, line 36: Previously this was not checked because we didn't want to hide quota issues due to wrong frequency/interval checks on a higher level. If this is armed we won't know that something is failing because it is hitting the quota limit for unexpected reasons. :thinking: Was this discussed with someone?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/SubmissionQuota.kt
```diff
@@ -86,6 +86,10 @@ class SubmissionQuota @Inject constructor(
 
     companion object {
         @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        /**
+         * This quota changes when using ExposureWindow mode from 20 to 6 per day
+         * See: https://developers.google.com/android/exposure-notifications/release-notes
+         */
```
d4rken, line 92: Why did we not change this from 20 to 6 in code?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -1,26 +1,48 @@
...
 
 @Singleton
-class DefaultRiskLevels @Inject constructor() : RiskLevels {
+class DefaultRiskLevels @Inject constructor(
+    private val appConfigProvider: AppConfigProvider
+) : RiskLevels {
+    private var appConfig: ConfigData
+
+    init {
+        runBlocking {
```
d4rken, line 39: No `runBlocking` in `init { }` calls. This would be called on injection which can happen on App init, i.e. on the main thread, and could freeze the whole app.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -1,26 +1,48 @@
...
+import org.joda.time.Instant
 import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
-import kotlin.math.round
 
 @Singleton
-class DefaultRiskLevels @Inject constructor() : RiskLevels {
+class DefaultRiskLevels @Inject constructor(
+    private val appConfigProvider: AppConfigProvider
```
d4rken, line 34: What do you think about passing the app config to each function that needs it, removing this dependency?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -1,26 +1,48 @@
...
+) : RiskLevels {
+    private var appConfig: ConfigData
```
d4rken, line 36: As the class is a singleton, and we are saving the app config only once, it could lead to future calculations with an outdated app config.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -1,26 +1,48 @@
...
+
+    init {
+        runBlocking {
+            appConfig = appConfigProvider.getAppConfig()
+        }
+
+        appConfigProvider.currentConfig
+            .onEach { if (it != null) appConfig = it }
```
d4rken, line 44: This code won't do anything, the flow doesn't have a terminal operator starting it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+        exposureWindow: ExposureWindow,
+        attenuationFilters: List<RiskCalculationParametersOuterClass.MinutesAtAttenuationFilter>
+    ) =
+        attenuationFilters.any { attenuationFilter ->
+            // Get total seconds at attenuation in exposure window
+            val secondsAtAttenuation = exposureWindow.scanInstances
+                .filter { attenuationFilter.attenuationRange.inRange(it.typicalAttenuationDb) }
+                .fold(0) { acc, scanInstance -> acc + scanInstance.secondsSinceLastScan }
+
+            val minutesAtAttenuation = secondsAtAttenuation / 60
```
d4rken, line 181: `Range` is a double AFAICT, don't we want `minutesAtAttenuation` to also be float or double and not int?

e.g.
`val minutesAtAttenuation: Double = secondsAtAttenuation / 60f`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+            return attenuationFilter.dropIfMinutesInRange.inRange(minutesAtAttenuation)
+        }
+
+    private fun determineTransmissionRiskLevel(
```
d4rken, line 185: We could make all these functions working on `ExposureWindow` extension functions on it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+        transmissionRiskLevelFilters.any {
+            it.dropIfTrlInRange.inRange(transmissionRiskLevel)
+        }
+
+    private fun determineWeightedSeconds(
+        exposureWindow: ExposureWindow,
+        minutesAtAttenuationWeight: List<RiskCalculationParametersOuterClass.MinutesAtAttenuationWeight>
+    ): Double =
+        exposureWindow.scanInstances.fold(.0) { seconds, scanInstance ->
+            val weight =
```
d4rken, line 225: Please add an explicit type here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+    ): RiskResult? {
+        if (dropDueToMinutesAtAttenuation(exposureWindow, appConfig.minutesAtAttenuationFilters)) {
+            Timber.d(
+                "%s dropped due to minutes at attenuation filter",
+                exposureWindow
+            )
+            return null
+        }
+
+        val transmissionRiskLevel = determineTransmissionRiskLevel(
```
d4rken, line 253: Please add an explicit type here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+        if (dropDueToTransmissionRiskLevel(transmissionRiskLevel, appConfig.transmissionRiskLevelFilters)) {
+            Timber.d(
+                "%s dropped due to transmission risk level filter, level is %s",
+                exposureWindow,
+                transmissionRiskLevel
+            )
+            return null
+        }
+
+        val transmissionRiskValue =
```
d4rken, line 267: Please add an explicit type here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+            transmissionRiskLevel * appConfig.transmissionRiskLevelMultiplier
+
+        Timber.d(
+            "%s's transmissionRiskValue is: %s",
+            exposureWindow,
+            transmissionRiskValue
+        )
+
+        val weightedMinutes = determineWeightedSeconds(
```
d4rken, line 276: Please add an explicit type here. If we expect a fractional number, also force it by specifying `60f` or `60.0`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+            appConfig.minutesAtAttenuationWeights
+        ) / 60
+
+        Timber.d(
+            "%s's weightedMinutes are: %s",
+            exposureWindow,
+            weightedMinutes
+        )
+
+        val normalizedTime = transmissionRiskValue * weightedMinutes
```
d4rken, line 287: Please add an explicit type here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+        Timber.d("Aggregating result for date $dateMillisSinceEpoch - ${Instant.ofEpochMilli(dateMillisSinceEpoch)}")
+
+        // 3. Determine `Risk Level per Date`
+        val riskLevel = try {
+            appConfig.normalizedTimePerDayToRiskLevelMappingList
+                .filter { it.normalizedTimeRange.inRange(normalizedTime) }
+                .map { it.riskLevel }
+                .first()
+        } catch (e: Exception) {
+            throw ApplicationConfigurationInvalidException(
```
d4rken, line 413: Maybe this should be checked by the config mapper. If the config is invalid if it doesn't contain it, and we throw during mapping (parsing), then the Config logic has the chance to provide an alternative config from storage or default.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -216,16 +159,313 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
...
+        )
+    }
+
+    private fun minimumDistinctEncountersForRisk(
+        exposureWindowsAndResultForDate: Map<ExposureWindow, RiskResult>,
+        riskLevel: RiskCalculationParametersOuterClass.NormalizedTimeToRiskLevelMapping.RiskLevel
+    ): Int =
+        exposureWindowsAndResultForDate
+            .filter { it.value.riskLevel == riskLevel }
+            .map { "${it.value.transmissionRiskLevel}_${it.key.calibrationConfidence}" }
```
d4rken, line 451: Can you explain the idea behind this?

d4rken (commented):
Good stuff :heart: 

About 50% through.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/ExposureSummaryRepository.kt
```diff
@@ -1,70 +1,30 @@
...
         private val internalMatchedKeyCount = MutableStateFlow(0)
         val matchedKeyCount: Flow<Int> = internalMatchedKeyCount
 
         private val internalDaysSinceLastExposure = MutableStateFlow(0)
         val daysSinceLastExposure: Flow<Int> = internalDaysSinceLastExposure
     }
 
-    suspend fun getExposureSummaryEntities() = exposureSummaryDao.getExposureSummaryEntities()
-        .map { it.convertToExposureSummary() }
+    private var entities = emptyList<ExposureWindow>()
```
d4rken, line 21: This was for the debug UIs right? We could make this a `MutableStateFlow` and then the UI could automatically update on new data.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProviderTest.kt
```diff
@@ -62,135 +51,64 @@ class DefaultDiagnosisKeyProviderTest : BaseTest() {
...
-                listOf(exampleKeyFiles[1]), fallback, exampleToken
-            )
-            submissionQuota.consumeQuota(2)
-        }
-    }
-
-    @Test
-    fun `quota is consumed silenently`() {
-        coEvery { googleAPIVersion.isAtLeast(GoogleAPIVersion.V16) } returns true
+    fun `quota is consumed silently`() {
```
d4rken, line 85: The quota is not consumed silently (passively, not interrupting), if the quota is exceeded, it returns `false` and we abort (in current code)?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProviderTest.kt
```diff
@@ -62,135 +51,64 @@ class DefaultDiagnosisKeyProviderTest : BaseTest() {
     )
 
     @Test
-    fun `legacy key provision is used on older ENF versions`() {
-        coEvery { googleAPIVersion.isAtLeast(GoogleAPIVersion.V16) } returns false
+    fun `key provision is used on older ENF versions`() {
```
d4rken, line 54: Missleading function name, we are checking that we DO NOT submit keys if the ENF version is too low.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProviderTest.kt
```diff
@@ -62,135 +51,64 @@ class DefaultDiagnosisKeyProviderTest : BaseTest() {
...
-            provider.provideDiagnosisKeys(exampleKeyFiles, exampleConfiguration, exampleToken)
-        }
+        runBlocking { provider.provideDiagnosisKeys(exampleKeyFiles) } shouldBe true
 
         coVerify(exactly = 1) {
-            googleENFClient.provideDiagnosisKeys(any(), any(), any())
-            googleENFClient.provideDiagnosisKeys(
-                exampleKeyFiles, exampleConfiguration, exampleToken
-            )
+            googleENFClient.provideDiagnosisKeys(any<List<File>>())
```
d4rken, line 78: `any()`  should suffice

d4rken (commented):
Rest of the code looks good too :slightly_smiling_face:

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProviderTest.kt
BMItter: nope, won't suffice.
there are two differnt provideDiagnosisKeys from API side..
you have to use  Any<List<bla>> ()

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: i was not actually sure from the tech spec if this should be rounded down or not, test cases all passed with this implementation but i will ask @mlenkeit to make sure

mlenkeit: Technically, secondsAtAttenuation is always divisible by 60 (as per EN spec). Anyway, I would still recommend not to make any implicit data type conversions...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/ExposureSummaryRepository.kt
chris-cwa: will change it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
BMItter: _4. Determine Minimum Distinct Encounters With Low Risk per Date : for all Exposure Windows of the same Date , count the
distinct combinations of Transmission Risk Level , and Callibration Confidence of all Exposure Windows with Risk
Level calculated as Low_ (max docu)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
vaubaehn: Is it ok to leave my 2 cents here?

l29: `// Actually this shouldn't happen`
This could happen when users are reactivating old phones to provide CWA to family members without phone.

Why not lifting UI/UX to the next level, and present a nice 'information (error) card' in the home fragment for known problems like API 17 or API 39508 instead of throwing an exception? Much work, yes, but as #1615 is WIP, probably a good point in time to think about it?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
kolyaopahle: done in #1656

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
vaubaehn: see https://github.com/corona-warn-app/cwa-app-android/pull/1643#discussion_r526281904

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
BMItter: I'm also not sure about this.. the check feels like a second or third check in dept.. even with an exception the user would be left a little alone here. Sure, better than crashing, but imho it makes more sense to check all the API/enf requirements while the app is starting, what do y think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/SubmissionQuota.kt
BMItter: _long answer_: when it was implemented, it was unclear whether the old mode / provideDiagnosisKey has to be supported too.
_short answer_: adjusted (PR #1656 )

chris-cwa:
This Preview PR will soon be closed. Actual PR for the Feature is https://github.com/corona-warn-app/cwa-app-android/pull/1659.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
d4rken: AFAICT (please confirm), if we throw here, we crash the task, and the crashed task will be caught by `TaskController` which will call `Error.report...` which will cause the "popular" error dialog to show, it will not crash the app. So my favored approach would be to just throw an exception here with a specific error message such that it's clear to the user what happened. Something like "Your systems exposure notification service are too old". This would be an easy fix and enough to let users reactivating old devices what is happening, which is not impossible, but probably rare.

I don't think this warrants more elaborate UI elements. Although I don't like our error dialog, and @BMItter has already been working on a nicer solution, it's not ready yet due to shifting priorities.

Thoughts @BMItter @vaubaehn ?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProviderTest.kt
BMItter: changed name to fun `provide diagnosis key when quota is empty`()

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProviderTest.kt
BMItter: changed to fun _provide diagnosis keys with outdated ENF versions_

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
chris-cwa: please continue discussion at https://github.com/corona-warn-app/cwa-app-android/pull/1659/files#r526719742

chris-cwa:
continued at https://github.com/corona-warn-app/cwa-app-android/pull/1659

PR #1644, opened 2020-11-17
Removed enf v1 exposure summary and tokens (EXPOSUREAPP-3538)

chris-cwa:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -106,8 +110,21 @@ class DefaultRiskLevels @Inject constructor(
 
         val aggregatedResult = aggregateResults(riskResultsPerWindow)
 
-        return aggregatedResult.totalRiskLevel ==
+        exposureSummaryRepository.exposureWindowEntities = Pair(exposureWindows, aggregatedResult)
```
BMItter, line 113: I'm not sure whether its neccessary to persist or "to repo" the exposureWindows... 
On my current point of view, only the Result is needed.

chris-cwa: renamed it, because it is not a repo anymore - as dicussed

BMItr (approved):

kolyaopahle (approved):

PR #1646, opened 2020-11-18
Update privacy_tr.html (EXPOSUREAPP-3675)

janetback:


ralfgehrer (approved):
Looks good, but noted that this one has no underlined section as other privacy statements @janetback.

harambasicluka (approved):

PR #1648, opened 2020-11-18
Remove manual refresh of last fetched timestamp (EXPOSUREAPP-3843)

ralfgehrer:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -399,6 +401,12 @@ object LocalData {
         return Date(time)
     }
 
+    fun lastTimeDiagnosisKeysFromServerFetchFlow() =
+        getSharedPreferenceInstance()
+            .createFlowPreference<Long?>(CoronaWarnApplication.getAppContext()
+                .getString(R.string.preference_timestamp_diagnosis_keys_fetch), 0L).flow
```
chris-cwa, line 407: why are we using a key from string resources here, not const val?

ralfgehrer: That seems to be the way we deal with all keys in LocalData. Just using what we already have in place  - see e..g the getter above.

chris-cwa (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
harambasicluka: Got this questions a few weeks ago also from @d4rken , in the beginning we followed the example from the docu: https://developer.android.com/training/data-storage/shared-preferences#GetSharedPreferences

image

But I would say that's something for a future refactoring.

Oliver-Zimmerman (approved):

harambasicluka (approved):

PR #1649, opened 2020-11-18
Fix config default values not being used if protobuf is missing the data structures (EXPOSUREAPP-3856)

d4rken:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/AppConfigSource.kt
```diff
@@ -24,6 +24,8 @@ class AppConfigSource @Inject constructor(
         if (localConfig != null && localConfig.isValid(timeStamper.nowUTC)) {
             Timber.tag(TAG).d("Returning local config, still valid.")
             return localConfig
+        } else {
+            Timber.tag(TAG).d("Local app config was unavailable(${localConfig == null} or invalid.")
```
chris-cwa, line 28: i would prefer severity 'warn' not 'debug'

chris-cwa (approved):

harambasicluka (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/AppConfigSource.kt
d4rken: Maybe invalid is too strong of a word here, the local config is invalid if it older than the CDNs cache header specifies, which is currently just 5min, so we would get the warning a lot.

PR #1655, opened 2020-11-18
Release 1.7 to 1.8

chris-cwa:


harambasicluka (approved):

gradle.properties
```diff
@@ -20,4 +20,4 @@ org.gradle.dependency.verification.console=verbose
 VERSION_MAJOR=1
 VERSION_MINOR=8
 VERSION_PATCH=0
-VERSION_BUILD=3
+VERSION_BUILD=5
```
harambasicluka, line 23: could set this to 1, then we have it directly available :)

chris-cwa: but please not in the merge commit, that would be evil

PR #1656, opened 2020-11-18
Started to address some comments from the sneak preview PR (EXPOSUREAPP-3456)

kolyaopahle:
### Description
This PR adresses some of the comments made in the #1643 PR


build.gradle
```diff
@@ -13,7 +13,7 @@ buildscript {
         }
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.0.0'
+        classpath 'com.android.tools.build:gradle:4.1.1'
```
chris-cwa, line 16: do we really want to update this in the feature branch? i would prefer to change this in its own branch, to (if something is not working anymore) we can decide to not yet update.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.1.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-all.zip
```
chris-cwa, line 3: do we really want to update this in the feature branch? i would prefer to change this in its own branch, to (if something is not working anymore) we can decide to not yet update.

build.gradle
kolyaopahle: no idea where this came from, removed it

gradle/wrapper/gradle-wrapper.properties
kolyaopahle: no idea where this came from, removed it

BMItr (approved):

PR #1657, opened 2020-11-18
Fix RiskLevelCalculationTask not being executed if no new match was found (EXPOSUREAPP-3849)

d4rken:
The RiskLevelCalculationTask was not being executed if the provided key files didn't result in a new exposure summary state.

For some reasons this was not a visible issue before, but unrelated changes have made this issue visible now.

## Testing
* Fresh install (time set in the past)
* Card is grey
* Timetravel exactly 1 day
* Card is green!


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/exception/reporting/ExceptionReporter.kt
```diff
@@ -16,17 +16,19 @@ import de.rki.coronawarnapp.util.tryHumanReadableError
...
+fun Throwable.report(exceptionCategory: ExceptionCategory) {
     this.report(exceptionCategory, null, null)
+}
 
 fun Throwable.report(
     exceptionCategory: ExceptionCategory,
     prefix: String?,
     suffix: String?
 ) {
+    reportProblem(tag = prefix, info = suffix)
```
BMItter, line 28: :ok_hand:

harambasicluka (approved):
LGTM.

Oliver-Zimmerman (approved):
lgtm

PR #1660, opened 2020-11-18
Fix google api token being overwritten despite aborted exposure detection (EXPOSUREAPP-3869)

d4rken:
We generated a new exposure detection token for every DownloadTask execution, even if we only download files and do not submit to the ENF, as we abort gracefully and do not throw, there is no rollback that sets the original token again.

This leads to RiskLevelTasks being executed that use empty ExposureSummary's turning red cards into green cards.


BMItr (approved):
nothing to complain

chris-cwa (approved):

PR #1668, opened 2020-11-19
Prevent premature task abortions, Store timestamp first (EXPOSUREAPP-3878)

chris-cwa:
### Steps:

1. time travel to t-2
2. open app
3. complete onboarding
4. wait for files to download
5. risk card is gray (initial)
6. kill app
7. time travel to t-1 at 23:00
8. open app
9. wait for files to download
10. risk card is gray (initial)
11. kill app
12. time travel to t-0 at 5:00
13. open app
14. wait for files to download
15. risk card is gray (initial)

### Expected result:
steps 10 and 15 should not be gray but green or red

not reproducable on clean install - only when app is reseted from menu

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
@@ -82,14 +82,16 @@ class DownloadDiagnosisKeysTask @Inject constructor(
                 return object : Task.Result {}
             }
 
+            // EXPOSUREAPP-3878 just to be sure that risk level calculation can take place
             if (TimeVariables.getLastTimeDiagnosisKeysFromServerFetch() != null) {
                 if (wasLastDetectionPerformedRecently(now, exposureConfig, trackedExposureDetections)) {
                     // At most one detection every 6h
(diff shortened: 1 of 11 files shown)


harambasicluka (dismissed):

harambasicluka (dismissed):

kolyaopahle (dismissed):
LGTM 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -82,18 +81,15 @@ class DownloadDiagnosisKeysTask @Inject constructor(
                 return object : Task.Result {}
             }
 
-            // EXPOSUREAPP-3878 just to be sure that risk level calculation can take place
-            if (TimeVariables.getLastTimeDiagnosisKeysFromServerFetch() != null) {
```
chris-cwa: why was the additional condition removed? it doesn't do any harm, does it? but it assures that there is actually data for the risk task

kolyaopahle (approved):

mlenkeit (changes requested):
Fixes the problem but comes with new ones 😄  See comments in Jira for logs, etc.

harambasicluka (approved):

mlenkeit (approved):
Original issue is fixed by the PR 👍

PR #1669, opened 2020-11-19
Fix update of last submission date (EXPOSUREAPP-3874)

d4rken:
* We were using a mix of different preferences objects when updating the last download timestamp which could cause them to go out of  sync.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
@@ -383,45 +383,23 @@ object LocalData {
      * SERVER FETCH DATA
      ****************************************************/
 
-    /**
-     * Gets the last time the server fetched the diagnosis keys from the server as Date object
-     * from the EncryptedSharedPrefs
-     *
-     * @return timestamp as Date
-     */
-    // TODO should be changed to Long as well to align with other timestamps
-    fun lastTimeDiagnosisKeysFromServerFetch(): Date? {
-        val time = getSharedPreferenceInstance().getLong(
-            CoronaWarnApplication.getAppContext()
-                .getString(R.string.preference_timestamp_diagnosis_keys_fetch),
-            0L
-        )
-        if (time == 0L) return null
-
-        return Date(time)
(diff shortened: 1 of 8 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -383,76 +382,29 @@ object LocalData {
...
-            putLong(
-                CoronaWarnApplication.getAppContext()
-                    .getString(R.string.preference_timestamp_diagnosis_keys_fetch),
-                value?.time ?: 0L
-            )
+    var lastTimeDiagnosisKeysFromServerFetch: Date?
+        get() = lastTimeDiagnosisKeysFetchedFlowPref.value.let {
+            if (it != 0L) Date(it) else null
         }
-    }
```
chris-cwa, line 424: unrelated to the bug; i don't think this the right time for refactoring (including regression bugs from it)

chris-cwa (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
d4rken: This is not refactoring, but a bugfix for:

> We were using a mix of different preferences objects when updating the last download timestamp which could cause them to go out of sync.

d4rken:
Need to change the timestamp again. Draft for 20min.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
d4rken: See new changes. Hope it's okay now.

chris-cwa: but here one preference only was used: key = getString(R.string.preference_timestamp_diagnosis_keys_fetch), you just made it a flow, or am i missing something?

kolyaopahle (approved):
LGTM 👍

chris-cwa (approved):

mlenkeit (approved):
Retested and seems to solve the problem

PR #1670, opened 2020-11-19
Fix hourly pkg expectancy check (EXPOSUREAPP-3875)

d4rken:
If we turn the device off and then on on the next day, in the same exact hour as on the previous day, the `HourPackageSyncTool`  would think there is no new hour data as it only compared the `hourOfDay`.


kolyaopahle (approved):
LGTM 👍

harambasicluka (approved):

PR #1677, opened 2020-11-20
Migration to google ENF client v2 (window mode) (EXPOSUREAPP-3456)

chris-cwa:


chris-cwa:
successor of https://github.com/corona-warn-app/cwa-app-android/pull/1659

d4rken (approved):
lgtm

BMItter:
have to check testmenu changes according to last commit changes

harambasicluka:
Not sure how to handle this edge case: I traveled back in time, imho the active tracing time displayed should be zero. But what happen if a user travels 500 days in future does a risk calculation and goes back to the current date then he would need 500 days to reach the correct state.
image

harambasicluka:
I went 4 days back on WRU-XD reset the data from OS level. I went 23h in future -> unknown risk, plus 2 -> low risk (visible in ENF menu in OS), traveled 3h in future -> no update, still on green & so on over multiple days. I'll try to reproduce it and see when it occurred, but currently it says `Begegnungen wurden noch nicht überprüft` even if there are exposure checks in the ENF settings on OS level. 

Will now check the logs.

On both screenshots the current date is 19.11.20.
image
image

harambasicluka:
> I went 4 days back on WRU-XD reset the data from OS level. I went 23h in future -> unknown risk, plus 2 -> low risk (visible in ENF menu in OS), traveled 3h in future -> no update, still on green & so on over multiple days. I'll try to reproduce it and see when it occurred, but currently it says Begegnungen wurden noch nicht überprüft even if there are exposure checks in the ENF settings on OS level.

Should be fixed with the latest update from `release/1.7.x`. Has to be retested after bot is merged in `release/1.8.x`

harambasicluka (approved):

PR #1678, opened 2020-11-20
updated test api fragment to display exposure windows

chris-cwa:
Screenshot_20201119-220355

### DIFF
--- a/Corona-Warn-App/src/main/res/values-bg/strings.xml
+++ b/Corona-Warn-App/src/main/res/values-bg/strings.xml
@@ -1259,16 +1259,6 @@
     <!-- NOTR -->
     <string name="test_api_exposure_summary_headline">"Exposure Windows"</string>
     <!-- NOTR -->
-    <string name="test_api_body_daysSinceLastExposure">"Days since last exposure: %1$s"</string>
-    <!-- NOTR -->
-    <string name="test_api_body_attenuation">"Attenuation Durations in Minutes: %1$s"</string>
-    <!-- NOTR -->
-    <string name="test_api_body_summation_risk">"Summation Risk Score: %1$s"</string>
-    <!-- NOTR -->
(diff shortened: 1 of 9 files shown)


d4rken (approved):

BMItter:
ExposureWindow is already shown in new ENFv2 Test Fragment.
I would really recommend to remove it from there(test api) and have all needed in ENFv2 Test Fragment.. This also saves us time

mlenkeit:
I haven't tested the EN v2 branch yet... anyway: I would suggest not to duplicate information on the tester menu (if that's the case here). Just because certain pieces of information existed in a certain place in v1 doesn't necessarily mean we need to have them in the same place again. But I can build this branch and take a look myself...

d4rken (approved):

Corona-Warn-App/src/main/res/values-bg/strings.xml
```diff
@@ -1257,17 +1257,7 @@
...
-    <string name="test_api_body_daysSinceLastExposure">"Days since last exposure: %1$s"</string>
-    <!-- NOTR -->
-    <string name="test_api_body_attenuation">"Attenuation Durations in Minutes: %1$s"</string>
-    <!-- NOTR -->
-    <string name="test_api_body_summation_risk">"Summation Risk Score: %1$s"</string>
-    <!-- NOTR -->
-    <string name="test_api_body_matchedKeyCount">"Matched key count: %1$s"</string>
-    <!-- NOTR -->
-    <string name="test_api_body_maximumRiskScore">"Maximum risk score %1$s"</string>
+    <string name="test_api_exposure_summary_headline">"Exposure Windows"</string>
```
BMItter, line 1260: better hardcode testmenu stuff to avoid translations.

BMItter:
sometimes - less is more :ok_hand:
PS: I also would remove the Exposure Windows title later. but low prio right now.

PR #1681, opened 2020-11-20
Resolve ENFv2 PR comments

d4rken:
Resolving comments on https://github.com/corona-warn-app/cwa-app-android/pull/1659
and https://github.com/corona-warn-app/cwa-app-android/pull/1677

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
@@ -34,8 +34,9 @@ class DefaultDiagnosisKeyProvider @Inject constructor(
         }
 
         if (!submissionQuota.consumeQuota(1)) {
-            Timber.w("No key files submitted because not enough quota available.")
-            return false
+            Timber.e("No key files submitted because not enough quota available.")
+            // Needs discussion until armed, concerns: Hiding other underlying issues.
+//            return false
         }
 
         return suspendCoroutine { cont ->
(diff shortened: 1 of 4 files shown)


kolyaopahle (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeyprovider/DefaultDiagnosisKeyProvider.kt
```diff
@@ -34,8 +34,9 @@ class DefaultDiagnosisKeyProvider @Inject constructor(
         }
 
         if (!submissionQuota.consumeQuota(1)) {
-            Timber.w("No key files submitted because not enough quota available.")
-            return false
+            Timber.e("No key files submitted because not enough quota available.")
```
chris-cwa, line 37: where do we keep track of this, besides the back of our minds?

d4rken: The error? That's just for the log, it's not used, but when it will be used, we already know the correct quota.
If you mean the current state `ENFClientLocalData`

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/windows/entities/cases/TestCase.kt
```diff
@@ -6,9 +6,9 @@ data class TestCase(
     @SerializedName("description")
     val description: String,
     @SerializedName("expAgeOfMostRecentDateWithHighRisk")
-    val expAgeOfMostRecentDateWithHighRisk: Long?,
+    val expAgeOfMostRecentDateWithHighRiskInDays: Long?,
```
chris-cwa, line 9: minor: if it is days, Int would suffice, i believe?

chris-cwa (approved):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/nearby/windows/entities/cases/TestCase.kt
d4rken: I agree, doesn't matter though for the test.

BMItr (approved):

PR #1686, opened 2020-11-20
Grey card undetermine state without wifi (DEV, EXPOSUREAPP-3892)

BMItr:
Quick fix for grey card state (1.8.x RC)


BMItter:
issue is under investigation. (DRAFT)

kolyaopahle (dismissed):
Worked on my device 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
```diff
@@ -7,7 +7,8 @@ import kotlinx.coroutines.flow.MutableStateFlow
 
 object RiskLevelRepository {
 
-    private val internalRisklevelScore = MutableStateFlow(getLastSuccessfullyCalculatedScore().raw)
+    private val internalRisklevelScore = MutableStateFlow(if (getLastSuccessfullyCalculatedScore() !=
```
ralfgehrer, line 10: I'd recommend moving this into a function call.

Also, getLastSuccessfullyCalculatedScore() is called twice, please "cache" it in a val.

Most importantly, I don't see how why this value has to be overwritten in this particular Flow initializer. getLastSuccessfullyCalculatedScore() solely accesses the LocalData. UNDETERMINED  is a default value if no data have been found/stored or the risk value integer value cannot be mapped to a constant. IMHO this is not addressing the issue in the right place. How does this relate to the no-wifi problem?

Lastly, please ensure it does not break the fixed behavior of EXPOSUREAPP-3723.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
```diff
@@ -7,7 +7,8 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
+        RiskLevel.UNDETERMINED) getLastSuccessfullyCalculatedScore().raw else RiskLevel.UNKNOWN_RISK_INITIAL.raw)
```
ralfgehrer, line 11: I'd recommend moving this into a function call.

Also, getLastSuccessfullyCalculatedScore() is called twice, please "cache" it in a val.

Most importantly, I don't see how why this value has to be overwritten in this particular Flow initializer. getLastSuccessfullyCalculatedScore() solely accesses the LocalData. UNDETERMINED is a default value if no data have been found/stored or the risk value integer value cannot be mapped to a constant. IMHO this is not addressing the issue in the right place. How does this relate to the no-wifi problem?

Lastly, please ensure it does not break the fixed behavior of EXPOSUREAPP-3723.

ralfgehrer (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
BMItter: there were just some minutes left on Friday for trying to fix this. I'm going to optimize the code. Will also check Exposureapp-3723 for collisions. If it does, further investigations by the original author are required.

Oliver-Zimmerman: I think it should be if(getLastSuccessfullyCalculatedScore().raw != RiskLevel.UNDETERMINED) getLastSuccesfullCalculatesScore.raw() else RiskLevelConstants.UNKNOWN_RISK_INITIAL

This more closely aligns with the original and the changes made here:
https://github.com/corona-warn-app/cwa-app-android/pull/1597/files

I can test when I get a chance

Oliver-Zimmerman: Although I think they effectively do the same thing.

Oliver-Zimmerman (dismissed):
Works for me and seems to resolve EXPOSUREAPP-3892 as well. 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
BMItter: same effect.

kolyaopahle (dismissed):

Oliver-Zimmerman (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
```diff
@@ -7,14 +7,28 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
     private val internalRiskLevelScoreLastSuccessfulCalculated =
         MutableStateFlow(LocalData.lastSuccessfullyCalculatedRiskLevel().raw)
     val riskLevelScoreLastSuccessfulCalculated: Flow<Int> =
         internalRiskLevelScoreLastSuccessfulCalculated
 
+    /**
+     * Set the internal risk level score [internalRisklevelScore]
+     */
+    private fun setInternalRiskLevelScore(): MutableStateFlow<Int> {
+        val lastSuccessfullyCalculatedScore = getLastCalculatedScore().raw
```
ralfgehrer, line 22: Thanks! I still don't get though, why we do this here in the Repo and not at the source in  LocalData.lastCalculatedRiskLevel().
Currently, we have overwriting behavior in multiple places.

chris-cwa (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
Oliver-Zimmerman: No, I think you're correct and I have adjusted the code to match this.

BMItter:
device Tests looking good so far (with & without connection). We still need to check for sideeffects...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
BMItter: @ralfgehrer @InterestedParties 
what do you think about the UNDETERMINED state now?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
```diff
@@ -14,7 +14,6 @@ object RiskLevelRepository {
         MutableStateFlow(LocalData.lastSuccessfullyCalculatedRiskLevel().raw)
```
ralfgehrer, line 14: Here we access "LocalData.lastSuccessfullyCalculatedRiskLevel().raw" directly whereas we use "getLastSuccessfullyCalculatedScore().raw" in line 10 
We should align this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/RiskLevelRepository.kt
ralfgehrer: @BMItter not quite sure I understand your question.

This is what it says in our code. Hence, initializing sth with that value is probably not the best idea.
`  
    // mapped to no UI state
    // this should never happen
    UNDETERMINED(RiskLevelConstants.UNDETERMINED);
`

harambasicluka:
In https://github.com/corona-warn-app/cwa-app-android/pull/1699 the UNKOWN_RISKS ARE removed, so do we even need this PR?

@BMItter @SamuraiKek @ralfgehrer @d4rken

d4rken:
Maybe wait with this until #1705 is merged. The whole `RiskLevelRepository` will be removed. There will be single source of truth for all risk level things.

d4rken:
After #1705 and #1699 is merged, we should reevaluate this bug again. Most of the bugs of this type were caused by weird states due to some values being stale, most of that will be fixed with these 2 PRs.

harambasicluka:
Ticket was set to "obsolete", will close this PR.

PR #1690, opened 2020-11-23
Risk Calculation does not work because DiagnosisKeysDataMapping is not applied (EXPOSUREAPP-3895)

kolyaopahle:
### Description
This PR includes config checking for the DiagnosisKeysDataMapping and applies it to the enfclient if configuration is deemed to have changed


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -60,4 +64,19 @@ class ConfigChangeDetector @Inject constructor(
             RiskLevelRepository.setRiskLevelScore(RiskLevel.UNDETERMINED)
         }
     }
+
+    @VisibleForTesting
+    internal suspend fun checkForDiagnosisKeysDataMapping(exposureWindowRiskCalculationConfig: ExposureWindowRiskCalculationConfig) {
+        val oldDiagnosisKeyDataMapping = enfClient.getDiagnosisKeysDataMapping()
+        val newDiagnosisKeyDataMapping = exposureWindowRiskCalculationConfig.diagnosisKeyDataMapping
+
+        if (oldDiagnosisKeyDataMapping != newDiagnosisKeyDataMapping) {
```
mlenkeit, line 73: Does this compare object references or compare by value?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -60,4 +64,19 @@ class ConfigChangeDetector @Inject constructor(
             RiskLevelRepository.setRiskLevelScore(RiskLevel.UNDETERMINED)
         }
     }
+
+    @VisibleForTesting
+    internal suspend fun checkForDiagnosisKeysDataMapping(exposureWindowRiskCalculationConfig: ExposureWindowRiskCalculationConfig) {
```
mlenkeit, line 69: Is this also invoked if the app config cannot be loaded initially and the client falls back to the default app config?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -60,4 +64,19 @@ class ConfigChangeDetector @Inject constructor(
...
+        val oldDiagnosisKeyDataMapping = enfClient.getDiagnosisKeysDataMapping()
```
mlenkeit, line 70: This is the _current_ diagnosisKeysDataMapping, right? Might be worth adjusting the variable name accordingly 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ExposureWindowRiskCalculationConfig.kt
```diff
@@ -13,6 +14,7 @@ interface ExposureWindowRiskCalculationConfig {
         List<RiskCalculationParametersOuterClass.NormalizedTimeToRiskLevelMapping>
     val normalizedTimePerDayToRiskLevelMappingList:
         List<RiskCalculationParametersOuterClass.NormalizedTimeToRiskLevelMapping>
+    val diagnosisKeyDataMapping: DiagnosisKeysDataMapping
```
mlenkeit, line 17: Typo: key**s** (plural)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -60,4 +64,19 @@ class ConfigChangeDetector @Inject constructor(
...
+        val newDiagnosisKeyDataMapping = exposureWindowRiskCalculationConfig.diagnosisKeyDataMapping
+
+        if (oldDiagnosisKeyDataMapping != newDiagnosisKeyDataMapping) {
+            try {
+                Timber.i("New DiagnosisKeysDataMapping differs from last one, applying.")
+                enfClient.setDiagnosisKeysDataMapping(newDiagnosisKeyDataMapping)
```
mlenkeit, line 76: This can throw exceptions for various reasons, at least:
1. rate limit error
1. invalid param

Only the rate limit error should be silent, other errors should not (otherwise, we wouldn't figure out that there's an error)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
kolyaopahle: should be an equal comparison, not 100% sure if its enough though, will do some checking

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/ExposureWindowRiskCalculationConfigMapper.kt
```diff
@@ -34,10 +41,24 @@ class ExposureWindowRiskCalculationConfigMapper @Inject constructor() :
             normalizedTimePerExposureWindowToRiskLevelMapping = riskCalculationParameters
                 .normalizedTimePerEWToRiskLevelMappingList,
             normalizedTimePerDayToRiskLevelMappingList = riskCalculationParameters
-                .normalizedTimePerDayToRiskLevelMappingList
+                .normalizedTimePerDayToRiskLevelMappingList,
+            diagnosisKeysDataMapping = rawConfig.diagnosisKeysDataMapping()
         )
     }
 
+    private fun AppConfigAndroid.ApplicationConfigurationAndroid.diagnosisKeysDataMapping():
```
d4rken, line 49: Why are we creating the mapping, and then taking the mapping, from protobuf class to protobuf class?
Could we use a data class here, to keep the protobuf classes hidden?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
```diff
@@ -0,0 +1,57 @@
...
+                if (e.statusCode == FAILED_RATE_LIMITED) {
+                    Timber.e(e, "Failed to setDiagnosisKeysDataMapping due to rate limit ")
+                } else {
+                    throw e
+                }
+            }
+        }
+    }
+
+    private fun DiagnosisKeysDataMapping.hasChanged(old: DiagnosisKeysDataMapping?): Boolean {
```
d4rken, line 53: Please add unit-tests for this.
* Direct matching
* Empty matching
* Null matching
* whole map changed
* one key in a map changed
* one value in a map changed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    private suspend fun setDiagnosisKeysDataMapping(diagnosisKeysDataMapping: DiagnosisKeysDataMapping) =
+        suspendCoroutine<Unit> { cont ->
+            client.setDiagnosisKeysDataMapping(diagnosisKeysDataMapping)
+                .addOnSuccessListener { cont.resume(Unit) }
+                .addOnFailureListener { cont.resumeWithException(it) }
+        }
+
+    override suspend fun updateDiagnosisKeysDataMapping() {
+        val currentDiagnosisKeysDataMapping = getDiagnosisKeysDataMapping()
```
d4rken, line 35: Assume that `getDiagnosisKeysDataMapping` can fail too, try catch + log it and early abort.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
```diff
@@ -0,0 +1,57 @@
...
+        val newDiagnosisKeysDataMapping = appConfigProvider
```
d4rken, line 36: I think it's nicer if we pass `ExposureWindowRiskCalculationConfig` via function parameter. Updating the mapping is always tied to a submission, as the app config can change asynchronously, this could lead to 2 different configs being used in the task and here.

Task gets Appconfig -> ENFClient provide diagnosisKeys(config, files)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
```diff
@@ -0,0 +1,57 @@
...
+            .getAppConfig().diagnosisKeysDataMapping
+
+        if (newDiagnosisKeysDataMapping.hasChanged(currentDiagnosisKeysDataMapping)) {
+            try {
+                Timber.i("New DiagnosisKeysDataMapping differs from last one, applying.")
```
d4rken, line 41: Let's log the old and the new config if this happens.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/ExposureWindowRiskCalculationConfigMapper.kt
kolyaopahle: the second class is not a protobuf but the internal enf config used by google, we could also implement a data class here, i oriented the initial implementation against the old mapRiskScoreToExposureConfiguration https://github.com/corona-warn-app/cwa-app-android/pull/1677/files#diff-0b01d7464a4e38df4e564db97b132516563e7b4ff1d004f4b6b0110d193f79edL67

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
kolyaopahle: Comparison has been changed and is now unittested

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
kolyaopahle: config change detector is no longer used

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
kolyaopahle: fixed in new place

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ExposureWindowRiskCalculationConfig.kt
kolyaopahle: fixed typo

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
kolyaopahle: new implementation rethrows error if status code is not Rate Limit

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
kolyaopahle: Unit tests have been added

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
kolyaopahle: Invocation has been moved to provide diagnosis key task

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
kolyaopahle: should now be caught and handled appropriately

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/diagnosiskeysdatamapper/DefaultDiagnosisKeysDataMapper.kt
kolyaopahle: has been added

BMItr (approved):
lgtm :ok_hand:

chris-cwa (approved):

PR #1694, opened 2020-11-23
Ignore negative secondsSinceLastScan (EXPOSUREAPP-3897)

chris-cwa:
The API can return a negative value for secondsSinceLastScan (seen on my device). Most likely, this is related to time travel and should not happen in production. Anyway, to avoid potential issues from this in the first place - regardless of whether it's PROD or testing - the client should handle this properly.


mlenkeit (approved):
LGTM

BMItr (approved):
looks good to me. thx for adding adjusten test.json

PR #1696, opened 2020-11-23
Exposure result repo (EXPOSUREAPP-3910)

chris-cwa:


chris-cwa:
not doing it this way

PR #1698, opened 2020-11-23
Retain the cache after connection timeout (EXPOSUREAPP-3912)

ralfgehrer:
If we run into a connection time while downloading the hour index, we must not delete the cache but return an unsuccessful result instead.


chiljamgossow (approved):

harambasicluka:
Has to be decided if it is 1.7.1 relevant.

harambasicluka (approved):

d4rken (approved):

Issue #1081, opened 2020-08-27
"Risk determination not possible" - wrong info


Jo-Achim:
_I've read this topic here before - but I can't find it right now, sorry._

Android 10 with Android security update: July 1st, 2020
Nokia 7.1, Build: 00WW_4_15C_SP02
NetGuard: 2.287
Internet: WLAN (only)
CWA version: 1.2.1

Problem: "Risk determination not possible" is displayed although the risk determination is active; see screenshots.

General condition: The last CWA data exchange took place on 08/24/2020 at 7:38 am (thereafter no more internet connections until 08/26/2020).
The next time the CWA was started (with internet connection) on August 26, 2020 at around 7:08 pm / 7:09 pm, the incorrect risk determination information was displayed:
![Screenshot_20200826-190939](https://user-images.githubusercontent.com/21124877/91381912-ca625800-e828-11ea-8a1c-608b342f1c06.png)

After [->] the updated, correct data including the risk assessment were displayed:
![Screenshot_20200826-190953](https://user-images.githubusercontent.com/21124877/91381931-d77f4700-e828-11ea-8ff6-5b0f85df2306.png)

![Screenshot_20200826-191006](https://user-images.githubusercontent.com/21124877/91381940-dcdc9180-e828-11ea-94e2-24aee72d30d1.png)

Best regards.

MikeMcC399:
Hi @Jo-Achim,

The issue is that the status "RISIKO-ERMITTLUNG AKTIV" is contradictory to the button "RISIKO-ERMITTLUNG EINSCHALTEN". If Exposure Logging is already active then there is no need to switch Exposure Logging on and the user should not be prompted to do so. That presents a confusing message to the user.

It's OK for a temporary status "Risiko-Ermittlung nicht möglich" if the app cannot pull the diagnosis keys from the server when it tries. That status was cleared and replaced by "Niedriges Risiko" after the diagnosis keys were fetched and compared to the keys on the mobile device.

Jo-Achim:
Hi @MikeMcC399, 

yes, on the one hand I wanted to draw your attention to the fact that at least the information "switch on risk determination" when "risk determination is switched on" is wrong.

On the other hand, I believe that the message "Risk determination not possible" / "Risiko-Ermittlung nicht möglich" should not appear at all if the diagnostic keys were exchanged in the background **at the same time**. Because **in this case** the info is irrelevant. 
[...]

PR #1699, opened 2020-11-23
Removed traces of grey risk card (EXPOSUREAPP-3051)

SamuraiKek:
* Removes the initial "unknown risk" grey card
* Adds a new card for "Failed because no internet"
* Refactored the RiskLevel to be determined by two other values. Our "RiskLevel" is now based on the whether "does an aggregated result exist" and "a failure reason which must exist if there is no aggregated result". This makes us very flexible to add new failure states the the risk level calculation process. The `FailureReason`s are enums stored in a database. More robust than the previous plain numbers.
* Simplified the RiskCardState by dropping manual key retrieval time an using "isBackgroundJobEnabled" directly.

## Testing 
Try to produce all different card states :wink:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevel.kt
```diff
@@ -58,7 +56,6 @@ enum class RiskLevel(val raw: Int) {
             )
         private val HIGH_RISK_LEVELS = arrayOf(INCREASED_RISK)
         private val LOW_RISK_LEVELS = arrayOf(
-            UNKNOWN_RISK_INITIAL,
             NO_CALCULATION_POSSIBLE_TRACING_OFF,
             LOW_LEVEL_RISK,
             UNKNOWN_RISK_OUTDATED_RESULTS,
```
mlenkeit, line 61: What do we need this for?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/tracing/card/TracingCardState.kt
```diff
@@ -272,13 +269,15 @@ data class TracingCardState(
     fun formatRiskLevelHeadline(c: Context, riskLevelScore: Int): String {
         return if (tracingStatus != GeneralTracingStatus.Status.TRACING_INACTIVE) {
             when (riskLevelScore) {
-                RiskLevelConstants.INCREASED_RISK -> R.string.risk_card_increased_risk_headline
-                RiskLevelConstants.UNKNOWN_RISK_OUTDATED_RESULTS -> R.string.risk_card_outdated_risk_headline
+                RiskLevelConstants.INCREASED_RISK ->
+                    R.string.risk_card_increased_risk_headline
+                RiskLevelConstants.UNKNOWN_RISK_OUTDATED_RESULTS,
+                RiskLevelConstants.UNKNOWN_RISK_OUTDATED_RESULTS_MANUAL ->
```
mlenkeit, line 275: What do we need this for?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/tracing/common/BaseTracingState.kt
```diff
@@ -29,10 +29,8 @@ abstract class BaseTracingState {
         return if (tracingStatus != GeneralTracingStatus.Status.TRACING_INACTIVE) {
             when (riskLevelScore) {
                 RiskLevelConstants.INCREASED_RISK -> R.color.colorSemanticHighRisk
-                RiskLevelConstants.UNKNOWN_RISK_OUTDATED_RESULTS,
-                RiskLevelConstants.NO_CALCULATION_POSSIBLE_TRACING_OFF -> R.color.colorSemanticUnknownRisk
                 RiskLevelConstants.LOW_LEVEL_RISK -> R.color.colorSemanticLowRisk
-                else -> R.color.colorSemanticNeutralRisk
+                else -> R.color.colorSemanticUnknownRisk
```
mlenkeit, line 33: What do we need this for?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/riskdetails/DefaultRiskDetailPresenterTest.kt
```diff
@@ -12,7 +12,6 @@ class DefaultRiskDetailPresenterTest {
         DefaultRiskDetailPresenter().apply {
             Assert.assertFalse(isAdditionalInfoVisible(RiskLevelConstants.LOW_LEVEL_RISK, 0))
             Assert.assertTrue(isAdditionalInfoVisible(RiskLevelConstants.LOW_LEVEL_RISK, 1))
-            Assert.assertFalse(isAdditionalInfoVisible(RiskLevelConstants.UNKNOWN_RISK_INITIAL, 0))
             Assert.assertFalse(isAdditionalInfoVisible(RiskLevelConstants.INCREASED_RISK, 0))
             Assert.assertFalse(isAdditionalInfoVisible(RiskLevelConstants.UNKNOWN_RISK_OUTDATED_RESULTS, 0))
```
mlenkeit, line 16: This state should also be gone I think?!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevel.kt
SamuraiKek: We were using it for displaying the grey risk cards. Basically before the risk calculation could take place after the onboarding, the initial risk score was attributed this constant value. By removing the grey card, I don't see a need for this constant anymore as the initial risk value is attributed to the LOW_RISK constant now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/tracing/card/TracingCardState.kt
SamuraiKek: While these two are also used in relation to the grey risk card, they are also needed for displaying the "Risiko-Ermittlung
nicht möglich" card which is a white card from what I can see in the Android Figma file. AFAIK there are instances where the risk calculation can fail after it's been calculated at least once, so these constants should be relevant for the cards with the white background (at least for now).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/tracing/common/BaseTracingState.kt
SamuraiKek: Opposite to what the names suggest, the colorSemanticNeutralRisk is the grey color used for the grey card, and colorSemanticUnknownRisk is white. The fallback color was grey before and I've now changed it to white since there won't be a grey card anymore.
Maybe these color names should be interchanged but I didn't want to introduce more unnecessary modifications for now.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/riskdetails/DefaultRiskDetailPresenterTest.kt
SamuraiKek: But I've removed it already.

harambasicluka:
@mlenkeit should all grey cards be removed or only the initial one?

harambasicluka:
Agreed with @mlenkeit that this is fine, but @SamuraiKek could you verify when `UNKNOWN_RISK_OUTDATED_RESULTS` would be visible? Not sure when i have seen it the last time? I think it's in but technically overruled by the white card.

SamuraiKek:
> Agreed with @mlenkeit that this is fine, but @SamuraiKek could you verify when `UNKNOWN_RISK_OUTDATED_RESULTS` would be visible? Not sure when i have seen it the last time? I think it's in but technically overruled by the white card.

I'm not sure when it would be visible. I think the values are assigned in RiskLevelTask, when the calculationNotPossibleBecauseOfOutdatedResults() method is called. Now whether this method is ever called, or if the values are overwritten someplace else, I'm not really sure right now.  I'll try and look more into it when working on #1615 but for now, I don't see why keeping them in would cause any issue.

harambasicluka (dismissed):
To resolve the merge conflicts it should be fine if you just add the removed strings again and only remove them in german.

Oliver-Zimmerman:
If we remove the UNKNOWN state then this PR becomes obsolete:
https://github.com/corona-warn-app/cwa-app-android/pull/1686

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/legacy/RiskLevelResultMigrator.kt
```diff
@@ -67,9 +69,24 @@ class RiskLevelResultMigrator @Inject constructor(
...
+        override val failureReason: RiskLevelResult.FailureReason? = null
         override val aggregatedRiskResult: AggregatedRiskResult? = null
         override val exposureWindows: List<ExposureWindow>? = null
         override val matchedKeyCount: Int = 0
         override val daysWithEncounters: Int = 0
     }
+
+    companion object {
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal fun mapRiskLevelConstant(value: Int): RiskLevel {
```
harambasicluka, line 81: why is this mapping necessary?

harambasicluka (commented):
Looks really good! I just got one question, will go to device testing now.

harambasicluka (dismissed):
I tested the following states, please correct me if I missed one:

- [x] INCREASED_RISK -> via test menu
- [x] LOW_LEVEL_RISK -> initially at the start
- [x] UNKNOWN_RISK_OUTDATED_RESULTS -> disabled tracing
- [ ] UNKNOWN_RISK_OUTDATED_RESULTS_MANUAL -> not able to get into this state, but I'm also not sure when I've seen it the last time, should be investigated in a follow up ticket
- [x] UNKNOWN_RISK_NO_INTERNET -> turned wifi off, time traveled, forced risk calculation via tester menu

image

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/legacy/RiskLevelResultMigrator.kt
d4rken: We move to a new risklevel/failure data state, but on the first update of the user we still need to know what the old 1,2,3,4 risk level constants mean, to provide an initial value. We could omit the whole migration if we are okay that the users have their card switch from red to green, if they can't calculate the first risklevel after the update.

harambasicluka: Thanks!

d4rken (approved):

harambasicluka (approved):

vaubaehn:
> 
> 
> > Agreed with @mlenkeit that this is fine, but @SamuraiKek could you verify when `UNKNOWN_RISK_OUTDATED_RESULTS` would be visible? Not sure when i have seen it the last time? I think it's in but technically overruled by the white card.
> 
> I'm not sure when it would be visible. I think the values are assigned in RiskLevelTask, when the calculationNotPossibleBecauseOfOutdatedResults() method is called. Now whether this method is ever called, or if the values are overwritten someplace else, I'm not really sure right now. I'll try and look more into it when working on #1615 but for now, I don't see why keeping them in would cause any issue.

Iirc it's also visible in the current version when there is a failure with exposure checking for more than 24 hrs (API 39508, API 10 no public key, BackgroundWorker does not start...), but tracing is on - see also #1081 (EXPOSUREAPP-2796).  For now, failure and tracing off will result in the same risk card. But I understood
> We move to a new risklevel/failure data state [...]

like this will be addressed elsewhere?

PR #1706, opened 2020-11-24
Test feature to fake high/low risk encounters via exposure windows (DEV)

d4rken:
As it's currently way too complicated to get the app into risk states with encounters to test various app behaviors, I wrote this test feature which I think would be useful for everyone else too.

The `deviceForTesters` flavor has a different `ExposureWindowsProvider`. 
Depending on which debug option was chosen exposure windows window data will be loaded from json files and then returned via the `ENFClient`.

From the apps perspective it will look like Googles exposure detection calculated windows with low/high risk encounters after the last submission.

Based on @BMItter `ExposureWindowProvider` and @mlenkeit ExposureWindow test data concepts.

### Testing
* Check that in `deviceForTesters` the option is there, and that in `device` the whole code is not included.

Screenshot from 2020-11-24 09-20-23


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/nearby/modules/exposurewindow/FakeExposureWindowProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+    @AppContext val context: Context
+) {
+
+    fun getExposureWindows(testSettings: FakeExposureWindowTypes): List<ExposureWindow> {
+        val jsonInput = when (testSettings) {
+            FakeExposureWindowTypes.INCREASED_RISK_DEFAULT -> "exposure-windows-increased-risk-random.json"
+            FakeExposureWindowTypes.LOW_RISK_DEFAULT -> "exposure-windows-lowrisk-random.json"
+            else -> throw NotImplementedError()
+        }.let { context.assets.open(it) }.readBytes().toString(Charsets.UTF_8)
+        val jsonWindows: List<JsonWindow> = Gson().fromJson(jsonInput)
```
ralfgehrer, line 27: Minor: replace Gson() by constructor  inject

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/nearby/modules/exposurewindow/FakeExposureWindowProvider.kt
```diff
@@ -0,0 +1,62 @@
...
+
+        return jsonWindows.map { jWindow ->
+            ExposureWindow.Builder().apply {
+                setDateMillisSinceEpoch(
+                    Instant.now().minus(Duration.standardDays(jWindow.ageInDays.toLong())).millis
```
ralfgehrer, line 32: Minor: use timestamper instead of Instant.now()

ralfgehrer (approved):

BMItter:
tested on Pixel 2 Device. Code is excluded to deviceForTesters correctly but TestSettings.kt (necessarily).

One optional suggestion:
Could we change the list order to assending? (Disabled->Low->High)

d4rken:
> One optional suggestion:
> Could we change the list order to assending? (Disabled->Low->High)

Sure maybe not now? Don't want to delay the PR. The next person that adds a new fake data, can change the orders I just added some defaults, might not cover all UI states we need.

BMItr (approved):

Issue #1609, opened 2020-11-16
Application 1.6.0 crashes on startup - MediaTek bug


openmindculture:
The current android version keeps crashing on startup.
Steps to reproduce: touch the app icon. After showing a splash screen for a few seconds, the app closes without further notice.
Earlier versions of the app did work on the same phone.
Android systems settings report that corona risk detection is active and that the German Corona Warn App is assigned to manage them. Bluetooth and location service is activated.
Error keeps recurring also after restarting the device.

App version: 1.6.0 (updated using google play store on 9 November 2020).
Android version: 7.0 (security patch 1 December 2018) Kernel 3.18.35+jslave@WUH1000074104 \#3
Hardware: Huawei Honor JMM-L22
Build JMM-L22C432B136
EMU-Version 5.1.3

## Avoid duplicates
* [ ] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [ ] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ ] Bug is not already reported in another issue

## Technical details

- Mobile device:
- Android version:

## Additional context

---
Internal Tracking ID: EXPOSUREAPP-3847

JHKobarg:
Hi, I observed the same behaviour with the 1.6.0 version of the app on both my parent’s HTC Desire 12 (with Android 7.1.1, Kernel 4.4.22+); one device failed to start Friday, the other since today; the App shows the logo/splash screen and then crashes directly; the symptom appears to be identical to the description from #1053 but I tried to execute the countermeasure described there and they are not effective.

The system log begins with 
```
F/libc (339): unable to stat "/proc/self/exe": Permission denied
F/libc (339): Fatal signal 6 (SIGABRT), code -6 in tid 339 (init.lct.bootch)
```
If you need more information I will try to provide that.

vaubaehn:
Hi @JHKobarg and @openmindculture ,
your issues are most likely not connected to #1053 (with #642 as the underlying root cause). However, the problem is addressed in PRs #1612 and #1604. Interpreting #1620 , a hotfix v1.6.1 should be available in the Google Play Store soon.

@d4rken : FYI. In case you need more logs, @JHKobarg could probably help out here?

d4rken:
A system log (logcat) from a few seconds before the app is opened till after it crashes would be very welcome.

[...]

PR #1707, opened 2020-11-24
Simplify polling based flows (DEV, EXPOSUREAPP-3847)

d4rken:
Simplify the flows.

* We don't need the callback flows
* Handle `CancellationException` better
* Possible positive effects towards #1609


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/BackgroundModeStatus.kt
```diff
@@ -21,41 +21,41 @@ class BackgroundModeStatus @Inject constructor(
...
             }
-
-            if (!isActive) break
-
-            delay(POLLING_DELAY_MS)
         }
     }
         .distinctUntilChanged()
+        .onCompletion {
+            if (it != null) Timber.w(it, "autoModeEnabled failed.")
```
ralfgehrer, line 57: How can the value ever be `null`?

d4rken: If we cancel the coroutine, we just break the `while(true)` and the `flow { }` ends, then we would complete without exception.

The implicit signature here is

```kotlin
.onCompletion { it: Throwable? ->

}
```

ralfgehrer (approved):

harambasicluka (approved):

PR #1708, opened 2020-11-24
Screen: Andere warnen --> Button Einverstanden greyed out (EXPOSUREAPP-3770)

kolyaopahle:
### Description
This pr refactors the SubmissionResultPositiveOtherWarningViewModel to fix the issues with an old (pre app resest) task being used to display the ui state for the submission task. Additionally it removes the dependency on ApiRequestState

### Steps to reproduce
1. Register a positive test
2. Complete the submission and return to homescreen
3. Reset app trough the settings menu (without restarting the app!)
4. Register a positive test
5. Try to complete the submission
6. If the Einverstanden button is greyed out the fix did not work


Oliver-Zimmerman (approved):
LGTM 👍 tested on Samsung S9 Plus and Pixel 5. Thanks again for help on this ticket

harambasicluka:
Can be merged in release/1.8.x

harambasicluka:
If we don't release a hotfix this will go in 1.8.0 if we create a hotfix this will go in 1.7.1. This depends on the outcome of EXPOSUREAPP-3937.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/warnothers/WarnOthersState.kt
```diff
@@ -1,17 +1,16 @@
 package de.rki.coronawarnapp.ui.submission.warnothers
 
+import de.rki.coronawarnapp.task.TaskState
 import de.rki.coronawarnapp.ui.Country
-import de.rki.coronawarnapp.ui.submission.ApiRequestState
 
 data class WarnOthersState(
-    val apiRequestState: ApiRequestState,
+    val submitTaskState: TaskState?,
```
ralfgehrer, line 7: Could this be implemented without a nullable state?

ralfgehrer (approved):
Tested on Pixel (Android 10) and the fix works nicely.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/warnothers/WarnOthersState.kt
kolyaopahle: hmm, im not sure, as we might not have a started task yet and in that case the task list will return null, we could map this to something else but i don't really see the benefit of it

PR #1709, opened 2020-11-24
Update strings.xml (EXPOSUREAPP-3887)

janetback:
Added some punctuation marks


Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -345,9 +345,9 @@
     <!-- XHED: risk details - infection period logged information body, below behaviors -->
     <string name="risk_details_information_body_period_logged_assessment">"Für Ihre Risiko-Ermittlung wird nur der Zeitraum der letzten 14 Tage betrachtet. In diesem Zeitraum war Ihre Risiko-Ermittlung für eine Gesamtdauer von %1$s Tagen aktiv. Ältere Tage werden automatisch gelöscht, da sie aus Sicht des Infektionsschutzes nicht mehr relevant sind."</string>
     <!-- XHED: risk details - how your risk level was calculated, below behaviors -->
-    <string name="risk_details_subtitle_infection_risk_past">"So wurde Ihr Risiko ermittelt"</string>
+    <string name="risk_details_subtitle_infection_risk_past">"So wurde Ihr Risiko ermittelt."</string>
     <!-- XHED: risk details - how your risk level will be calculated, below behaviors -->
-    <string name="risk_details_subtitle_infection_risk">"So wird Ihr Risiko ermittelt"</string>
```
chris-cwa, line 350: no ':' here? because the explanation then follows....

ralfgehrer (approved):

chiljamgossow (approved):

Corona-Warn-App/src/main/res/values-de/strings.xml
janetback: hi @chris-cwa I put a colon in the first instance because it is followed by a bulleted list - I believe in this second instance it is perhaps more a matter of taste, but will double-check with my colleague tomorrow.

janetback: hi @chris-cwa - I have since checked with my colleague Daniela Vogler - the text that follows in the second instance is more of an explanation of your current risk status rather than the an explanation of  how exposure logging work. So if anything we should probably adapt the existing subtitle text slightly.
Sor for now and for this bug, we agreed that '.' here is the better option.

PR #1711, opened 2020-11-24
Consent status custom view (EXPOSUREAPP-3737)

AlexanderAlferov:
Consent status custom view

Screenshot 2020-11-24 at 17 00 02


Corona-Warn-App/src/main/res/layout/view_consent_status.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="utf-8"?>
```
chiljamgossow, line 1: Can you use a ConstraintLayout to flatten the view hierarchy?

AlexanderAlferov: Done ✅

chiljamgossow (dismissed):

Corona-Warn-App/src/main/res/layout/view_consent_status.xml
```diff
@@ -0,0 +1,76 @@
...
+<androidx.constraintlayout.widget.ConstraintLayout
```
d4rken, line 2: This currently uses 3-nested ConstraintLayouts, please make it use only one.

Corona-Warn-App/src/main/res/layout/view_consent_status.xml
```diff
@@ -0,0 +1,76 @@
...
+            android:layout_height="wrap_content"
+            android:layout_marginTop="@dimen/spacing_mega_tiny"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintVertical_bias="0.0">
+
+            <ImageView
+                style="@style/icon"
+                android:layout_width="@dimen/icon_size_risk_details_behavior"
```
d4rken, line 40: This custom view has no relationship with risk details behavior. I would prefer if we either add a new dimension (if there are other similar views with such icons as this one, or just hardcode it here as it a custom view.

d4rken (changes requested):

d4rken (dismissed):

chiljamgossow (dismissed):

BMItr (dismissed):
nice custom view :ok_hand:
just resolve last conflict and then give it a go.

BMItr (approved):

chiljamgossow (approved):

PR #1714, opened 2020-11-25
Make worker type visible on log output (DEV)

d4rken:
Testers had trouble recognizing the workers as the implicit logtag generation is not useful due to coroutines, before:

```
1605821280113  D/CoroutineWorker$startWork: 90bfd423-bf13-4701-9e0e-77f4383c3106: doWork() started. Run attempt: 0
```


ralfgehrer (dismissed):
Nice, thx. Maybe we can also change the epoch millis to a proper timestamp?

d4rken:
> Nice, thx. Maybe we can also change the epoch millis to a proper timestamp?

#1719

harambasicluka (dismissed):

ralfgehrer (approved):

chris-cwa (approved):

harambasicluka (approved):

PR #1715, opened 2020-11-25
Improve Gson deserialization handling (DEV)

d4rken:
Move Gson deserialization related sanity checks directly to the `fromJson(file: File)` function.

* Force the deserialized type to be not nullable
* If the file doesn't exist, the value doesn't exist and return null
* If the file exists, but is empty check for a deserialized null, delete the empty file and then return null. (An empty file to represent a nullable json should never have been written.
* Add a unit test that checks for exactly this behavior.


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/serialization/GsonExtensionsTest.kt
```diff
@@ -0,0 +1,51 @@
...
+    @Test
+    fun `serialize and deserialize`() {
+        val testData = TestData(value = UUID.randomUUID().toString())
+        gson.toJson(testData, testFile)
+
+        gson.fromJson<TestData>(testFile) shouldBe testData
+    }
+
+    @Test
+    fun `deserialize an empty file`() {
```
ralfgehrer, line 41: Would it make sense to also add a test case for a malformed JSON file that cannot be deserialized properly?

ralfgehrer (dismissed):

ralfgehrer (approved):

kolyaopahle (approved):

PR #1716, opened 2020-11-25
Prevent accidental use of protobuf default value for app features (DEV)

d4rken:
Make use of `hasAppFeatures` to prevent accidental use of default values.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -30,11 +30,23 @@ class CWAConfigMapper @Inject constructor() : CWAConfig.Mapper {
             else -> supportedCountriesList
         }
 
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun ApplicationConfigurationAndroid.mapAppFeatures(): List<AppFeaturesOuterClass.AppFeature> =
+        if (!hasAppFeatures()) {
```
ralfgehrer, line 35: very minor: Personally, I would prefer a non-negated if statement for easier readability

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -3,24 +3,24 @@ package de.rki.coronawarnapp.appconfig.mapping
...
 @Reusable
 class CWAConfigMapper @Inject constructor() : CWAConfig.Mapper {
-    override fun map(rawConfig: AppConfigAndroid.ApplicationConfigurationAndroid): CWAConfig {
+    override fun map(rawConfig: ApplicationConfigurationAndroid): CWAConfig {
         return CWAConfigContainer(
             latestVersionCode = rawConfig.latestVersionCode,
             minVersionCode = rawConfig.minVersionCode,
             supportedCountries = rawConfig.getMappedSupportedCountries(),
-            appFeatures = rawConfig.appFeatures
+            appFeatures = rawConfig.mapAppFeatures()
```
ralfgehrer, line 18: minor: one could add a test to verify `mapAppFeatures()` is being called

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
ralfgehrer: Verifying the actual call would require loosening up the visibility. The call is implicitly verified already. I rather opted for reducing visibility to private.

ralfgehrer (approved):

kolyaopahle (approved):

PR #1717, opened 2020-11-25
Improve TaskController logging and limit state history (DEV)

d4rken:
Fixes a few things I noticed after looking through multiple logs:

* Logging was a bit too verbose and could be better formatted.
* We have no limit on the states in the TaskController's history. To not get into memory troubles at some point, we should limit it. The limit is now `50`, and we only remove task states that are `finished`. Background operation would cause about 12 entries per day (6x download+risklevel).


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -137,7 +136,19 @@ class TaskController @Inject constructor(
             this.putAll(it)
         }
 
-        Timber.tag(TAG).v("Tasks after processing: %s", this.values)
+        if (size > TASK_HISTORY_LIMIT) {
+            Timber.v("Enforcing history limits (%d), need to remove %d.", TASK_HISTORY_LIMIT, size - TASK_HISTORY_LIMIT)
+            values
+                .filter { it.isFinished }
+                .sortedBy { it.startedAt }
```
ralfgehrer, line 143: minor: with a history of 50 probably just an edge case but wouldn't be sorted by finishedAt be more beneficial? In case we have long and short running tasks?

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
d4rken: Good point. Done.

ralfgehrer (approved):

kolyaopahle (approved):

PR #1720, opened 2020-11-25
Log cause for HTTP errors (DEV)

d4rken:
* When http errors occur, try to log the error message.
* If the http error request fails without status code, log the cause.


harambasicluka (approved):

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/http/HttpErrorParser.kt
```diff
@@ -34,43 +35,54 @@ class HttpErrorParser : Interceptor {
     override fun intercept(chain: Interceptor.Chain): Response {
         try {
             val response = chain.proceed(chain.request())
+
+            val message: String? = try {
+                if (response.isSuccessful) {
+                    null
+                } else {
+                    response.message
```
reschke, line 43: Is that the HTTP reason phrase (RFC 7230, Section 3.1.2)? In which case note that it doesn't exist in HTTP/2.

PR #1729, opened 2020-11-26
Update symptom cancellation card (EXPOSUREAPP-3747)

d4rken:
* New texts
* Remember the last picked date when going one step back


chiljamgossow (approved):
LGTM, also tested on device

janetback (approved):
UA reviewed

PR #1732, opened 2020-11-26
Fix notification issues by introducing IDs (EXPOSUREAPP-3971)

ralfgehrer:
This PR fixes the notification part of EXPOSUREAPP-3971.

So far, every notification had its own random ID which is bad for two reasons:
- if triggered multiple times, notifications for the same event (type) pile up
- we cannot cancel notifications without knowing their id

This PR introduces two fixed notification types 
`NEW_MESSAGE_RISK_LEVEL_SCORE_NOTIFICATION_ID` and `NEW_MESSAGE_TEST_RESULT_NOTIFICATION_ID`

Moreover, if a new test result notification is scheduled, all risk level score notifications are canceled.
_EDIT_: Additional testing showed that the cancelation of notifications is not working reliably. Will be further investigated and addressed in an upcoming release.

**How to test:**
- put the app in a high-risk state --> a notification should show up (don't click on it  nor dismiss it - keep it in the notification drawer)
- scan a QR code with a positive test result 
Expected  result: the notification should be dismissed

More advanced:
- put the app in a high-risk state (see above)
- don't dismiss the notification
- generate a  QR Code that is pending
- Scan it with the app
- but the app in the background
- eventually, a notification should remind you that you received a new test  result 
Expected result: There is only one notification in the navigation drawer.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationConstants.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationConstants.kt
@@ -39,14 +39,4 @@ object NotificationConstants {
      * Notification channel description String.xml path
      */
     const val CHANNEL_DESCRIPTION = R.string.notification_description
-
-    /**
-     * Risk changed notification title String.xml path
-     */
-    const val NOTIFICATION_CONTENT_TITLE_RISK_CHANGED = R.string.notification_headline
-
-    /**
-     * Risk changed notification content text String.xml path
-     */
-    const val NOTIFICATION_CONTENT_TEXT_RISK_CHANGED = R.string.notification_body
 }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationConstants.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationConstants.kt
@@ -17,6 +17,9 @@ object NotificationConstants {
     val POSITIVE_RESULT_NOTIFICATION_INITIAL_OFFSET: Duration = Duration.standardHours(2)
     val POSITIVE_RESULT_NOTIFICATION_INTERVAL: Duration = Duration.standardHours(2)
 
+    const val NEW_MESSAGE_RISK_LEVEL_SCORE_NOTIFICATION_ID = 110
+    const val NEW_MESSAGE_TEST_RESULT_NOTIFICATION_ID = 120
+
     /**
      * Notification channel id String.xml path
      */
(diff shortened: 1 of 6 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationHelper.kt
```diff
@@ -94,6 +94,11 @@ object NotificationHelper {
         Timber.v("Canceled future notifications with id: %s", notificationId)
     }
 
+    fun cancelCurrentNotification(notificationId: Int) {
+        NotificationManagerCompat.from(CoronaWarnApplication.getAppContext()).cancel(notificationId)
```
harambasicluka, line 98: Has a check to be added if the notification exists are is this handled by the `NotificationManagerCompat`?

harambasicluka (commented):
Looks fine, only would add one log and I have a question regarding the cancelation of notifications.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationHelper.kt
ralfgehrer: Is handled by the manager.

kolyaopahle (approved):
LGTM 👍 
qick note for a later refactoring, maybe replace the generic `NotificationHandler` with specifict notification data-classes that define their own notification id and text used that the app can instantiate and pass to the NotificationHandler, this would be especially usefull when displaying and keeping track of the same kind of notifications in multiple places

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationConstants.kt
```diff
@@ -17,6 +17,9 @@ object NotificationConstants {
     val POSITIVE_RESULT_NOTIFICATION_INITIAL_OFFSET: Duration = Duration.standardHours(2)
     val POSITIVE_RESULT_NOTIFICATION_INTERVAL: Duration = Duration.standardHours(2)
 
+    const val NEW_MESSAGE_RISK_LEVEL_SCORE_NOTIFICATION_ID = 110
+    const val NEW_MESSAGE_TEST_RESULT_NOTIFICATION_ID = 120
```
chris-cwa, line 21: how about defining an enum type for the notification id? this way we would have a defined set of values and it is guaranteed that caller of NotificationHelper.send / cancel passes only valid numbers

ralfgehrer: Unfortunately, we still have the debug notifications which generate random ids. Afaik, they have been removed in 1.8. Therefore, I'd suggest postponing that refactoring to 1.8.

chris-cwa (approved):

BMItr (approved):

ralfgehrer:
> LGTM 👍
> qick note for a later refactoring, maybe replace the generic `NotificationHandler` with specifict notification data-classes that define their own notification id and text used that the app can instantiate and pass to the NotificationHandler, this would be especially usefull when displaying and keeping track of the same kind of notifications in multiple places

Created a follow-up ticket for this and @chris-cwa's comment targeting 1.8.

PR #1735, opened 2020-11-26
Use inclusive language (EXPOSUREAPP-3074)

chiljamgossow:
Replace master with main where possible
remove pronouns


Corona-Warn-App/src/main/res/values-bg/strings.xml
```diff
@@ -619,7 +619,7 @@
     <string name="settings_notifications_illustration_description_inactive">"Жена, която е дезактивирала известията от приложението Corona-Warn-App."</string>
     <!-- XBUT: settings - go to reset application -->
     <string name="settings_reset_title">"Изчистване на данните в приложението"</string>
-    <!-- XTXT: settings(reset) - explains the user what do expect when he navigates to reset -->
+    <!-- XTXT: settings(reset) - explains the user what do expect when she or he navigates to reset -->
```
BMItter, line 622: just a short question: Was it recommended to replace it with "she or he"?
It's a little bit strange to read as the user is a "general term"..
Maybe something like: "explains the user what to do when navigating to reset" would also be fine..
But I really don't know, just thinking..

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -620,7 +620,7 @@
     <string name="settings_notifications_illustration_description_inactive">"Eine Frau hat die Mitteilungen ihrer Corona-Warn-App ausgeschaltet."</string>
     <!-- XBUT: settings - go to reset application -->
     <string name="settings_reset_title">"Anwendung zurücksetzen"</string>
-    <!-- XTXT: settings(reset) - explains the user what do expect when he navigates to reset -->
+    <!-- XTXT: settings(reset) - explains the user what do expect when she or he navigates to reset -->
```
harambasicluka, line 623: To reduce merge conflicts only update something in german, it will be shipped to the other languages with the next translation system. 
Maybe we could get ride completely of genders in this spot:
```suggestion
    <!-- XTXT: settings(reset) - explains the user what to expect when navigating to the reset settings -->
```

harambasicluka (changes requested):
Please make changes to the language file only in german.

Corona-Warn-App/src/main/res/values-bg/strings.xml
chiljamgossow: a user can be male or female, he would only include men

chiljamgossow: It's strange because you are used to language that only addresses men

harambasicluka (approved):

janetback (approved):
UA approved

PR #1744, opened 2020-11-27
Merges recent fixes from 1.7.x into 1.8.x (DEV)

chiljamgossow:
Merges recent fixes from 1.7.x into 1.8.x

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationHelper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationHelper.kt
@@ -16,12 +16,12 @@ import androidx.core.app.NotificationCompat.PRIORITY_HIGH
 import androidx.core.app.NotificationManagerCompat
 import de.rki.coronawarnapp.BuildConfig
 import de.rki.coronawarnapp.CoronaWarnApplication
+import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.notification.NotificationConstants.NOTIFICATION_ID
 import de.rki.coronawarnapp.ui.main.MainActivity
 import org.joda.time.Duration
 import org.joda.time.Instant
 import timber.log.Timber
-import kotlin.random.Random
 
 /**
  * Singleton class for notification handling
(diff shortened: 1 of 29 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationHelper.kt
```diff
@@ -194,19 +199,39 @@ object NotificationHelper {
      * @param pendingIntent: PendingIntent
      */
     fun sendNotification(
-        title: String,
+        title: String = CoronaWarnApplication.getAppContext().getString(R.string.notification_name),
```
ralfgehrer, line 202: This default value should not be set

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
```diff
@@ -56,7 +57,9 @@ class RiskLevelChangeDetector @Inject constructor(
             Timber.d("Notification Permission = ${notificationManagerCompat.areNotificationsEnabled()}")
 
             if (!foregroundState.isInForeground.first()) {
-                NotificationHelper.sendNotification("", context.getString(R.string.notification_body), true)
```
ralfgehrer, line 59: IMHO the if clause could be removed and  `NotificationHelper .sendNotificationIfAppIsNotInForeground` be called instead.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/NotificationHelper.kt
chiljamgossow: Why? This way we make sure that there is no notification without title.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
d4rken: Then we depend on static access inside `NotificationHelper` instead of being able to check the injected `foregroundState` in unit tests.

d4rken (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
chiljamgossow: In general, it is easier to understand code if methods either check conditions or execute a task, but not a combination

ralfgehrer (approved):

PR #1748, opened 2020-11-27
Use key package day/hour to determine data staleness (EXPOSUREAPP-4042)

d4rken:
* Use the actual files in the key pkg store to determine whether our results are outdated, not the last time the key files were fetched.
* If no key files are available yet treat it as "outdated" and abort, instead of throwing an exception.

## Testing
Do some time traveling. If you time travel 48h past the last time for which packages are available, we display the error state card. "Risiko-Ermittlung nicht möglich".


BMItter:
Screenshot_20201129-191928
little issues regarding error state. I'll send you details

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -103,22 +108,26 @@ class RiskLevelTask @Inject constructor(
...
-        val timeSinceLastDiagnosisKeyFetchFromServer =
-            TimeVariables.getTimeSinceLastDiagnosisKeyFetchFromServer()
-                ?: throw RiskLevelCalculationException(
-                    IllegalArgumentException("Time since last exposure calculation is null")
-                )
-        /** we only return outdated risk level if the threshold is reached AND the active tracing time is above the
-        defined threshold because [UNKNOWN_RISK_INITIAL] overrules [UNKNOWN_RISK_OUTDATED_RESULTS] */
-        return (timeSinceLastDiagnosisKeyFetchFromServer.millisecondsToHours() >
-            TimeVariables.getMaxStaleExposureRiskRange() && isActiveTracingTimeAboveThreshold()).also {
+    private suspend fun areKeyPkgsOutDated(nowUTC: Instant): Boolean {
```
chris-cwa, line 111: <3

chris-cwa (approved):

chris-cwa (approved):

BMItr (approved):
already deep checked last friday.
code lgtm.
card issue will probably be solved with #1751 .
As this issue is unrelated a little here,.. approved

PR #1749, opened 2020-11-27
fix typo in strings.xml

JapuDCret:
### Description

When you have a stale test and the App informs you, that you can safely delete it, there's a small typo.

### Steps to reproduce

1. Have the Android CoronaWarnApp
2. Have a test registered
3. Wait for the test to become "stale"
4. The infotext about the "stale" test contains a typo (text instead of test)


harambasicluka:
Hi @JapuDCret,

thank's for your contribution! :) Sadly we can't include your PR... Our source language is german, all other languages are translated from experts and when they are done we receive a `Translation Delivery` via PR. That means we can't change anything in this files as it would be overwritten by the next delivery or we would have to resolve this conflict every time manually. And the changes also need to be reflected in `values-en/strings.xml`.

However I'll forward this to our translators! Thanks for bringing this to our attention.

Cheers,
Luka

PR #1750, opened 2020-11-27
Fix possible notification loop (EXPOSUREAPP-4036)

d4rken:
This fixes the issue that if no new risk level calculation happened, and the last one triggered a notification, killing and restarting the app would also trigger a notification.

To test this you'd have to get to a state where the last risk level calculations were LOW_RISK -> HIGH_RISK, and then just kill the app process and open the app again.

Also: While debugging this I noticed a few cryptic log messages which needed explicit tags.


chris-cwa (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
```diff
@@ -47,6 +48,13 @@ class RiskLevelChangeDetector @Inject constructor(
         val oldResult = changedLevels.first()
         val newResult = changedLevels.last()
 
+        val lastCheckedResult = riskLevelSettings.lastChangeCheckedRiskLevelTimestamp
+        if (lastCheckedResult == newResult.calculatedAt) {
+            Timber.d("We already checked this risk level change, skipping further checks.")
+            return
+        }
+        riskLevelSettings.lastChangeCheckedRiskLevelTimestamp = newResult.calculatedAt
+
```
ritsam, line 57: is there any advantage to create the variable _lastCheckedResult_ ? can we not directly compare _riskLevelSettings.lastChangeCheckedRiskLevelTimestamp_ in the condition?

The comment is more for an information seeking purpose rather that a suggestion.

```suggestion
        if (iskLevelSettings.lastChangeCheckedRiskLevelTimestamp == newResult.calculatedAt) {
            Timber.d("We already checked this risk level change, skipping further checks.")
            return
        }
        riskLevelSettings.lastChangeCheckedRiskLevelTimestamp = newResult.calculatedAt

```

d4rken: Readability, inlining the variable exceeds our maximum line length.

BMItr (approved):
code lgtm, nothing to complain.
passed local testing
image

ralfgehrer (commented):
Code looks fine. Testing now. Should also fix another issue (EXPOSUREAPP-4050)

ralfgehrer (approved):
works like a charm

PR #1754, opened 2020-11-27
Don't bubble up exceptions on day sync caused by no internet (DEV)

d4rken:
If we fail to download the current days, i.e. fail to sync the day index for the specified region, don't bubble up an exception, treat it as unsuccessful, but not as a failed task execution. If there are connection problems we don't want the user to get an error dialog popup everytime.


harambasicluka (commented):
LGTM, is there a way how to test this on device?

chiljamgossow:
> LGTM, is there a way how to test this on device?

I think it's the error dialog you get when going into airplane mode while it tries to sync
device-2020-11-30-160207

chiljamgossow (approved):
Tested on device, cannot trigger the error dialog any more

BMItr (approved):

PR #1763, opened 2020-11-30
Adopt TRL encoding to ENF restrictions (EXPOSUREAPP-4039)

mlenkeit:
keys with report type RECURSIVE are dropped by Android in Exposure
Window mode. The new encoding uses RECUSRIVE only with TRL 1 and 2
which are dropped anyway.

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4039
https://github.com/corona-warn-app/cwa-server/pull/1021


BMItr (approved):
checked code & tests.

tklingbeil (approved):
Checked the calculations. Looks all good to me.

PR #1766, opened 2020-11-30
Add missing commit history to 1.8.x (DEV)

d4rken:
This re-adds the commit history from 1.7.x due to a squashed commit, while it does not fix the the actual history (i.e. `blame`) this resolves the conflicts in case we have to release a hotfix.

Force-push is not feasible as the problematic commit is already part of several PRs.


d4rken:
Pushed via FF

PR #1751, opened 2020-11-27
Refactor riskcard & states; Chapter1 (EXPOSUREAPP-4049, EXPOSUREAPP-3834)

d4rken:
* Removed `RiskLevelConstants` which we not useful anymore and just unnecessary complicated
* Simplified `RiskLevel` to `RiskState` which just has 3 states now `INCREASED_RISK`, `LOW_RISK`, `FAILED`
* Simplified state behavior of the risk cards on the homescreen.
    * Disabled tracing overrides all other cards, as it's a current state we want the user to fix
    * Otherwise if the last calculation failed (for any reason), we show the "tracing failed" card
    * If the last calculation did not fail we display either a high or low risk card

## TODO
* [x] No data yet, fresh install, airplane mode, manually press "restart", card seems to be just empty (light mode)
* [x] No data yet, fresh install, airplane mode, manually press "restart", card seems to be just empty, now enable WIFI, card is empty then green card appears (light mode)
* [x] Tracing disabled, light mode, divider is missing or invisible
* [x] Missing "contacts" when displaying low risk card
* [x] Negative days when time traveling backwards
* [x] Remove "minimum tracing time active" restrictions which should be removed according to EXPOSUREAPP-3051 (moved to extra PR #1762)

## Testing
* Make sure all 4 cards look correct (compare with figma, though it's not 100% current)
   * Red card (high rish)
   * Green card (low risk)
   * Tracing stopped
   * Tracing failed
* Check all designs in dark mode too

PR #1772, opened 2020-12-01
Fix refresh check and improve test menu (EXPOSUREAPP-4049)

d4rken:
Remove "lastTimeDiagnosisKeysFromServerFetch" and replace it with less missleading data.

While it was called "lastTimeDiagnosisKeysFromServerFetch" it was actually "last time we submitted keys for exposure detection". Which was also missleading because it does not directly relate to what the risk card is displaying. It's neither the last download (because it only get's set on submission) and it get's set on start of submission so it likely does not represent the latest risk level until the second risk level calculation has been executed.

* To decide whether to refresh in "onResume", we now use "has there been any submission to the ENF?" (exposure detection `startedAt`).
* To display a timestamp on the risk card, we take the last successful submission (`finishedAt`) to the ENF as the risk card displays the calculation results, which in turn are based on the latest submission. While we could use the last calculated risk level result timestamp, we currently also trigger risk calculations if there are no new submissions to the ENF, which would mean the timestamp is updated even though the result is not based on new data.
    * When we deem it no longer necessary to launch the risk level task in `onResume` we should consider displaying the last risk level result timestamp on the card. This is only viable though when the risk level task only get's launched as reaction to finished exposure detections.

I've also fixed the test fragment button behavior and added descriptions, the "Reset risk level" button surfaced the initial issue because it behaved like a "do a 75% data reset" button. 

@harambasicluka the bug you described [here](https://github.com/corona-warn-app/cwa-app-android/pull/1751#issuecomment-735906177) was likely caused by the incorrectly working test menu button "Reset risklevel", did more than just reset the risk level, it reset `LocalData`. This lead to downloaded diagnosis keys existing such that no new download was deemed necessary. As no new downloaded happened the "lastTimeDiagnosisKeysFromServerFetch" used for detecting new installs was never set again despite being reset by "Reset risk level", this meant that `val wasNotYetFetched = LocalData.lastTimeDiagnosisKeysFromServerFetch() == null` always returned true and let the risk level calculation be launched.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/ExposureDetectionTrackerExtensions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/ExposureDetectionTrackerExtensions.kt
@@ -5,10 +5,10 @@ import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.map
 
 suspend fun ExposureDetectionTracker.lastSubmission(
-    onlySuccessful: Boolean = true
+    onlyFinished: Boolean = true
 ): TrackedExposureDetection? = calculations
     .first().values
-    .filter { it.isSuccessful || !onlySuccessful }
+    .filter { it.isSuccessful || !onlyFinished }
     .maxByOrNull { it.startedAt }
 
 fun ExposureDetectionTracker.latestSubmission(
@@ -20,4 +20,3 @@ fun ExposureDetectionTracker.latestSubmission(
     .map { detections ->
         detections.maxByOrNull { it.startedAt }
     }
-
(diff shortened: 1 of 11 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/detectiontracker/ExposureDetectionTrackerExtensions.kt
```diff
@@ -0,0 +1,22 @@
+package de.rki.coronawarnapp.nearby.modules.detectiontracker
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+
+suspend fun ExposureDetectionTracker.lastSubmission(
```
ralfgehrer, line 7: I would not mind moving this into the same file as the interface itself. Was shortly confused about the location of `lastSubmission` when reviewing this PR

ralfgehrer (approved):

harambasicluka (approved):

PR #1775, opened 2020-12-01
No background fetch of test result after removing first test (EXPOSUREAPP-3816)

kolyaopahle:
### Description
This PR sets isTestResultNotificationSent to false on test removal allowing further added tests to also send notifications for results.

### Steps to reproduce
1. Pair a test
2. Get a negative result notification
3. Remove the test
4. Pair a new test
5. No result notification will arrive


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -126,6 +126,7 @@ class HomeFragmentViewModel @AssistedInject constructor(
         SubmissionRepository.deleteRegistrationToken()
```
kolyaopahle, line 126: Maybe we should move all this logic to the `SubmissionRepository` to deduplicate the deletion logic

d4rken: Sounds reasonable, could we then also remove the static access and add a test for it?

kolyaopahle: done, still some kind of static access to LocalData, will try to move to SubmissionSettings in an additional pr 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -474,38 +474,6 @@ object LocalData {
 
     fun submissionWasSuccessful(): Boolean = numberOfSuccessfulSubmissions() >= 1
 
-    fun testGUID(): String? = getSharedPreferenceInstance().getString(
```
chiljamgossow, line 477: Are these not needed anymore (testGUID, authCode, teletan)? I do not see an replacement

chiljamgossow (commented):
Looks much cleaner now

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
d4rken: Seems they were previously already unused. `testGUID` was still being called but only with `testGUID(null)`.

Screenshot from 2020-12-07 19-37-07

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -474,38 +474,6 @@ object LocalData {
...
-        CoronaWarnApplication.getAppContext()
-            .getString(R.string.preference_test_guid),
-        null
-    )
-
-    fun testGUID(value: String?) {
-        getSharedPreferenceInstance().edit(true) {
-            putString(
-                CoronaWarnApplication.getAppContext()
-                    .getString(R.string.preference_test_guid),
```
d4rken, line 487: If they are unused, can we remove them from the `values-de/strings.xml` such that are removed from all files on the next translation update?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
```diff
@@ -524,17 +492,6 @@ object LocalData {
         )
     }
 
-    fun teletan(value: String?) = getSharedPreferenceInstance().edit(true) {
```
d4rken, line 527: Can you explain all of these can be removed now? Was it just dead code?

d4rken (commented):
Love the clean up :heart:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
kolyaopahle: yes, these have been removed by a prior refactoring and were only used by the tests and resetting them to null, i stumbled upon this while i was fixing unit tests and decided to remove them

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
kolyaopahle: sure, will do :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/LocalData.kt
kolyaopahle: teletan and testGUID were both used when test scanning / tan entering and test registration were different steps and ui screens, since they have been merged into their respective single screen we do not need to persist this data anymore but we still did for quite a while without ever using it

d4rken (approved):

chiljamgossow (approved):

BMItr (approved):

PR #1776, opened 2020-12-01
Don't show last risk contact on "tracing disabled" card (DEV)

d4rken:
Mock ups show that we shouldn't show the "last contact" on the "tracing disabled" card.

## Test
* Generate green card
* Disable tracing


harambasicluka (approved):

ralfgehrer (approved):
Tested on Pixel (Android 10)

PR #1787, opened 2020-12-02
Adapted singular string encounters for high and low risk (EXPOSUREAPP-4093)

harambasicluka:
Get a high and low risk with 1 day since last encounters.

image

image


ralfgehrer (approved):

janetback (approved):
UA reviewed

PR #1788, opened 2020-12-02
Fixed: tasks are run (EXPOSUREAPP-3868, EXPOSUREAPP-3988)

chris-cwa:
- take a device with red card (high risk) on 1.7.x
- install 1.8
- red card is still there

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
@@ -65,6 +65,9 @@ open class QueueingTask @Inject constructor() : Task<DefaultProgress, QueueingTa
 
         override val collisionBehavior: TaskFactory.Config.CollisionBehavior =
             TaskFactory.Config.CollisionBehavior.ENQUEUE
+
+        override val preconditions: List<suspend () -> Boolean>
+            get() = listOf { preconditionIsMet }
     }
 
     class Factory @Inject constructor(
@@ -73,5 +76,9 @@ open class QueueingTask @Inject constructor() : Task<DefaultProgress, QueueingTa
 
         override suspend fun createConfig(): TaskFactory.Config = Config()
         override val taskProvider: () -> Task<DefaultProgress, Result> = { taskByDagger.get() }
+    }
+
+    companion object {
+        var preconditionIsMet = true
     }
 }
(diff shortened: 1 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
```diff
@@ -74,4 +77,11 @@ open class QueueingTask @Inject constructor() : Task<DefaultProgress, QueueingTa
         override suspend fun createConfig(): TaskFactory.Config = Config()
         override val taskProvider: () -> Task<DefaultProgress, Result> = { taskByDagger.get() }
     }
+
+    companion object {
+        suspend fun isPreconditionMet(): Boolean{
+            return preconditionIsMet
+        }
+        var preconditionIsMet = true
```
ralfgehrer, line 85: I don't quite follow. Can you please elaborate on why this is needed?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
```diff
@@ -74,4 +77,11 @@ open class QueueingTask @Inject constructor() : Task<DefaultProgress, QueueingTa
         override suspend fun createConfig(): TaskFactory.Config = Config()
         override val taskProvider: () -> Task<DefaultProgress, Result> = { taskByDagger.get() }
     }
+
+    companion object {
+        suspend fun isPreconditionMet(): Boolean{
```
ralfgehrer, line 82: This is not a suspending function

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
ralfgehrer, line 85: Moreover, this is only a `var` instead of a `val` becaus of the Test afaik? How about annotating it accordingly.

d4rken (approved):

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/example/QueueingTask.kt
chris-cwa: Queuing task is used for test only (should be move to test directory soon). To test whether task controller skips tasks which preconditions are not yet met, we use this flag to manually indicate in test whether preconditions for this task shall be considered fulfilled

PR #1794, opened 2020-12-02
Submission upload (EXPOSUREAPP-3748)

d4rken:


ralfgehrer:
Android vs iOS Difference 
Screenshot 2020-12-03 at 18 25 11

harambasicluka:
image
start submission flow -> enter valid positive tan -> click on x -> see popup -> click on warn

get redirected to this screen:
image

I would expect that the popup is closed and stays on the same screen, that's at least the behaviour in other parts.

ralfgehrer:
Flags are off - should be within the border 

Screenshot 2020-12-03 at 18 35 23

ritsam:
> Flags are off - should be within the border
> 
> Screenshot 2020-12-03 at 18 35 23

Hi Ralf,

What I understood by borders you mean within those two lines right? IF that's the case, the designs had a mixed UI representation, perhaps more clarification from the design side is needed.

harambasicluka:
The flow isn't correct.

image

Positive test result on home > warn others > give ENF consent > go ahead with symptoms

Expected: go directly to the calendar view symptom screen

image

Actual: 

image

ralfgehrer:
I have scanned a positive QR Code with given consent, went into the symptoms screen aborted, the keys were submitted and the thank you screen was shown. So far so good.
**BUT, now I can navigate back (via androids back button/gesture) and end up on the symptom screen again.**
Furthermore, aborting the symptoms flow results in a 400 error as we try to upload the keys again.

harambasicluka:
> The flow isn't correct.
> 
> image
> 
> Positive test result on home > warn others > give ENF consent > go ahead with symptoms
> 
> Expected: go directly to the calendar view symptom screen
> 
> image
> 
> Actual:
> 
> image

isn't bug, the implementation is correct

ralfgehrer:
"Andere Warnen" screen: flags are missing entirely while the flags are present on the "Ihr Einverständnis" screen (hence, not a config issue)

Screenshot 2020-12-03 at 19 10 27

ralfgehrer:
I have scanned a positive QR Code with consent aborted, refused to submit keys, ended up on the home screen.
Clicked "Andere Warnen" on the home screen, consented, shared, ended on the thank you screen (fragment), submitted keys without symptoms.

BUT, now I can navigate back (via androids back button/gesture) and end up on the thank you screen again.

ralfgehrer (approved):

Oliver-Zimmerman (approved):

PR #1797, opened 2020-12-03
Submission - Fix to include headline on TestResultSection 

Oliver-Zimmerman:
### Description

The heading of the test result section card was missing, this is now being set.

### Steps to reproduce
Get a test result and open screen, the heading should now be included.


AlexanderAlferov (dismissed):

janetback (dismissed):
Hi, both of these changes look to me like an old version of hte scren texts is being used.
Zufallscode is definitely an obsolete term. For the first change I will need to see if I can find some documentation  on when this was changed.
Could you please also review from your side. Thanks.

AlexanderAlferov (approved):

chris-cwa (approved):

PR #1798, opened 2020-12-03
Update legal_strings.xml (EXPOSUREAPP-3995)

janetback:
### Description
Added English version of legal texts


chiljamgossow (dismissed):

Corona-Warn-App/src/main/res/values/legal_strings.xml
```diff
@@ -14,23 +14,23 @@
...
     <!-- XHED: Page subheadline for consent sub section your consent  -->
-    <string name="submission_consent_your_consent_subsection_headline" translatable="false"></string>
```
chiljamgossow, line 17: We nned to keep the translatable= false here as we do not provide all languages

chiljamgossow (commented):
We need to keep the translatable= false here as we do not provide all languages

ralfgehrer:
> We need to keep the translatable= false here as we do not provide all languages

Fixed it.

ralfgehrer (approved):

harambasicluka (approved):

PR #1806, opened 2020-12-04
Submission - Negative Test Result text changes (EXPOSUREAPP-4131)

Oliver-Zimmerman:
### Description
Negative test result text changes to more closely align with Figma design. 

### Steps to reproduce
Receive a negative test result and admire text changes - they should be exactly like the Figma designs now.


BMItr (dismissed):

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -876,7 +876,7 @@
     <!-- XBUT: test result pending : remove the test button -->
     <string name="submission_test_result_pending_remove_test_button">"Test entfernen"</string>
     <!-- XHED: Page headline for negative test result next steps  -->
-    <string name="submission_test_result_negative_steps_negative_heading">"Ihr Testergebnis"</string>
+    <string name="submission_test_result_negative_steps_negative_heading">"Befund Negativ"</string>
```
SabineLoss, line 879:
```suggestion
    <string name="submission_test_result_negative_steps_negative_heading">"Befund negativ"</string>
```
I can't see a reason for a capital "n" in this case, but I can't double check in the mockups. @janetback , can you please double check?

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1216,6 +1216,7 @@
         <item>"Sie haben trotzdem gesundheitliche Beschwerden? Wenn Sie sich sehr krank fühlen und/oder Ihre Beschwerden sich verschlechtert haben, wenden Sie sich bitte an Ihre Hausarztpraxis."</item>
         <item>"Bleiben Sie bis zu Ihrer Genesung trotzdem zu Hause. Falls Sie sich durch eine andere Infektion geschwächt mit dem Coronavirus (SARS-CoV-2) infizieren, kann dies zu schwereren Krankheitsverläufen führen."</item>
         <item>"Gehen Sie nicht krank zur Arbeit, um andere Personen nicht zu gefährden. Falls sich Ihre Symptome verschlechtern, kann die Notwendigkeit eines weiteren SARS-CoV-2-Tests bestehen."</item>
+        <item>"Falls sich Ihre Symptome verschlechtern, kann die Notwendigkeit eines weiteres SARS-CoV-2-Tests bestehen."</item>
```
SabineLoss, line 1219: As this is now a separate item, I assume that the sentence "Falls sich Ihre Symptome verschlechtern ..." needs to be removed from the previous item in the list.

Oliver-Zimmerman: You are correct. I will make this adjustment now.

SabineLoss (approved):
Thanks! :-) UA approved

BMItr (approved):

PR #1808, opened 2020-12-04
Fixed: fields without values are visible (EXPOSUREAPP-3988)

chris-cwa:
Directly after migration to 1.8 (enf v2) diagnosis keys must be provided.

When updating to 1.8.x after 1.7.x there is no data initially available that the `RiskLevelTask` can use, executing it early would lead to a previous increased-risk from V1 (ExposureSummary) being low-risk in V2 (ExposureWindows).

We solve this by detecting the version upgrade and skipping the otherwise evaluated abort conditions for the DiagnosisKeysDownloadTask.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -72,21 +72,24 @@ class DownloadDiagnosisKeysTask @Inject constructor(
...
             val now = timeStamper.nowUTC
 
             if (exposureConfig.maxExposureDetectionsPerUTCDay == 0) {
                 Timber.tag(TAG).w("Exposure detections are disabled! maxExposureDetectionsPerUTCDay=0")
                 return object : Task.Result {}
             }
 
-            if (wasLastDetectionPerformedRecently(now, exposureConfig, trackedExposureDetections)) {
+            val trackedExposureDetections = enfClient.latestTrackedExposureDetection().first()
+            val isUpdateToEnfV2 = settings.isUpdateToEnfV2
```
ralfgehrer, line 83: Minor readability improvement:
```
if (!settings.isFirstUpdateToEnfV2()) {
  if (wasLastDetectionPerformedRecently(now, exposureConfig, trackedExposureDetections)) {
   ....
  } 
  if (hasRecentDetectionAndNoNewFiles(now, keySyncResult, trackedExposureDetections)) {
    ....
  }
} else {
Timber.tag(TAG).d("is first update ToEnfV2")
}

```

ralfgehrer (approved):

d4rken (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
d4rken: Will do for 1.9.x

janetback (commented):
UA reviewed

PR #1813, opened 2020-12-04
Fix PopupTo Backstack issue (EXPOSUREAPP-4123)

AndroidMedaGalaxy:
### Description
Back navigation popupstack fixed

#### Testing with release 1.9.x

Positive QR Code scanned >  consent aborted by refusing to submit keys > ending up on the home screen. >
Click "Andere Warnen" on the home screen > consented > shared keys > Submit keys without symptoms > come back on the screen by selecting yes in the popup  > The app should close on Android's back key press. but it comes back to the thank you screen again.

This behaviour is the same for all the screens in Submission No consent Keys flow ( only till thank you screen)

#### Testing with this PR
Perform all the actions as above but when clicked back button on home screen, the App should close now instead of going back to the previous fragment.

### Changes
nav_graph has been updated where  _app:popUpTo="@id/mainFragment"_ is replaced with  _app:popUpTo="@id/nav_graph"_

#### Reason for this change
Somehow there could be multiple instances of mainfragment in the backstack. we're clearing the backstack completely to make a fresh start each time we navigate to mainfragment


chiljamgossow (dismissed):

d4rken (commented):
Did not work for all screens.

* Scan positive QR code
* Give consent
* Click "Finish" without symptom input
* We end up on the home screen but on "back button" we still end up on the test result.

Same when ending the submission from `SubmissionSymptomIntroductionFragment`.

I've added the proposed solution with `app:popUpTo="@id/nav_graph"` to all navgraph actions leading from submission to the home fragment and it works for me now.

Please retest/review this @ritsam @chiljamgossow

chiljamgossow:
I cleaned up the graph. The app should only close after the mainFragment is shown. Many actions popUpTo mainFragment with popUpInclusive = true, so the mainFragment gets removed from the stack and any following popUpTo mainFragment has no effect.

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -271,16 +271,9 @@
         android:name="de.rki.coronawarnapp.ui.submission.tan.SubmissionTanFragment"
         android:label="fragment_submission_tan"
         tools:layout="@layout/fragment_submission_tan">
-        <action
-            android:id="@+id/action_submissionTanFragment_to_submissionDispatcherFragment"
```
chiljamgossow, line 275: This was unused

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -271,16 +271,9 @@
...
-            app:destination="@id/submissionDispatcherFragment"
-            app:popUpTo="@id/submissionDispatcherFragment"
-            app:popUpToInclusive="true" />
         <action
             android:id="@+id/action_submissionTanFragment_to_submissionResultFragment"
-            app:destination="@id/submissionResultFragment"
-            app:popUpTo="@id/submissionResultFragment"
```
chiljamgossow, line 282: This has no effect, so I removed it

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -321,23 +314,23 @@
         android:label="SubmissionQRCodeScanFragment">
         <action
             android:id="@+id/action_submissionQRCodeScanFragment_to_submissionDispatcherFragment"
-            app:destination="@id/submissionDispatcherFragment"
```
chiljamgossow, line 324: To ga back, one needs to pop the current fragment

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -321,23 +314,23 @@
...
-            app:popUpTo="@id/submissionDispatcherFragment"
+            app:popUpTo="@id/submissionQRCodeScanFragment"
             app:popUpToInclusive="true" />
         <action
             android:id="@+id/action_submissionQRCodeScanFragment_to_submissionResultFragment"
             app:destination="@id/submissionResultFragment"
-            app:popUpTo="@id/submissionResultFragment">
```
chiljamgossow, line 330: Back navigation should not be allowed here, so all fragments need to be popped except mainFragment

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -321,23 +314,23 @@
...
                 android:name="skipInitialTestResultRefresh"
                 android:defaultValue="true"
                 app:argType="boolean" />
         </action>
-        <action
+         <action
             android:id="@+id/action_submissionQRCodeScanFragment_to_submissionTestResultAvailableFragment"
             app:destination="@id/submissionTestResultAvailableFragment"
             app:popUpTo="@id/mainFragment"
-            app:popUpToInclusive="true" />
```
chiljamgossow, line 340: We need to leave the mainFragment on the stack, so going back brings us there before closing the app

chiljamgossow (approved):

Oliver-Zimmerman (approved):

ralfgehrer (approved):

PR #1818, opened 2020-12-07
After long update no exposure, additonal logs & potential fix (EXPSOUREAPP-4120)

BMItr:
#During analysis of 4120 & 4087 I spotted two points which should be adjusted 

This PR addresses (DRAFT):
- Exception while using the work:work-runtime-ktx:2.5.0-beta01 during runtime -> Downgrade to latest stable (2.4.0), which solves the Exception
- Added additonal log to DefaultDiagnosisKeyProvider for deeper analysis
- Avoid multiple downloadtasks by using startedAt as fallback when finishedAt is null to avoid new download in TracingRepository


Corona-Warn-App/build.gradle
```diff
@@ -318,7 +318,7 @@ dependencies {
     implementation 'androidx.navigation:navigation-ui-ktx:2.2.2'
     implementation 'androidx.legacy:legacy-support-v4:1.0.0'
     implementation 'androidx.preference:preference:1.1.1'
-    implementation 'androidx.work:work-runtime-ktx:2.5.0-beta01'
+    implementation 'androidx.work:work-runtime-ktx:2.4.0'
```
d4rken, line 321: Are both issues related or why are both changes in the same PR?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -153,11 +153,10 @@ class TracingRepository @Inject constructor(
         val taskLastFinishedAt = try {
             taskController.tasks.first()
                 .filter { it.taskState.type == DownloadDiagnosisKeysTask::class }
-                .mapNotNull { it.taskState.finishedAt }
-                .sortedDescending()
-                .first()
+                .mapNotNull { if (it.taskState.finishedAt != null) it.taskState.finishedAt else it.taskState.startedAt }
```
d4rken, line 156: Maybe `it.taskState.finishedAt ?: it.taskState.startedAt`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -153,11 +153,10 @@ class TracingRepository @Inject constructor(
...
+                .maxOrNull()!!
```
d4rken, line 157: Doesn't this now throw a `KotlinNullpointerException` which will no longer be caught by `catch(e: NoSuchElementException)`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
BMItter: good point. adjusted.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
BMItter: adjusted to catch Nullpointer

Corona-Warn-App/build.gradle
BMItter: tickets were declared to related inside jira.
I can separte them if needed.

ralfgehrer (approved):

d4rken (approved):

PR #1821, opened 2020-12-07
Fix tracing switch in settings (DEV)

chiljamgossow:
Fixes issues with turning tracing on and off

1. error case
a. Deny tracing and background processing in onboarding
b. Try to turn tracing on in the settings screen

2. error case
a. Go to tracing settings and turn tracing on by pressing on the switch (not the row) and deny the popup for consent
b. state is inconsistent (switch is on, but tracing is actually off)


d4rken (approved):
Can't reproduce the bugs anymore :+1: 

The whole `SettingsTracingFragment` is refactoring worthy, but that's for another PR :grin:

ralfgehrer (approved):

PR #1825, opened 2020-12-07
New Submission Flow: new test result notification (EXPOSUREAPP-4134)

AlexanderAlferov:
New notification for submission flow.

Notification creation kept the same. Main difference is in intent that leads to different screens depending on test result.
For quick intent test new button was introduced in test menu (Test for API).

For better testability NotificationHelper was refactored from object to injectable class with further refactoring of some classes and tests 🤖 

Screenshot 2020-12-07 at 18 28 35


chiljamgossow:
How can we test this?

AndroidMedaGalaxy (dismissed):
Note: Checked with the Test Menu.
1. Tested with positive Result : Deep linking works and the Submission Flow starts when clicked on the notification
2. Tested with Negative REsult : Deep linking works here again and Result Page is shown after clicking on the notification. of course, the result visible in the screen is negative as it should.

Long story short, All looks good to me.  (. _ .)/ \\(.__.)

Thanks

chiljamgossow:
When I tested with the test menu, screen were incomplete. I do not know whether it is fine in a more realistic scenario.

ralfgehrer (dismissed):
Code looks good.

AndroidMedaGalaxy (approved):
Tested

Scenario 1: Positive test  with cancelled submission flow after QR scan 
    - > Positive notification fired from Test API - > Clicked on notification - >  Submission flow start screen  "Your test is available" was opened

Scenario2 : Negative  test  
    - > Negative notification fired from Test API - > Clicked on notification - >  Test Result screen was opened

Tried interchanging the Positive and Negative notifications buttons in Test API regardless of the test result


All works good. ヽ(ヅ)ノ

ralfgehrer (approved):

PR #1840, opened 2020-12-08
Fix issues with TEK submission flow (EXPOSUREAPP-4189, EXPOSUREAPP-4215)

d4rken:
This addresses a problematic TEK submission flow that allowed to you to reach TEK submission screens without passing a TEK retrieval screen, i.e. you were never asked to grant access to your TEK history before performing the submission.

The core issue here was the previously used test result fragment. Before the new submission flow it was used for both positive, negative and pending tests. Then it was just used for pending and negative test results, the positive was split into it's own. If the test was pending, the "legacy" test result fragment did not navigate correctly to the screen that you need to pass before viewing your positive test result, instead it directly forwarded you to the symptoms.

This PR splits the legacy fragment again such that we now have one fragment per state (positive, negative, pending, invalid). The `SubmissionTestResultPendingFragment` will automatically forward the user to the correct screen.

## TODO
- [x]  Add checks for `isTracingEnabled` before attempting to get the TEK history
- [ ] Coordinate with @Oliver-Zimmerman regarding the home screen card for "test result available"
- [x] Clean up test result pending layout

## Testing
* Use the CLI to scan to get a pending test result that later turns into either a positive or a negative result and check that the submission flow behaves correctly.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/tekhistory/DefaultTEKHistoryProvider.kt
```diff
@@ -18,25 +18,25 @@ class DefaultTEKHistoryProvider @Inject constructor(
...
-            getTEKHistory()
-            true
-        } catch (e: ApiException) {
-            if (e.statusCode == ExposureNotificationStatusCodes.RESOLUTION_REQUIRED) {
-                false
-            } else {
-                throw e
-            }
+    override suspend fun isTEKAccessPermissionGranted(): Boolean = try {
+        getTEKHistory()
```
mlenkeit, line 22: Doesn't this prompt for the native popup where the user has to grant access?!

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -272,8 +289,8 @@
         android:label="fragment_submission_tan"
         tools:layout="@layout/fragment_submission_tan">
         <action
-            android:id="@+id/action_submissionTanFragment_to_submissionResultFragment"
-            app:destination="@id/submissionResultFragment">
+            android:id="@+id/action_submissionTanFragment_to_submissionTestResultPendingFragment"
+            app:destination="@id/submissionTestResultPendingFragment">
```
chiljamgossow, line 293: Here we should pop the stack to prevent back navigation

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/invalid/SubmissionTestResultInvalidFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    @Inject lateinit var viewModelFactory: CWAViewModelFactoryProvider.Factory
+    private val vm: SubmissionTestResultInvalidViewModel by cwaViewModels { viewModelFactory }
+
+    private val binding: FragmentSubmissionTestResultInvalidBinding by viewBindingLazy()
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val backCallback: OnBackPressedCallback = object : OnBackPressedCallback(true) {
```
chiljamgossow, line 30: We do not need to override the back navigation behavior unless we need to perform a specific task, like showing a dialog. The back stack should be cleared in the action leading to this fragment, eg. popping up to main fragment, so pressing back naturally lands on main fragment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/negative/SubmissionTestResultNegativeFragment.kt
```diff
@@ -0,0 +1,67 @@
...
+
+    @Inject lateinit var viewModelFactory: CWAViewModelFactoryProvider.Factory
+    private val vm: SubmissionTestResultNegativeViewModel by cwaViewModels { viewModelFactory }
+
+    private val binding: FragmentSubmissionTestResultNegativeBinding by viewBindingLazy()
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val backCallback: OnBackPressedCallback = object : OnBackPressedCallback(true) {
```
chiljamgossow, line 30: same here, not needed if the action clears the stack

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/negative/SubmissionTestResultNegativeViewModel.kt
```diff
@@ -0,0 +1,62 @@
...
+    }.asLiveData(context = dispatcherProvider.Default)
+
+    fun onBackPressed() {
+        routeToScreen.postValue(
+            SubmissionTestResultNegativeFragmentDirections
+                .actionSubmissionTestResultNegativeFragmentToHomeFragment()
+        )
+    }
+
+    fun deregisterTestFromDevice() {
```
chiljamgossow, line 40: I saw a method for this in the repo

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -0,0 +1,153 @@
...
+                submissionTestResultSection.setTestResultSection(result.deviceUiState, result.testResultReceivedDate)
+
+                submissionTestResultSpinner.setInvisible(hasResult)
+
+                submissionTestResultContent.setInvisible(!hasResult)
+                buttonContainer.setInvisible(!hasResult)
+            }
+        }
+
+        val backCallback: OnBackPressedCallback = object : OnBackPressedCallback(true) {
```
chiljamgossow, line 61: again, not necessary

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/tekhistory/DefaultTEKHistoryProvider.kt
d4rken: Correct, it was a leftover, my initial API design added it, then due to the concerns you mentioned it was never used. Removed it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/negative/SubmissionTestResultNegativeViewModel.kt
d4rken: Was merged in a different PR, I've changed the code to use that new method. :+1:

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
d4rken:
```xml
        <action
            android:id="@+id/action_submissionTanFragment_to_submissionTestResultPendingFragment"
            app:destination="@id/submissionTestResultPendingFragment"
            app:popUpTo="@id/mainFragment"
            app:popUpToInclusive="false">
```
Right?

chiljamgossow: Yes

chris-cwa (approved):
code looks good. tested it! thank you!

chiljamgossow (approved):
Back navigation looks fine

ralfgehrer (approved):

PR #1847, opened 2020-12-09
Strings for Contact Diary card on the homescreen (EXPOSUREAPP-4153)

SamuraiKek:
New strings for the Contact Diary card on the home screen.


ralfgehrer (approved):

janetback (approved):
UA reviewed

PR #1849, opened 2020-12-09
Fix countries missing in onboarding after in-app reset (EXPOSUREAPP-4203)

d4rken:
Remove the extra flow that would cache the countries. The app config itself is cached already.
When accessing the country list they will be mapped from the latest available app config.

### Testing
* Check the country list in on-boarding and interop settings
* Reset the app via in-app data reset
* Check the country list again


ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/interoperability/InteroperabilityConfigurationFragmentViewModel.kt
```diff
@@ -1,28 +1,33 @@
...
 import de.rki.coronawarnapp.storage.interoperability.InteroperabilityRepository
+import de.rki.coronawarnapp.util.coroutine.DispatcherProvider
 import de.rki.coronawarnapp.util.ui.SingleLiveEvent
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
 import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
 
 class InteroperabilityConfigurationFragmentViewModel @AssistedInject constructor(
-    private val interoperabilityRepository: InteroperabilityRepository
-) : CWAViewModel() {
+    private val interopRepo: InteroperabilityRepository,
```
SamuraiKek, line 12: Why did you rename this val? I thought we weren't using shortened words.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/warnothers/SubmissionResultPositiveOtherWarningViewModel.kt
```diff
@@ -47,14 +47,14 @@ class SubmissionResultPositiveOtherWarningViewModel @AssistedInject constructor(
             }
         }
 
-    val uiState = combineTransform(
+    val uiState = combine(
         currentSubmission,
-        interoperabilityRepository.countryListFlow
+        interoperabilityRepository.countryList
```
SamuraiKek, line 52: I see you've kept its name here. In all other places it's changed.

chiljamgossow (dismissed):
I tested the reset, and there were flags shown in the onboarding.

SamuraiKek:
I've observed a strange behaviour while testing:

My app has been installed for about 2 weeks on my Huawei device on which I've tested and the first time I've tried to reset the app, the country list still didn't show up in the onboarding. 
I went to interop country list and it showed up in there. After that I reset the app again and nothing happened for a good 5 seconds until I was brought to the onboarding screen. 
The onboarding screen still didn't have the country list and after completing the onboarding I went to the interop  settings where the list showed up.

After this I deleted my app data and tried it again.
This time country list also showed up in onboarding but as well as the interop settings. But the behaviour of the app hanging after resetting the app a second time still happens.

Not sure what to make of this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/interoperability/InteroperabilityConfigurationFragmentViewModel.kt
d4rken: I didn't like it spilling over into a new line. I've changed it back.

d4rken:
> After this I deleted my app data and tried it again.
This time country list also showed up in onboarding as well as the interop settings. But the behaviour of the app hanging after resetting the app a second time still happens.

I'm looking into it, thanks.

d4rken:
I think what @SamuraiKek is experiencing is related to a blocking call on the mainthread for which I will open an extra PR.

PR #1852, opened 2020-12-09
Implement Card on Homescreen "Ihr Testergebnis liegt vor" (EXPOSUREAPP-4212)

Oliver-Zimmerman:
### Description

Pull request to add the missing home code from new submission flow. 

The new card explains that the user's test result is ready, rather than displaying the actual result. This should only be shown if the user has not yet seen their result. After the user has seen their result, the old test result content cards are shown. 

The only exception to this is in the case of a negative test result, in which case the old test result content card should be down regardless of whether or not the user has viewed their test result yet or not. 

The risk card above should now also be visible, regardless of the result, if this card is shown (originally hidden if positive)

### Steps to reproduce
1. Scan a QR code
2. If positive, do not complete flow (don't view result) and go back to the main screen. Click the x before providing consent and going to result. The new card should be shown under risk card.
3. Once you view the result, the normal content card should be shown. 
4. If a negative QR code is scanned, the usual content card should be shown. 

Screenshot_20201209-134919_CWA TEST


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionSettings.kt
```diff
@@ -39,6 +39,11 @@ class SubmissionSettings @Inject constructor(
         defaultValue = false
     )
 
+    val hasViewedTestResult = prefs.createFlowPreference(
```
chiljamgossow, line 42: This needs to be reset when tests are removed or the app is reset via settings. I did not see that in this PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/SubmissionRepository.kt
```diff
@@ -57,6 +57,7 @@ class SubmissionRepository @Inject constructor(
 
     // to be used by new submission flow screens
     val hasGivenConsentToSubmission = submissionSettings.hasGivenConsent.flow
+    val hasViewedTestResult = submissionSettings.hasViewedTestResult.value
```
chiljamgossow, line 60: Why is that not a flow?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionSettings.kt
Oliver-Zimmerman: This happens in SubmissionSettings with the clear() function

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/SubmissionRepository.kt
Oliver-Zimmerman: I originally had it as flow, and read it in SubmissionCardState as: 
    private val testViewed = submissionRepository.hasViewedTestResult.asLiveData()

But the value kept returning null. Perhaps implementation was incorrect? I can have a look again.

chiljamgossow: LiveData needs an observer to emit data. We should expose the flow in the repository, otherwise it is not observable.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/SubmissionCardState.kt
```diff
@@ -17,16 +18,33 @@ import de.rki.coronawarnapp.util.NetworkRequestWrapper.Companion.withSuccess
...
+    val isDeviceRegistered: Boolean,
+    val submissionRepository: SubmissionRepository
 ) {
 
-    fun isRiskCardVisible(): Boolean =
-        deviceUiState.withSuccess(true) {
-            when (it) {
-                PAIRED_POSITIVE, PAIRED_POSITIVE_TELETAN, SUBMITTED_FINAL -> false
-                else -> true
+    private val testViewed = submissionRepository.hasViewedTestResult
```
chiljamgossow, line 25: You need an observer here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/SubmissionCardState.kt
```diff
@@ -17,16 +18,33 @@ import de.rki.coronawarnapp.util.NetworkRequestWrapper.Companion.withSuccess
...
+
+    fun isRiskCardVisible(): Boolean {
```
chiljamgossow, line 27: We could expose LiveData here to make the view reactive. This would work well with the flow from the repository.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionSettings.kt
chiljamgossow: Ok, that's true

chiljamgossow: What is missing is resetting is when the test is removed. That does not call clear() as other settings should remain.

Oliver-Zimmerman: Good point. I will implement this.

chiljamgossow: Yes,  I think removing a test should not call clear() but have a more explicit behavior

chiljamgossow (changes requested):
I tested as described, but saw the positive test result right away on the home screen without viewing it in the flow

Oliver-Zimmerman:
@chiljamgossow  As discussed on a call, this doesn't happen normally. Only when updating with a pre-existing test which is why I missed it. I will look into this now.

ralfgehrer (commented):
Tested all scenarios, works as expected.

ralfgehrer (dismissed):

ralfgehrer (dismissed):
lgtm

d4rken (dismissed):

chiljamgossow (dismissed):

d4rken (dismissed):

ralfgehrer (approved):

d4rken (approved):

PR #1854, opened 2020-12-09
Fix long loading times for legal texts

fynngodau:
For myself and others I've asked, long texts like the privacy policy and the licenses take long to load for no good reason. The application "hangs" for a few seconds because some action (namely, rendering the layout – the `Html.fromHtml` call itself is a lot quicker) is performed on the main thread.

I used the profiler to determine that most of the time gets wasted for measuring layouts, which is an action that the `ConstraintsLayout` performs. I don't see a reason to use `ConstraintsLayout` in this location, as it is a quite simple layout which is able to render a lot faster when using `LinearLayout`.

Therefore, I propose to use `LinearLayout` in this location. It should also improve the loading times for other screens, but it is most noticeable with the long legal texts.

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/fragment_information_privacy.xml
+++ b/Corona-Warn-App/src/main/res/layout/fragment_information_privacy.xml
@@ -10,62 +10,39 @@
 
     </data>
 
-    <androidx.constraintlayout.widget.ConstraintLayout
+    <LinearLayout
         android:id="@+id/information_privacy_container"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
+        android:orientation="vertical"
         android:focusable="true"
         android:contentDescription="@string/information_privacy_title">
 
         <include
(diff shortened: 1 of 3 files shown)


fynngodau:
I noticed there was further potential for improvememt when additionally modifying `fragment_information_privacy` and `fragment_information_technical` and therefore added to my PR. That reduces the number of nested ("stacked") `ConstraintLayout`s from originally 4 to now 0 on these two screen, which is beneficial for performance while the appearance remains unchanged (unless I missed something).

d4rken:
Unfortunately time did not allow to include this in the `1.9.x`.
Your profiling is on point though and the changes improve layouting performance by a lot, on first glance I'd have bet that the html parsing would be the bottle neck, but it's not :+1: 

I'm closing this in favor of #1882 where I cherry picked your commits and went through all other fragments too. Would love your review on that PR @fynngodau .

PR #1856, opened 2020-12-10
Allow back navigation on symptom intro and calendar(EXPOSUREAPP-4201)

chiljamgossow:
Removes abort dialog on back press to allow back navigation
Replaces app bar icon with X to indicate abort, while keeping the dialog here


ralfgehrer:
Screenshot_20201210-112921
All works except that I would have expected that pop-up - when going back by android's back button - on the first screen of the symptom flow instead of on top of the test result. Is this intended?

chiljamgossow:
The idea is to allow the user to go back as long as the flow allows and it does not cause an issue. So going back from the intro is fine, while going back from test result is not.

ralfgehrer (approved):

d4rken (approved):
:+1:

PR #1857, opened 2020-12-10
Move update check off mainthread (EXPOSUREAPP-4203)

d4rken:
Follow up @SamuraiKek  reporting that the app hangs #1849, not 100% that this is the issue he reported, but this is a definitely an issue in itself.

We shouldn't access the update checker on the main thread as it will call for the app config which may be blocking.


ralfgehrer (dismissed):
Works well on Pixel (Android 10).
Also thanks for introducing a VM for the Activitiy.

ralfgehrer (approved):

AndroidMedaGalaxy (approved):
Tested on S9+ Android 10
All looked good.  ┌(ㆆ㉨ㆆ)ʃ

PR #1858, opened 2020-12-10
Fixed legacy config (EXPOSUREAPP-4195)

chris-cwa:
The problem was that the legacy config had a wrong time stamp and therefor was sometimes not considered outdated.
This leads to some Apps being updated from <= 1.6 to >= 1.7 to not always download remote config.
Test:

1. install 1.6.x
2. update to 1.7.x
3. config should be downloaded on first access (for instance when displaying interop countries)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
@@ -15,8 +15,10 @@ data class ConfigDataContainer(
 ) : ConfigData, ConfigMapping by mappedConfig {
     override val updatedAt: Instant = serverTime.plus(localOffset)
 
-    override fun isValid(nowUTC: Instant): Boolean {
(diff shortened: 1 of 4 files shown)


ralfgehrer (commented):
IMHO we should merge this into release/1.8.x as we won't do a hotfix for 1.7 anymore

chris-cwa:
> IMHO we should merge this into release/1.8.x as we won't do a hotfix for 1.7 anymore

the plan is to merge 1.7 -> 1.8 afterwards

d4rken (approved):
LGTM

d4rken:
> IMHO we should merge this into release/1.8.x as we won't do a hotfix for 1.7 anymore

Doesn't hurt to merge it here, and if there EVER is a hotfix for 1.7.1, then this should definitely be included. If not, still no harm done. As long as 1.8 isn't out, we can't guarantee that there will be no hotfix for 1.7.1 :grin:

mlenkeit:
Please only set the Jira item to Implemented (or something similiar) once the fix is in 1.8 and 1.9

BMItr (approved):
Code lgtm, Devicetest (Pixel 2) successful.

PR #1874, opened 2020-12-11
Adjust "abort warn others" dialog (EXPOSUREAPP-4253)

d4rken:
We no longer leave the screen if we cancel the cancel dialog :smile:.

* Change dialog cancel behavior. Stay on screen.
* Switch dialog button. The "positive" action is always in reference to the dialog's question. So positive here means to not warn others, while "warn others" is the negative action that cancels the dialog.
* Also removed some unused codes and did a little code cleanup.

## Testing
* Basically what is described in Jira: Positive TAN, press cancel on the test result screen, cancel the dialog.


ralfgehrer (approved):
Tested on Pixel (Android 10)

Oliver-Zimmerman (approved):
Tested on Pixel 5, works as expected. Code looks good to me.

PR #1875, opened 2020-12-11
Home screen only shows FAQ when leaving Consent (EXPOSUREAPP-4275)

Oliver-Zimmerman:
### Description

Small condition change for the positive card to be shown. Short oversight on my part. Apologies. 

Positive test result _**should**_ be shown _**if test result has been seen**_. Adjusted the state function for this logic to be true. 

Adjusted tests to match. 

### Steps to reproduce
1.Scan positive QR code
4.Tap on "Continue" on the "Test Result is Available" screen
5.Tap on "Don't share" on the OS consent screen
6.Tap on "X" on the "Positive Test Result (no consent)" screen
8.Tap on "Don't warn" in the confirmation pop-up
9. _**Positive Test Result card should now be shown**_


ralfgehrer (approved):
Tested on Pixel (Android 10).
Could also recover a previous install in the wrong state.

chris-cwa (approved):

PR #1877, opened 2020-12-11
Fix incorrect TRL data being applied to submitted TEKs (EXPOSUREAPP-4260)

d4rken:
Because we stored the information "we have so far" on each symptom step, aborting early could lead to the constellation of `Symptoms(startOf=null,indication=POSITIVE)`. The new submission flow behavior was missunderstood.

This PR fixes the behavior by switching to the `1.8.x` style of symptom indication:
* Pass the indication from the indication fragment to the calendar fragment using navigation arguments
* Only when the symptom data is completed on the calendar fragment, is the internal `SubmissionSettings.currentSymptoms` value updated.

## Test
* The original issue can be reproduced by performing a positive submission, and on the symptom indication screen, select "yes", but then abort symptom indication via the `X` in the toolbar.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/TransmissionRiskVectorDeterminator.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/TransmissionRiskVectorDeterminator.kt
@@ -49,15 +49,12 @@ class TransmissionRiskVectorDeterminator @Inject constructor(
                 is StartOf.NoInformation -> intArrayOf(5, 6, 8, 8, 8, 7, 5, 3, 2, 1, 1, 1, 1, 1, 1)
                 is StartOf.OneToTwoWeeksAgo -> intArrayOf(1, 1, 1, 1, 2, 3, 4, 5, 6, 6, 7, 7, 6, 6, 4)
                 else -> {
-                    try {
-                        throw IllegalStateException("Positive indication without startDate is not allowed: $symptoms")
-                    } catch (e: IllegalStateException) {
-                        e.reportProblem(
+                   IllegalStateException("Positive indication without startDate is not allowed: $symptoms")
+                   .reportProblem(
                             tag = "TransmissionRiskVectorDeterminator",
                             info = "Symptoms has an invalid state."
                         )
                         intArrayOf(1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
-                    }
                 }
             }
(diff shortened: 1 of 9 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
```diff
@@ -48,17 +50,29 @@ class SubmissionSymptomCalendarViewModel @AssistedInject constructor(
...
 
     fun onDone() {
-        Timber.d("onDone() clicked on calender screen.")
+        if (symptomStartInternal.value == null) {
+            try {
+                throw IllegalStateException("Can't finish symptom indication without symptomStart value.")
+            } catch (e: IllegalStateException) {
+                e.reportProblem(tag = TAG, "UI should not allow symptom submission without start date.")
+                return
+            }
```
chris-cwa, line 67:
```suggestion
           IllegalStateException("Can't finish symptom indication without symptomStart value.").reportProblem(tag = TAG, "UI should not allow symptom submission without start date.")
                return
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/TransmissionRiskVectorDeterminator.kt
```diff
@@ -47,9 +48,20 @@ class TransmissionRiskVectorDeterminator @Inject constructor(
...
+                else -> {
+                    try {
+                        throw IllegalStateException("Positive indication without startDate is not allowed: $symptoms")
+                    } catch (e: IllegalStateException) {
+                        e.reportProblem(
+                            tag = "TransmissionRiskVectorDeterminator",
+                            info = "Symptoms has an invalid state."
+                        )
+                        intArrayOf(1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
+                    }
```
chris-cwa, line 60:
```suggestion
                   IllegalStateException("Positive indication without startDate is not allowed: $symptoms").reportProblem(
                            tag = "TransmissionRiskVectorDeterminator",
                            info = "Symptoms has an invalid state."
                        )
                        intArrayOf(1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1)
```

ralfgehrer (approved):

AndroidMedaGalaxy (approved):

PR #1879, opened 2020-12-11
Display country list from config (EXPOSUREAPP-4278)

chiljamgossow:
Removes redundant method call that delayed display of countries


d4rken (approved):

ralfgehrer (approved):
Lgtm and could not reproduce the issue on a Pixel (Android 10).

PR #1881, opened 2020-12-11
Fix age restriction text color in light mode (DEV)

d4rken:
Screenshot from 2020-12-11 17-07-20

Screenshot from 2020-12-11 17-12-58


BMItr (approved):
I red the whole code over & over again.
Still looks good to me.

ralfgehrer (approved):

Issue #1855, opened 2020-12-09
Reduce use of ConstraintLayout


fynngodau:
When investigating why the privacy policy screen is taking so long to load, I've learned that `ConstraintLayout`s appear to be quite inefficient (in the mentioned case, replacing it with a `LinearLayout` has helped significantly, see #1854).

There is oftentimes a better alternative to `ConstraintLayout` which also makes the layout easier to read and maintain (see, for example, #1740). Therefore, I would like to suggest avoiding the use of `ConstraintLayout` and to prefer using simpler layouts like `LinearLayout` or `RelativeLayout` unless there is a good reason not to.

Should I come across other locations where there is a particular benefit to avoiding `ConstraintLayout`, I will of course open further PRs.

In specific, I suspect that startup time could be improved by a more simple home layout.

d4rken:
> In specific, I suspect that startup time could be improved by a more simple home layout.

I agree. We plan to flatten the home layout and remove all uses of "visible"/"gone" on included layouts by using a `RecyclerView` to display the elements.

fynngodau:
That's a great idea, `RecyclerView` is very well fit for this purpose.

For dealing with `RecyclerView`s with different items that must be rendered as different views, I wrote [a utility library](https://codeberg.org/fynngodau/TypedRecyclerView) that allows for a more object-oriented approach over passing the `RecyclerView.Adapter` superclass different `int`s for each view (which usually results in one single rather ugly class that handles all the different views). I find it useful for cases like these in my own projects, but maybe it's not well fit for this project, I'm not sure.

d4rken:
I've written a `ModularAdapter` pattern a few times already which I will likely re-use here as I know it plays nicely with `DiffUtil`, which we will need to update cards independently.

..hm don't have it in an open repo to show, but the basic idea is to allow adding modules to the adapter, where a module can be  something that binds data to a viewholder, or creates a viewholder or maps clicks/longclicks to viewholders.

e.g.
```kotlin
        modules.add(TypedVHCreator(0, { data[it].config is AppRestoreConfig }) { AppConfigVH(it) })
        modules.add(TypedVHCreator(1, { data[it].config is FilesRestoreConfig }) { FileConfigVH(it) })
```

I'm not stoked about the "ID" assignment for VHs in my solution though.
[...]

PR #1882, opened 2020-12-11
Polish information fragments (DEV/Community)

d4rken:
This PR polishes the information fragments under `Menu > App Information`.

* Contains the cherry picked commits from @fynngodau that improve loading performance (see #1854) , especially of fragments that display `.html` files from assets. 
* Refactor the formatter helper to remove static context access and make the html parser reusable
* Flatten a few more layouts by 1-2 levels
* Align headline sizes
* Fix illustration display, previously there was white space at the start and end due to the illustrations dimensions, now they are touching the screen edge so they have the desired "effects" (i.e. hand reaching in).

Screenshot from 2020-12-11 18-26-46

Ref #1855

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
+++ b/Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
@@ -14,8 +14,8 @@
         android:id="@+id/information_terms_container"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        android:focusable="true"
-        android:contentDescription="@string/information_terms_title">
+        android:contentDescription="@string/information_terms_title"
+        android:focusable="true">
 
         <include
             android:id="@+id/information_terms_header"
@@ -46,7 +46,7 @@
                     layout="@layout/include_information_details"
                     android:layout_width="0dp"
                     android:layout_height="wrap_content"
-                    app:body="@{FormatterHelper.formatStringAsHTMLFromLocal(@string/information_terms_html_path)}"
+                    app:body="@{FormatterHelper.parseHtmlFromAssets(context, @string/information_terms_html_path)}"
                     app:headline="@{@string/information_terms_headline}"
                     app:illustration="@{@drawable/ic_information_illustration_terms}"
                     app:illustrationDescription="@{@string/information_terms_illustration_description}"
--- a/Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
+++ b/Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
@@ -10,75 +10,37 @@
 
     </data>
 
-    <androidx.constraintlayout.widget.ConstraintLayout
+    <LinearLayout
         android:id="@+id/information_terms_container"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
         android:contentDescription="@string/information_terms_title"
-        android:focusable="true">
+        android:focusable="true"
+        android:orientation="vertical">
 
         <include
             android:id="@+id/information_terms_header"
             layout="@layout/include_header"
-            android:layout_width="0dp"
+            android:layout_width="match_parent"
             android:layout_height="wrap_content"
             app:icon="@{@drawable/ic_back}"
-            app:layout_constraintEnd_toEndOf="parent"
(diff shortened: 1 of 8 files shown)


Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
```diff
@@ -10,75 +10,37 @@
...
-            android:layout_width="0dp"
-            android:layout_height="0dp"
-            android:fillViewport="true"
-            app:layout_constraintBottom_toBottomOf="@+id/guideline_bottom"
-            app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintStart_toStartOf="parent"
-            app:layout_constraintTop_toBottomOf="@+id/information_terms_header">
+            android:layout_width="match_parent"
+            android:layout_height="match_parent"
+            android:fillViewport="true">
```
fynngodau, line 32: You might be missing a `paddingBottom` here if you want the same behavior as previously.

```suggestion
            android:paddingBottom="@dimen/guideline_bottom"
            android:fillViewport="true">
```

Or, if you prefer:


```suggestion
            android:paddingBottom="@dimen/spacing_small"
            android:fillViewport="true">
```

fynngodau (dismissed):
Looks good! I totally overlooked the terms layout because I removed that view from our fork, so I'm glad you made the according changes there, too.

Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
fynngodau: Regarding this padding (which the other layouts have, by the way): are you aware of the property `clipToPadding`? Setting it to false might be an effect that you desire, as the `ScrollView` would still have empty space below its last view but wouldn't clip the content to that padding anymore, allowing content to be rendered in those last 16dp of the screen as well.

d4rken: I did not check for that specifically, I just scanned the layout after refactoring to make sure it still looks good.

In terms of solutions i generally prefer `android:paddingBottom="@dimen/spacing_small"`, I don't like all the indirections with guidelines, includes, dimens we currently have, one screen refactoring at a time :wink: 

@ralfgehrer Do you spot the difference with padding here?

ralfgehrer (dismissed):

Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
ralfgehrer: Looks all the same with and without padding on my device. Hence, I'd leave it as is. Do you agree fynngodau?

ralfgehrer (dismissed):

ralfgehrer:
This, unfortunately, needs to be rebased to `release/1.11.x`

Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
fynngodau: I was pretty sure that this change would have the same effect as the guideline that is being removed from lines 75-80 had. I'll have another look.

ralfgehrer (approved):

chiljamgossow (approved):

Corona-Warn-App/src/main/res/layout/fragment_information_terms.xml
fynngodau: I was correct that this change would make it uniform with all the other views; you are correct that it is not useful. See the attached screenshots (with layout bounds):

tos-withoutpadding
tos-withpadding

The padding adds the white space that clips the text at the bottom. All other screens have that clipping.

Like I mentioned, I suppose the intended effect is that text doesn't align to the bottom without any margin, though like I also said above, this could be achieved in such a way that it doesn't clip the text (which I recommend).

Of course, here, the text has enough empty lines at the bottom that it has margin to the bottom anyway.

You will have to decide whether you want to make it uniform to other views.

d4rken: I don't think the clipping with whitespace looks good.
I'll merge this now and I'd be open to follow up PRs that introduce uniform whitespace at the bottom without clipping text, e.g. some "overscroll".

PR #1883, opened 2020-12-11
Fix app crash because of wrong technical.html path (EXPOSUREAPP-4298)

ralfgehrer:
While testing #1882, I noticed that for some languages a non-existing `technical_en.html` is linked - which leads to an app crash :(


Corona-Warn-App/src/main/res/values-ro/strings.xml
```diff
@@ -702,7 +702,7 @@
     <!-- XACT: describes illustration -->
     <string name="information_technical_illustration_description">"O mână ține un smartphone pe al cărui ecran este afișat un corp mare de text. Lângă text este afișat un cântar care simbolizează mențiunile legale."</string>
     <!-- XTXT: Path to the full blown legal html, to translate it exchange "_de" to "_en" and provide the corresponding html file -->
-    <string name="information_technical_html_path" translatable="false">"technical_en.html"</string>
```
d4rken, line 705: technical**s** ?

d4rken (changes requested):

d4rken (approved):

chiljamgossow (approved):

PR #1886, opened 2020-12-14
Fix crash when accessing TEK history without tracing enabled (EXPOSUREAPP-4284)

d4rken:
We crashed when tracing was disabled and tried to update the TEK history during submission.

This PR:
* Refactors the tracing permission helper to offer more reusability
* The `TEKHistoryUpdater` will now use the `TracingPermissionHelper` to request tracing permission to be enabled when we want to update the TEK history
* Besides fixing the crash it has the added benefit that the user no longer has to leave the submission flow if they have had tracing disabled.

## Testing
* Enable / disable tracing where ever possible
* Perform submission with tracing enabled and disabled
* If the user never gave consent during onboarding, the tracing consent dialog needs to be shown every time we want to enable tracing on any screen until consent was given once, and tracing has been enabled

### DIFF
--- a/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/submission/ui/SubmissionTestFragment.kt
+++ b/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/submission/ui/SubmissionTestFragment.kt
@@ -9,6 +9,7 @@ import androidx.recyclerview.widget.LinearLayoutManager
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.databinding.FragmentTestSubmissionBinding
 import de.rki.coronawarnapp.test.menu.ui.TestMenuItem
+import de.rki.coronawarnapp.tracing.ui.TracingConsentDialog
 import de.rki.coronawarnapp.util.di.AutoInject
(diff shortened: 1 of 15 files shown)


d4rken:
Did I cover all edge-cases in `TracingPermissionHelperTest` and `TEKHistoryUpdaterTest`?

ralfgehrer (commented):
Very nice refactoring with the dedicated helper. Device testing now.

Oliver-Zimmerman (dismissed):
LGTM. Turned off tracing, scanned positive result. With consent to warn others turned on, the pop-up asking to turn on tracing appeared. With consent manually turned off in the screen that follows the positive result scan, the flow works as usual with no tracing related pop-up appearing.

Code looks good too 👍

AlexanderAlferov (dismissed):
LGTM.
Small remark: shouldn't we cover TEKHistoryUpdater.handleActivityResult and TracingPermissionHelper.handleActivityResult with tests as public function that are being used in multiple places?

ralfgehrer (dismissed):
Cannot reproduce the crash anymore in any scenario I could think of.

ralfgehrer (approved):

AlexanderAlferov (approved):
LGTM. Thanks for adding more tests! 🚀

PR #1887, opened 2020-12-14
Fix symptom indication not being set correctly if NEGATIVE (EXPOSUREAPP-4286)

d4rken:
For the case "Indication = NEGATIVE" we did not update the persisted symptom indication used by the submission task.

## Testing
Perform submission with negative symptom submission and check the used TRLs.


mlenkeit (approved):
LGTM, tested on a Moto G7.

ralfgehrer (approved):
Tested on Pixel (Android 10): works nicely

Oliver-Zimmerman (approved):
Tested on Pixel 5. LGTM 👍

PR #1902, opened 2020-12-15
Made the whole diary card clickable (EXPOSUREAPP-4152)

SamuraiKek:
This is a small PR that also makes the contact diary card clickable, not just the button from the card.


AlexanderAlferov (approved):

BMItr (approved):
looking good so far. tested on pixel 2 device.
please recheck the navigation after merge to 4125-contact-diary (activity is call correctly)

Oliver-Zimmerman (approved):
LGTM, card is now clickable but navigation isn't working. I assume this needs to be merged with another open PR

SamuraiKek:
> LGTM, card is now clickable but navigation isn't working. I assume this needs to be merged with another open PR

Yes, that is correct.

PR #1912, opened 2020-12-16
Refactor home-screen and tracing details fragment (EXPOSUREAPP-4049)

d4rken:
* Use `RecyclerView` for dynamic display of list items
* Remove nested/included layouts and make Android Studio layout previews functional
* Split list items into independent elements (positive/negative/pending)
* Layout cleanups & flattening
* Small padding and alignment fixes

## TODO
- [x] Fix tests
- [x] Fix progress behavior in tracing details view
- [x] Cleanup state providers in the home and tracing details viewmodel
- [x] Check if we can make the tracing status bar on the home screen use scroll behavior

## Testing
* Check the UI states on the home screen and tracing details screen against FIGMA mockups
* Check the conditional logic in `SubmissionStateProvider`
* Check the conditional logic in `TracingStateProvider`


d4rken:
Ping @fynngodau

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestInvalidCard.kt
```diff
@@ -0,0 +1,25 @@
...
+import de.rki.coronawarnapp.databinding.HomeSubmissionStatusCardInvalidBinding
+import de.rki.coronawarnapp.submission.ui.homecards.TestInvalidCard.Item
+import de.rki.coronawarnapp.ui.main.home.HomeAdapter
+
+class TestInvalidCard(parent: ViewGroup) : HomeAdapter.HomeItemVH<Item, HomeSubmissionStatusCardInvalidBinding>(
+    R.layout.home_card_container_layout, parent
+) {
+
+    override val viewBinding = lazy {
+        HomeSubmissionStatusCardInvalidBinding.inflate(layoutInflater, itemView.findViewById(R.id.card_container), true)
```
fynngodau, line 14: This looks similar across all / most subclasses of `HomeAdapter.HomeItemVH`… I think it should be prettier to declare them in the superclass instead, but somehow that doesn't seem to be possible?

fynngodau (commented):
I looked at the PR partially so far

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeAdapter.kt
```diff
@@ -0,0 +1,62 @@
...
+            add(TypedVHCreatorMod({ data[it] is TestSubmissionDoneCard.Item }) { TestSubmissionDoneCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestInvalidCard.Item }) { TestInvalidCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestErrorCard.Item }) { TestErrorCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestFetchingCard.Item }) { TestFetchingCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestPositiveCard.Item }) { TestPositiveCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestNegativeCard.Item }) { TestNegativeCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestReadyCard.Item }) { TestReadyCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestPendingCard.Item }) { TestPendingCard(it) })
+            add(TypedVHCreatorMod({ data[it] is TestUnregisteredCard.Item }) { TestUnregisteredCard(it) })
+        }
```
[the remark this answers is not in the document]
d4rken: Done :ok_hand:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/tracing/details/TracingDetailsAdapter.kt
```diff
@@ -0,0 +1,56 @@
...
+            add(TypedVHCreatorMod({ data[it] is TracingDisabledBox.Item }) { TracingDisabledBox(it) })
+            add(TypedVHCreatorMod({ data[it] is TracingProgressBox.Item }) { TracingProgressBox(it) })
+            add(TypedVHCreatorMod({ data[it] is DetailsFailedCalculationBox.Item }) { DetailsFailedCalculationBox(it) })
+            add(TypedVHCreatorMod({ data[it] is DetailsIncreasedRiskBox.Item }) { DetailsIncreasedRiskBox(it) })
+            add(TypedVHCreatorMod({ data[it] is DetailsLowRiskBox.Item }) { DetailsLowRiskBox(it) })
+            add(TypedVHCreatorMod({ data[it] is PeriodLoggedBox.Item }) { PeriodLoggedBox(it) })
+            add(TypedVHCreatorMod({ data[it] is BehaviorIncreasedRiskBox.Item }) { BehaviorIncreasedRiskBox(it) })
+            add(TypedVHCreatorMod({ data[it] is BehaviorNormalRiskBox.Item }) { BehaviorNormalRiskBox(it) })
+            add(TypedVHCreatorMod({ data[it] is AdditionalInfoLowRiskBox.Item }) { AdditionalInfoLowRiskBox(it) })
+        }
```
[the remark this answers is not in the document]
d4rken: :+1:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestInvalidCard.kt
d4rken: Hm, only if we could call somehow pass the binding class and call that in the superclass with the necessary arguments, don't see it either, the inflation is via static methods, I'd not like to use reflection.

The base inflation into `itemView.findViewById(R.id.card_container)` is currently all the same, but there will be future features that have a home item that is not the default card.

So maybe we are okay with just leaving it as it is.

fynngodau: Ah, yes, reflection should work, but that wouldn't be too much better either.

> there will be future features that have a home item that is not the default card

That wouldn't be an issue if it could simply be overwritten.

It also looks to me like there is no common superclass offering the `inflate` method. If view binding wasn't used, one would call `LayoutInflater.inflate`, and there one could unify that by having sublasses provide just the layout ID to their superclass, but that wouldn't work here either…

So yes, it's probably best left as is, though I'm not very happy with it personally.

fynngodau (commented):
Some more thoughts.

Corona-Warn-App/src/main/res/layout/home_fragment_layout.xml
```diff
@@ -0,0 +1,116 @@
...
+
+        <include
+            android:id="@+id/main_tracing_divider"
+            layout="@layout/include_divider"
+            android:layout_width="match_parent"
+            android:layout_height="wrap_content" />
+
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/recycler_view"
+            android:layout_width="match_parent"
```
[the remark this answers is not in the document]
d4rken: :heart:

Corona-Warn-App/src/main/res/layout/home_fragment_layout.xml
```diff
@@ -0,0 +1,116 @@
...
+
+    <LinearLayout
+        android:id="@+id/container"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:contentDescription="@string/main_title"
+        android:orientation="vertical"
+        tools:context=".ui.main.MainActivity">
+
+        <androidx.constraintlayout.widget.ConstraintLayout
```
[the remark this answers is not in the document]
d4rken: I'd like to replace the custom layout with a normal android toolbar at somepoint, but as this is reused through out all screens, that will be in an extra PR, this one is already huge. :ghost:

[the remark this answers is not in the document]
fynngodau: Hm, that's a good idea, too. I'm not sure whether the default toolbar supports images like used here that are larger than an icon, but we don't have to worry about that right now.

fynngodau:
I am observing the cards getting elevated additionally and then de-elevated again whenever they are loaded (visual artifacts / yellow lines are due to my rendering somehow, but the effect I'm talking about is still visible):

cwaelevation

Additionally, there are now two visual effects when touching a card on top of each other, which is also visible in my GIF:

* The ripple effect
* The change in background color through the background tint color state list

I'd suggest to remove the latter.

ralfgehrer:
First of all: Amazing work!

Further improvements (preferably in a future PR) would be nice to have:
- The "Wurden Sie getestet" card flickers and refreshes its content when returning from the card details view of if the card itself refreshes
- The card details screen briefly shows the previous state and refreshes afterward. E.g.: User has a green card, go to the details view and notice the green card, too. Now turn off tracing, notice the white card on the home screen. Go to the card details --> the previous green card is shown very briefly before the white card shows up.

Corona-Warn-App/src/main/res/layout/home_fragment_layout.xml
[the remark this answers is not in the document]
d4rken: It does support images, I changed it to two toolbars and made the lower one react to scroll . :sunglasses:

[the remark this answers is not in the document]
fynngodau: Now the two toolbars are also elevated properly :+1: very cool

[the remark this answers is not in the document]
fynngodau: However this change adds a new issue: there is a white background behind the menu now, causing an attentive user to see a visual artefact while the menu is opening or closing.

This is caused by the `background` that the toolbar's style has.

I took this screenshot while opening the menu:


menuopening

d4rken:
> However this change adds a new issue: there is a white background behind the menu now, causing an attentive user to see a visual artefact while the menu is opening or closing.

Fixed it, though rather workaroundish, need to look into what's going on with the themes and styles., overriding `colorPrimary` for that style, the artifact doesn't appear when using `tint` instead of `background`.

> Additionally, there are now two visual effects when touching a card on top of each other, which is also visible in my GIF:

Good catch! I agree with your recommendation and replaced the color state list.

> The "Wurden Sie getestet" card flickers and refreshes its content when returning from the card details view of if the card itself refreshes

The diff-util did not recognize it as the same data item, should be fixed now.

> The card details screen briefly shows the previous state and refreshes afterward. E.g.: User has a green card, go to the details view and notice the green card, too. Now turn off tracing, notice the white card on the home screen. Go to the card details --> the previous green card is shown very briefly before the white card shows up.

Still looking into this one, will likely fix in a separate PR.

Corona-Warn-App/src/main/res/layout/home_card_container_layout.xml
```diff
@@ -0,0 +1,12 @@
...
+    android:id="@+id/card_container"
+    style="@style/cardNoPadding"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:layout_marginStart="@dimen/spacing_small"
+    android:layout_marginEnd="@dimen/spacing_small"
+    android:layout_marginTop="@dimen/spacing_tiny"
+    android:layout_marginBottom="@dimen/spacing_tiny"
+    android:foreground="?selectableItemBackground"
+    android:focusable="true" />
```
fynngodau, line 12: This change replaces the `FrameLayout` with a `CardLayout`.

```suggestion
<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
    xmlns:app="http://schemas.android.com/apk/res-auto"
    android:id="@+id/card_container"
    android:layout_width="match_parent"
    android:layout_height="wrap_content"
    android:background="@color/colorSurface1"
    android:layout_marginStart="@dimen/spacing_small"
    android:layout_marginEnd="@dimen/spacing_small"
    android:layout_marginTop="@dimen/spacing_tiny"
    android:layout_marginBottom="@dimen/spacing_tiny"
    android:foreground="?selectableItemBackground"
    android:focusable="true"
    app:cardElevation="@dimen/elevation_strong"
    app:cardCornerRadius="@dimen/radius_card" />
```

Here's a screenshot:

cwahome

Notice how the green card has rounded corners again.

Corona-Warn-App/src/main/res/layout/tracing_content_failed_view.xml
```diff
@@ -0,0 +1,93 @@
...
+    <data>
+        <variable
+            name="state"
+            type="de.rki.coronawarnapp.tracing.states.TracingFailed" />
+    </data>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:backgroundTint="@color/colorSurface1"
```
fynngodau, line 15: This tint has no effect.

```suggestion

```

fynngodau (changes requested):
I noticed two more issues:

* The risk card has no rounded corners. This is also visible in the video you posted above. It is caused by the `background` color which overwrites the `@drawable/card` in the background, thereby not respecting the rounded corners.
* The ripple effect does not respect the corners.

I looked into the first one for a bit in particular, and I think both are quite tricky: you used `backgroundTint` previously (and you still are in  `tracing_content_failed_view` without any effect, by the way), but that wouldn't work as is because then you'd need to declare `background` in the same layout, which would work on the home screen as expected, but not in the risk detail fragment where the same layout is used again.

The second issue happens because the `foreground` does not care about the background and draws within the full bounds of the layout.

Both of these would be fixed if the `home_card_container_layout` was actually clipping to the `4dp` corner radius. The `CardView` layout does that, so I would replace the `FrameLayout` with a `CardView`.

d4rken:
> Both of these would be fixed if the home_card_container_layout was actually clipping to the 4dp corner radius. The CardView layout does that, so I would replace the FrameLayout with a CardView.

That's a nice solution. Still having some trouble with the card background in darkmode. Seems to be related to `background` vs `backgroundTint` on the cardview.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestInvalidCard.kt
kolyaopahle: hmm, what about `DataBindingUtil.inflate`? we already pass the binding type as a generic argument

fynngodau: @kolyaopahle That's a good idea as this class is made for that purpose and it looks to me like it would work. We'd need to pass the layout ID to the superclass so that it can inflate it, though.

Corona-Warn-App/src/main/res/layout/home_card_container_layout.xml
```diff
@@ -0,0 +1,15 @@
...
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    android:backgroundTint="@color/colorSurface1"
+    android:layout_marginStart="@dimen/spacing_small"
+    android:layout_marginEnd="@dimen/spacing_small"
+    android:layout_marginTop="@dimen/spacing_tiny"
+    android:layout_marginBottom="@dimen/spacing_tiny"
+    android:foreground="?selectableItemBackground"
+    android:focusable="true"
+    app:cardElevation="@dimen/elevation_strong"
```
fynngodau, line 14: I think I introduced an inconsistency with the intended elevation here, as I spotted some code that explicitly set the risk card to `10dp` (e.g. `TracingDisabledBox`). I added this line because the code didn't work anymore (without me knowing that it existed and wondering a little), because it would have to set `cardElevation` for the `CardView` instead of `elevation`. Maybe you could try to find out what the intended elevation is, and make sure to set it correctly.

chiljamgossow (dismissed):
I think this is a major improvement & and it works well enough to merge. Polishing can be done in small separate PRs.

SamuraiKek:
Looking great! I love the fact that the UI doesn't flicker anymore when returning to the home screen. One small thing I have noticed is that the header with the cwa logo now stays while scrolling whereas, on the old version of the app, it's scrollable. I'm not sure what is the actual intended behaviour but I like it more the way it is right now. It feels more consistent with the rest of the screens where the header remains on top while scrolling.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/settings/SettingsTracingFragment.kt
```diff
@@ -41,8 +41,8 @@ class SettingsTracingFragment : Fragment(R.layout.fragment_settings_tracing), Au
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        vm.tracingDetailsState.observe2(this) {
-            binding.tracingDetails = it
+        vm.logginPeriod.observe2(this) {
```
SamuraiKek, line 44: I think this refers to the logging* period.

d4rken:
> Looking great! I love the fact that the UI doesn't flicker anymore when returning to the home screen. One small thing I have noticed is that the header with the cwa logo now stays while scrolling whereas, on the old version of the app, it's scrollable. I'm not sure what is the actual intended behaviour but I like it more the way it is right now. It feels more consistent with the rest of the screens where the header remains on top while scrolling.

It wasn't clearly defined in mockups AFAICT, I agree that it feels more consistent (within our app, and with regards to other apps) if it does not scroll and would be happy to keep it this way.

Corona-Warn-App/src/main/res/layout/tracing_details_fragment_layout.xml
```diff
@@ -0,0 +1,138 @@
...
+    xmlns:bind="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
+
+    <data>
+        <variable
+            name="tracingDetailsState"
+            type="de.rki.coronawarnapp.tracing.ui.details.TracingDetailsState" />
+    </data>
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
BMItter, line 13: ConstraintLayout stacking.
No urgent, but for keeping in mind / future PR.

BMItr (approved):
:heart: Nice.
Code lgtm. Also good to read.
Tested on Pixel 2 & Samsung S10 Device.
Also noticeable performance increase while navigating!
Code is very welcome.
Little layout issues, but no urgent. Also have a look at the id handling (+id/id) when using e.g. constraints (e.g. in Chiljas PR), if you like once.

PR #1931, opened 2020-12-17
Show onboarding screen only once (EXPOSUREAPP-4367)

axelherbstreith:
This PR addresses the persistance handling of the onboaring info screen.
ExposureApp-4152.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/ContactDiarySettings.kt
```diff
@@ -0,0 +1,29 @@
+package de.rki.coronawarnapp.submission
```
BMItter, line 1: please recheck package name

axelherbstreith: Done

BMItr (approved):

Oliver-Zimmerman (approved):

Corona-Warn-App/src/main/res/layout/contact_diary_activity.xml
```diff
@@ -11,7 +11,6 @@
         android:name="androidx.navigation.fragment.NavHostFragment"
         android:layout_width="match_parent"
         android:layout_height="match_parent"
-        app:defaultNavHost="true"
-        app:navGraph="@navigation/contact_diary_nav_graph" />
```
chiljamgossow, line 15: We don't need this?

axelherbstreith: Will be set programmatically in the ContactActivity. This allows to dynamically set the first fragment based on the data from the Shared Preferences

SamuraiKek: Looks like it works without it but the main activity xml has it so I'm not sure...

PR #1943, opened 2020-12-18
Code smells diary (EXPOSUREAPP-4152)

BMItr:
this PR addresses little code smells and comments for the ContactDiary.
ExposureApp-4152

PR #1949, opened 2020-12-18
Fix date format on overview page (EXPOSUREAPP-4404)

ralfgehrer:
Screenshot 2020-12-18 at 13 46 49

two digit year instead of four digits.


ralfgehrer:
Meanwhile obsolete -  allready addressed by https://github.com/corona-warn-app/cwa-app-android/pull/1943.

PR #1953, opened 2020-12-18
Improve risklevel result data access performance (DEV)

d4rken:
Got pinged by testers about the app taking longer to load, further investigation showed several issues that this PR fixes. The underlying issues are not present (or only to a lot lesser degree) on production:
* Exposure Window data was being explicitly logged in tester builds, due to much larger scan instance counts, the logging slowed down data emission by `Flow` that printed the log output.
    * I've changed the data class to exclude scan instance output via `toString`, reducing it to only show the item count.

* RiskLevel result history size is much larger for the tester builds, and tester builds store the exposure window data. We only had a single way of accessing the stored risk result which meant that the access triggered retrieval of all results (and required mapping matching exposure windows to each result).
   * There is a new `latestRiskLevelResults` that uses SQL queries to prefilter the data (faster than in kotlin) a return only the latest two results (which is what the UI is mostly interested in, except for test screens)

* Multiple data access was happening, where we could have reused the results, switching from home to the risk details screen, caused 2 sepperate retrieval events.
    * We are now using kotlin `stateIn` to share the flow subscription. Currently it's  only `WhileSubscribed` with a replay value of `0ms` but due to async viewmodel teardown when leaving the home screen this is already enough to prevent double flow subscription.

## Testing
* Pay  specific attention to the SQL queries used: https://github.com/corona-warn-app/cwa-app-android/pull/1953/files#diff-6ea921ce2a6fafac7cebb28c46196a4253ddae3c5e0283a996af07414e525e42
* Their tests: https://github.com/corona-warn-app/cwa-app-android/pull/1953/files#diff-b5e6e55b5b230901e3f416ad888b9facb41ef08809cbbbb31736f1b2ffae17c7R18
* The mapping: https://github.com/corona-warn-app/cwa-app-android/pull/1953/files#diff-edd0be10604d896c0dfd9383bc23e4e1185ba1cca660156aec9c9d9b91fdc7edR30

Also do a lookup in Android Studio on where each `Flow` is accessed from and whether it's the correct data being supplied.


BMItr (dismissed):
I'm unsure whether these changes will be sufficent for the mentioned problem.
Issue should be kept in mind. @kolyaopahle 
Code lgtm.

d4rken:
> I'm unsure whether these changes will be sufficent for the mentioned problem.

I concur that we should keep an :eye: on this.

I added some debug code that multiplied the processed windows by 100 and it was better, there may be issues where the amount of scan instances is high. The next step would probably be to created combined SQL queries, this was problematic at first because we only care about `ExposureWindow`s on tester builds :thinking: 

There may be further optimizations by increasing the "replay duration" i.e. how long the timeout is during which the last value will be replayed, but that's a different topic, where we have to check if there are certain constellations where the data is accessed via `.first()` and replaying values could cause subtle bugs.

ralfgehrer (dismissed):
lgtm

d4rken:
* Pay  specific attention to the SQL queries used: https://github.com/corona-warn-app/cwa-app-android/pull/1953/files#diff-6ea921ce2a6fafac7cebb28c46196a4253ddae3c5e0283a996af07414e525e42
* Their tests: https://github.com/corona-warn-app/cwa-app-android/pull/1953/files#diff-b5e6e55b5b230901e3f416ad888b9facb41ef08809cbbbb31736f1b2ffae17c7R18
* The mapping: https://github.com/corona-warn-app/cwa-app-android/pull/1953/files#diff-edd0be10604d896c0dfd9383bc23e4e1185ba1cca660156aec9c9d9b91fdc7edR30

Also do a lookup in Android Studio on where each `Flow` is accessed from and whether it's the correct data being supplied.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/internal/RiskResultDatabase.kt
```diff
@@ -41,9 +41,15 @@ abstract class RiskResultDatabase : RoomDatabase() {
...
     @Dao
     interface RiskResultsDao {
-        @Query("SELECT * FROM riskresults")
+        @Query("SELECT * FROM riskresults ORDER BY calculatedAt DESC")
         fun allEntries(): Flow<List<PersistedRiskLevelResultDao>>
 
+        @Query("SELECT * FROM riskresults ORDER BY calculatedAt DESC LIMIT :limit")
+        fun latestEntries(limit: Int): Flow<List<PersistedRiskLevelResultDao>>
+
+        @Query("SELECT * FROM (SELECT * FROM riskresults ORDER BY calculatedAt DESC LIMIT 1) UNION ALL SELECT * FROM (SELECT * FROM riskresults where failureReason is null ORDER BY calculatedAt DESC LIMIT 1)")
```
ralfgehrer, line 50: are the outer `SELECT * FROM` on both sides of  UNION ALL really needed?

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/internal/RiskResultDatabase.kt
d4rken: I would have said "no" before trying, but I can't get it to work without. Room throws errors on compile. @BMItter Any idea?

BMItter: Its a specification for sqlite/room; only possible with subqueries as done.
so all right here

ralfgehrer (dismissed):

BMItter:
looking good so far.
current branch doesn't allow me to run RiskResultDatabaseTest (due some other testproblems (HomeFragmentTest?))
could you check this at last?

d4rken:
> current branch doesn't allow me to run RiskResultDatabaseTest (due some other testproblems (HomeFragmentTest?))
could you check this at last?

Done. The HomeFragment test does not pass since the big refactoring of the home screen, but the other tests can be run again.

BMItr (approved):
:ok_hand:

ralfgehrer (approved):

PR #1954, opened 2020-12-18
Run db access on io thread (DEV)

chiljamgossow:
db access on io thread
addresses review comment


ralfgehrer (approved):

BMItr (approved):
I'm not really sure why this is necessary because room creates a co-routines itself when suspending.
Also calling launchIn on a Flow would make sense to me.

chiljamgossow:
I did not find an issue here, both versions work fine.

d4rken:
There was confusion over the scopes(threads) being implicitly used. TLDR:

* Room internally switches to a Room specific scope with extra (non-main) threads.
* Live Data does not and executes on the main thread (scope)

The database query would not have been an issue but the sorting (any future repository operations) would be running on the main thread.
By expclitly telling livedata to run on a different context, as you did here, we avoid any situations where internal changes or sorting of large data sets lags the UI.

It could be discussed whether `IO` or `Default` would be the better dispatcher choice, but either has the desired effect.

All is okay :+1:

PR #1967, opened 2020-12-21
 Provide Darkmode Icon/Layout (EXPOSUREAPP-4237)

axelherbstreith:
-  Provide Dark mode Icon/Layout (EXPOSUREAPP-4237)
-  Fix issue on P30 on which the contact journal could be opened several times by multi-tap on the home screen's contact journal card


Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -80,6 +80,7 @@
             android:name=".contactdiary.ui.ContactDiaryActivity"
             android:exported="false"
             android:screenOrientation="portrait"
+            android:launchMode= "singleTop"
```
SamuraiKek, line 83: Why is this necessary?

axelherbstreith: Some devices allow to open the contact diary multiple times when clicking fast on the Home Screen card (tested on Huawei P30). SingleTop solves this behaviour

SamuraiKek:
I got this working on my emulator - Android 10, but both my Nokia 3 - Android 9 and Huawei Mate 8 - Android 6, are not able to open the day fragment after pressing on a specific day in the list. I get a white page for a few seconds and then I'm returned to the home screen. As a sidenote, none of my physical phones have dark mode option.

SamuraiKek:
Thre's still some minor differences from the design in dark mode. 
Screenshot_1608551694
The text for the selected tabs is white in the design while it's blue in the current version.
The text 'Noch keine Personen vorhanden' is white it the design while it's grey in the current version.
The same observations go to the 'Orte' page as well.
Screenshot_1608551934
The entry text is white in the design but grey in the current version.
Same goes for the text in 'Orte'.
The rest seems to be fine.

BMItr (dismissed):

ralfgehrer (approved):

BMItr (approved):

PR #1972, opened 2020-12-21
Fix for not showing keyboard on all devices on edittext focus (EXPOSUREAPP-4152)

BMItr:
This PR addresses the no-keyboard-showing-up automatically on some device when edittext has focus.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
@@ -2,9 +2,8 @@ package de.rki.coronawarnapp.contactdiary.util
 
 import android.content.Context
 import android.view.View
+import android.view.ViewTreeObserver
 import android.view.inputmethod.InputMethodManager
-import android.widget.EditText
-import androidx.fragment.app.Fragment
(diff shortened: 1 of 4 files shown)


ralfgehrer (approved):
Works nicely on a Pixel and a Huawei P10.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
```diff
@@ -40,3 +28,32 @@ fun String.formatContactDiaryNameField(maxLength: Int): String {
     }
     return newName.take(maxLength)
 }
+
+fun View.focusAndShowKeyboard() {
+    /**
+     * This is to be called when the window already has focus.
+     */
+    fun View.showTheKeyboardNow() {
```
chiljamgossow, line 36: Why is this nested?

axelherbstreith (approved):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
BMItter: member functions can be accessed without creating an object and for good overview

PR #1984, opened 2020-12-22
Rely on native  `observe` from LiveData instead of `observe2` (DEV)

mtwalli:
Replace `observe2` with `LiveDate.observe`. 
Since kotlin 1.4 it not necessarily to write the old syntax of observing `LiveData` in the Fragments or Activities.

old syntax: `viewModel.liveData.observe(this, Observer { // TODO})`

new syntax: `viewModel.liveData.observe(this) {// TODO}`

Therefore we don't have to write our own extension function to achieve that or use extension function from lifecycle library that was doing the same.

For more information please have a look at the following resources:
1- https://developer.android.com/jetpack/androidx/releases/lifecycle#2.3.0-beta01
2- https://android-review.googlesource.com/c/platform/frameworks/support/+/1413815

- UPDATE  - 
I agreed with @d4rken to keep `observe2` for Fragments as is


d4rken:
Thanks for your PR @mtwalli .

While I like that the syntax is now nicer by default, I don't like that we would have to specify `viewLifecycleOwner` everywhere, how can we make this nicer?

d4rken (approved):
LGTM

PR #1985, opened 2020-12-22
Check device time (EXPOSUREAPP-2998)

d4rken:
Incorrect device time handling.

* Evaluate current device time on each config update
* Show popup dialog on incorrect device
* Show notification if device time is incorrect and app is in background
* Download but do not submit keys if device time is incorrect
* Don't perform risk calculation if time is incorrect

## Testing
* Play with the device time (timetravel)
* Force app config refreshes via test menu
* Filter logcat for `AppConfigSource` output
* Comapre against techspec


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/AppConfigSource.kt
```diff
@@ -25,19 +28,28 @@ class AppConfigSource @Inject constructor(
...
         }
 
         val remoteConfig = remoteAppConfigSource.getConfigData()
 
         return when {
             remoteConfig != null -> {
                 Timber.tag(TAG).d("Returning remote config.")
+                if (!remoteConfig.isDeviceTimeCorrect) {
+                    Timber.tag(TAG).w(
+                        "Device time is incorrect, offset=%s", remoteConfig.localOffset.standardMinutes
```
mlenkeit, line 41: Might make sense to add a unit to the log statement...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -3,18 +3,22 @@ package de.rki.coronawarnapp.appconfig.mapping
...
+    private val testSettings: TestSettings
+) : CWAConfig.Mapper {
+
     override fun map(rawConfig: ApplicationConfigurationAndroid): CWAConfig {
         return CWAConfigContainer(
             latestVersionCode = rawConfig.latestVersionCode,
             minVersionCode = rawConfig.minVersionCode,
             supportedCountries = rawConfig.getMappedSupportedCountries(),
-            appFeatures = rawConfig.mapAppFeatures()
+            isDeviceTimeCheckEnabled = !rawConfig.isDeviceTimeCheckDisabled()
```
ralfgehrer, line 21: minor: as far as I could see, in all other instances the disabled case (`isDeviceTimeCheckDisabled`)was used for all other booleans - why switch to enabled `isDeviceTimeCheckEnabled `here?

d4rken: Here we evaluate whether the config contains the "disable the check flag".
It doesn't tell us whether it should be enabled (which is also affected by other things like caching) it just tells us if we should disable it.

I use "isDeviceTimeCheckEnabled" in all other cases because there we can make the decision whether the check is enabled or not.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
```diff
@@ -13,12 +13,15 @@ data class ConfigDataContainer(
     override val localOffset: Duration,
     override val configType: ConfigData.Type
 ) : ConfigData, ConfigMapping by mappedConfig {
+
+    override val isDeviceTimeCorrect: Boolean
+        get() = !isDeviceTimeCheckEnabled || localOffset.abs() < ConfigData.DEVICE_TIME_GRACE_RANGE
```
ralfgehrer, line 18: disabled but corect time: true | true -> true 
disabled but wrong time: true | true -> true 
enabled but correct time: false | true -> true 
enabled but wrong time: false | false -> false

Looks good - am just wondering how this plays out if the testers disable the clock-check the tester menu. What's the priority then?

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/appconfig/ui/AppConfigTestFragmentViewModel.kt
```diff
@@ -11,12 +12,16 @@ import timber.log.Timber
...
     dispatcherProvider: DispatcherProvider,
-    private val appConfigProvider: AppConfigProvider
+    private val appConfigProvider: AppConfigProvider,
+    private val testSettings: TestSettings
 ) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
 
     val currentConfig = appConfigProvider.currentConfig.asLiveData()
     val errorEvent = SingleLiveEvent<Exception>()
 
+    val deviceTimeCheckDisabled = testSettings.isDeviceTimeCheckDisabled.flow
```
ralfgehrer, line 22: just for clarification: this is solely used for the test menu and has no impact on any other behaviour in the app - e.g. preventing key download/submission or alike?

d4rken: No, this affects the whole app. Enabling this will fake the existence of the `disable-time-check` feature flag in the AppConfig:

https://github.com/corona-warn-app/cwa-app-android/blob/bd2f7bfa69ad287e93010da469120b70a904e885/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt#L35-L39

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
d4rken: If the test menu option "disable time check" is enabled, then `isDeviceTimeCheckEnabled` will return `false`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/risk/RiskLevelTaskTest.kt
```diff
@@ -86,4 +90,17 @@ class RiskLevelTaskTest : BaseTest() {
 //
 //        verify { riskLevelSettings.lastUsedConfigIdentifier = "config-identifier" }
```
ralfgehrer, line 91: on purpose? Test is now empty.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/appconfig/ui/AppConfigTestFragmentViewModel.kt
ralfgehrer: thanks

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/risk/RiskLevelTaskTest.kt
d4rken: Basically an unrelated TODO from the RiskLevelTask refactoring. We should increase test coverage for this task in another PR.

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_appconfig.xml
```diff
@@ -73,6 +73,25 @@
...
+                style="@style/body1"
+                android:layout_width="match_parent"
+                android:layout_height="0dp"
+                android:layout_marginTop="@dimen/spacing_tiny"
+                android:layout_weight="1"
+                android:text="Disable device time check"
+                android:theme="@style/switchBase" />
+
+        </LinearLayout>
+
```
mtwalli, line 94: can you just use `Switch` without `LinearLayout`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/devicetime/DeviceTimeHandler.kt
```diff
@@ -0,0 +1,54 @@
...
+        appConfigProvider.currentConfig
+            .onStart { Timber.tag(TAG).d("Observing device time.") }
+            // If we don't delay emissions, we will consume the event before the UI "wasDeviceTimeIncorrectAcknowledged"
+            // ForegroundState.isInForeground has a day when opening the app.
+            .sample(5000)
+            .onEach {
+                Timber.tag(TAG).v("Current device time offset is: %dms", it.localOffset.millis)
+                if (it.isDeviceTimeCorrect) {
+                    Timber.v("Dismissing any notification, device time is correct again.")
+                    notification.dimiss()
```
ralfgehrer, line 34: I think we also need to reset the "notification was shown" flag

"Once the device time has been corrected, the persisted information about the already shown notification is reset, so that the next time the device time is wrong again, the user gets again a notification"

If I time travel, I receive the notification once. After reset to the current time, I let the app be in that state for 10 minutes, then time-traveled again but did not receive a notification/pop-up.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/sources/remote/RemoteAppConfigSource.kt
```diff
@@ -48,7 +48,7 @@ class RemoteAppConfigSource @Inject constructor(
     }
 
     fun clear() {
-        server.clearCache()
+        Timber.tag(TAG).v("clear()")
```
mtwalli, line 51: `clear` now  not really  clearing either to change the name or document why it has to be like that

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -238,6 +245,17 @@ class HomeFragmentViewModel @AssistedInject constructor(
...
 
+    private var wasDeviceTimeDialogShown = false
+    val showIncorrectDeviceTimeDialog = appConfigProvider.currentConfig
+        .filter {
+            !it.isDeviceTimeCorrect && !wasDeviceTimeDialogShown && !cwaSettings.wasDeviceTimeIncorrectAcknowledged
+        }
+        .onEach {
+            wasDeviceTimeDialogShown = true
+        }
+        .map { }
```
mtwalli, line 256: Please remove the unused `map`

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_appconfig.xml
d4rken: We could, but I'm sure there will be more switches in the future to toggle other app config features, the whole box is there to contain those switches. We could maybe add a title to it, but IMHO it is okay as it is.

ralfgehrer (dismissed):

PR #1994, opened 2020-12-23
Announce type for location and person - accessibility (EXPOSUREAPP-4459-4458-4456)

chiljamgossow:
Announce page title, type of list item (person or location), state (selected or unselected) and change of state
Add action description on edit screen
Can be tested with Talkback


mtwalli (dismissed):

mtwalli:
LGTM

Corona-Warn-App/src/main/res/layout/contact_diary_location_list_item.xml
```diff
@@ -4,6 +4,7 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/contact_diary_location_list_line"
```
axelherbstreith, line 7: May rename the ids to "[...]_list_item" instead of "[...]_list_line to be aligned with the other files

chiljamgossow: done

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/location/ContactDiaryLocationListViewModel.kt
```diff
@@ -50,18 +70,21 @@ class ContactDiaryLocationListViewModel @AssistedInject constructor(
...
     interface Factory : CWAViewModelFactory<ContactDiaryLocationListViewModel> {
         fun create(selectedDay: String): ContactDiaryLocationListViewModel
     }
 }
+
+private val TAG = ContactDiaryLocationListViewModel::class.java.simpleName
+private const val SELECTED_CONTENT_DESCRIPTION = R.string.accessibility_location_selected
+private const val UNSELECTED_CONTENT_DESCRIPTION = R.string.accessibility_location_unselected
+private const val SELECT_ACTION_DESCRIPTION = R.string.accessibility_action_select
+private const val DESELECT_ACTION_DESCRIPTION = R.string.accessibility_action_deselect
```
mtwalli, line 90: R.string.* are actually constants, any reason why it should be assigned to another constant?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/person/ContactDiaryPersonListViewModel.kt
```diff
@@ -56,12 +74,14 @@ class ContactDiaryPersonListViewModel @AssistedInject constructor(
...
     interface Factory : CWAViewModelFactory<ContactDiaryPersonListViewModel> {
         fun create(selectedDay: String): ContactDiaryPersonListViewModel
     }
 }
+
+private val TAG = ContactDiaryPersonListViewModel::class.java.simpleName
+private const val SELECTED_CONTENT_DESCRIPTION = R.string.accessibility_person_selected
+private const val UNSELECTED_CONTENT_DESCRIPTION = R.string.accessibility_person_unselected
+private const val SELECT_ACTION_DESCRIPTION = R.string.accessibility_action_select
+private const val DESELECT_ACTION_DESCRIPTION = R.string.accessibility_action_deselect
```
mtwalli, line 87: Same here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/StringProvider.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import android.content.Context
+import androidx.annotation.StringRes
+
+interface StringProvider {
+    fun getString(@StringRes resId: Int, formatArg: Any? = null): String
+}
+
+class ContextStringProvider(private val context: Context) : StringProvider {
+    override fun getString(@StringRes resId: Int, formatArg: Any?): String {
```
mtwalli, line 11: To make it flexible for other cases  `varargs formatArg: Any?` as in `context.getString`

Corona-Warn-App/src/main/res/layout/contact_diary_edit_locations_fragment.xml
```diff
@@ -16,7 +16,9 @@
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toTopOf="parent"
             app:navigationIcon="@drawable/ic_close"
-            app:title="@string/contact_diary_edit_locations_title" />
+            app:navigationContentDescription="@string/accessibility_close"
+            app:title="@string/contact_diary_edit_locations_title"
+            android:contentDescription="@string/contact_diary_edit_locations_title"/>
```
mtwalli, line 21: Please format? Normally `android:` attributes are before `app:`

Corona-Warn-App/src/main/res/layout/contact_diary_edit_persons_fragment.xml
```diff
@@ -16,7 +16,9 @@
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toTopOf="parent"
             app:navigationIcon="@drawable/ic_close"
-            app:title="@string/contact_diary_edit_persons_title" />
+            app:navigationContentDescription="@string/accessibility_close"
+            app:title="@string/contact_diary_edit_persons_title"
+            android:contentDescription="@string/contact_diary_edit_persons_title"/>
```
mtwalli, line 21: same here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/location/ContactDiaryLocationListViewModel.kt
chiljamgossow: I do it because it is easier to change/ maintain

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/StringProvider.kt
chiljamgossow: Our current lint setup does not allow varargs.

Corona-Warn-App/src/main/res/layout/contact_diary_edit_locations_fragment.xml
chiljamgossow: If we want that we should do that automatically/ check with lint.

It's not done anywhere in the app.

axelherbstreith (dismissed):
LGTM

mtwalli (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt
```diff
@@ -63,4 +65,7 @@ class AndroidModule {
     @Provides
     @Singleton
     fun activityManager(@AppContext context: Context): ActivityManager = context.getSystemService()!!
+
+    @Provides
+    fun stringProvider(@AppContext context: Context): StringProvider = ContextStringProvider(context)
```
d4rken, line 70: I'm not a fan using the application context, isn't it better to use the activity/fragment/view context to resolve the strings, and resolve them as close as possible to UI?

We have a similar pattern in the app, see `LazyString`, which reverses the responsibility, instead of using a provider to resolve using the application context, we create a getter where the caller has to input their own context to get the string.

d4rken: Something maybe like

```kotlin
fun Int.toResolvingString(vararg formatArgs: Any): LazyString = object : LazyString {
    override fun get(context: Context): String = context.getString(this@toResolvingString, *formatArgs)
}

Item(
    R.string.my_string_resource.toResolvingString()
)

````

File: Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/StringProvider.kt (hunk 4 above)
Comment (d4rken) [reply Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/StringProvider.kt:11]: Would be okay here to use `Suppress`, the performance impact of `varargs` shouldn't be an issue in this case. Maybe we should relax the LINT check for this, this is rarely  problematic and Kotlin `varargs` is really nice :smiley: .

File: Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt (hunk 7 above)
Comment (chiljamgossow) [reply Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt:70]: This is used to resolve accessibility strings for the list items. Strings need to be resolved for each element, depending on states only known in the viewmodel. I did not want to move that much logic into the fragment. 
I saw the LazyString but did not find it useful in this particular case.
Is there a problem with using the app context?

Comment (chiljamgossow) [reply Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt:70]: I can use an assisted inject with the fragment's context

Comment (d4rken) [reply Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt:70]: > Strings need to be resolved for each element, depending on states only known in the viewmodel. I did not want to move that much logic into the fragment.

The logic would still be in the viewmodel, just the resolving would be moved to the fragment. It would look like:

```kotlin
SelectableItem(
    true,
    contactDiaryLocation,
    SELECTED_CONTENT_DESCRIPTION.toResolvingString(contactDiaryLocation.locationName),
    (...)
)

// Later in UI
it.contentDescription = item.onClickDescription.get(context)
```

> Is there a problem with using the app context?

It's good practice to not use the app context for UI related operations. The internal call will perform `context.resources.getString(...)` and `resources` can be different between app context and activity context, resources is affected by theming and other options that can be different per activity.

d4rken: > I can use an assisted inject with the fragment's context

That would be okay too. Not sure about testing. Up to you.

chiljamgossow: I added what you proposed

axelherbstreith (approved):

d4rken (approved):
Neat :ok_hand:

PR #1996, opened 2020-12-23
Fix wrong icon in settings when bluetooth is disabled (EXPOSUREAPP-3563)

d4rken:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/settings/start/SettingsTracingState.kt
```diff
@@ -15,11 +15,10 @@ sealed class SettingsTracingState {
 
     object BluetoothDisabled : SettingsTracingState() {
         override fun getTracingIconColor(c: Context): Int =
-            c.getColor(R.color.colorTextPrimary3)
+            c.getColor(R.color.colorTextSemanticRed)
```
mtwalli, line 18: Please rename `c`to `context`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/settings/start/SettingsTracingState.kt
```diff
@@ -15,11 +15,10 @@ sealed class SettingsTracingState {
...
 
-        // TODO Why are we using the active indicator for Bluetooth?
         override fun getTracingIcon(c: Context): Drawable? =
-            c.getDrawable(R.drawable.ic_settings_tracing_active_small)
+            c.getDrawable(R.drawable.ic_settings_tracing_bluetooth_inactive)
```
mtwalli, line 21: same here

mtwalli (dismissed):
LGTM, just renaming if you agree

ralfgehrer (approved):

mtwalli (approved):

PR #1999, opened 2020-12-23
Fix deadman notification text not being expandable and clean up code (EXPOSUREAPP-4251)

d4rken:
Setting the style to `BigTextStyle` to make the notification expandable and refactor duplicate code away.

## Testing
View the deadman notification on a device with a small screen, try to expand it, and click it.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt
```diff
@@ -37,6 +38,12 @@ class AndroidModule {
         @AppContext context: Context
     ): NotificationManagerCompat = NotificationManagerCompat.from(context)
 
+    @Provides
+    @Singleton
+    fun notificationManager(
+        @AppContext context: Context
+    ): NotificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+
```
mtwalli, line 46: There is a handy extension for that in `core-ktx`
`context.getSystemService< NotificationManager>`

mtwalli (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/AndroidModule.kt
ralfgehrer: I preferred the original solution without the `!!`

mtwalli: I understand, but it has the same result at the end. `getSystemService` returns a nullable `NotificationManager`casting it to a NonNullable type would throw a null exception pointer too.

mtwalli (approved):

ralfgehrer (approved):

PR #2004, opened 2020-12-28
Manual tan submission talkback error (EXPOSUREAPP-4099)

SamuraiKek:
This PR fixes a talkback issue where entering a full wrong TAN would make talkback skip one of the error messages.
This was happening because both textviews in the fragment were having the "accessibilityLiveRegion" property set to "assertive" and because the second textview only shows up when entering a full wrong TAN, it was hitching the focus from the other textview before that one got to be interpreted by talkback. 

Also, I had to change the order in which the two error message textviews are displayed because the invisible one that shows up after entering the full wrong TAN  will be last in the talkback stack when becoming visible. So even though in the UI, it comes in before the second textview, talkback was actually reading the second textview first and then the first one.

To test the fix, enable talkback in your settings and then enter a full wrong TAN inside the application.


mtwalli (approved):
LGTM

chiljamgossow (approved):

PR #2006, opened 2020-12-29
Contact Journal Day date is not announced (EXPOSUREAPP-4457)

axelherbstreith:
- Fixed issue that prevented the title of the day screen to be read by the screen reader
- Fixed issue where dates were read in the wrong format when using TalkBack


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
```diff
@@ -17,7 +18,9 @@ fun ViewPager2.registerOnPageChangeCallback(cb: (position: Int) -> Unit) {
 }
 
 // According to tech spec german locale only
-fun LocalDate.toFormattedDay(): String = toString("EEEE, dd.MM.yy", Locale.GERMAN)
+fun LocalDate.toFormattedDay(): String = toString("EEEE, dd.MM.yy", ConfigurationCompat.getLocales(Resources.getSystem().configuration)[0])
```
d4rken, line 21: Why are you using the systems `Configuration` object here and not our apps?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/ContactDiaryDayFragment.kt
```diff
@@ -83,4 +86,9 @@ class ContactDiaryDayFragment : Fragment(R.layout.contact_diary_day_fragment), A
             }
         }
     }
+
+    override fun onResume() {
+        super.onResume()
+        binding.contentContainer.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
+    }
```
chiljamgossow, line 93: What is the benefit of having two announcements?

axelherbstreith: removed the duplicated one

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
axelherbstreith: Reverted this change back to Locale.GERMAN since Localization of date format is already addressed in #2021 (https://github.com/corona-warn-app/cwa-app-android/pull/2021). To match accessibility, I will include the changes from #2021 once the PR is merged so that the dates are read in the correct language and format when using TalkBack.

d4rken (approved):
LGTM :+1:

chiljamgossow (approved):

PR #2008, opened 2020-12-29
Background/Async TEK submission (EXPOSUREAPP-3620)

d4rken:
This PR adds background submission of TEKs to increase chance of submission if the user "forgets" or in some other capacity allows submission but does not finish it.

Once consent is given, and TEKs are available there will be a periodic background worker "lurking", waiting for ~30min of user inactivity. Once 30min is reached, or if the user completes submission themselves, or if the app is restarted, submission is performed.

There are two time values to note, the 15min interval at which we can execute the submission worker, and the 30min of user inactivity that is required before we automatically submit the keys.

Interesting classes:
* [`AutoSubmission.kt`](https://github.com/corona-warn-app/cwa-app-android/pull/2008/files#diff-0f1b55ca0920e5b6101962e69d8b989c08129ebea0196a8a51b9d89e0d664d60)
* [`SubmissionTask.kt`](https://github.com/corona-warn-app/cwa-app-android/pull/2008/files#diff-7c8f249bf3704756dd510cab7ec704ec5af690d1c2633b2c020870f6964877d8)

### Testing
For testing the 30min inactivity timeout may be reduced to 15min for testing (see `SubmissionTask.USER_INACTIVITY_TIMEOUT`).

* Normal submission should work as before, direct and blocking, on submission success, home screen shows "thank you" card etc. (monitor logcat, there should be no more SubmissionTask executions.)
* Normal submission, but server throws error, user should land on home screen, home screen should show positive test result, when the worker runs the next time (every 15min) and user activity has timed out, keys should be submitted and home screen should show "thank you" card.
* Normal submission, server throws error, user should land on home screen, but user re-enters submission flow, automatic submission should be delayed again until 30min of inactivity.
* Device on desk, user in submission flow (e.g. Symptom input), nothing happening (e.g. AFK), after 30 minutes submission should happen anyways. User should stay in submission flow, and when submitting, no error should be shown (because submission already happened).
* User is in submission flow and has given consent, but app is put into background and then killed by the system (e.g. swipe it away), submission should happen in the background after the inactivity timeout is reached and when the background worker is executed the next time.
* User is in submission flow, has given consent, puts app in background (swipe app away), open it again, "positive test result" should be shown, but if the user does nothing, automatic submission should happen once the user inactivity timeout is reached and the background worker is executed the next time.
* Submission was attempted, but failed or user swiped app away, app is reopened, positive test result is still shown, but app will try to submit in the background, if the user goes back into the submission flow and removes consent, the task will execute, but not perform submission and just deactivate automatic submission.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarFragment.kt
```diff
@@ -117,4 +117,9 @@ class SubmissionSymptomCalendarFragment : Fragment(R.layout.fragment_submission_
             }
         }
     }
+
+    override fun onResume() {
+        super.onResume()
+        viewModel.onNewUserActivity()
```
ralfgehrer, line 123: I suppose we have to call the same in onResume of `SubmissionTestResultAvailableFragment` (and `SubmissionConsentFragment`) - see the dark blue box on Figma.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
```diff
@@ -0,0 +1,144 @@
...
+
+        val result = taskController.submitBlocking(
+            DefaultTaskRequest(
+                type = SubmissionTask::class,
+                arguments = SubmissionTask.Arguments(checkUserActivity = false),
+                originTag = TAG
+            )
+        )
+        if (result.isSuccessful) {
+            Timber.tag(TAG).i("Blocking submission was successful.")
```
ralfgehrer, line 74: call `isableAutoSubmission()`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/task/SubmissionTask.kt
```diff
@@ -0,0 +1,216 @@
...
+        checkCancel()
+
+        Timber.tag(TAG).d("Submitting %s", submissionData)
+        playbook.submit(submissionData)
+
+        Timber.tag(TAG).d("Submission successful, deleting submission data.")
+        tekHistoryStorage.clear()
+        submissionSettings.symptoms.update { null }
+
+        autoSubmission.updateMode(AutoSubmission.Mode.DISABLED)
```
ralfgehrer, line 144: In `AutoSubmission` you reset all those values
```
 submissionSettings.apply {
            autoSubmissionEnabled.update { false }
            lastSubmissionUserActivityUTC.update { Instant.EPOCH }
            autoSubmissionAttemptsCount.update { 0 }
            autoSubmissionAttemptsLast.update { Instant.EPOCH }
        }
```
Should we do the same here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
```diff
@@ -0,0 +1,144 @@
...
+
+@Singleton
+class AutoSubmission @Inject constructor(
+    private val timeStamper: TimeStamper,
+    private val submissionSettings: SubmissionSettings,
+    private val workManager: WorkManager,
+    private val taskController: TaskController
+) {
+
+    private var currentMode: Mode = Mode.DISABLED
```
ralfgehrer, line 32: `curentMode` isn't used (anymore), is it?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
```diff
@@ -0,0 +1,144 @@
...
+
+        // Data is available, but user may not be finished, submit on timeout
+        MONITOR,
+
+        // Data is available, assume user finished, try async submission ASAP, reset user activity timeout.
+        SUBMIT_ASAP
+    }
+
+    companion object {
+        private const val AUTOSUBMISSIO_WORKER_TAG = "AutoSubmissionWorker"
```
ralfgehrer, line 141: very minor: Typo in AutosubmissioN

ralfgehrer:
Very nice and readable code! Double checked all scenarios I could come up with (cold start, resuming the app, initial states, second submissions) from a coding POV. Device testing still ongoing.

ralfgehrer (dismissed):
Device test successful on a Pixel (Android 10).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
```diff
@@ -0,0 +1,144 @@
...
+            updateMode(Mode.SUBMIT_ASAP)
+        }
+    }
+
+    private fun scheduleWorker() {
+        Timber.tag(TAG).v("scheduleWorker(): Creating periodic worker request for submission.")
+
+        val request = PeriodicWorkRequestBuilder<SubmissionWorker>(15, TimeUnit.MINUTES).apply {
+            addTag(AUTOSUBMISSIO_WORKER_TAG)
+            setBackoffCriteria(BackoffPolicy.EXPONENTIAL, BackgroundConstants.BACKOFF_INITIAL_DELAY, TimeUnit.MINUTES)
```
BMItter, line 86: short question:
setting a BackOffPolicy here, does it have any effect? (worker doesn't use Result.retry() or something)
Not sure about that...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
```diff
@@ -0,0 +1,144 @@
...
+        } else {
+            Timber.tag(TAG).w(result.error, "Blocking submission was not successful, enabling auto submission.")
+            updateMode(Mode.SUBMIT_ASAP)
+        }
+    }
+
+    private fun scheduleWorker() {
+        Timber.tag(TAG).v("scheduleWorker(): Creating periodic worker request for submission.")
+
+        val request = PeriodicWorkRequestBuilder<SubmissionWorker>(15, TimeUnit.MINUTES).apply {
```
BMItter, line 84: 15 Minutes...
Do you think it could affect battery consumption?

BMItr (dismissed):
Code lgtm, tested on pixel 2 device. 
Little questions, but no urgent.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
d4rken: True

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
d4rken: Probably, but it's a short task. And there is really no alternative. I'd actually like to have it quicker, but 15 is the minimum. If we increase this, we raise the chance that the user does something to prevent us from submitting the keys ever.

e.g. get a call during submission, app in background, app is killed, user thinks "positive test :frowning:", forgets about the submission, maybe even uninstalls the app...

d4rken: The worker is only scheduled though once there is positive test result, so until that, there should be no impact.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/auto/AutoSubmission.kt
d4rken: I've changed the worker to return `retry()` on exceptions.

BMItter:
Please adjust _SubmissionWorkerTest_ at last 
Will reapprove then.
thx for good work ☕

BMItr (dismissed):

ralfgehrer (dismissed):

ralfgehrer (approved):

BMItr (approved):

PR #2015, opened 2020-12-30
Create Instrumentation tests & screenshots for the Onboarding flow (EXPOSUREAPP-4478)

mtwalli:
- Add instrumentation tests for Onboarding flow to capture screenshots.
- Tests can be filtered using Screenshot custom annotation to run exclusively.
For Testing:

1- Install fastlane
2- Assemble testing APK `./gradlew assembleDebug assembleAndroidTest`
3- Open emulator or a physical device (make sure it connected properly)
3- Run fastlane Screengrab `fastlane screengrab`

after that Fastlane should start pushing APK and run tests that capture screenshots


fastlane/Screengrabfile
```diff
@@ -0,0 +1,8 @@
+locales ['de-DE', 'en-US', 'tr-TR', 'bg-BG', 'pl-PL', 'ro-RO']
+use_adb_root true
+clear_previous_screenshots true
+app_package_name 'de.rki.coronawarnapp.test'
+launch_arguments ["annotation testhelpers.Screenshot"]
+test_instrumentation_runner 'testhelpers.TestApplicationUIRunner'
+app_apk_path './Corona-Warn-App/build/outputs/apk/deviceForTesters/debug/Corona-Warn-App-1.11.0-RC1-deviceForTesters-debug.apk'
```
d4rken, line 7: Hm, this will break on newer versions

Server-Protocol-Buffer/build.gradle
```diff
@@ -5,6 +5,9 @@ sourceSets {
     main.java.srcDirs += "${protobuf.generatedFilesBaseDir}/main/javalite"
 }
 
+sourceCompatibility = "8"
```
d4rken, line 8: Maybe

```
        sourceCompatibility JavaVersion.VERSION_1_8
        targetCompatibility JavaVersion.VERSION_1_8
```

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/onboarding/OnboardingDeltaInteroperabilityFragmentTest.kt
```diff
@@ -35,6 +55,15 @@ class OnboardingDeltaInteroperabilityFragmentTest : BaseUITest() {
     fun launch_fragment() {
         launchFragment<OnboardingDeltaInteroperabilityFragment>()
     }
+
+    @Screenshot
+    @Test
+    fun capture_screenshot() {
+        launchFragmentInContainer<OnboardingDeltaInteroperabilityFragment>(themeResId = R.style.AppTheme)
```
d4rken, line 62: Should we introduce a parent abstract class that sets this? Or maybe an extension function with default arguments?

When would we want to use a different theme?

d4rken:
The usb debugging notification is always visible for, any idea how to fix that?

[OnboardingDeltaInteroperabilityFragment_1609763759874](https://user-images.githubusercontent.com/1439229/103536116-323d4500-4e92-11eb-8fa7-b72caa91a3ab.png)

d4rken:
Otherwise :ok_hand: , screenshots work nicely.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/onboarding/OnboardingDeltaInteroperabilityFragmentTest.kt
mtwalli: I introduced  a new inline function `launchFragmentInContainer2` defaults to `AppTheme`. We need to specify the AppTheme here to avoid showing: 1- ActionBar 2- Squeezing some UI elements. I mentioned also this in the docs for the function mentioned above

mtwalli:
> The usb debugging notification is always visible for, any idea how to fix that?

I tried cleaning it as in here https://docs.fastlane.tools/actions/screengrab/#clean-status-bar . 
Unfortunately, it did not clear it. in the emulator, I was able to disable it manually which is not the case for physical devices

d4rken (dismissed):
Nice :sunglasses:

chiljamgossow:
I ran it on a Xiaomi device and had an abort:
[11:07:26]: ▸ de.rki.coronawarnapp.ui.onboarding.OnboardingDeltaInteroperabilityFragmentTest:
[11:07:26]: ▸ Error in capture_screenshot(de.rki.coronawarnapp.ui.onboarding.OnboardingDeltaInteroperabilityFragmentTest):
[11:07:26]: ▸ java.lang.AssertionError: Activity never becomes requested state "[DESTROYED, STARTED, RESUMED, CREATED]" (last lifecycle transition = "PRE_ON_CREATE")

jurajkusnier:
> The usb debugging notification is always visible for, any idea how to fix that?
> 
> [OnboardingDeltaInteroperabilityFragment_1609763759874](https://user-images.githubusercontent.com/1439229/103536116-323d4500-4e92-11eb-8fa7-b72caa91a3ab.png)

It's not visible in "Do Not Disturb" mode.

mtwalli:
@chiljamgossow No Idea why is this happening on your device. According to this 
https://forums.raywenderlich.com/t/problem-with-testing-fragments-in-androidx/58035
the Gradle dep is fine and it should work. Could you please post the full stack and what configurations do you have on the device?

d4rken:
> It's not visible in "Do Not Disturb" mode.

Can this be set from test or fastlane?

jurajkusnier:
> > It's not visible in "Do Not Disturb" mode.
> 
> Can this be set from test or fastlane?

This code works
```kotlin
val  notificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
notificationManager.setInterruptionFilter(NotificationManager.INTERRUPTION_FILTER_NONE)
```
but only with `android.permission.ACCESS_NOTIFICATION_POLICY` permission and user have to grand to Do Not Disturb Access manually via settings > apps & notifications > special app access > Do Not Disturb access
I don't know of other options

jurajkusnier (approved):
I had a problem to make it run on some devices but it's working reliably on emulator

d4rken (approved):
Nice solution with the `SystemUIDemoMode`

d4rken:
> I had a problem to make it run on some devices but it's working reliably on emulator

If it works on the emulators / Pixel devices, it would be okay to merge for now. Additional fixes in an extra PR.

It would be nice to have this PR merged for working on running it in Firebase Testlab via CircleCI.
(Once that is setup we may need to do a few more tweaks too)

mtwalli:
> > > It's not visible in "Do Not Disturb" mode.
> > 
> > 
> > Can this be set from test or fastlane?
> 
> This code works
> 
> ```kotlin
> val  notificationManager = getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
> notificationManager.setInterruptionFilter(NotificationManager.INTERRUPTION_FILTER_NONE)
> ```
> 
> but only with `android.permission.ACCESS_NOTIFICATION_POLICY` permission and user have to grand to Do Not Disturb Access manually via settings > apps & notifications > special app access > Do Not Disturb access
> I don't know of other options

Unfortunately, this solution did not work for me. Despite that, I grant the permission using `GrantPermissionRule`. it did throw this Security Exception. `java.lang.SecurityException: Notification policy access denied`

jurajkusnier (approved):

PR #2019, opened 2020-12-31
Better handling of issues caused by time traveling (RiskResults) (DEV)

d4rken:
Change primary key for risk result database to a monotonic ID.

The `RiskLevelTask` calculates a `RiskResult`, that's stored in a database.
The UI uses the "latest result" and "latest successful result" when displaying the risk/tracing card on the home screen.
If the device travels into the future, calculates a result, and then returns back to "our" time, then the risk card always has a "newer" result until we have caught up to the future date at which the result was calculated.

This is also likely not the cause for #2001, but looking deeper into that issue caused me to think it's nice to have this additional resilience against time-traveling issues. This also makes it easier for developers/users/testers to debug issues by changing the devices clock, this was previously possible before the move to ENFv2.

The PR migrates the database by creating a new table that has the autoincrementing field `monotonicId` and then moving the old data into it. The SQL queries are adjusted to sort my `monotonicId` instead of `calculatedAt`.

## Testing
* Use the app without this PR, fake a few different risk results through the test menu
* Update to this PR's version
* Everything should look the same

* Use this app without this PR, time travel and generate a risk result in the future
* Return to normal time, notice that any risk result changes (e.g through the test menu), will not replace the result (card state) from the future
* Upgrade to this PR's version, the risk result card on the home screen should now show the last calculated result (which is from your time now, and not the one from time traveling)


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/risk/storage/RiskResultDatabaseMigrationTest.kt
```diff
@@ -0,0 +1,117 @@
...
+            2,
+            true,
+            RiskResultDatabaseMigrations.MIGRATION_1_2
+        )
+
+        val daoDb = RiskResultDatabase.Factory(
+            context = ApplicationProvider.getApplicationContext()
+        ).create(databaseName = DB_NAME)
+
+        val allEntries = runBlocking { daoDb.riskResults().allEntries().first() }
```
mtwalli, line 69: `runBlockingTest`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/internal/RiskResultDatabase.kt
```diff
@@ -81,11 +82,12 @@ abstract class RiskResultDatabase : RoomDatabase() {
...
 
-        fun create(): RiskResultDatabase {
+        fun create(databaseName: String = DATABASE_NAME): RiskResultDatabase {
             Timber.d("Instantiating risk result database.")
             return Room
-                .databaseBuilder(context, RiskResultDatabase::class.java, DATABASE_NAME)
-                .fallbackToDestructiveMigrationFrom()
+                .databaseBuilder(context, RiskResultDatabase::class.java, databaseName)
+                .addMigrations(RiskResultDatabaseMigrations.MIGRATION_1_2)
+                .fallbackToDestructiveMigrationFrom(3)
```
mtwalli, line 90: How this will behave as I did not see a schema with version 3? 
would `fallbackToDestructiveMigration()` be an option in this case?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/risk/storage/RiskResultDatabaseMigrationTest.kt
d4rken: What would be the effective difference here?

mtwalli: I think it executes the test body on an immediate dispatcher, especially if there are `delay` or `async` blocks inside.

d4rken: I'd like to leave it on `runBlocking` as we don't need any `delay` handling here and `runBlockingTest` returns `Unit` while `runBlocking` returns `T` (last type in scope) which makes the test more readable.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/internal/RiskResultDatabase.kt
d4rken: I've removed it as it was not needed here, thx!

d4rken:
I've added exception handling to the migration that drops the table and creates the v2 schema.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/risk/storage/RiskResultDatabaseMigrationTest.kt
mtwalli: Sure

mtwalli (approved):
LGTM

BMItr (approved):

Issue #1958, opened 2020-12-19
Transnational country list names displayed in English when system is German


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Samsung Galaxy A50 SM-A505FN
- Android version: 10
- App version: 1.9.1

## Describe the bug

The country names are shown in English when the system language is set to Deutsch (Deutschland).
This is a regression from version 1.7.1 (see screenshots from previous version in https://github.com/corona-warn-app/cwa-app-android/issues/1872#issuecomment-743124411)

## Steps to reproduce the issue

1. Set Android system language to Deutsch (Deutschland)
2. Tap RISIKO-ERMITTLUNG AKTIV
3. Tap Länderübergreifende Risiko-Ermittlung
4. Scroll down to heading "Derzeit nehmen die folgenden Länder an der länderübergreifenden Risiko-Ermittlung teil:"

![Country_list_1 9 1_de](https://user-images.githubusercontent.com/66998419/102689485-c5be1700-41fe-11eb-9ada-2222266c7de9.jpg)

## Expected behaviour

The names of the countries should be shown in German when the system language is set to German.

## Possible Fix

Use German country names when the UI is set to German and sort the names according to their German names.

## Additional context

See also https://github.com/corona-warn-app/cwa-app-android/issues/1957 regarding sorting.

---
Internal Tracking ID: [EXPOSUREAPP-4416](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4416)

dsarkar:
ok, thanks, in iOS is not an issue. Thanks, DS

Jo-Achim:
Android 10 with Android security update: November 1st, 2020
Nokia 7.1, Build: 00WW_4_16B_SP01
CWA version: 1.9.1
ENF: 18204713000

No problems here:

![CWA 1 9 1 - Country-Info](https://user-images.githubusercontent.com/21124877/102690124-7f1eeb80-4203-11eb-82a0-418a89333659.jpg)

Best regards.

dsarkar:
@Jo-Achim, ok! Interesting. Thanks for the information, will be forwarded to devs. DS

[...]

PR #2021, opened 2021-01-01
Localize dates in contact diary (EXPOSUREAPP-4485) (closes #2009)

### LINKED ISSUE #2009: Contact Journal dates in German with English locale
## Avoid duplicates

## Technical details

- Device name: Android Pixel 3a emulator
- Android version: 11
- App version: 1.10.1

## Describe the bug

The dates in the Contact Journal appear in German when the device is set to English.

1. Settings > System > Languages and input > Languages
2. Set or add English (United Kingdom) as first language in list.
3. Open CWA
4. Scroll down and tap "MANAGE JOURNAL"
5. Note that the names of the days are German and the date format is German.

image

## Expected behaviour

When the UI is displaying text in English, the dates in the Contact Journal should also be shown in English.

## Additional context

There are some other localisation issues with the Contact Journal, which I will put into separate issues.

---
Internal Tracking ID: [EXPOSUREAPP-4485](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4485)

PhilippNowak96:
### Description

As shown in #2009 dates in the contact diary are not localized. The fix addresses this issue for the contact diary overview as well as for the header of the single day view. The exported data stays the same as before.

Please note: Some languages do not support short forms of the year and therefore the used `shortDate()` method returns the full year. I wrote a unit test for every currently supported language (and additionally covered US and UK for english) of the CWA to make the differences clear and cover all the cases.

#### Before fix (with en-US)
4485_english_old

#### After fix (with en-US)
4485_english

#### After fix (with de-DE)
4485_german

### Steps to reproduce

(Taken from #2009):

1. Settings > System > Languages and input > Languages
2. Set or add English (United Kingdom) as first language in list.
3. Open CWA
4. Scroll down and tap "MANAGE JOURNAL"
5. Note that the names of the days are German and the date format is German.


PhilippNowak96:
Looks like the new unit tests do not run correctly on CircleCI. Can anyone with access rights provide a stacktrace? On my local machine everything is running fine.

dsarkar:
@PhilippNowak96 Thanks again .... Forwarded info about PR to devs. DS

---

axelherbstreith (commented):
Nice one! Tested it on Huawei P30 and Nokia 7.1 (both Android 10) and it works as expected, the weekdays and date format are both adapted to the system locale while the export function remains in german format. 

However, while testing with English language (UK), I noticed that on my physical devices the year is displayed completely e.g. 01/01/2021 instead of 01/01/21. When testing on virtual devices, the date was displayed correctly. 

overview_screen
day_screen

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryExtensions.kt
```diff
@@ -16,8 +17,12 @@ fun ViewPager2.registerOnPageChangeCallback(cb: (position: Int) -> Unit) {
     })
 }
 
-// According to tech spec german locale only
-fun LocalDate.toFormattedDay(): String = toString("EEEE, dd.MM.yy", Locale.GERMAN)
+fun LocalDate.toFormattedDay(): String {
+    // Use two different methods to get the final date format (Weekday, Shortdate)
+    // because the custom pattern of toString() does not localize characters like "/" or "."
+    return "${toString("EEEE", Locale.getDefault())}, " +
+        DateTimeFormat.shortDate().withLocale(Locale.getDefault()).print(this)
```
d4rken, line 24: Locale via context instead of static?

d4rken: Align with #2006 which also changes this:

https://github.com/corona-warn-app/cwa-app-android/pull/2006/files#diff-8961af63058c902361512c62b3c1a1764c30f1158d977c682c7e9ca244273c87R21

axelherbstreith: @PhilippNowak96 feel free to tackle this change if you want to :) #2006 also addresses the same changes, but I only included it as a small fix (also not 100% finished yet) which is not directly related to #2006. I will exclude the changes from my PR for now.

PhilippNowak96: @axelherbstreith Okay, I will have a look at it later in the evening :)

BMItter: short Note:
according to the current version of the contact diary tech spec, it has to be: withLocale(Locale.Germany...).
"Texts and locale for e.g. date formatting are in German only. "
This is something @PhilippNowak96 you probably can't know in lack of access. :pensive:
I will ask the responsible person for an exception...

BMItter: short Note 2: 
After talk back with our architect, its can stay in this way (locale.getDefault) :blush:
locale.germany only for the export function.
So nothing to complain here from my side

PhilippNowak96: @d4rken Good point, thank you! I replaced the static Locale.getDefault() by the already existing SystemInfoProvider and made it using the provided context instead of the system resources. I struggled a little bit with all that Dagger stuff because I didn't deep dive into how you use Dagger and everything related to it yet but I hope I get everything. If anything looks wrong, please let me know! 😃 

@BMItter Thanks for clarifying this. According to #2009 at least iOS already had the "after-fix"-behavior, or do they use a different tech spec? 🤔 Anyways, nice that we can fix this for Android! 😄 

@axelherbstreith Maybe you can adjust your implementation for the accessibility part to also match the visible one and benefit from using the selected locale.

PhilippNowak96:
@axelherbstreith I have the full 4 digit year on a virtual device (API 29) with en-UK selected. Interesting that you get the short year on an AVD 🤔 Not sure if we need to put more effort into this or just rely on the implementation of `shortDate()`?

axelherbstreith:
@PhilippNowak96 My bad, I checked the language settings on the virtual device again and they were set to en-US, thought I put them on en-GB. So the 4 digit year occurs on both my physical and virtual devices (only) when setting the locale to en-GB. Sorry for the confusion. 
To align with the contact diary requirements it is needed to always have a two digit year afaik but I will clarify this tomorrow.

PhilippNowak96:
@axelherbstreith Also see the unit tests, there are more locales using the full year (not sure if it maybe isn't common to use 2-digit years in these countries?):

- UK
- Poland
- Romania
- Turkey

axelherbstreith:
@PhilippNowak96  simpleDate() is fine! Your mentioned countries don't support a 2 digit year officially (according to Wikipedia https://en.wikipedia.org/wiki/Date_format_by_country), I guess that's why simpleDate always displays them with 4 digits. I clarified it with the team and simpleDate() can be used without any adaptions from your side 👍 

Thanks for your effort :)

axelherbstreith (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/DefaultSystemInfoProvider.kt
```diff
@@ -1,19 +1,23 @@
...
-import android.content.res.Resources
 import android.os.Build
+import de.rki.coronawarnapp.util.di.AppContext
 import java.util.Locale
 import javax.inject.Inject
+import javax.inject.Singleton
 
-class DefaultSystemInfoProvider @Inject constructor(context: Context) : SystemInfoProvider {
+@Singleton
+class DefaultSystemInfoProvider @Inject constructor(
```
d4rken, line 11: I would prefer an extension function here.

Something like

```kotlin
val Context.locale ...
// or
fun Context.getLocale
```

the system info provider is IMHO overkill and also tied to the application context while `context.resources` is affected by theming  and configuration changes, so it could differ from activity to activity.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/contactdiary/util/ContactDiaryDateFormatterExtensionTest.kt
```diff
@@ -0,0 +1,90 @@
...
+        clearAllMocks()
+    }
+
+    @Test
+    fun `format bulgarian date`() {
+        every { defaultSystemInfoProvider.locale } returns Locale("bg", "BG")
+
+        val testDate = LocalDate("2021-01-01")
+
+        Assert.assertEquals("петък, 1.01.21 г.", testDate.toFormattedDay(defaultSystemInfoProvider))
```
d4rken, line 34: Please use the kotest matchers as in other tests ( e.g. `shouldBe`).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/ContactDiaryDayViewModel.kt
```diff
@@ -15,14 +16,15 @@ import org.joda.time.LocalDate
...
+    systemInfoProvider: DefaultSystemInfoProvider,
     @Assisted selectedDay: String
 ) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
     private val displayedDay = MutableStateFlow(LocalDate.parse(selectedDay))
 
     val routeToScreen: SingleLiveEvent<ContactDiaryDayNavigationEvents> = SingleLiveEvent()
 
     val uiState = displayedDay.map { day ->
-        UIState(dayText = day.toFormattedDay())
+        UIState(dayText = day.toFormattedDay(systemInfoProvider))
```
d4rken, line 27: You can use this pattern here or create a variant of it to resolve the string with a context shortly before displaying it:

https://github.com/corona-warn-app/cwa-app-android/blob/349c7072282dcce3cd87bab1b1224146ba9aa87b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/LazyString.kt#L5-L7

d4rken (changes requested):
Formatting looks good but would like a little refactoring on the locale access.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/ContactDiaryDayViewModel.kt
PhilippNowak96: I'm not sure if I got your point here. You want to put a context into the ViewModel?

d4rken: Ideally we want to use the view's context to resolve the date. As close to the place where we display the date as possible.

* We could inject the application context into the viewmodel, but that's the application context again, same issue as previously.
* We could use assisted inject to make the fragment's context accessible from within the viewmodel but that's not a nice separation of concerns.
* We could change the `UIState.dayText` into a function that takes a context and returns a formatted day.
    * Could do that by just creating a lambda, e.g. `dayText: (Context) -> String`
    * Or if this is something that would be used else were too, created a new implementation of the interface `LazyString` (see other implementations in that file).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/DefaultSystemInfoProvider.kt
PhilippNowak96: Fair point, done.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/ContactDiaryDayViewModel.kt
PhilippNowak96: Thank you for the detailed explanation! I decided to use the lambda because I was afraid of introducing side effects with `LazyString` like in #1958 and I think it does not make a big difference at this point. Also reverted all the SystemInfoProvider stuff (the cleanup of its current usages should be done in a separate PR I guess to not mix up too much in this one).

d4rken (approved):
lgtm

axelherbstreith (approved):

PR #2023, opened 2021-01-01
Removed Kotlin Android Extensions plugin (EXPOSUREAPP-4495)(closes #2022)

### LINKED ISSUE #2022: Remove old Kotlin Android Extensions plugin
## Avoid duplicates
* [x] Enhancement request is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)
* [x] If you are proposing a new feature, please do so in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)

## Current Implementation
The CWA currently uses the Kotlin Android Extensions plugin at some places for Kotlin synthetics and for Kotlin Parcelize.

## Suggested Enhancement
Due to the deprecation of the Kotlin Android Extensions plugin it should be removed. The usages of Kotlin synthetics can be replaced by the already introduced Jetpack view binding (usages in custom views by findViewById respectively) and for the Kotlin Parcelize part there is a standalone plugin now. See: https://developer.android.com/topic/libraries/view-binding/migration

## Expected Benefits
Future proof implementation and unification of view binding technologies (KS was only used at some very few places).

## Additional notes
I will open a PR for this soon and link it here.

PR: #2023 

---
Internal Tracking ID: EXPOSUREAPP-4495

PhilippNowak96:
### Description

The Kotlin Android Extensions plugin is deprecated and should be removed for the future. Because the CWA uses synthetics only at some very few spots, which can be replaced by already used Jetpack view binding, and the Parcelize feature has now an own standalone plugin this was no big deal.

For further information see: https://developer.android.com/topic/libraries/view-binding/migration and #2022 

---
Internal Tracking ID: EXPOSUREAPP-4495


dsarkar:
@PhilippNowak96 Thanks, devs are informed about PR. DS

---

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/TestResultSectionView.kt
```diff
@@ -26,30 +27,39 @@ constructor(
...
+    private lateinit var testResultSectionStatusIcon: ImageView
 
     init {
         inflate(context, R.layout.view_test_result_section, this)
         context.withStyledAttributes(attrs, R.styleable.TestResultSection) {
-            test_result_section_headline.text =
+            testResultSectionHeadline = findViewById(R.id.test_result_section_headline)
+            testResultSectionContent = findViewById(R.id.test_result_section_content)
+            testResultSectionRegisteredAt = findViewById(R.id.test_result_section_registered_at_text)
+            testResultSectionStatusIcon = findViewById(R.id.test_result_section_status_icon)
```
mtwalli, line 41: `TestResultSectionViewBinding.bind(this)`might be helpful here to avoid this boilerplate code

PhilippNowak96: I was not sure if view binding works the same way for custom views. Thanks for this! :) I will update it in the afternoon.

PhilippNowak96: Adjusted it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/TestResultSectionView.kt
```diff
@@ -26,30 +27,32 @@ constructor(
     attrs: AttributeSet? = null,
     defStyleAttr: Int = 0
 ) : ConstraintLayout(context, attrs, defStyleAttr) {
+    private val binding = ViewTestResultSectionBinding.inflate(LayoutInflater.from(context), this)
```
mtwalli, line 30: I'm afraid that `inflate` is inflating another view that is not attached to the current layout. Please use `bind(this)`
instead.

PhilippNowak96: `CircleProgress` for example is using `inflate` too. But I can change it later

PhilippNowak96: Done

mtwalli (approved):
Thanks! LGTM

d4rken (approved):
:+1:

PR #2024, opened 2021-01-01
Reset the initial registration date on test removal correctly (EXPOSUREAPP-4484) (closes #2005)

### LINKED ISSUE #2005: test registration date remains from deleted test after adding a new test
## Avoid duplicates
* [ ] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [ ] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ ] Bug is not already reported in another issue

## Technical details

- Device name: Samsung Galaxy S10
- Android version: 10
- App version: 1.10.1

1. I added a test via QR Code on Dec 25. Test result page correctly showed "Registered on 12/25/20" (Test returned negative on Dec 26)
2. Today (Dec 29) I deleted the old test to add a new one.
3. Afterwards (Dec 29), I registered that new test via QR Code. The result page still shows "Registered on 12/25/20"

tempFileForShare_20201229-103344

## Expected behaviour

Test registration date should correctly be 29 Dec 2020, not the registration date of the last test.

## Additional context

---
Internal Tracking ID: [EXPOSUREAPP-4484](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4484)

PhilippNowak96:
### Description

As reported in #2005 the initial registration date of a test stays the same even after deleting the test and adding a new one on another day. This fix cares about the correct reset of the initial registration date when removing a test.

### Steps to reproduce

1. Add a test via TAN or QR code and see the "Registered on" date
2. Delete the test manually
3. Add a second test one day later (so the dates are not the same)
4. See that the "Registered on" date matches the one of the first added test

### Additional notes

I was not sure if I am able to test that in some way but I guess the code speaks for itself. The reset of that date was just missing. I also adjusted the unit test and added a missing whitespace in the `CwaSecurityException` which I encountered while adjusting the unit test.

---
Internal Tracking ID: EXPOSUREAPP-4484


dsarkar:
@PhilippNowak96 Thanks and a Happy New Year! Forwarded this to the developers. Best, DS. 

---

crcsn:
Will this also take care of those users who already removed their test result in the current (or prior) version of the app or will they continue to experience the bug when adding a new test? 

And while I'm here: Happy New Year, everyone!

PhilippNowak96:
@crcsn Good point! This fix won't tackle the issue until users remove a test again (I just added the correct resetting when removing a test). I'm not sure if there is a chance to fix this, especially with a new test added (otherwise we could check for the other values being reset and then reset this specific timestamp too).

ralfgehrer (approved):
Can be tested and validated by time-traveling into the future (change the date settings in Android) before registration of a second test. Thanks for the fix @PhilippNowak96.

d4rken (approved):
LGTM

PR #2029, opened 2021-01-05
Add debug log feature for non-tester builds (EXPOSUREAPP-4451)

d4rken:
This adds the option for users to record debug logs in production builds.
To help investigate issues, fix bugs, or even if you are just curious about the app.

New entry under: `Menu > App Information > Error report`
The entry opens a new screen with a short explanation and controls to start/stop a debug log and share it.

* Sharing is manual and can happen through the native "Share" dialog through an app of your choice that supports sharing .zip file (e.g. email)
* Controls:
    * Pressing "start" will plant a `DebugLogTree` that logs all statements in a buffered `Flow`
    * `DebugLogger` will subscribe to this and write each line to a file
    * Before each line is written there is the chance for modules of type `BugCensor` to anonymize certain values
    * Pressing "delete" will stop and delete the current recording. It's not possible to stop recording without deleting the logs.
    * While logging active you can share the current log file. It's zipped and exposed via URI from a specific subdirectory `cache/debuglog/shared`. Sharing again will delete the previously shared file. Stopping the recording will also delete the zip file.
* The implementation differs a bit from the rest of the app (injection/singletons) to allow for more useful debug logs
   * To be able to trigger recordings across app restarts and to be independent of other app logic, we save the desired recording state with a trigger file, and we check for it's existence
   * To be able to record bugs caused by injeciton, early in the apps lifecycle, `DebugLogger` is an `object` and setup before we setup injection. We need access to some injected classes though for allowing value censoring, so logging will be buffered for a short while and only written to disk once Dagger is initialized

### TODO
- [x] Tests

### Testing
* See package `de.rki.coronawarnapp.bugreporting.debuglog`
* Test app kills, device restarts
* Check logout on release builds
* Check log sharing
* Check log deletion on stop


Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -733,6 +733,27 @@
     <!-- XACT: describes illustration -->
     <string name="information_legal_illustration_description">"Eine Hand hält ein Smartphone mit viel Text, daneben ist ein Paragraphenzeichen als Symbol für das Impressum."</string>
 
+    <!-- XHED: Headline for debug log screen -->
+    <string name="debugging_debuglog_title">"Fehlerbericht"</string>
+    <!-- YTXT: Description for the debug option to record log files -->
+    <string name="debugging_debuglog_intro_explanation">"Durch diese Option wird das App Verhalten in einer Text-Datei protokolliert. Ist die Option aktiviert bevor ein Fehler auftritt, können Sie diesen Bericht zur Verfügung stellen um die Entwickler bei der Problemlösung zu unterstützen.\nEine dauerhafte Aktivierung dieser Option führt zu hohem Speicherbedarf. Der Fehlerbericht wird durch deaktivieren der Option gelöscht."</string>
```
PhilippNowak96, line 739: - `zur Verfügung Stellen, um die Entwickler` (Comma)
- Not 100% sure, but I guess `deaktivierung` should also be `Deaktivierung` (Substantivierung)

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -733,6 +733,27 @@
...
+    <!-- YTXT: Warning regarding downsides of recording a log file -->
+    <string name="debugging_debuglog_intro_warning">"Bitte beachten Sie das in Fehlerberichten Information über Risikoberechnung und Testergebnis enthalten sein können, aus diesem Grund sollten Sie Fehlerberichte nicht öffentlich teilen."</string>
```
PhilippNowak96, line 741: - `Bitte beachten Sie, dass`
- `enthalten sein können. Aus diesem Grund`

Corona-Warn-App/src/main/res/layout/bugreporting_debuglog_fragment.xml
```diff
@@ -0,0 +1,114 @@
...
+                    android:visibility="gone"
+                    app:layout_constraintBottom_toBottomOf="@+id/debuglog_status_secondary"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toTopOf="@+id/debuglog_status_primary"
+                    tools:visibility="visible" />
+
+                <TextView
+                    android:id="@+id/debuglog_status_primary"
+                    style="@style/body1"
+                    android:layout_width="0dp"
```
PhilippNowak96, line 66: Some layouts use `@dimen/match_constraint` here. Is this still a thing? Then it should be used here and for the following views too.

Corona-Warn-App/src/deviceForTesters/AndroidManifest.xml
```diff
@@ -1,21 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
-<manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
+<manifest xmlns:tools="http://schemas.android.com/tools"
```
PhilippNowak96, line 2: This flavored manifest seems to be needless after moving out the provider

ralfgehrer:
The deletion seems to not delete the entire log. Some non-critical information is kept.

Screenshot 2021-01-06 at 10 11 08

The file size is not 0B (as expected) but contains the following content:
`2021-01-07T08:51:18.400Z  I/DebugLogger: stop()`

Corona-Warn-App/src/main/res/values-de/strings.xml
d4rken: :+1:

Corona-Warn-App/src/main/res/values-de/strings.xml
d4rken: :+1:

Corona-Warn-App/src/main/res/layout/bugreporting_debuglog_fragment.xml
d4rken: While we are still discussing use of `spacing_*`, I'm all for removing `@dimen/match_constraint`.
Neither plays well with the constraint layout editor.

I'll follow up with a PR that removes `@dimen/match_constraint` from the project.

Corona-Warn-App/src/deviceForTesters/AndroidManifest.xml
d4rken: Good point. Removed :+1:

d4rken:
@ralfgehrer  Can you still reproduce it?

This should fix it:

https://github.com/corona-warn-app/cwa-app-android/pull/2029/commits/a301e0b8892c8c9e679f0354814a30f1447e691c

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/sharing/FileSharing.kt
```diff
@@ -0,0 +1,49 @@
...
+        path
+    )
+
+    fun getIntent(
+        path: File,
+        title: String,
+        body: String? = null
+    ): Intent = Intent(Intent.ACTION_SEND).apply {
+        type = determineMimeType(path)
+        addFlags(Intent.FLAG_GRANT_READ_URI_PERMISSION)
```
ralfgehrer, line 31: In the contact journal, we used the `ShareCompat.IntentBuilder` instead. See `ContactDiaryOverviewFragment.exportLocationsAndPersons`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -0,0 +1,161 @@
...
+        logTree?.let { Timber.uproot(it) }
+
+        DebugLogTree().apply {
+            Timber.plant(this)
+            logTree = this
+
+            if (!runningLog.exists()) {
+                runningLog.parentFile?.mkdirs()
+                if (runningLog.createNewFile()) {
+                    Timber.tag(TAG).i("Log file didn't exist and was created.")
```
ralfgehrer, line 83: Would be cool if you were to log the 
- app version
- ENF version
- device details, such as Android version, device type, manufacturer, ... 

at the beginning of each log.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -0,0 +1,161 @@
...
+                    Timber.tag(TAG).i("Logging was canceled.")
+                } catch (e: Exception) {
+                    Log.e(TAG, "Failed to call appendLogLine(...)", e)
+                }
+            }
+        }
+
+        if (!triggerFile.exists()) {
+            Timber.tag(TAG).i("Trigger file created.")
+            triggerFile.createNewFile()
```
ralfgehrer, line 106: Shouldn't we catch the exception?

d4rken: Good point. Especially the init on app start.
I've changed it to catch the exception on app init, and to display the exception if it happens on the debug log screen.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
d4rken: Device infos are already logged when the app process is started.
I've changed it to also log them again when the debug log is started manually, otherwise they would only be visible after the next app start.
I've also added the ENF version there, which we do not print on app process start due to being more expensive and prone to side effects.

ralfgehrer (dismissed):
Tested on Pixel (Android 10). 
<3 Love it.

ralfgehrer (approved):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/LogLine.kt
```diff
@@ -0,0 +1,28 @@
...
+
+data class LogLine(
+    val timestamp: Long,
+    val priority: Int,
+    val tag: String?,
+    val message: String,
+    val throwable: Throwable?
+) {
+
+    fun format(context: Context): String {
```
BMItter, line 15: context is not required here :snowman:

BMItr (approved):
Code LGTM, first Device Test passed (Pixel 2). (still doing some lookup)
very helpful :baby_chick:
if a little percentage of the people use this feature, let's say a million, and they want to share their zips... Maybe we should put in your EmailAdress by default? :wink:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/ui/DebugLogViewModel.kt
```diff
@@ -0,0 +1,115 @@
...
+import java.io.File
+
+class DebugLogViewModel @AssistedInject constructor(
+    private val debugLogger: DebugLogger,
+    dispatcherProvider: DispatcherProvider,
+    private val timeStamper: TimeStamper,
+    private val fileSharing: FileSharing,
+    private val enfClient: ENFClient
+) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
+    private val ticker = flow {
```
BMItter, line 32: there is a nice tickerchannel in kotlin. has some little benefits.. (highly optional but :cherries:)
have a look here:
https://kotlinlang.org/docs/reference/coroutines/channels.html#ticker-channels

d4rken: I tried that but it seems this is deprecated/experimental?
I got warnings and would have had to annotate using this.
I'll look into it again.

vaubaehn:
Hi @d4rken ,
I'm a little bit confused - shouldn't that feature go to the production builds? I don't have any menu entry 😳

PhilippNowak96:
@vaubaehn looks like it is not yet clarified when to release this. The visibility of the entry is set to gone for none tester builds in `release/1.11`.

https://github.com/corona-warn-app/cwa-app-android/blob/a204c6ada1d05ba26ca54c341e69f4ff416f2ca8/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/information/InformationFragment.kt#L60-L61

d4rken:
@vaubaehn We are sadly not allowed to release this yet. There are concerns due to  "health data" being shared.

vaubaehn:
Hi @PhilippNowak96 and @d4rken ,

thanks a lot for the information, pointer to the code and explanation.
In my profession I'm also handling with health data, and data privacy is an ongoing big issue. From that perspective I have an idea of what is blocking the release of that feature.
From a user perspective I'm quite sad that this feature couldn't be released yet...

But I'm staying optimistic that debug logging may be released at a later time, maybe after some nescessary adjustments.
Best wishes to you all, stay healthy!

@PhilippNowak96 : I found many constructive contributions from you to the project - as a simple user I'm very thankful for your engagement :heart:

PhilippNowak96:
@vaubaehn Thank you for your kind words! 
I think especially from a developer perspective this feature could help in some situations so much. But of course I also understand why this might be critical to release and needs to be discussed a lot (at least from a German point of view 😅 😁)

PR #2034, opened 2021-01-06
Deletes wrong Romanian translation (EXPOSUREAPP-4310)

chiljamgossow:
Delete wrong translation to trigger new translation


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -9,6 +9,7 @@ import android.widget.Toast
 import androidx.appcompat.app.AppCompatActivity
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentManager
+import androidx.lifecycle.observe
```
ralfgehrer, line 12: needed?

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
chiljamgossow: Yes, AS complains

PR #2036, opened 2021-01-06
Fix calendar calculation tests for non-English locales (COMMUNITY)

fynngodau:
On a non-English system, some tests in `CalendarCalculationTest` are failing because the expected result is localized, for example like this:

```
Expected :"December 2021 - January 2022"
Actual   :"Dezember 2021 - Januar 2022"
```

Therefore, I propose setting the locale to English using this line:

    Locale.setDefault(Locale.ENGLISH)

This causes the test to pass on non-English systems.


ralfgehrer (dismissed):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/calendar/CalendarCalculationTest.kt
```diff
@@ -3,13 +3,22 @@ package de.rki.coronawarnapp.ui.calendar
...
 class CalendarCalculationTest {
 
     private var pattern = "dd.MM.yyyy"
     private val formatter = DateTimeFormat.forPattern(pattern)
 
+
+    @Before
+    fun setUp() {
+        // Always return English results no matter the locale of the test environment host
+        Locale.setDefault(Locale.ENGLISH)
```
PhilippNowak96, line 19: IMO it would be good to store the initial locale in some variable and reset it after the tests finished because `setLocale()` changes the locale of the entire JVM instance currently used. In addition `@BeforeClass` and `@AfterClass` would fit better here I think as this needs to be done only once for the whole test class.

PhilippNowak96 (changes requested):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/calendar/CalendarCalculationTest.kt
fynngodau: Another alternative I considered was to mock `getDefault()` to return `Locale.ENGLISH`.

d4rken: It would be ideal if this doesn't require static access at all, but I'm not sure what the effort here would be.

Thinking out loud:
* Pass context to `CalendarCalculation`
* Use locale from `context.resources`
* Pass mock context in tests

PhilippNowak96: @d4rken or even use locale from the new `Context.getLocale()` extension function 😎

d4rken: @fynngodau What do you think about the above ideas?

fynngodau: Hi again, sorry for the inactivity here, I have unfortunately been busy with other things.

I don't quite see the benefit in avoiding a static `getLocale` access inside `CalendarCalculation` yet… would that maybe mean that this particular call would respond to changes in the system language settings quicker? I'm not quite sure what changes would have to be done to this class for it to be able to access a `Context`, I suppose it should use dependency injection?

I'd already have a stale commit floating around somewhere for mocking `getDefault`, though it sounds like you'd prefer the alternative that you (@d4rken) proposed.

d4rken: > Hi again, sorry for the inactivity here, I have unfortunately been busy with other things.

No worries  :hugs: .

> Would that maybe mean that this particular call would respond to changes in the system language settings quicker?

Yes. An app may have different configurations per activity, which means that `context.resources.configuration.locale` may change, so it's good practice to get the locale from the context we are currently in (i.e. activity, or even better from the view).

> I'm not quite sure what changes would have to be done to this class for it to be able to access a Context, I suppose it should use dependency injection?

I think I'd go for just passing the context.

> I'd already have a stale commit floating around somewhere for mocking getDefault, though it sounds like you'd prefer the alternative that you (@d4rken) proposed.

I'll keep this open until this evening when I'll make a new PR with my proposed solution.
Will ping @fynngodau  and @PhilippNowak96 , always happy for feedback.

fynngodau: > I'll make a new PR with my proposed solution.

Great that you're taking care of it (:

d4rken:
Closing in favor of #2139

PR #2039, opened 2021-01-06
Fix crash when displaying test result (DEV)

chiljamgossow:
fix crash when trying to display test result
some other issues that popped up in AS

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/TestResultSectionView.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/TestResultSectionView.kt
@@ -26,7 +26,7 @@ constructor(
     attrs: AttributeSet? = null,
     defStyleAttr: Int = 0
 ) : ConstraintLayout(context, attrs, defStyleAttr) {
-    private var binding: ViewTestResultSectionBinding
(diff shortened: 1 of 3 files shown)


Oliver-Zimmerman (dismissed):

ralfgehrer (approved):
Tested on a Pixel (Android 10)

Oliver-Zimmerman (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/TestResultSectionView.kt
```diff
@@ -26,10 +26,11 @@ constructor(
     attrs: AttributeSet? = null,
     defStyleAttr: Int = 0
 ) : ConstraintLayout(context, attrs, defStyleAttr) {
-    private val binding = ViewTestResultSectionBinding.bind(this)
+    private val binding: ViewTestResultSectionBinding
 
     init {
         inflate(context, R.layout.view_test_result_section, this)
+        binding = ViewTestResultSectionBinding.bind(this)
```
kolyaopahle, line 33: could be replaced by just calling

`binding = ViewTestResultSectionBinding.inflate(this)`

PR #2040, opened 2021-01-06
Use close button on warn others screen (EXPOSUREAPP-4257) 

chiljamgossow:
Exchange back button with close button on warn others screen
Use toolbar


mtwalli (approved):
👍🏼

d4rken (approved):
:+1:

PR #2042, opened 2021-01-07
Set up and configure fastlane to take screenshots - Home Screen(EXPOSUREAPP-2950)

mtwalli:
1- Add screenshot tests for Home Screen
2- Create a screenshot shell script to run `screengrab` and zip screenshots afterward

Enhancement: Added delay time of 2 sec before every screenshot statement to avoid blank screens

Aside from the ticket:
1- Fixed crash in DI that was happening for UI tests on pre-Android 11 API `onPreCreateActivity` was not called for all versions
2- Fixed Images are not showing in UI tests for some cards in Home Screen (This might be a topic for our Tech talk)
    [Vector Drawable](https://developer.android.com/guide/topics/graphics/vector-drawable-resources#vector-drawables-backward-solution)

### Testing:
1. run `./gradlew clean assembleDebug assembleAndroidTest` for clean state
2. run `bash screenshot.sh` in case there is APK already built using the command in point 1 you can start from here.

Screenshots and screenshots.zip are located under `fastlane/metadata/android`


jurajkusnier:
what do you think about more descriptive naming? Not just "HomeFragment" but also app state like "submission done", "test negative/positive/invalid", etc?

mtwalli:
> what do you think about more descriptive screenshots file naming? Not just "HomeFragment" but also app state like "submission done", "test negative/positive/invalid", etc?

Make sense to identify the screen's case when something is odd

jurajkusnier (dismissed):

ralfgehrer:
If I run it on my usual test device, the icon color is black instead of white.


HomeFragment_test_error_1610530272705


HomeFragment_submission_done_1610530305941

mtwalli:
> If I run it on my usual test device, the icon color is black instead of white.
> 
> HomeFragment_test_error_1610530272705
> 
> HomeFragment_submission_done_1610530305941

Thx, fixed

ralfgehrer (approved):

jurajkusnier (approved):

PR #2043, opened 2021-01-07
Skeletons for showing statistics in the home screen feed (EXPOSUREAPP-3674)

d4rken:
Introduce base for showing statistics in the home screen feed to enable parallel work on each screen.
Adds a nested horizontally scrolling RecyclerView that supports dynamic content updates.

### Testing
* It should not crash
* The code should not have any broad flaws that will be a problem later on

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/source/StatisticsProvider.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/source/StatisticsProvider.kt
@@ -1,0 +1,102 @@
+package de.rki.coronawarnapp.statistics.source
+
+import de.rki.coronawarnapp.server.protocols.internal.stats.KeyFigureCardOuterClass
+import de.rki.coronawarnapp.statistics.IncidenceStats
+import de.rki.coronawarnapp.statistics.InfectionStats
+import de.rki.coronawarnapp.statistics.KeySubmissionsStats
+import de.rki.coronawarnapp.statistics.StatisticsData
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.GlobalScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.launch
+import org.joda.time.Instant
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class StatisticsProvider @Inject constructor() {
+
+    private val currentInternal = MutableStateFlow(StatisticsData(items = emptyList()))
+    val current: Flow<StatisticsData> = currentInternal.filterNotNull()
+
+    init {
+        // Mock data
+        GlobalScope.launch(context = Dispatchers.IO) {
+            val statisticsData = StatisticsData(
+                items = listOf(
+                    InfectionStats(
+                        updatedAt = Instant.ofEpochMilli(1604839761),
+                        keyFigures = listOf(
+                            KeyFigureCardOuterClass.KeyFigure.newBuilder().apply {
+                                rank = KeyFigureCardOuterClass.KeyFigure.Rank.PRIMARY
+                                value = 14714.0
+                                decimals = 0
+                                trend = KeyFigureCardOuterClass.KeyFigure.Trend.UNSPECIFIED_TREND
+                                trendSemantic =
+                                    KeyFigureCardOuterClass.KeyFigure.TrendSemantic.UNSPECIFIED_TREND_SEMANTIC
+                            }.build(),
+                            KeyFigureCardOuterClass.KeyFigure.newBuilder().apply {
+                                rank = KeyFigureCardOuterClass.KeyFigure.Rank.SECONDARY
+                                value = 11981.0
+                                decimals = 0
+                                trend = KeyFigureCardOuterClass.KeyFigure.Trend.INCREASING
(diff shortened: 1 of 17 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/source/StatisticsProvider.kt
```diff
@@ -0,0 +1,102 @@
...
+
+@Singleton
+class StatisticsProvider @Inject constructor() {
+
+    private val currentInternal = MutableStateFlow(StatisticsData(items = emptyList()))
+    val current: Flow<StatisticsData> = currentInternal.filterNotNull()
+
+    init {
+        // Mock data
+        GlobalScope.launch(context = Dispatchers.IO) {
```
mtwalli, line 26: it is better to inject the dispatcher

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsCardPaddingDecorator.kt
```diff
@@ -0,0 +1,38 @@
...
+                if (itemPosition == it.itemCount - 1) {
+                    outRect.right = resources.getDimensionPixelSize(endPadding)
+                } else {
+                    val distance = resources.getDimensionPixelSize(cardDistance)
+                    outRect.left = distance / 2
+                    outRect.right = distance / 2
+                }
+            }
+        }
+    }
```
mtwalli, line 37: maybe:
```
when(itemPosition){
}
``` 
is more readable

Corona-Warn-App/src/main/res/layout/home_statistics_cards_basecard_layout.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<androidx.cardview.widget.CardView xmlns:android="http://schemas.android.com/apk/res/android"
```
mtwalli, line 2: [Optional] there is MaterailCardView , something to keep in mind towards MDC

Corona-Warn-App/src/main/res/layout/home_statistics_cards_incidence_layout.xml
```diff
@@ -0,0 +1,24 @@
...
+            android:layout_height="wrap_content"
+            android:layout_marginStart="24dp"
+            android:layout_marginTop="24dp"
+            android:focusable="false"
+            android:text="Incidence"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent" />
+    </androidx.constraintlayout.widget.ConstraintLayout>
+
+</layout>
```
mtwalli, line 24: is layout tag needed here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/source/StatisticsProvider.kt
d4rken: Agree, but this is all just placeholders so multiple people can start working on this, I'll rewrite the whole class next week.

Corona-Warn-App/src/main/res/layout/home_statistics_cards_basecard_layout.xml
d4rken: I think for that we also need the MaterialTheme (using that for the diary), not AppCompatTheme (rest of the app).
Can we use the MaterialCardView without that?
We have a ticket for refactoring the theming open.

Corona-Warn-App/src/main/res/layout/home_statistics_cards_incidence_layout.xml
d4rken: In the current form, no, but it's a placeholder. The final layout probably needs it, but the person that will work on the ticket for this card will have to decide that.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsCardPaddingDecorator.kt
d4rken: Not sure due to the `adapter.?` but will try. :ok_hand:

d4rken: Done.

Corona-Warn-App/src/main/res/layout/home_statistics_cards_basecard_layout.xml
mtwalli: Not sure, let us keep for the refactoring ticket then

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/source/StatisticsProvider.kt
mtwalli: ok

Corona-Warn-App/src/main/res/layout/home_statistics_cards_incidence_layout.xml
mtwalli: make sense

mtwalli (approved):
👍🏼

ralfgehrer (approved):

PR #2048, opened 2021-01-07
Fix missing timestamp on TracingFailed card (DEV)

d4rken:
Just noticed while working on something else.

The card looked like this:
Screenshot from 2021-01-07 11-50-06

I see no reason to not display the timestamp here. Seems to be an oversight from refactoring.
The TracingDisabled card also did not have this restriction to LOW/INCREASED risk levels.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
@@ -131,13 +131,13 @@ data class TracingFailed(
 
     val showRestartButton: Boolean = !isInDetailsMode
 
-    fun getTimeFetched(c: Context): String = if (lastExposureDetectionTime != null) {
-        c.getString(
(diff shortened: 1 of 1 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
```diff
@@ -131,18 +131,13 @@ data class TracingFailed(
...
-                c.getString(
-                    R.string.risk_card_body_time_fetched,
-                    formatRelativeDateTimeString(c, lastExposureDetectionTime)
-                )
-            } else {
-                c.getString(R.string.risk_card_body_not_yet_fetched)
-            }
-        }
-        else -> ""
+    fun getTimeFetched(c: Context): String = if (lastExposureDetectionTime != null) {
```
mtwalli, line 134: please rename c to context

ralfgehrer (dismissed):
works nicely on Pixel (Android 10)

ralfgehrer (approved):

mtwalli (approved):

PR #2050, opened 2021-01-07
Remove legacy FileLogger in favor of new DebugLogger (DEV)

d4rken:
The new `DebugLogger` can fully replace the previous `FileLogger`, it's not necessary to maintain both and it's desirable that we "dogfood" the production variant internally too.

* Removed old logger
* Trigger automatically on tester builds
* Replace test menu option with a button taking you to the new screen and an explanation for testers

### Testing
* Check that it's enabled by default for test builds, but that it's not for production builds
* Review a logfile to check that recording works correctly

Screenshot from 2021-01-07 12-57-19


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -44,8 +45,19 @@ object DebugLogger : DebugLoggerBase() {
...
 
         try {
-            if (triggerFile.exists()) {
-                Timber.tag(TAG).i("Trigger file exists, starting debug log.")
+            val startLogger = when {
+                triggerFile.exists() -> {
+                    Timber.tag(TAG).i("Trigger file exists, starting debug log.")
+                    true
+                }
+                CWADebug.isDeviceForTestersBuild -> {
```
ralfgehrer, line 53: Now we don't have the option anymore to disable logging on deviceForTestersBuild. Every restart will overwrite the last setting and automatically turn on logging. Might be problematic for low device storage tests or alike.

d4rken: Has this been the case yet? Even low end devices have a few 100MB storage. Test devices often have no other apps installed. I'd say we fix this it becomes an issue for testers.

ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
vaubaehn: Just my small 2 cents for future enhancement:
What about letting the logger for production and DeviceForTesters builds query the available disk space from time to time, and automatically stop logging when it gets below a threshold of 100MB?

chiljamgossow (dismissed):

ralfgehrer (approved):

chiljamgossow (approved):

PR #2053, opened 2021-01-07
Transnational country list names not changing when language is changed (EXPOSUREAPP-4416)

axelherbstreith:
This PR fixes an issue where the country names in the transnational country list are not changed when changing the system language.
This only occurred when this particular screen is currently opened or was opened before. This behaviour was traced back to the usage of CachedStrings for the country names.

While testing the changes, no performance issue was identified.


d4rken:
If this may be an issue on other screens too, you could change the LazyString type for the country names to an implementation that just does not cache the result, but keeps the indirection to use the context from where it's resolved.

axelherbstreith:
@d4rken good point! I will check if there are any other screen that are effected

mtwalli (approved):

BMItr (approved):

PR #2055, opened 2021-01-07
Remove @dimen/match_constraint as it is replaced by tooling (DEV)

d4rken:
Replace `@dimen/match_constraint` with hardcoded `0dp`.
There is no reason this would be changed, and while it may be nice for some to understand what `0dp` means, it conflicts with using the built-in layout editor which when used via GUI and not XML code, will replace it with `0dp` everytime leading to a mix of both in the project.

 I just did a search+replace to change this.

### Testing
* Check that it still compiles.
* Run it on your device, visit a few screens, in tester and non-tester builds variants.


ralfgehrer (dismissed):
Thx. Tested and confirmed on Pixel (Android 10).

mtwalli (dismissed):

Issue #2037, opened 2021-01-06
Ripple effect going beyond background box


fynngodau:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: HUAWEI HONOR 5X
- Android version: 9
- App version: 1.11.0 (self-compiled)

## Describe the bug

The ripple effect for the boxes that one can click in "Retrieve Test Result" goes beyond the rounded corners in the background, like seen in this screenshot:

![image](https://user-images.githubusercontent.com/16943720/103771522-313e1c00-5028-11eb-89ce-397cebe1d757.png)

I recorded this screenshot while holding my finger on the upper card.

## Steps to reproduce the issue

1. Click the card "Have you been tested?"
2. Hold one of the three cards ("Test with QR Code" / "Enter TAN" / "No TAN yet?")
3. Observe the ripple effect

## Expected behaviour

The ripple effect should be clipped to the rounded corners.

## Possible Fix

One could use a `CardView` that has zero elevation, like I suggested here already: https://github.com/corona-warn-app/cwa-app-android/pull/1912#pullrequestreview-555027177

## Additional context

This issue was introduced in #1912 when the ripple effect was introduced.

---
Internal Tracking ID: [EXPOSUREAPP-4518](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4518)

dsarkar:
@fynngodau. Thanks for your contribution. 

Info is forwarded to devs.

Best wishes, 
DS 
  
--- 
Corona-Warn-App Open Source Team 
Internal Tracking ID: EXPOSUREAPP-4518

MikeMcC399:
"Adjust ripple area on dispatcher screen (EXPOSUREAPP-4518)" https://github.com/corona-warn-app/cwa-app-android/pull/2058 is included in [v1.11.0-RC1](https://github.com/corona-warn-app/cwa-app-android/releases/tag/v1.11.0-RC1).

PR #2058, opened 2021-01-08
Adjust ripple area on dispatcher screen (EXPOSUREAPP-4518)

chiljamgossow:
Adjust ripple area on dispatcher screen


axelherbstreith (approved):
LGTM

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_appconfig.xml
```diff
@@ -14,7 +14,7 @@
         android:paddingBottom="32dp">
 
         <LinearLayout
-            style="@style/card"
+            style="@style/Card"
```
mtwalli, line 17: just a question, I saw some styles names start with lower case letters and some with capital case letters. 
what is our convention?

chiljamgossow: As far as I can see, there is none. I would prefer them upper case, because GreyCard.Ripple looks better to me.

mtwalli: I like that way too.

mtwalli (approved):

fynngodau:
Just to explicitly mention it here, this is a fix for #2037.

PR #2060, opened 2021-01-08
Additional debug log censoring (EXPOSUREAPP-4451)

d4rken:
Adds censoring for diary locations, persons and the QRCode GUID.

Because diary data is stored asynchronously, we need to delay logging by a little bit such that the data (what to censor) is available to the censor modules.

To improve "dogfooding" for the test builds, I've changed the code to execute what happens in non-tester builds and then just overwrite the censored line again.

## Testing
* Check censoring
    * Scan a QRCode check GUID censoring
    * Create diary persons, should be censored as `Person#<Number>`
    * Create diary locations, should be censored as `Location#<Number>`
    * Export the diary, any logged export data should also be censored
* Check logging of uncaught exceptions again


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/bugreporting/censors/DiaryPersonCensorTest.kt
```diff
@@ -0,0 +1,86 @@
...
+            tag = "I'm a tag",
+            throwable = null
+        )
+        instance.checkLog(censorMe) shouldBe censorMe.copy(
+            message = "Person#2 needs more coffee, but Person#3 has had enough for today."
+        )
+
+        every { CWADebug.isDeviceForTestersBuild } returns true
+        instance.checkLog(censorMe) shouldBe censorMe.copy(
+            message = "Ralf needs more coffee, but Matthias has had enough for today."
```
ralfgehrer, line 69: 😄

ralfgehrer:
Minor: Noticed the double-
Screenshot 2021-01-08 at 19 10 02
line on the regular build

ralfgehrer (commented):
Persons are censored but Locations are not.

Excerpt of the log:
```
Die nachfolgende Liste dient dem zuständigen Gesundheitsamt zur Kontaktnachverfolgung gem. § 25 IfSG.

09.01.2021 Person#2
09.01.2021 Person#1
09.01.2021 Berlin
)
```

harambasicluka:
@d4rken can this be merged? So is currently everything for production disabled?

d4rken:
> @d4rken can this be merged? So is currently everything for production disabled?

Need to fix the comments, but yes after that can be merged.
For production it needs to be started via menu and you can't get to the menu on production currently.

d4rken:
> Minor: Noticed the double-line on the regular build

Addressed.

> Persons are censored but Locations are not.

Fixed, censor modules were not all applied if multiple modules match the same line.

ralfgehrer (dismissed):

chiljamgossow (commented):
Tested on Xiaomi device.
I found registration tokens in the OkHttp and SubmissionTask logging. Do we need to censor that as well?

d4rken:
> I found registration tokens in the OkHttp and SubmissionTask logging. Do we need to censor that as well?

Yes but was it in the debug console or the actual text file.
In tester builds we still have the normal timer debug log panted in `CWADebug`. The censoring only happens for the written file and only in `device` builds.

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -107,8 +108,14 @@ object DebugLogger : DebugLoggerBase() {
                         while (!isDaggerReady) {
                             yield()
                         }
-                        val censoredLine = bugCensors.get().mapNotNull { it.checkLog(rawLine) }.firstOrNull()
-                        appendLogLine(censoredLine ?: rawLine)
+                        launch {
+                            // Censor data sources need a moment to know what to censor
```
BMItter, line 112: I don't like the delay part too much (feels hacky)
but I dont know how to improve it right now

d4rken: I also don't like it, but don't have a better solution currently either :shrug:

BMItr (approved):

d4rken:
There will be a follow up PR to checkout registration token censoring, @chiljamgossow  found an issue.

PR #2063, opened 2021-01-11
Pump versions - Android API 30 (EXPOSUREAPP-4562)

mtwalli:
First of all, Please FOCUS on non-test dependencies. I updated the tests and made sure it is working.

- Pump dependencies versions
- Target Android 30
- Remove duplicate `coroutines-testing` dependency
- Remove `Kotlin-stdlib`  dependency as it is added by default since 1.4 ([Look here](https://kotlinlang.org/docs/reference/whatsnew14.html#dependency-on-the-standard-library-added-by-default))

Changelogs:
- Dagger(2.28.2 -> 2.30.1) the versions in between are mostly addressing Hilt issues which we don't use yet.
      - A new Dagger compiler option was added to allow disabling transitive validation of component dependency
      - Breaking change https://github.com/google/dagger/commit/08193ad70bd78904e56bb262e924bc692422066e
      - Here is the full log: https://github.com/google/dagger/releases

- Core-ktx (1.3.0 -> 1.3.2) bug fixes: https://developer.android.com/jetpack/androidx/releases/core#core_and_core-ktx_version_132_2

- MDC (1.1.0 -> 1.21) 
   - New compoents Transition/Motion, Slider, ShapeableImageView and bug fixes 
   - https://github.com/material-components/material-components-android/releases/tag/1.2.1

- ConstrainLayout (1.1.3 -> 2.0.4)  Two years of changes, It is a new ConstaintLayout :D
- Preferences (Same version) just using `-ktx` version
- Navigation (2.2.2 -> 2.3.2) Mostly bug fixes, Major changes https://developer.android.com/jetpack/androidx/releases/navigation#2.3.0
  
Changes due to the update:
- Fix `app:lint` errors
- Move to `ContextCompat` and `ResourcesCompat` versions of `getDrawable`
    more information regarding that can be found here: https://developer.android.com/jetpack/androidx/releases/appcompat#1.2.0-beta01
- Use `Handler(looper)` constructor instead of deprecated `Handler()`
- Fix Nullable types where NonNullable are expected 

### Android 11 Compatibility Test
https://developer.android.com/about/versions/11/test-changes


chiljamgossow (dismissed):
LGTM
Tested on Xiaomi

ralfgehrer (changes requested):
On a fresh install, the 3rd onboarding screen is empty and a notification is falsely shown.

Screenshot 2021-01-11 at 12 57 24
Screenshot 2021-01-11 at 12 57 31

PS: I double-checked on the release/1.11.x branch. I don't experience this issue there.

d4rken:
> and a notification is falsely shown.

Not sure if that is a false notification, your device shows "Tuesday" today is "Monday" so the time would be wrong.

ralfgehrer:
> > and a notification is falsely shown.
> 
> Not sure if that is a false notification, your device shows "Tuesday" today is "Monday" so the time would be wrong.

True, the notification is correct. The empty screen issue, however, remains with the correct time.

ralfgehrer:
Tested extensively on Pixel (Android), e.g.
- Contact Journal with all functionalities
- Test Results: Pending, Negative and Positive with Symptoms via QR code scanning
- App Reset via Settings
- Same as above.

LGTM

ralfgehrer (dismissed):

mtwalli:
> On a fresh install, the 3rd onboarding screen is empty and a notification is falsely shown.
> 
> Screenshot 2021-01-11 at 12 57 24
> 
> Screenshot 2021-01-11 at 12 57 31
> 
> PS: I double-checked on the release/1.11.x branch. I don't experience this issue there.

Thanks for the feedback, It should be working now. I am really wondering how this working before, the content was aligned to the back button which is not a sibling view to the content (i.e both are in the same ConstraintLayout ).

ralfgehrer (approved):

axelherbstreith:
LGTM! Tested on Huawei P30 with Android 10

chiljamgossow (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DataBindingAdapters.kt
```diff
@@ -29,7 +33,7 @@ fun setAnimation(view: LottieAnimationView, animation: Int?) {
         val type = appContext.resources.getResourceTypeName(animation)
 
         if (type == DRAWABLE_TYPE) {
-            view.setImageDrawable(appContext.getDrawable(animation))
+            view.setImageDrawable(appContext.getDrawableCompat(animation))
```
BMItter, line 36: here you could use the context of the view itself

BMItter:
I recommend deeper testing for an SDK change to 30 from several persons or several devices.
Please also make a ticket or link it to the title.
adjustments are looking good so far :princess:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DataBindingAdapters.kt
mtwalli: Good point, I think I changed this in one of my screenshots PRs. I will check it there again as there are unit tests covering such change which I already modified #2042

mtwalli: this commit addressed the change .https://github.com/corona-warn-app/cwa-app-android/pull/2042/commits/6cbd6b1abca296fb5e70252b597094364ed6d513

SamuraiKek:
LGTM. Smoke tested on Nokia 3 running Android 9 and Huawei Mate 8 running Android 6.

chiljamgossow:
Tested on Nokia 7 Plus, Android 10

jurajkusnier:
LGTM. Tested on OnePlus X (Android 6.0.1), Pixel 4a (Android 11)

chris-cwa (approved):
looking good on pixel 3a / Android 10

AlexanderAlferov:
Samsung S8, Android 9 - looking good!

LukasLechnerDev:
Google Pixel 3A, Android 11, No issues found

BMItr (approved):

PR #2066, opened 2021-01-11
Switched from large to xlarge for faster builds and checks (DEV)

AlexanderAlferov:
Switched circle ci containers from large to xlarge for faster builds and checks.


ralfgehrer (approved):

d4rken:
Closing this after discussion with @AlexanderAlferov I'm already editing the same file and will add these changes to prevent conflicts.

PR #2068, opened 2021-01-11
Contact Journal - Fast double click leads to double entry in contact journal (EXPOSUREAPP-4561)

axelherbstreith:
This PR fixes an issue where a person/location was displayed duplicated in the contact journal overview screen when quickly pressing on the respective item in the day screen.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/person/ContactDiaryPersonListAdapter.kt
```diff
@@ -22,10 +23,13 @@ internal class ContactDiaryPersonListAdapter(
 
     override fun onBindBaseVH(holder: CachedPersonViewHolder, position: Int, payloads: MutableList<Any>) {
         val item = data[position]
+        val codeThrottle = CodeThrottle()
```
ralfgehrer, line 26: I'd prefer an extension function on View. E.g. `fun View.setOnThrottlesClicklistener()` or, alternatively, an ThrottledClickListener (overriding the regular one).

d4rken: Thinking out loud:
* `fun View.setOnClickListenerDebounced { }`
* `fun View.clicks(): Flow<View> -> View.clicks().debounced()` not sure if it would be a burden or an asset to have this as `Flow` :thinking:

d4rken: Maybe `debounced` is the wrong name, what the code currently does is `ThrottleFirst`, e.g. http://reactivex.io/documentation/operators/sample.html which seems to be what we want.

AlexanderAlferov:
Bug is no longer reproducible. Samsung S8 Android 9. Please, address code comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/person/ContactDiaryPersonListAdapter.kt
axelherbstreith: @d4rken I implemented it as an extension function now as @ralfgehrer proposed. I didn't use `Flow` since this would add more complexity in my opinion

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/recyclerview/ThrottledClickListener.kt
```diff
@@ -0,0 +1,22 @@
+package de.rki.coronawarnapp.util.recyclerview
+
+import android.view.View
+
+/*  Prevents a recyclerview item to be double selected when double clicking on it  */
+class ThrottledClickListener(
+    private val interval: Long,
```
ralfgehrer, line 7: you can set a default of 300 so you don't have to set it when calling the function above.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/ViewExtensions.kt
```diff
@@ -12,3 +13,6 @@ fun View.setGone(gone: Boolean) {
 fun View.setInvisible(invisible: Boolean) {
     visibility = if (invisible) View.INVISIBLE else View.VISIBLE
 }
+
+fun View.setOnClickListenerThrottled(interval: Long, listenerBlock: (View) -> Unit) =
```
ralfgehrer, line 17: you can set an interval default value so you don't have to necessarily set it in the caller.

axelherbstreith: Done 👍

ralfgehrer (approved):
Tested on Pixel (Android 10)

AlexanderAlferov (approved):

PR #2074, opened 2021-01-12
Show pending card when offline/ request fails (EXPOSUREAPP-4507)

chiljamgossow:
When a request fails, the state should switch to pending and not stay in fetching as this suggests an ongoing process.
Another bug that was fixed: the error dialog (eg. when offline) was triggered multiple times when the screen was reopened, either by going back or putting the app to background


LukasLechnerDev:
Branch name should be **fix**/4507-offline-test-result-pending instead of **feature**/4507-offline-test-result-pending, right? 

But I don't know if it's worth fixing ...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -43,7 +45,8 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
         pendingViewModel.testState.observe2(this) { result ->
             result.deviceUiState.withFailure {
                 if (it is CwaWebException) {
-                    DialogHelper.showDialog(buildErrorDialog(it))
+                    dialog?.dismiss()
+                    dialog = DialogHelper.showDialog(buildErrorDialog(it))
```
LukasLechnerDev, line 49: We observe a failure here twice and therefore there is flickering, because the dialog is quickly shown, dismissed and shown again. 

So I would suggest:

```suggestion
                    if (dialog == null || dialog?.isShowing == false) {
                        dialog = DialogHelper.showDialog(buildErrorDialog(it))
                    }
```

The better solution though would be if we wouldn't receive the failure here twice. But this would be needed to fix in another layer of the app I guess.

chiljamgossow: There are different error dialogs, otherwise your suggestion would solve the issue. I will investigate why the state is issued twice.

LukasLechnerDev: Are you sure there are different error dialogs? 

Since we have the `if (it is CwaWebException)` condition before showing a dialog here, `buildErrorDialog(it)` will always build the same dialog, or am I missing something?

d4rken:
> Branch name should be fix/4507-offline-test-result-pending instead of feature/4507-offline-test-result-pending, right?
> But I don't know if it's worth fixing ...

Not worth changing once a PR is open. Just a pattern to help us devs when checking out branches.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
d4rken: I also see the duplicate dialogs. I think it happens because the `SubmissionRepository` still holds the previous data to which we subscribe when opening the screen, but then opening the screen also causes a refresh too, via `refreshDeviceUIState`.

> Since we have the if (it is CwaWebException) condition before showing a dialog here, buildErrorDialog(it) will always build the same dialog, or am I missing something?

We can probably remove the `else {` branch, the only difference seems to be the web-exception status code and that is available in any case via the base exception class `CwaWebException`.

The `else` branch may be a left over from previous submission flow refactoring.

d4rken (commented):
This solves the described issue. Would approve once the open discussion is resolved.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: I adjusted the logic so that there is only one error dialog, no old values trigger the dialog when opening the screen multiple times/ putting app to background

chiljamgossow: The root cause it that the deviceUIStateFlow is a mixture of the result of a specific request (success or failure) and the current state of the corona test.

LukasLechnerDev (dismissed):

LukasLechnerDev:
LGTM, Tested on Pixel 3A, Android 11

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -33,27 +33,20 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
 
     private var skipInitialTestResultRefresh = false
 
+    private var errorDialog: AlertDialog? = null
+
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        pendingViewModel.consentGiven.observe2(this) {
+        pendingViewModel.consentGiven.observe(viewLifecycleOwner) {
```
d4rken, line 41: Why did you change this?

d4rken, line 41: Why did you change this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -33,27 +33,20 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
...
-        pendingViewModel.testState.observe2(this) { result ->
-            result.deviceUiState.withFailure {
-                if (it is CwaWebException) {
-                    DialogHelper.showDialog(buildErrorDialog(it))
-                }
-            }
-
-            val hasResult = result.deviceUiState.withSuccess(false) { true }
-
+        pendingViewModel.testState.observe(viewLifecycleOwner) { result ->
```
d4rken, line 45: Same here, why the change away from `observe2`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -98,8 +91,17 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
         super.onResume()
         binding.submissionTestResultContainer.sendAccessibilityEvent(AccessibilityEvent.TYPE_ANNOUNCEMENT)
         pendingViewModel.refreshDeviceUIState(refreshTestResult = !skipInitialTestResultRefresh)
-
         skipInitialTestResultRefresh = false
+        pendingViewModel.cwaWebExceptionLiveData.observe(this.viewLifecycleOwner) { exception ->
```
d4rken, line 95: There is the extension function `observeOnce`, would that be suitable here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -98,8 +91,17 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
...
+            pendingViewModel.cwaWebExceptionLiveData.removeObservers(this.viewLifecycleOwner)
+            handleError(exception)
+        }
+    }
+
+    override fun onPause() {
+        super.onPause()
```
d4rken, line 102: Won't cause an issue, here, but FYI, it's considered best practice to call super.<incoming> (onCreate, onResume) before your own code, and super.<outgoing> (onPause, onDestroy) after your own code.

https://groups.google.com/g/android-developers/c/ndsrBsIchFc?pli=1

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -118,33 +120,42 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
...
+            R.string.submission_error_dialog_web_generic_network_error_body,
+            exception.statusCode
+        ),
+        R.string.submission_error_dialog_web_generic_error_button_positive,
+        null,
+        true,
+        ::navigateToMainScreen
+    )
+
+    private val genericErrorDialog: DialogHelper.DialogInstance
```
d4rken, line 151: In which circumstance do we need this?

The "non-generic" one only has a different  error message, one that takes a statuscode, but every exception will have that statuscode, so can't we use the same dialog for all `CwaWebException`?

d4rken (commented):
Nothing is a blocker but I'm curious about some changes.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: I do not understand why we use a wrapper here. It's just confusing to see for everyone new to the project.

d4rken: * To have shorter code, initially it removed the need for `(object : Observer...` which is no longer needed since newer androidx versions, but `this` is still shorter than `viewLifecycleOwner`.
* To reduce the chance someone writing `observe(this)`  and using the wrong lifecycle owner.

If it were possible I'd like to actually make it just `observe()` in fragments but I don't think Kotlin allows for that.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: yes, replaced it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: switched the order

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: The generic is shown for exceptions which are not CwaClientError or CwaServerError. Why? I do not know, maybe someone decided that the user should not get more details in these cases.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: reverted

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
chiljamgossow: reverted

d4rken:
If I press "Zurück" on the dialog, I get back to the home screen.
If I press the back button on the device, I close the dialog and stay on an empty screen.

Screenshot from 2021-01-14 16-23-31

chiljamgossow:
Yes, that is the default behavior and also what happens with all our dialogs. What would you expect?

d4rken:
> Yes, that is the default behavior and also what happens with all our dialogs., same for clicking outside the dialog. What would you expect?

I would have thought that if the dialogs only action is "Zurück", that dismissing the dialog would also act like "Zurück".

d4rken (approved):

d4rken (approved):
Fixes the original issue, other inconsistencies can be picked up in an extra PR.

LukasLechnerDev (approved):

Issue #2038, opened 2021-01-06
Likely incorrect card elevation in 1.11


fynngodau:
This issue is about the elevation of the main screen cards in `release/1.11.x` after #1912.

* [x] None of the other issue templates apply to this issue.

See https://github.com/corona-warn-app/cwa-app-android/pull/1912#pullrequestreview-555954874.

> I think I introduced an inconsistency with the intended elevation […], as I spotted some code that explicitly set the risk card to `10dp` (e.g. `TracingDisabledBox`). […] Maybe you could try to find out what the intended elevation is, and make sure to set it correctly.

The current behaviour, to my knowledge, is that all cards in the main screen of `release/1.11.x` have a very strong `10dp` elevation. I suggest you check again what the correct elevation would be, as I see reason to assume it should actually be `4dp` for most of the cards.

--- 
Internal Tracking ID: [EXPOSUREAPP-4519](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4519)

dsarkar:
@fynngodau Thanks. Devs are informed.

Best wishes, 
DS 
  
--- 
Corona-Warn-App Open Source Team 
Internal Tracking ID: [EXPOSUREAPP-4519](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4519)

MikeMcC399:
"Fixed card elevation (EXPOSUREAPP-4519)" https://github.com/corona-warn-app/cwa-app-android/pull/2075 is included in [v1.11.0-RC2](https://github.com/corona-warn-app/cwa-app-android/releases/tag/v1.11.0-RC2).

fynngodau:
I'm repeating what I said in #2075 here so it doesn't get lost in a closed PR:

The PR is probably only a partial fix to what I explained here – see for example this code in `TracingDisabledBox`:

    override val viewBinding = lazy {
        TracingContentDisabledViewBinding.inflate(
            layoutInflater,
            itemView.findViewById(R.id.box_container),
            true
        ).also {
            it.root.elevation = resources.getDimension(R.dimen.elevation_strong)
        }
    }

This code does not work, IIRC because it sets the wrong attribute (`app:cardElevation` vs. `android:elevation`).

The existence of this code implies to me that *certain* cards *should* have `elevation_strong`, which was the reason I created this issue – either this and similar setters need to be fixed, or this code needs to be removed.

dsarkar:
[...]

PR #2075, opened 2021-01-12
Fixed card elevation (EXPOSUREAPP-4519)

chris-cwa:
Screenshot 2021-01-12 at 13 52 47

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/home_card_container_layout.xml
+++ b/Corona-Warn-App/src/main/res/layout/home_card_container_layout.xml
@@ -11,5 +11,5 @@
     android:layout_marginBottom="@dimen/spacing_tiny"
     android:foreground="?selectableItemBackground"
     android:focusable="true"
-    app:cardElevation="@dimen/elevation_strong"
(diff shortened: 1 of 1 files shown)


ralfgehrer (approved):

AlexanderAlferov (approved):
LGTM 🚀  Tested on Samsung S8 Android 9

fynngodau:
I think that this is only a partial fix to what I mentioned in #2038, see for example this code in `TracingDisabledBox`:


    override val viewBinding = lazy {
        TracingContentDisabledViewBinding.inflate(
            layoutInflater,
            itemView.findViewById(R.id.box_container),
            true
        ).also {
            it.root.elevation = resources.getDimension(R.dimen.elevation_strong)
        }
    }

This code does not work, IIRC because it sets the wrong attribute (`app:cardElevation` vs. `android:elevation`).

The existence of this code implies to me that *certain* cards *should* have `elevation_strong`, which was the reason I created #2038 – either this and similar setters need to be fixed, or this code needs to be removed.

PR #2078, opened 2021-01-12
Contact Journal clickable cards

SamuraiKek:
This PR makes the whole contact diary cards clickable on the overview page. Up until now, only the header of the cards was clickable when when the persons and places lists were populated. 

How to test:

- Navigate to the contact diary. 
- On the Contact Journal page click anywhere on a date entry that has some people and/or places assigned. You should be able to navigate to the next screen no matter where you click on the card.
- If none of the dates have any people assigned, first tap on one of the dates to navigate to the people/places screen and enter some data, then go back to the Contact Journal and try tapping anywhere on the date card with data.


axelherbstreith (approved):
LGTM👍 Tested on Huawei P30 with Android 10

AlexanderAlferov (approved):
LGTM 🚀  Tested on Samsung S8 Android 9.

PR #2094, opened 2021-01-13
Add snapping scroll behavior to the statistic cards (EXPOSUREAPP-4520)

ralfgehrer:
When scrolling horizontally through the statistic cards on the home screen, they should snap to the center of the respective card.


axelherbstreith (approved):
LGTM 👍  tested on Huawei P30 and Nokia 7.1, both Android 10

mtwalli (approved):

PR #2095, opened 2021-01-13
Fix missing submission consent when using TANs (EXPOSUREAPP-4588)

d4rken:
The consent screen within the TAN submission flow did not set the consent value before letting the user proceed.
This caused the user to be able to start submission without consent being set to `true` which causes the submission task to abort.

### Testing (taken from ticket)
 * Register a TAN (within the 1.11.0 (RC1) 
 * press "Abbrechen"
 * in the CWA popup press "don't warn" 
 * Homescreen with positive test result is shown
 * press "andere warnen" 
 * press "einverstanden"
 * in the OS opup press "share"
 * now warn others e.g. via "beenden" > Symptonerfassung abbrechen? JA

*Expected Result*
 * it's possible to warn others

*Actual Result*
 * the positive TAN stucks in the app, it's impossible to warn others


chiljamgossow (approved):
LGTM

ralfgehrer (approved):
Tested on Pixel (Android 10) and WRU-XD.

PR #2102, opened 2021-01-14
Instrumentation tests on CI (EXPOSUREAPP-4480,DEV)

d4rken:
* Clean up CircleCI config to be a bit more readable
* Split into two workflows for different buildtypes to be able to restart them individually if failed
* Cleaned up a few scripts that were re-usable
* Added jobs for instrumentation tests that run on firebase, one for just tests and one for just screenshots

TODO:
* We locally use `fastlane screengrab` but I can't get it to work via firebase so far
* Some tests were just failing, likely due to previously not being maintained as they were not regularly executed. Need to look into those (possibly fix in an extra PR).


harambasicluka:
Please add this to `release/1.12.x` as we don't want to introduce new features for 1.11.0.

d4rken:
Superseeded by #2160

PR #2103, opened 2021-01-14
Layout for Risk History (EXPOSUREAPP-4577)

axelherbstreith:
Adds layouts needed to display the risk history within the contact diary

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/contact_diary_overview_nested_list_item_risk.xml
+++ b/Corona-Warn-App/src/main/res/layout/contact_diary_overview_nested_list_item_risk.xml
@@ -20,81 +20,68 @@
     </data>
 
     <androidx.constraintlayout.widget.ConstraintLayout
-        android:id="@+id/contact_diary_overview_element_nested_container"
+        android:id="@+id/contact_diary_overview_risk_item"
         android:layout_width="match_parent"
         android:layout_height="wrap_content"
+        android:layout_marginVertical="@dimen/spacing_tiny"
         android:focusable="true"
+        android:paddingBottom="@dimen/spacing_tiny"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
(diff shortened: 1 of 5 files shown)


BMItter:
looking good so far for first version.
could you provide a string res too(or did the process changed)? maybe first version according to ticket or our designtool

axelherbstreith:
> looking good so far for first version.
> could you provide a string res too(or did the process changed)? maybe first version according to ticket or our designtool

Yes, currently working on that 👍

BMItr (dismissed):
all right for first version. may be pollished later.
Thx

Corona-Warn-App/src/main/res/layout/contact_diary_overview_nested_list_item_risk.xml
```diff
@@ -0,0 +1,101 @@
...
+        android:id="@+id/contact_diary_overview_element_nested_container"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:focusable="true"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent">
+
+        <androidx.constraintlayout.widget.ConstraintLayout
```
ralfgehrer, line 32: No need to nest a constraint layout within a constraint layout - especially if the inner one is the only child, you can drop the outer one.

BMItr (approved):

Corona-Warn-App/src/main/res/layout/contact_diary_overview_nested_list_item_risk.xml
axelherbstreith: Good point! Is removed 👍

ralfgehrer (approved):

janetback (approved):
UA approved

PR #2109, opened 2021-01-15
Contact diary date formatting (DEV)

chris-cwa:
1. Double checking the date time patterns given by the underlying implementation
2. accept multiple correct values for printed dates with non-ascii chars to avoid unit tests failing because of wrong encoding

Unit tests should neither be failing in circle ci nor on local machine.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/CWADateTimeFormatPatternFactory.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/CWADateTimeFormatPatternFactory.kt
@@ -7,11 +7,15 @@ object CWADateTimeFormatPatternFactory {
 
     fun Locale.shortDatePattern(): String {
         DateTimeFormat.patternForStyle("S-", this).also {
+            /**
+             * since results differ depending on the implementation of the underlying platform,
+             * we make sure that we get, what we expect
+             */
             return when {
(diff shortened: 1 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/util/CWADateTimeFormatPatternFactory.kt
```diff
@@ -0,0 +1,32 @@
+package de.rki.coronawarnapp.contactdiary.util
+
+import org.joda.time.format.DateTimeFormat
+import java.util.Locale
+
+object CWADateTimeFormatPatternFactory {
+
+    fun Locale.shortDatePattern(): String {
+        DateTimeFormat.patternForStyle("S-", this).fixYear().also {
```
d4rken, line 9: Can't we just hardcode the patterns i.e. `"dd/MM/yyyy"` for each case? The current implementation seems very fragile.

chris-cwa: sure, I can change it. But then wen we would be forced to change code anytime we want to support a new region.

axelherbstreith (dismissed):

d4rken (dismissed):
LGTM

axelherbstreith (approved):

ralfgehrer (approved):

PR #2112, opened 2021-01-15
Fix foreground state emissions (DEV)

d4rken:
While working on the statistics I noticed that the foreground detection just wasn't working.
We did not emit the values and relied on setting the static property.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/ProcessLifecycle.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/ProcessLifecycle.kt
@@ -1,0 +1,8 @@
+package de.rki.coronawarnapp.util.di
+
+import javax.inject.Qualifier
+
+@Qualifier
+@MustBeDocumented
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ProcessLifecycle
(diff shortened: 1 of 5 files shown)


ralfgehrer (approved):
LGTM.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/di/ProcessLifecycle.kt
```diff
@@ -0,0 +1,8 @@
+package de.rki.coronawarnapp.util.di
+
+import javax.inject.Qualifier
+
+@Qualifier
+@MustBeDocumented
+@Retention(AnnotationRetention.RUNTIME)
+annotation class ProcessLifecycle
```
chris-cwa, line 8: why do we need this?

d4rken: `ProcressLifecycleOwner.get(): LifecycleOwner`, the interface is generic and there are multiple types of `LifecycleOwner`, so they don't get mixed up in the DI graph, we qualify this specific type of `LifecycleOwner` similar to how we handle `Context`.

d4rken:
```kotlin
class Example constructor(@ProcessLifecycle val lifecycleOwner: LifecycleOwner)
```

chris-cwa (approved):

PR #2115, opened 2021-01-15
Load statistics data from server (EXPOSUREAPP-4514)

d4rken:
This replaces the mock data with data loaded from the servers.

* Respect cache header from server
* Verify data signature
* Use local fallback cache
* Refresh on app foreground/background state
* Filter out unsupported data types
* Sort data by card order received from server
* Check card data for validity (are all key figures there?)

### TODO
- [x] Complete unit  tests

### Testing
* Only DEV and INT currently has statistics data
* Filter logcat for "Statistics"
* Check against TechSpec

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
@@ -1,7 +1,8 @@
 package de.rki.coronawarnapp.statistics
 
-import de.rki.coronawarnapp.server.protocols.internal.stats.KeyFigureCardOuterClass
+import de.rki.coronawarnapp.server.protocols.internal.stats.KeyFigureCardOuterClass.KeyFigure
 import org.joda.time.Instant
+import timber.log.Timber
 
 data class StatisticsData(
     val items: List<StatsItem> = emptyList()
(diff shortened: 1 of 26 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
```diff
@@ -1,30 +1,117 @@
...
 }
 
-sealed class StatsItem(val cardId: Int) {
+sealed class StatsItem(val cardType: Type) {
     abstract val updatedAt: Instant
-    abstract val keyFigures: List<KeyFigureCardOuterClass.KeyFigure>
+    abstract val keyFigures: List<KeyFigure>
+
+    enum class Type(val id: Int) {
+        INFECTION(1),
```
ralfgehrer, line 22: minor: might be a little odd to start with 1 instead of 0

ralfgehrer: Or is this referring to the `cardIdSequence` with values `[1,2,3,4]` of the tech spec?

d4rken: Yes. These are the `cardId`'s defined by the tech spec.

BMItr (approved):
lgtm 👌

ralfgehrer (approved):
LGTM

PR #2118, opened 2021-01-15
Create NumberFormatter for Statistics Values (EXPOSUREAPP-4652)

LukasLechnerDev:
Implementation of the NumberFormatter + Tests

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/util/StatisticsNumberValueFormatter.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/util/StatisticsNumberValueFormatter.kt
@@ -20,11 +20,9 @@ object StatisticsNumberValueFormatter {
         }
 
         return when (decimals) {
-            (-1) -> DecimalFormat("#", DecimalFormatSymbols(locale))
-            0 -> DecimalFormat("#,###", DecimalFormatSymbols(locale))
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/util/StatisticsNumberValueFormatter.kt
```diff
@@ -0,0 +1,28 @@
+package de.rki.coronawarnapp.statistics.util
+
+import java.text.DecimalFormat
+import java.text.DecimalFormatSymbols
+import java.util.Locale
+
+object StatisticsNumberValueFormatter {
```
kolyaopahle, line 7: I'd prefer not having an object, either just define the func by itself or use a class and inject it with dagger

kolyaopahle (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/util/StatisticsNumberValueFormatter.kt
LukasLechnerDev: I changed it to a top-level function.

jurajkusnier (approved):
LGTM

kolyaopahle (approved):

PR #2123, opened 2021-01-18
Date formatter for statistics (EXPOSUREAPP-4654)

jurajkusnier:
- Implement a date formatter used in statistics cards


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/formatter/FormatterStatisticsHelperTest.kt
```diff
@@ -0,0 +1,72 @@
...
+            every { getString(R.string.statistics_primary_value_until_yesterday) } returns R.string.statistics_primary_value_until_yesterday.toString()
+        }
+    }
+
+    @Test
+    fun `formatStatisticsDate test`() {
+        formatStatisticsDate(
+            context = context,
+            localDate = today
+        ) shouldBe context.getString(R.string.statistics_primary_value_today)
```
mtwalli, line 44: I don't think, it is really testing what you want. if you check the output of this function it will be "2131886709".
Basically taking the resource Id value and converting it to a String. I would recommend to hard code the expected result if you are testing for German locale so you what to expect.

mtwalli, line 44: I don't think, it is really testing what you want. if you check the output of this function it will be "2131886709".
Basically taking the resource Id value and converting it to a String. I would recommend to hard code the expected result if you are testing for German locale so you what to expect.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/formatter/FormatterStatistics.kt
```diff
@@ -0,0 +1,41 @@
...
+    return when (localDate) {
+        today -> {
+            context.getString(R.string.statistics_primary_value_until_today)
+        }
+        yesterday -> {
+            context.getString(R.string.statistics_primary_value_until_yesterday)
+        }
+        else -> {
+            localDate.toFormattedDay(context.getLocale())
+        }
```
mtwalli, line 39: For readability: you don't need {}  if it just one statement

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/formatter/FormatterStatistics.kt
```diff
@@ -0,0 +1,41 @@
+@file:JvmName("FormatterStatistics")
```
mtwalli, line 1: is it needed? we don't have a Java code as far as I know it help when you to use the code from Java

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/formatter/FormatterStatisticsHelperTest.kt
jurajkusnier: I'm testing here only if methods use the right string resources, I don't need the real value. Like for example in `FormatterSubmissionHelperTest`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/formatter/FormatterStatistics.kt
jurajkusnier: yes, as it's intended to be used with ViewBinding

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/formatter/FormatterStatisticsHelperTest.kt
LukasLechnerDev: I also think that hardcoded values are better because then you have real values e.g. in test error logs instead of just numbers.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/formatter/FormatterStatistics.kt
```diff
@@ -0,0 +1,41 @@
...
+    val yesterday = today.minusDays(1)
+    return when (localDate) {
+        today -> {
+            context.getString(R.string.statistics_primary_value_until_today)
+        }
+        yesterday -> {
+            context.getString(R.string.statistics_primary_value_until_yesterday)
+        }
+        else -> {
+            localDate.toFormattedDay(context.getLocale())
```
LukasLechnerDev, line 38: Are you sure that we only need to show the date without a prefix like "until/bis" here?

LukasLechnerDev (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/formatter/FormatterStatistics.kt
jurajkusnier: I'm not sure, still waiting for clarification.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/formatter/FormatterStatistics.kt
jurajkusnier: true

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/formatter/FormatterStatisticsHelperTest.kt
jurajkusnier: updated for better readability

mtwalli (approved):
LGTM

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1308,6 +1308,10 @@
     <string name="statistics_primary_value_until_today">"Bis heute"</string>
     <!-- XTXT: Timestamp refers to yesterday's date in incident card -->
     <string name="statistics_primary_value_until_yesterday">"Bis gestern"</string>
+    <!-- XTXT: Timestamp refers to today's date in Reproduction Number card -->
+    <string name="statistics_primary_value_curret">"Aktuell"</string>
```
LukasLechnerDev, line 1312: typo: current

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1316,6 +1316,10 @@
     <string name="statistics_primary_value_until_today">"Bis heute"</string>
     <!-- XTXT: Timestamp refers to yesterday's date in incident card -->
     <string name="statistics_primary_value_until_yesterday">"Bis gestern"</string>
+    <!-- XTXT: Timestamp refers to today's date in Reproduction Number card -->
+    <string name="statistics_primary_value_curret">"Aktuell"</string>
```
LukasLechnerDev, line 1320: typo: current

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/formatter/FormatterStatisticsHelperTest.kt
```diff
@@ -20,49 +26,68 @@ class FormatterStatisticsHelperTest {
     @MockK
     private lateinit var context: Context
 
-    private val today = LocalDate()
-    private val yesterday = LocalDate().minusDays(1)
+    private val slot = slot<String>()
```
LukasLechnerDev, line 29: `slot` doesn't need to be a property and can be defined as a local variable in `setUp()`

LukasLechnerDev (approved):
LGTM! Just some minor things.

Corona-Warn-App/src/main/res/values/strings.xml
jurajkusnier: fixed, thanks

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/util/formatter/FormatterStatisticsHelperTest.kt
jurajkusnier: true

mtwalli (approved):

PR #2125, opened 2021-01-18
Samsung S6: freshly installed app does not open (EXPOSUREAPP-4647)

kolyaopahle:
enabling support vector drawables on devices pre api level 24 results in a crash on launch as we use some features in our drawables that are not supported by the support lib


BMItter:
tested on Android 6,7. looking good, no issues anmyore.

mtwalli (approved):
As I mentioned in the daily this might cause some issues for screenshots on some Android APIs, but more important the App. I just tested taking screenshots on Android 11 and seems fine. In my opinion it not a big issue for UI tests as long as we can take them on a specific API

BMItr (approved):

Issue #2093, opened 2021-01-13
Notification to open app from time to time comes back every few minutes


nicolaus-hee:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Google Pixel 4 XL
- Android version: 11
- App version: 1.10.1

## Describe the bug

This afternoon, I got a notification asking me to periodically open CWA (I assume to avoid Android from killing it in the background to save battery). Never got this notification type before. I couldn't get rid of the notification, though:

Clicked on it, CWA opened and notification disappeared, closed the app. A few minutes later, the notification came back. I repeated opening CWA, notification disappeared again, closed again, notification came back again. Seems to be endless. Had this four times in a row now.

![Screenshot_20210113-163200-2](https://user-images.githubusercontent.com/48563755/104473976-d4a9a680-55bd-11eb-9285-a15b715172f8.png)

## Steps to reproduce the issue

Didn't really cause this. Phone was lying around since >60mins when the first notification popped up. Had opened the app yesterday, phone not restarted since then. As described above, notification keeps coming back without me doing anything.

## Expected behaviour

Reacting to notification / clicking on it should make it disappear for a while (days?).

## Additional context

I'm also affected by #2081 

---
Possibly related:
#2081 
Internal Tracking ID: [EXPOSUREAPP-4594](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4594) duplicate of
Internal Tracking ID: [EXPOSUREAPP-4592](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4592)

---
Internal Tracking ID: [EXPOSUREAPP-4623](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4623)
PR: https://github.com/corona-warn-app/cwa-app-android/pull/2128

MikeMcC399:
@nicolaus-hee 
I expect your issue is directly related to https://github.com/corona-warn-app/cwa-app-android/issues/2081 which you referred to. I also got a monthly reminder from Google today, which I had never seen before. I'm also suffering from the problem which is stopping exposure checks running through.

The notification you are seeing is from
[...]

PR #2128, opened 2021-01-18
Notification to open app from time to time comes back every few minutes (EXPOSUREAPP-4623)

kolyaopahle:
changed ExistingPeriodicWorkPolicy to KEEP for deadman check task to stop notification loop on every app open

### Steps to reproduce
1. launch app without having done an exposure check in the last 35 hours (or set delay in de/rki/coronawarnapp/deadman/DeadmanNotificationScheduler.kt:22 to 0L)
2. wait for notification
3. open app via notification
4. a new notification should only appear after one hour or more not within minutes


ralfgehrer:
According to the ticket, this should go into 1.12. instead

kolyaopahle:
> According to the ticket, this should go into 1.12. instead

whoops, changed base and rebased commits

d4rken:
Can you elaborate on why the `REPLACE` causes it, and the `KEEP` fixes the behavior?

kolyaopahle:
Currently the app schedules a check for the deadman notification with a periodic worker that will check how long the last successful submission is in the past and will schedule a new one time worker with a run delay of the difference of that time and 35 hour's

If the last check was run more than 35 hours ago a one time worker to display the notification is scheduled immediately.

The periodic worker is scheduled on each app launch, which results in a run of the worker on each app start wenn using the REPLACE option, which in turn results in a scheduling of the one time worker to display the notification, which will keep you in the feedback loop.

Previously if you swiped away the notification you would also observe the 1 notification per hour behavior but opening the app via the notification would result in the notification spam

d4rken:
@kolyaopahle Thanks for the explanation. This to be an important part of the deadman mechanism, would it be possible to add a shorter explanation as comment to the code where the worker is setup?

kolyaopahle:
> @kolyaopahle Thanks for the explanation. This seems to be an important part of the deadman mechanism, would it be possible to add a shorter explanation as comment to the code where the worker is setup?

Done :)

AlexanderAlferov (approved):
LGTM 🚀

d4rken (approved):

MikeMcC399:
Link to #2093.

PR #2129, opened 2021-01-18
Contact Journal Location/Person not announced (EXPOSUREAPP-4459)

axelherbstreith:
Fixes an issue where only the names of locations and persons in the contact journal overview screen were announced but not the types e.g. "John Doe" instead of "Person John Doe".

To test this changes, activate TalkBack, navigate to the contact journal overview screen and focus a person or location within a day element. 

device-2021-01-18-143447


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/overview/adapter/ContactDiaryOverviewNestedAdapter.kt
```diff
@@ -39,6 +39,10 @@ class ContactDiaryOverviewNestedAdapter(
             { key, _ ->
                 contactDiaryOverviewElementImage.setImageResource(key.drawableId)
                 contactDiaryOverviewElementName.text = key.text
+                contactDiaryOverviewElementName.contentDescription = if (key.type == ListItem.Type.LOCATION)
```
ralfgehrer, line 42: Please use a `when` statement and list types explicitly (default is empty) so when we add additional types, they don't accidentally default to the peson type.

LukasLechnerDev: I also think that using a `when` is better here.

axelherbstreith: Good point, I will change it

LukasLechnerDev (dismissed):
LGTM, tested on Google Pixel 3a with Android 11

LukasLechnerDev (approved):

ralfgehrer (approved):

PR #2132, opened 2021-01-18
Add TrendArrowView (EXPOSUREAPP-4621)

chiljamgossow:
Implementation TrendArrowView


Corona-Warn-App/src/main/res/layout/statistics_trend_view.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<FrameLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
LukasLechnerDev, line 2: This additional nested `FrameLayout` is not necessary, because the custom view is a `FrameLayout` already.

AlexanderAlferov (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/TrendArrowView.kt
```diff
@@ -2,20 +2,47 @@ package de.rki.coronawarnapp.statistics.ui
...
         trend: KeyFigureCardOuterClass.KeyFigure.Trend,
         trendSemantic: KeyFigureCardOuterClass.KeyFigure.TrendSemantic
     ) {
-        Timber.d("setTrend(trend=$trend, trendSemantic=$trendSemantic)")
+        imageView.rotation = when (trend) {
+            INCREASING -> -45F
+            DECREASING -> 45F
+            else -> 0F
+        }
+
```
LukasLechnerDev, line 39: I think we also have to set a `contentDescription` here. The Story Specification says: 

> Meaningful icons like the "tendency" icons (red arrow up etc.) should have image/icon descriptions with their meaning (respective text from help/info dialog content legend)

Corona-Warn-App/src/main/res/layout/statistics_trend_view.xml
chiljamgossow: True, I removed it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/TrendArrowView.kt
chiljamgossow: I added the content description

PR #2136, opened 2021-01-18
Make talkback speak out german comma values appropriately (EXPOSUREAPP-4665)

LukasLechnerDev:
Decimal numbers are spoken out correctly when the locale is "en". 

But unfortunately, Talkback ignores the German "Komma" (,) Symbol and speaks out numbers like 134,4 as 1344. 

This is probably also not working for our other supported languages.

This PR is just about the translation for the decimal separator.

### DIFF
--- a/Corona-Warn-App/src/main/res/values-de/strings.xml
+++ b/Corona-Warn-App/src/main/res/values-de/strings.xml
@@ -1300,6 +1300,9 @@
     <!-- XTXT: Text displayed in the bottom of submission statistics card-->
     <string name="statistics_card_submission_bottom_text">"bundesweit warnende Personen"</string>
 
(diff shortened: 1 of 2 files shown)


janetback (approved):
UA reviewed the spoken comma :-).

d4rken:
> But unfortunately, Talkback ignores the German "Komma" (,) Symbol and speaks out numbers like 134,4 as 1344. 

Are we sure that this isn't a TTS setting issue? That we don't just need to set the locale correctly somewhere so the TTS works correctly?
Splitting numbers and manually injecting "Komma" into the content description seems not nice.

There is also `LocaleSpan` that we could maybe use to force a locale for the TTS?
https://developer.android.com/reference/android/text/style/LocaleSpan.html

LukasLechnerDev:
This PR isn't needed anymore as we found a solution that doesn't need translations.

PR #2138, opened 2021-01-18
Statistics 7-Day-R-Value (EXPOSUREAPP-4513)

chiljamgossow:
Implements 7-day-R-value card
Fixes padding and shadow on home screen

Missing: Info button navigation as screen is not merged


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsCardPaddingDecorator.kt
```diff
@@ -22,15 +22,17 @@ class StatisticsCardPaddingDecorator(
         val resources = parent.context.resources
 
         val adapter = parent.adapter
+        val distance = resources.getDimensionPixelSize(cardDistance)
```
d4rken, line 25: The card padding is larger than on the mockups now though. On the mockups the horizontal distance between cards is half of the distance between vertical cards. I think both look okay, and we can merge this because looks better than if it's inconsistent.

What do you think? @chiljamgossow

Corona-Warn-App/src/main/res/layout/home_statistics_cards_sevendayrvalue_layout.xml
```diff
@@ -57,16 +58,33 @@
...
 
         <TextView
-            android:id="@+id/secondary_value_label"
+            android:id="@+id/secondary_label"
             style="@style/StatisticsCardValueLabel"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
             android:layout_marginHorizontal="@dimen/spacing_normal"
-            android:layout_marginTop="@dimen/spacing_normal"
+            android:layout_marginVertical="@dimen/spacing_normal"
```
d4rken, line 66: We should maybe remove some margin/padding so that there is more space for text. Many people will use increased font-sizes on their devices.

Screenshot from 2021-01-18 18-51-54

d4rken (commented):
Looks good but I'd like to discuss a few things.

I think the info button graphic is also a bit too large, though this didn't change here.

Corona-Warn-App/src/main/res/layout/home_statistics_cards_sevendayrvalue_layout.xml
chiljamgossow: I reduced some of the margins

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsCardPaddingDecorator.kt
chiljamgossow: I corrected it

Corona-Warn-App/src/main/res/layout/home_statistics_cards_sevendayrvalue_layout.xml
```diff
@@ -4,11 +4,12 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <androidx.constraintlayout.widget.ConstraintLayout
-        android:layout_width="match_parent"
-        android:layout_height="wrap_content"
+        android:layout_width="300dp"
```
ralfgehrer, line 7: Why the fixed values?

ralfgehrer (approved):
Tested on Pixel (Android 10) in light and dark mode.

Corona-Warn-App/src/main/res/layout/home_statistics_cards_sevendayrvalue_layout.xml
chiljamgossow: They are fixed, so all cards have the same size. I put them here so the preview is correct. Should move them to dimens though.

PR #2144, opened 2021-01-19
New release info screen - UI (EXPOSUREAP-4591)

SamuraiKek:
This PR adds the UI for the new release info screen. Logic will be implemented in a different PR.


d4rken (approved):
lgtm

d4rken:
If the "info" items always look the same, maybe in a future versions we could use a recyclerview to create the entries dynamically?

SamuraiKek:
> If the "info" items always look the same, maybe in a future versions we could use a recyclerview to create the entries dynamically?

Sound like a good idea. I guess we'll see with release 1.13.

Corona-Warn-App/src/main/res/values/styles.xml
```diff
@@ -257,6 +257,12 @@
         <item name="android:textColor">@color/colorTextPrimary1</item>
     </style>
 
+    <style name="subtitleBoldSixteen" parent="@style/TextAppearance.MaterialComponents.Subtitle1">
```
harambasicluka, line 260: Could you please clarify with the designers if we should introduce a new typographie?

harambasicluka: A new text style is wanted.

harambasicluka (approved):

Corona-Warn-App/src/main/res/values/styles.xml
SamuraiKek: Talked to the designers and it's indeed a new text style.

PR #2145, opened 2021-01-19
Make talkback speak out decimal values appropriately (EXPOSUREAPP-4665)

LukasLechnerDev:
Talkback only works appropriately in German with decimal numbers when setting the text as a localized spannable. 

This implementation provides a helper function that converts a string to a localized spannable.


ralfgehrer (approved):

ralfgehrer:
Code looks good. Hard to test - is it applied anywhere, yet?

LukasLechnerDev:
@ralfgehrer Unfortunately it is not applied yet so you can't test it yet.

d4rken (approved):

PR #2148, opened 2021-01-19
Privacy update for exposure history (EXPOSUREAPP-4590)

harambasicluka:
Updating english, turkish and german data privacy for the upcoming exposure history feature. 

This change also introduces some reformatting as the privacy texts were directly provided as HTML. Hopefully they will look the same for all future updates.


Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -199,7 +199,7 @@ <h2>
     Smartphone nicht über diese Zufalls-IDs identifiziert werden kann. Die von
     Ihrem Smartphone ausgesendeten eigenen Begegnungsdaten und die
     aufgezeichneten Begegnungsdaten der Personen, mit denen Sie Kontakt hatten,
-    werden auf Ihrem Smartphone gespeichert und jeweils nach 14 Tagen gelöscht.
+    werden auf Ihrem Smartphone gespeichert und jeweils nach 14 Tage gelöscht.
```
harambasicluka, line 202: @janetback I used the html file provided by the lawyers but I think that's wrong...

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -759,12 +795,9 @@ <h1>
     </li>
     <li>
         das Recht, den behördlichen
-		<a
-					href="https://www.rki.de/DE/Content/Institut/OrgEinheiten/Datenschutz/Datenschutz_node.html"
-				>
         Datenschutzbeauftragten des RKI
-		</a>
```
ralfgehrer, line 766: Closing tag shall not be deleted

ralfgehrer, line 766: Why was the link deleted? If I recall correctly, this was changed in the last update.

harambasicluka: I wasn't aware of this discussion, but also if it's included the link is still not clickable even if it looks like that.

So I would add the link again in the HTML, but we should still should check why it isn't clickable.

https://user-images.githubusercontent.com/64483219/105026600-018f0b00-5a4f-11eb-95c8-2f431a6aa6b2.mp4

ralfgehrer (dismissed):

Corona-Warn-App/src/main/assets/privacy_de.html
janetback: Hi @harambasicluka, hier sollte es tatsächlich 'Tagen' heißen.

janetback (approved):

ralfgehrer (approved):

BMItr (approved):

PR #2152, opened 2021-01-19
Contact Journal Edit Screen Title not accounced (EXPOSUREAPP-4456)

axelherbstreith:
Fixes a bug in the contact journal that announced the app name + the title when opening the edit screen views e.g. "Corona-Warn, Edit Places". Speech output is not just "Edit Places". 

In addition, this changed behavior is applied to all views within the contact journal that uses the default android `Toolbar` as header.

### DIFF
--- a/Corona-Warn-App/src/main/AndroidManifest.xml
+++ b/Corona-Warn-App/src/main/AndroidManifest.xml
@@ -81,6 +81,7 @@
             android:exported="false"
             android:screenOrientation="portrait"
             android:launchMode="singleTop"
+            android:label=""
             android:theme="@style/AppTheme.ContactDiary"
             android:windowSoftInputMode="adjustResize" />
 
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -81,6 +81,7 @@
             android:exported="false"
             android:screenOrientation="portrait"
             android:launchMode="singleTop"
+            android:label=""
```
ralfgehrer, line 84: Would be good to add a comment on why it is set to an empty string. Maybe use a string resource with a descriptive name.

ralfgehrer (commented):
Successfully tested on Pixel (Android 10) - German Talk Back

Corona-Warn-App/src/main/AndroidManifest.xml
axelherbstreith: Done 👍

ralfgehrer (approved):

BMItr (approved):

PR #2153, opened 2021-01-19
Screen 7-day Incidence (EXPOSUREAPP-4512)

LukasLechnerDev:
Implementation of the 7-day incidence screen.

Missing: Info Button not yet leading to info screen.
Known Issue: Text cut off when setting font size to very large


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/cards/IncidenceCard.kt
```diff
@@ -23,6 +27,23 @@ class IncidenceCard(parent: ViewGroup) :
...
-        item.stats as IncidenceStats
+
+        viewBinding.value.iconInfo.setOnClickListener {
+            item.onHelpAction.invoke(item.stats)
+        }
+
+        val stats = item.stats as IncidenceStats
+
+        val primaryFigure =
+            stats.keyFigures.firstOrNull { it.rank == KeyFigureCardOuterClass.KeyFigure.Rank.PRIMARY }
```
jurajkusnier, line 38: `IncidenceStats` has now getter `sevenDayIncidence` and it's always not null

chiljamgossow (approved):
LGTM, label and info button are addressed in other PRs

jurajkusnier (changes requested):
Looks good, just please update `IncidenceCard` according to comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/cards/IncidenceCard.kt
LukasLechnerDev: That's nice. I updated the PR.

jurajkusnier (approved):

PR #2160, opened 2021-01-19
CircleCI config overhaul, instrumentation tests & screenshot automation (EXPOSUREAPP-4480,DEV)

d4rken:
@mtwalli, @AlexanderAlferov and me overhauled the CircleCI config, added instrumentation testing and automatic screenshots.

Highlights:
* Refactored the config (yes you can refactor YAML :dizzy_face:) to make use more use of re-usable functions
* Use larger instances for faster test execution
* Adjusted gradle/JVM options to improve performance and use larger instances better
* Run instrumentation tests via macOS CircleCI executor
* Run automated screenshot routine via fastlane when a new release candidate is build
* Improved caching behavior
* Split works for better job restart and improved dependencies between tasks
* Renamed tasks and work flows to be easier to understand

We tried Firebase Testlab, but it was not working out, due to lack of control over the whole process (more of a black box) and inability to use fastlane with it. There were also issues with our mocking library which could fail on newer API emulators if not run on `x86_64` architecture emulators, and firebase testlab doesn't allow us to enforce it.

Because we previously didn't run instrumentation tests via CI, a few tests fell into disarray and will need fixing. The current config setup is already a huge improvement over the previous config and I think should merge it despite the failing instrumentation tests. Until we enforce the check to pass for branches, there should be no issue. A follow up PR can then be made stress free to fix the failing tests.

This was initially targeted for 1.12, but as this will require adjusting the branch restrictions, it would make sense to merge this into 1.11, we share the same branch restrictions between all branches. Then the only problematic branch would be 1.10.x which will soon be superseeded by 1.11.x.

Once the config is merged I will change the restrictions and all open PRs will have to pull the new config to run the correct tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -485,7 +485,6 @@ workflows:
           requires:
             - unit_tests_device_release
       - instrumentation_tests_device
-      - device_screenshots
 
   check_buildtype_device_for_testers:
     jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -257,9 +257,9 @@ jobs:
             - ./project
       - compress-path:
           input: ./Corona-Warn-App/build/test-results
-          output: /tmp/unit_tests_device_release-$CIRCLE_SHA1.zip
+          output: /tmp/unit_tests_device_release.zip
       - store_artifacts:
-          path: /tmp/unit_tests_device_release-$CIRCLE_SHA1.zip
+          path: /tmp/unit_tests_device_release.zip
           destination: zips
 
   unit_tests_device_for_testers_release:
@@ -328,9 +328,9 @@ jobs:
           destination: reports
       - compress-path:
           input: ./Corona-Warn-App/build/reports
-          output: /tmp/ktlint_device_release_check-$CIRCLE_SHA1.zip
+          output: /tmp/ktlint_device_release_check.zip
       - store_artifacts:
-          path: /tmp/ktlint_device_release_check-$CIRCLE_SHA1.zip
+          path: /tmp/ktlint_device_release_check.zip
           destination: zips
 
   ktlint_device_for_testers_release_check:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -260,7 +260,7 @@ jobs:
           output: /tmp/unit_tests_device_release.zip
       - store_artifacts:
           path: /tmp/unit_tests_device_release.zip
-          destination: zips
+          destination: zips/unit_tests_device_release.zip
 
   unit_tests_device_for_testers_release:
     executor: android/android
@@ -326,12 +326,6 @@ jobs:
       - store_artifacts:
           path: Corona-Warn-App/build/reports
           destination: reports
-      - compress-path:
-          input: ./Corona-Warn-App/build/reports
-          output: /tmp/ktlint_device_release_check.zip
-      - store_artifacts:
-          path: /tmp/ktlint_device_release_check.zip
-          destination: zips
 
   ktlint_device_for_testers_release_check:
     executor: android/android
@@ -471,6 +465,12 @@ jobs:
       - store_artifacts:
           path: fastlane/metadata/android
           destination: screenshots
+      - compress-path:
+          input: ./fastlane/metadata/android
+          output: /tmp/device_screenshots.zip
+      - store_artifacts:
+          path: /tmp/device_screenshots.zip
+          destination: zips/device_screenshots.zip
 
 
 #######################
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -434,6 +434,12 @@ jobs:
       - kill-all-emulators
       - store_test_results:
           path: Corona-Warn-App/build/outputs/androidTest-results
+      - compress-path:
+          input: ./Corona-Warn-App/build/outputs/androidTest-results
+          output: /tmp/instrumentation_tests_device.zip
+      - store_artifacts:
+          path: /tmp/instrumentation_tests_device.zip
+          destination: zips/instrumentation_tests_device.zip
 
   device_screenshots:
     macos:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,16 +2,23 @@ version: 2.1
 orbs:
   android: circleci/android@0.2.1
   sonarcloud: sonarsource/sonarcloud@1.0.2
+
+#######################
+# Commands section
+# For code reuse.
+#######################
 commands:
   install-ndk: android/install-ndk
   restore-android-build-cache: android/restore-build-cache
   save-android-build-cache: android/save-build-cache
   scan-sonar: sonarcloud/scan
+
   restore-gradle-cache:
     description: "Restore gradle caches"
     steps:
       - restore_cache:
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "Corona-Warn-App/build.gradle" }}-{{ checksum  "Server-Protocol-Buffer/build.gradle" }}
+
   save-gradle-cache:
     description: "Save gradle caches"
     steps:
@@ -19,17 +26,7 @@ commands:
           paths:
             - ~/.gradle
           key: jars-{{ checksum "build.gradle" }}-{{ checksum  "Corona-Warn-App/build.gradle" }}-{{ checksum  "Server-Protocol-Buffer/build.gradle" }}
-  require-version-bump:
-    description: "Require version bump for binary assembling"
-    steps:
-      - run:
-          name: "Check if assemble required"
-          command: |
-            last_commit=$(git log -1 --pretty=%B)
-            if [[ $last_commit != *"Version bump"* ]]; then
-              circleci-agent step halt
-              echo "Skipping job"
-            fi
+
   run-gradle-cmd:
     description: "Running gradle command with environment options"
     parameters:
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -276,22 +400,101 @@ jobs:
...
-      - ktlint_device_for_testers_release_check
-      - detekt
+      - quick_build_device_release_no_tests
+      - unit_tests_device_release
       - run_sonar:
           requires:
-            - device_release_unit_tests
+            - unit_tests_device_release
+      - instrumentation_tests_device
+      - device_screenshots
```
d4rken, line 488: Will be removed once the current run through passes again.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/contactdiary/DiaryData.kt
```diff
@@ -14,22 +14,26 @@ object DiaryData {
     val LIST_ITEMS = listOf(
         ListItem.Data(
             R.drawable.ic_contact_diary_person_item,
-            "Max Mustermann"
+            "Max Mustermann",
+            ListItem.Type.PERSON
```
d4rken, line 18: Caused instrumentation test compile error.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/util/security/DBPasswordTest.kt
```diff
@@ -27,10 +45,29 @@ class DBPasswordTest {
 
     @Before
     fun setUp() {
+        MockKAnnotations.init(this)
+        mockkObject(AppInjector)
```
d4rken, line 49: Was failing and easier to fix than others.

d4rken:
Example screenshot execution from another run:
https://43049-268027139-gh.circle-artifacts.com/0/screenshots/screenshots.html

d4rken:
* The screenshot job takes quite long ~50-100 minutes, only executed on releases though.
* Instrumentation test take around 17min. Normal tests are around 10 min. So this adds 7 min to our test pipeline time, but I think that is just the price to pay and will be well worth it. Maybe someone has a good idea on how we can further speed this up? The alternative would be to either live with it, or make the test result non-blocking... :thinking:

.circleci/config.yml
```diff
@@ -59,21 +65,144 @@ commands:
...
+    description: "Setup Google Cloud access."
+    steps:
+      - run:
+          name: Store Google Service Account
+          command: echo $GCLOUD_SERVICE_KEY_BASE64 | base64 -di > ${HOME}/gcloud-service-key.json
+      - run:
+          name: Authorize gcloud and set config defaults
+          command: |
+            sudo gcloud auth activate-service-account --key-file=${HOME}/gcloud-service-key.json
+            sudo gcloud --quiet config set project ${GOOGLE_PROJECT_ID}
```
mtwalli, line 102: Do we still need Google Cloud access?

.circleci/config.yml
```diff
@@ -59,21 +65,144 @@ commands:
...
-            GRADLE_OPTS: -Xmx1536m -XX:+HeapDumpOnOutOfMemoryError -Dorg.gradle.caching=true -Dorg.gradle.configureondemand=true -Dkotlin.compiler.execution.strategy=in-process -Dkotlin.incremental=false
+            JVM_OPTS: -Xmx4096m
+            GRADLE_OPTS: >
+              -Xmx1536m -XX:+HeapDumpOnOutOfMemoryError
+              -Dorg.gradle.caching=true
+              -Dorg.gradle.configureondemand=true
+              -Dkotlin.compiler.execution.strategy=in-process
+              -Dkotlin.incremental=false
+
+  skip-for-external-pull-requests:
```
mtwalli, line 81: I think this also is not needed, if we run it only for release build, right?

.circleci/config.yml
```diff
@@ -276,22 +400,100 @@ jobs:
...
+      #      - kill-all-emulators
+      #      - run-emulator-api23
+      #      - run-gradle-cmd:
+      #          desc: Run instrumentation test command
+      #          cmd: |
+      #            :Corona-Warn-App:connectedDeviceForTestersDebugAndroidTest \
+      #            --info \
+      #            --continue \
+      #            -Pandroid.testInstrumentationRunnerArguments.notAnnotation=testhelpers.Screenshot \
+      #            -Pandroid.testInstrumentationRunnerArguments.clearPackageData=true
```
mtwalli, line 434: Please remove it if not needed anymore?

.circleci/install-android-sdk.sh
```diff
@@ -0,0 +1,30 @@
...
+echo "d975f751698a77b662f1254ddbeed3901e976f5a" > "$ANDROID_SDK_ROOT/licenses/intel-android-extra-license"
+
+SDKMANAGER=$ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager
+
+$SDKMANAGER "platform-tools"
+$SDKMANAGER "platforms;android-29"
+$SDKMANAGER "build-tools;29.0.3"
+$SDKMANAGER "ndk-bundle"
+$SDKMANAGER "system-images;android-29;google_apis;x86_64"
+$SDKMANAGER "system-images;android-23;google_apis;x86_64"
```
mtwalli, line 27: API 23 is not needed, from what I see both instrumentation tests and screenshots test are running on API 29

Corona-Warn-App/build.gradle
```diff
@@ -196,6 +196,12 @@ android {
             includeAndroidResources = true
             returnDefaultValues = true
         }
+
+        // Using orchestration toegether with mockk on x86 (32bit) emulator images crashes
+        // Leaving this in here as reminder
+        // https://github.com/android/android-test/issues/352
+        // https://github.com/mockk/mockk/issues/466
+        // execution 'ANDROIDX_TEST_ORCHESTRATOR'
```
mtwalli, line 204: 👍🏼

.circleci/config.yml
d4rken: AFAICT no, I removed it.

.circleci/config.yml
d4rken: We need this to skip sonar for external pull requests. Otherwise the sonar check will fail and external PRs can only be merged with admin permission. By skipping the check and ending it with a positive result, others can merge external PRs too. Not ideal that sonarr is skipped but also not ideal that only admins can merge external PRs.

.circleci/config.yml
d4rken: I want to later add API23 tests, but you are right, we can re-add it later.

.circleci/install-android-sdk.sh
d4rken: Removed for now, will make execution a little faster, but later we will want api23 tests too.

.circleci/install-android-sdk.sh
```diff
@@ -0,0 +1,29 @@
...
+curl https://dl.google.com/android/repository/commandlinetools-mac-6858069_latest.zip -o cmdline-tools.zip
+
+unzip cmdline-tools.zip -d $ANDROID_SDK_ROOT/cmdline-tools
+mv $ANDROID_SDK_ROOT/cmdline-tools/cmdline-tools $ANDROID_SDK_ROOT/cmdline-tools/latest
+
+mkdir -p "$ANDROID_SDK_ROOT/licenses"
+
+echo "24333f8a63b6825ea9c5514f83c2829b004d1fee" > "$ANDROID_SDK_ROOT/licenses/android-sdk-license"
+echo "84831b9409646a918e30573bab4c9c91346d8abd" > "$ANDROID_SDK_ROOT/licenses/android-sdk-preview-license"
+echo "d975f751698a77b662f1254ddbeed3901e976f5a" > "$ANDROID_SDK_ROOT/licenses/intel-android-extra-license"
```
mtwalli, line 18: ❓ what these hex-strings mean?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/contactdiary/DiaryData.kt
mtwalli: Yeah, I think we should inform everyone to check  instrumentation tests on their machines after UI changes.

.circleci/install-android-sdk.sh
d4rken: A checksum related to the specific license that is accepted.

kaddaSz:
Could you also Zip the ScreenShots? So I could just donwload the zip. The idea is, to add this zip later also to the release notes.

d4rken:
This would be merged after we have finished the statistics stuff so there is no delay due to pipeline issues.

d4rken:
> Could you also Zip the ScreenShots? So I could just donwload the zip. The idea is, to add this zip later also to the release notes.

I think so. Will check. We could also link the artifacts on CircleCI directy, though the link will probably break after a few weeks? CircleCI artifact storage times out at some time.
 
> The same would make sense to the test results I guess.

What do you mean by that?

kaddaSz:
> > The same would make sense to the test results I guess.
> 
> What do you mean by that?

https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/8541/workflows/2a8cd323-f037-4369-818a-eaf96ad45373/jobs/43183/artifacts

There are also a lot of artifacts, which you maybe could zip? I could use these results for our internal reporting, but it would be easier to just download a zip.

d4rken:
Screenshots, Instrumentation tests (release), unit tests (release) are now uploaded as .zip too.

AlexanderAlferov (dismissed):
LGTM

Issue #2035, opened 2021-01-06
After Positive Test Result: Warn Others results in "4000 Etwas ist schiefgelaufen. Error during web request: code=400"


KlugB:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: LG G4
- Android version: 6.0
- App version: 1.10.1  - ENF Version: 18204516000

## Describe the bug

Error message "4000 Etwas ist schiefgelaufen. Error during web request: code=400"

## Steps to reproduce the issue

Andere Warnen - Einverstanden - Symptomerfassung - Symptom-Beginn - Fertig - "Datenabruf"

![image](https://user-images.githubusercontent.com/25726929/103761785-858cd000-5017-11eb-929b-e9e09989d85c.png)
![image](https://user-images.githubusercontent.com/25726929/103761860-a48b6200-5017-11eb-9db8-b9956960ea2d.png)

So it seems impossible on my device to share my "Zufalls-IDs" with others.

It is definitely not a network/firewall problem. Tried rebooting. All other apps have normal network access.

---

PR #2161

--- 
Internal Tracking ID: [EXPOSUREAPP-4515](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4515)

dsarkar:
@KlugB, Thanks for reporting. 

- Did you try this https://www.coronawarn.app/en/faq/#access-list?

MikeMcC399:
@dsarkar 
https://www.coronawarn.app/en/faq/#access-list is about http error 901 (see https://github.com/corona-warn-app/cwa-app-android/issues/785 and https://github.com/corona-warn-app/cwa-app-android/issues/1124)

Perhaps this error (code = 400) is a bad http request?

MikeMcC399:
@KlugB 
Sorry to read that you have been tested positive! I hope things go well for you!

You might want to try submitting your test result without inputting symptoms, in case there is something wrong in the symptom part of the code. That way you might at least be able to warn other people without delay. The only difference is that the calculation about how infectious you were at the time of each contact will not be so specific.

KlugB:
[...]

PR #2161, opened 2021-01-19
Error message after positive test result (EXPOSUREAPP-4515)

mtwalli:
Fix improper message in cases where:
🛠️   Tan can not be retrieved.
🛠️  Keys submission can be done.

In those two cases `BadRequestException` is wrapped into `TanPairingException`  that offers an insightful message to the end the user.

Unfortunately nothing can be done from the App side ,but to offer the end user a proper message. 
This was the case before somehow this message got lost due to some code refactoring.

### Testing
Keep trying the same follow a few times.
⚠️ *MAKE SURE* to reset the App to clear cached data.

device-2021-01-20-000424 device-2021-01-20-000440


LukasLechnerDev:
Code looks good! Is it somehow possible to test this?

mtwalli:
> Code looks good! Is it somehow possible to test this?

Yes, you should try to submit positive test result many times , at some point error will be shown. Please after each submission reset the App and try again.

LukasLechnerDev (approved):

chris-cwa (approved):
LGTM!

MikeMcC399:
Link to #2035.

PR #2167, opened 2021-01-20
Polishing statistics (EXPOSUREAPP-4693)

jurajkusnier:
- Unify statistics cards design


Corona-Warn-App/src/main/res/values/dimens.xml
```diff
@@ -145,6 +145,6 @@
     <!-- Contact Diary -->
     <dimen name="contact_diary_list_item">64dp</dimen>
     <dimen name="contact_diary_nested_recyclerview_margin">60dp</dimen>
-    <dimen name="statistics_card_height">328dp</dimen>
+    <dimen name="statistics_card_height">352dp</dimen>
```
harambasicluka, line 148: IMHO this only mitigates the problem, one pixel phones you are quite limited regarding the font size. I'll test it on a Samsung device.

jurajkusnier: still working on a better layout fix

harambasicluka:
Galaxy 8 with an strange font, biggest available font size and in large:
image

standard font with largest possible size:
image

Corona-Warn-App/src/main/res/values/dimens.xml
LukasLechnerDev, line 148: This is differs to the figma designs and uses too much whitespace when using regular font size. 

As discussed on Teams we should revert it and probably implement a dynamic card height until Friday.

LukasLechnerDev (changes requested):
Please revert card height, otherwise LGTM.

LukasLechnerDev (approved):

Corona-Warn-App/src/main/res/layout/home_statistics_cards_incidence_layout.xml
```diff
@@ -22,28 +22,19 @@
...
-            android:layout_height="wrap_content"
-            app:layout_constraintEnd_toEndOf="parent"
-            app:layout_constraintTop_toTopOf="parent" />
-
         <TextView
             android:id="@+id/value_label"
             style="@style/StatisticsCardValueLabel"
             android:layout_width="0dp"
             android:layout_height="wrap_content"
-            android:layout_marginTop="16dp"
```
chiljamgossow, line 38: I think @d4rken just changed that to be fixed

chiljamgossow (commented):
I believe you reverted some recent changes made for increased font size

jurajkusnier:
> I believe you reverted some recent changes made for increased font size

Yes, that's possible, I adjusted some changes to have the cards look the same. Font size adaptation will be part of another PR anyway.

PR #2175, opened 2021-01-21
Fix flickering of statistics cards when returning from statistics information screen (EXPOSUREAPP-4704)

LukasLechnerDev:
This PR overrides `equals()` of the Statistics Items so that it ignores the `onClickListener` that is reset when resuming from the information screen, causing the statistics cards to re-draw. 

The respective Jira Ticket contains more technical information about the bug.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsHomeCard.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsHomeCard.kt
@@ -47,10 +47,25 @@ class StatisticsHomeCard(
         }.let { statsAdapter.update(it) }
     }
 
+    @Suppress("EqualsOrHashCode") // ignore warning: custom implementation of hashCode() not needed
     data class Item(
         val data: StatisticsData,
         val onHelpAction: (StatsItem) -> Unit
     ) : HomeItem {
         override val stableId: Long = Item::class.java.name.hashCode().toLong()
+
+        // ignore onHelpAction so that view is not re-drawn when only the onHelpAction click listener is updated
+        // Auto-generated
+        override fun equals(other: Any?): Boolean {
+            if (this === other) return true
+            if (javaClass != other?.javaClass) return false
+
+            other as Item
+
+            if (data != other.data) return false
+            if (stableId != other.stableId) return false
+
+            return true
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


d4rken:
What do you think about the concerns that the clicklistener could become stale?
Is it possible that the we return to the home screen, the list data is still in the adapter, but we need to update the action?
Can the adapter data outlive the viewmodel?

AlexanderAlferov (approved):
No flickering 👍

LukasLechnerDev:
@d4rken 

> What do you think about the concerns that the clicklistener could become stale?

What exactly do you mean by "stale"?
 

> Is it possible that the we return to the home screen, the list data is still in the adapter, but we need to update the action?

I can't think of a scenario where this is the case. 


> Can the adapter data outlive the viewmodel?

The `HomeFragmentViewModel` has the same lifetime as the `HomeFragment` in which the `HomeAdapter` and the `StatisticsCardAdapter` reside. Therefore I don't think that the viewmodel can outlive the adapter.

d4rken:
> What exactly do you mean by "stale"?

If the fragment or rather the adapter data outlives the viewmodel, then the onclicklisteners would be stale, meaning that reference a viewmodel that no longer exist (well it does, as a memory leak).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsHomeCard.kt
```diff
@@ -47,10 +47,25 @@ class StatisticsHomeCard(
...
+    @Suppress("EqualsOrHashCode") // ignore warning: custom implementation of hashCode() not needed
     data class Item(
         val data: StatisticsData,
         val onHelpAction: (StatsItem) -> Unit
     ) : HomeItem {
         override val stableId: Long = Item::class.java.name.hashCode().toLong()
+
+        // ignore onHelpAction so that view is not re-drawn when only the onHelpAction click listener is updated
+        // Auto-generated
+        override fun equals(other: Any?): Boolean {
```
d4rken, line 59: This works, but if you override `equals` please also also override `hashcode`.

d4rken (changes requested):
Fixes the issue, but please override `hashcode` too, otherwise other implementations (recyclerview/adapter code) could cause subtle bugs due to the contract between equals and hashcode being broken.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsHomeCard.kt
LukasLechnerDev: Done.

d4rken (approved):

PR #2180, opened 2021-01-21
New release info screen logic (EXPOSUREAPP-4591)

SamuraiKek:
This PR adds the logic behind the new release info screen.

How to test:

- Start by installing an older version of the app.
- Install the new app with this branch.
- Upon opening the app the New Release info screen should show up.
- Pressing "Weiter" or the "X" button will take you to the home screen.
- Closing the app and opening it again should not show the New Release screen anymore.
- By going to the menu on the top right of the home screen -> App-Information, a new entry "Neue Funktionen" should be visible at the bottom of the list. Tapping it should open the New Release screen. 
- Pressing "Weiter" or the "X" button will take you back to the App-Information screen.

- Try deleting the app data and do the onboarding again.
- The New Release screen should NOT show up.


Corona-Warn-App/src/main/res/values-de/release_info_strings.xml
```diff
@@ -7,7 +7,7 @@
     <!-- XHED: Title for the release info screen -->
     <string name="release_info_header">"Neue Funktionen"</string>
     <!-- XHED: Version title for the release info screen -->
-    <string name="release_info_version_title">"Release <xliff:g id="release_app_version" example="1.12">"%1$d"</xliff:g></string>
+    <string name="release_info_version_title">"Release <xliff:g id="release_app_version" example="1.12">"%1$s"</xliff:g></string>
```
harambasicluka, line 10: I think this isn't needed or at least we don't use it everywhere else: `xliff:g`

harambasicluka (commented):
I have three small remarks, none of them is blocking.

The implementation works as described and looks exactly like in the mockups tested with a Samsung 8.

But IMHO the button don't make sense if you visit the screen from the menu. @SamuraiKek could you please get in touch with our designers and clarify this? It also don't fit the other information screens.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -109,9 +110,18 @@ class HomeFragmentViewModel @AssistedInject constructor(
                     }
                 }
             }
+            if (cwaSettings.lastAppVersion.value < BuildConfigWrap.VERSION_CODE &&
+                LocalData.isOnboarded() && LocalData.onboardingCompletedTimestamp()
```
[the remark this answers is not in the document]
SamuraiKek: Because isOnboarded() is set to true right after pressing the "Next" button from the last screen in the onboarding flow. When we reach the home fragment and we perform this check, the isOnboarded() value already returns true. Luckily I've found that we also stored the timestamp of the onboarding so I've used it to avoid any changes to the existing logic.

Corona-Warn-App/src/main/res/values-de/release_info_strings.xml
SamuraiKek: I thought so too, but I included it because I've found it in multiple places through our strings and wasn't sure if this was the convention or not. I will remove it.

SamuraiKek:
> I have three small remarks, none of them is blocking.
> 
> The implementation works as described and looks exactly like in the mockups tested with a Samsung 8.
> 
> But IMHO the button don't make sense if you visit the screen from the menu. @SamuraiKek could you please get in touch with our designers and clarify this? It also don't fit the other information screens.

Yes, this seemed weird for me too. I'll ask them.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
```diff
@@ -24,11 +25,18 @@ class CWASettings @Inject constructor(
         get() = prefs.getBoolean(PKEY_DEVICE_TIME_INCORRECT_ACK, false)
         set(value) = prefs.edit { putBoolean(PKEY_DEVICE_TIME_INCORRECT_ACK, value) }
 
+    val lastAppVersion = prefs.createFlowPreference(
```
d4rken, line 28: As you noticed, we already store multiple onboarding true/false, and timestamps. All of which were not quite right to know when to show the new release infos. There may be multiple things we want to do on version upgrades.

Should we make this more specific? So there isn't any confusion what this reflects.
What do you think about:

```kotlin
val lastChangelogVersion = .....( key = "update.changelog.lastversion")
```

i.e. it's the last version for which the change log was shown.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
```diff
@@ -24,11 +25,18 @@ class CWASettings @Inject constructor(
...
+    )
+
     fun clear() {
         prefs.clearAndNotify()
     }
 
     companion object {
         private const val PKEY_DEVICE_TIME_INCORRECT_ACK = "devicetime.incorrect.acknowledged"
+        private const val LAST_APP_VERSION = "last.installed.version"
+        private const val DEFAULT_APP_VERSION = 1.0.toLong()
```
d4rken, line 40: This `1.0.toLong()` is confusing to me. `2.0` is a `double` , but then we cast it to `long`? Let's just inline this and write `defaultValue= 1L`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -109,9 +110,18 @@ class HomeFragmentViewModel @AssistedInject constructor(
                     }
                 }
             }
+            if (cwaSettings.lastAppVersion.value < BuildConfigWrap.VERSION_CODE &&
```
d4rken, line 113: Why do we need to check the onboarding values? If we are on the home screen, onboarding was completed in any case, wasn't it? Could we just check the last version for which the changelog was shown?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragment.kt
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        vm.routeToScreen.observe2(this) {
+            if (it is NewReleaseInfoFragmentNavigationEvents.NavigateToMainActivity) {
+                doNavigate(NewReleaseInfoFragmentDirections.actionNewReleaseInfoFragmentPop())
+            }
+        }
+    }
+
+    fun userHasReadReleaseNotes() {
```
d4rken, line 53: Why is this not part of the viewmodel?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragmentViewModel.kt
```diff
@@ -0,0 +1,31 @@
...
+    @AppContext private val context: Context
+) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
+    val routeToScreen: SingleLiveEvent<NewReleaseInfoFragmentNavigationEvents> = SingleLiveEvent()
+
+    fun onNextButtonClick() {
+        routeToScreen.postValue(NewReleaseInfoFragmentNavigationEvents.NavigateToMainActivity)
+    }
+
+    val appVersion = flowOf(
+        context.getString(R.string.release_info_version_title).format(BuildConfig.VERSION_NAME)
```
d4rken, line 26: This is evaluated on construction so the the `.asLiveData(context = dispatcherProvider.Default)` does not make it async.

To make the string format not happen on the main thread, you could write:
```kotlin
flow {
 emit(context.getString(R.string.release_info_version_title).format(BuildConfig.VERSION_NAME))
}
```

there isn't a huge cost to the string formatting here though.
It would be a bit nicer to not use the app context for formatting strings.

@chiljamgossow  added an extension function you could use here, I think (see `LazyString.kt`):

```kotlin
flow {
    R.string.release_info_version_title.toResolvingString(BuildConfig.VERSION_NAME).let { emit(it) }
}
```

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -486,4 +492,15 @@
...
+
+    <!-- New Release -->
+    <fragment
+        android:id="@+id/newReleaseInfoFragment"
+        android:name="de.rki.coronawarnapp.ui.release.NewReleaseInfoFragment"
+        android:label="NewReleaseInfoFragment" >
+        <action
+            android:id="@+id/action_newReleaseInfoFragment_pop"
+            app:popUpTo="@id/newReleaseInfoFragment"
+            app:popUpToInclusive="true" />
```
d4rken, line 504: @chiljamgossow  Is this the right way? Could we just leave it out, as we `popBackStack()`?

d4rken (commented):
Looks good, maybe we can tweak a few things?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
SamuraiKek: Yes, that's a fair point. I'll have it changed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
SamuraiKek: I think we could get rid of the isOnboarded() check because as you said, this is always true anyways. But I'd leave the onboarding timestamp check in because that takes care of NOT displaying the release notes after the user completes onboarding.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragment.kt
SamuraiKek: Good question. I'll change it to have it called from the vm :D

AlexanderAlferov:
Tested on Samsung S8.
Found one issue:
First time (right after updating I have "**>**" at the end of release name:
Vuzix View_20210125_155004_750
But after opening it it from "Application details" menu - everything looks great.

SamuraiKek:
Added the "do not merge" label as I'm waiting for some updated strings and clarification from the designers. The statistics are going to be included in release 1.11 so the text for the release notes in 1.12 is no longer relevant.

SamuraiKek:
> Tested on Samsung S8.
> Found one issue:
> First time (right after updating I have "**>**" at the end of release name:
> But after opening it it from "Application details" menu - everything looks great.

That's weird. I haven't seen this so far but I'll investigate.

harambasicluka (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    private val vm: NewReleaseInfoFragmentViewModel by cwaViewModels { viewModelFactory }
+    private val binding: NewReleaseInfoScreenFragmentBinding by viewBindingLazy()
+    private val args: NewReleaseInfoFragmentArgs by navArgs()
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        binding.apply {
+            newReleaseInfoNextButton.setOnClickListener {
+                vm.userHasReadReleaseNotes()
```
d4rken, line 31: Is there any case where we need use `onNextButtonClick` without `userHasReadReleaseNotes` so it requires having this as two separate function calls

SamuraiKek: You're right, I don't think so. I'll put them under the same function.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragmentViewModel.kt
SamuraiKek: I tried using it with flow, the way you described it here but the thing is, I was observing it from the fragment using observe2 which requires LiveData. I wasn't really trying to make this string format async in the first place. As you said, it's not something that would actually give a performance hit so I think it's ok to live it as is.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragmentViewModel.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    val navigationIcon = flowOf(
+        context.getDrawableCompat(R.drawable.ic_back)
+    ).asLiveData(context = dispatcherProvider.Default)
+
+    fun userHasReadReleaseNotes() {
+        settings.lastChangelogVersion.update { BuildConfigWrap.VERSION_CODE }
+    }
+
+    fun setNextButtonVisibility(isInfoScreen: Boolean) = !isInfoScreen
```
d4rken, line 42: I think this is unused?

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
SamuraiKek: Still not sure about what should I do with this. I haven't changed it yet. It would work with popBackStack() also but since this is included in the navgraph features doesn't it make more sense to use it instead of pop backStack()?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/release/NewReleaseInfoFragmentViewModel.kt
SamuraiKek: Ah, you're right. Forgot to remove it.

SamuraiKek:
> > Tested on Samsung S8.
> > Found one issue:
> > First time (right after updating I have "**>**" at the end of release name:
> > But after opening it it from "Application details" menu - everything looks great.
> 
> That's weird. I haven't seen this so far but I'll investigate. There shouldn't be any difference between the one that shows up after the update and the one shown from "Application details" since it's exactly the same fragment. Only thing I could think of is the `BuildConfigWrap.VERSION_CODE` changing in the meantime.

Found the issue. I accidentally left a ">" symbol in the english strings file. Should be fixed now.

d4rken (approved):
lgtm, any open comments are not blockers for me.

ralfgehrer (approved):
Tested on Pixel (Android 10)

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
chiljamgossow: If popBackStack does the trick, use it.

chiljamgossow: This action is dependent on the stack and the order of items, which may result in unwanted behavior. What you did is correct, however it is more complicated and therefore more prone to bugs.

vaubaehn:
Hi @SamuraiKek & @d4rken ,
is the ENF version (#2794) still shown after this PR? On my app information screen it's gone. Could be related to #2302 , but my screen seems to be scrollable...?

PR #2181, opened 2021-01-21
Dagger2 Assisted Inject (Dev)

mtwalli:
Dagger2 supports AssistedInject now. This PR migrates square Assisted Injection to the default Dagger Inject.

Changes:
   - Adding new  imports for the already existing annotations `@Assisted`, `@AssistedInject`
     `import dagger.assisted.AssistedFactory`
     `import dagger.assisted.AssistedInject`

   - Replacing `@AssistedInject.Factory` with `@AssistedFactory`
    ` import dagger.assisted.AssistedFactory`
   - Delete `AssistedInjectModule` not needed anymore
   
   ### MainThread changes in Testing
It appears that this dagger version brought some changes from `androidx.lifecycle:lifecycle-*:2.3.0`. Starting from `androidx.lifecycle:lifecycle-*:2.3.0-alpha06`. There is a behaviour change in `LifecycleRegistry` . `LifecycleRegistry` now verifies that its methods are called on the main thread. Now remains another question unsolved whether the instrumentation tests are running in the MainThread? Apparently no otherwise it would have passed. It is using Instrumentation Thread where most of the tests run. As per documentation, `@UiThreadTest` should be used in this case (unfortunately
 did not fix issues here ). `runOnUiThread` stays a good solution ,but at least we understand why 

Why? I will quote the question and answer from Dagger dev site:

>What about AutoFactory and Square’s AssistedInject?
For Dagger users, we recommend using Dagger’s assisted injection rather than other assisted injection libraries like AutoFactory or square/AssistedInject. The existence of these libraries predate Dagger’s assisted injection. While they can be used with Dagger, they require a bit of extra boilerplate to integrate with Dagger.

Resources:
https://dagger.dev/dev-guide/assisted-injection
https://github.com/google/dagger/releases/tag/dagger-2.31


Corona-Warn-App/src/androidTest/java/testhelpers/viewmodels/MockViewModelModule.kt
```diff
@@ -1,35 +1,16 @@
...
-                    CREATORS,
-                    savedStateOwner,
-                    defaultSavedState,
-                    assistAction
-                )
-            }
-        }
-        return factory
-    }
+    fun viewModelCreators(): MutableMap<Class<out CWAViewModel>, CWAViewModelFactory<out CWAViewModel>> =
```
d4rken, line 12: What happened here?
How does it relate tor changing the assisted injection library?

mtwalli: Dagger was throwing a compile-time error: `Dagger does not support AssistedInject types` pointing to the `CWAViewModelFactoryProvider` constructor.  in UI tests we provide it through a module, while it is not the same in AppComponent. I assumed it does not work with @Provides. so instead of providing an instance of CWAViewModelFactoryProvider. I provided the creators map instead, and it seems to be fine. if you have a better way, I will be happy to do it.

mtwalli: Exact error `Dagger does not support providing @AssistedFactory types.`

Corona-Warn-App/build.gradle
```diff
@@ -329,10 +329,6 @@ dependencies {
     kaptTest "com.google.dagger:dagger-android-processor:$dagger_version"
     kaptAndroidTest "com.google.dagger:dagger-android-processor:$dagger_version"
 
-    def assisted_injection_version = "0.6.0"
```
harambasicluka, line 332: @AlexanderAlferov do we have something to do if we remove a dependency?

mtwalli: Good question I removed it from oss file. https://github.com/corona-warn-app/cwa-app-android/pull/2181/commits/83ea5cab1bc20aa5315a92f01137aa52726535f8

Corona-Warn-App/src/androidTest/java/testhelpers/viewmodels/MockViewModelModule.kt
d4rken: I'm just suprised this works, I prefer your solution, I'm sure I tried it previously though and couldn't get it to work.

So if it all still works, this is nice :+1:

d4rken (dismissed):
lgtm, thanks for the refactoring @mtwalli 

Please check that there are no new issues in the instrumentation tests once CI is done.

Corona-Warn-App/build.gradle
harambasicluka: @kaddaSz is aware of this and our pipeline should handle it.

d4rken (dismissed):

LukasLechnerDev (approved):
Nice work!

ralfgehrer (approved):

PR #2184, opened 2021-01-21
Optimize Accessibility for Retrieve Test Result Screen (EXPOSUREAPP-4460)

axelherbstreith:
Added screen reader support for the illustration on the "Retrieve Rest Result" screen when exploring it by touch.

To test this changes, activate TalkBack and open the "Have you been tested" Card on the Home Screen.

retrieve_test_result_content_description


SamuraiKek (approved):
At first I was testing it in English and it still said "Unlabelled", but then I realised we only had strings in German. Safe to say, the description works. Tested on Nokia 3 (Android 9).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/fragment/SubmissionDispatcherFragment.kt
```diff
@@ -52,7 +52,7 @@ class SubmissionDispatcherFragment : Fragment(R.layout.fragment_submission_dispa
 
     override fun onResume() {
         super.onResume()
-        binding.submissionDispatcherRoot.sendAccessibilityEvent(AccessibilityEvent.TYPE_ANNOUNCEMENT)
+        binding.submissionDispatcherRoot.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
```
harambasicluka, line 55: Why is this change necessary?

SabineLoss (approved):
UA reviewed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/fragment/SubmissionDispatcherFragment.kt
axelherbstreith: During testing I found out that the component wasn't focused by default when using `TYPE_ANNOUCEMENT` and therefore nothing was announced by the screen reader. Also settings android:focusable="true"  alone didn't work out. Explicitly setting `TYPE_ANNOUCEMENT` to the view fixes the issue.

harambasicluka (approved):

PR #2185, opened 2021-01-21
Fix invisible tracing settings screen if the last calculation failed (EXPOSUREAPP-4717)

d4rken:
TracingDetailsItemProvider is used for the tracing/risk details screen.
This has specific behavior that the logging period is not shown if the last calculation failed.
By reusing this data provider for the tracing settings, if the last calculation failed, we never set "loggingPeriod".
If loggingPeriod is null, the databinding causes the layout to be invisible.

### Testing
* Produce a "risk calculation failed" card on the homescreen. 
* Tap the risk status bar which should take you to the settings.
    * Before: Invisible layout
    * Now:  Visible layout :partying_face: 

[Screenshot from 2021-01-21 18-46-40](https://user-images.githubusercontent.com/1439229/105390185-099aa680-5c19-11eb-9dfc-625ecf982baa.png)
[Screenshot from 2021-01-21 18-46-30](https://user-images.githubusercontent.com/1439229/105390187-0a333d00-5c19-11eb-994e-a91b8d7b07b3.png)


ralfgehrer (approved):
Works nicely on Pixel (Android 10)

LukasLechnerDev (approved):
👍  Fixed. Tested on Samsung S20+, Android 11

PR #2202, opened 2021-01-25
Inconsistent ( X icon ) navigation behaviour (EXPOSUREAPP-4256)

mtwalli:
Change behaviour of `X` in SubmissionConsentFragment to go back.
According to: https://material.io/components/dialogs#full-screen-dialog 
- `X` and `<-` go back. 
- `X` should be used in cases where it acts as `dismiss` while ⬅️  should be not be used as it gives the user an impression that state is stored (This a different story and does not change the behaviour when they are clicked).

New behaviour:
https://user-images.githubusercontent.com/25054729/105712912-c8590e00-5f1a-11eb-82ad-faad5fb1dfa0.mp4

Similar case from Google Maps:
https://user-images.githubusercontent.com/25054729/105713193-1cfc8900-5f1b-11eb-8fbf-1949d9bfab76.mp4


AlexanderAlferov (approved):
Worked exactly like on video.

chiljamgossow (approved):

PR #2204, opened 2021-01-25
Test result no notification intended fix (EXPOSUREAPP-4625)

BMItr:
Details: Exposureapp-4625

Expected Result:
no notification about the changed Covid 19 test

Teststeps: 

- Einstellungen - Mitteilungen > Status Ihres COVID-19-Tests OFF
- Scan a Pending QR, go back to Homescreen of CWA
- put app in the background, put mobile away
- change pending -> invalid
- wait up to 6h and check lock screen of the mobile from time to time

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/TestResultAvailableNotificationService.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/TestResultAvailableNotificationService.kt
@@ -3,11 +3,13 @@ package de.rki.coronawarnapp.notification
 import android.content.Context
 import androidx.navigation.NavDeepLinkBuilder
 import de.rki.coronawarnapp.R
+import de.rki.coronawarnapp.storage.LocalData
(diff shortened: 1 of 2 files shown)


kolyaopahle (dismissed):
works on my machine 👍 
fix tests pls 💯

chris-cwa (dismissed):

kolyaopahle (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/TestResultAvailableNotificationService.kt
```diff
@@ -21,6 +23,11 @@ class TestResultAvailableNotificationService @Inject constructor(
     suspend fun showTestResultAvailableNotification(testResult: TestResult) {
         if (foregroundState.isInForeground.first()) return
 
+        if (!LocalData.isNotificationsTestEnabled) {
```
chiljamgossow, line 26: This should be moved to the worker so all abort conditions are checked at the same spot

chris-cwa (approved):

PR #2210, opened 2021-01-26
Wrong navigation in "Symptom-Erfassung" (EXPOSUREAPP-4703)

chris-cwa:
No Pop-up is to be shown, when user selects 'no information'.


LukasLechnerDev (approved):
LGTM, Tested on Samsung S20+, the dialog is not shown anymore when the user doesn't want to share symptoms.

ralfgehrer (approved):
Works nicely on Pixel (Android 10)

PR #2211, opened 2021-01-26
Exposure logging switch not working when screen reader is active (EXPOSUREAPP-2758)

LukasLechnerDev:
The problem was that Exposure logging cannot be turned off or on when TalkBack is activated. 

Although the switch state itself changes, the actual state of this setting does _not_. 

The problem was that there is this check in the `setOnCheckedChangeListener` of the `Switch`:
`if (switch.isPressed || row.isPressed)`

This check is needed in order to only execute code in the lambda of the listener when the `Switch` was actually pressed by the user and not set programmatically. 

The problem though is that when TalkBack is activated and the user double taps to change the `Switch` value, neither `switch.isPressed` nor `row.isPressed` is true and therefore the code in the lambda was never executed. 

My solution here is to set an `onClickListener` on the whole row, which is triggered when the user uses TalkBack but not triggered when we change the value of the `Switch` programmatically.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/settings/SettingsTracingFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/settings/SettingsTracingFragment.kt
@@ -51,7 +51,8 @@ class SettingsTracingFragment : Fragment(R.layout.fragment_settings_tracing), Au
                     TracingSettingsState.LocationDisabled -> setOnClickListener(null)
                     TracingSettingsState.TracingInactive,
                     TracingSettingsState.TracingActive -> setOnClickListener {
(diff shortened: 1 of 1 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/settings/SettingsTracingFragment.kt
```diff
@@ -114,9 +114,6 @@ class SettingsTracingFragment : Fragment(R.layout.fragment_settings_tracing), Au
     }
 
     private fun onTracingToggled(isChecked: Boolean) {
-        // Focus on the body text after to announce the tracing status for accessibility reasons
```
LukasLechnerDev, line 117: This event isn't needed.

chiljamgossow (approved):
Tested on Xiaomi, seems fine

d4rken (approved):
lgtm

PR #2213, opened 2021-01-26
Adds unit test for result worker (EXPOSUREAPP-4663)

chiljamgossow:
Adds unit test for DiagnosisTestResultRetrievalPeriodicWorker
Removes instrumentation test


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
```diff
@@ -54,7 +48,7 @@ class DiagnosisTestResultRetrievalPeriodicWorker @AssistedInject constructor(
         var result = Result.success()
         try {
 
-            if (abortConditionsMet()) {
+            if (abortConditionsMet(System.currentTimeMillis())) {
```
d4rken, line 51: We could inject `TimeStamper` into the worker as use it as time source, then we have full control over the time.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
```diff
@@ -84,7 +78,8 @@ class DiagnosisTestResultRetrievalPeriodicWorker @AssistedInject constructor(
         return result
     }
 
-    private fun abortConditionsMet(): Boolean {
+    @VisibleForTesting
+    fun abortConditionsMet(currentMillis: Long): Boolean {
```
d4rken, line 82: If we mock time using `TimeStamper`, we might not need `VisibleForTesting`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
```diff
@@ -75,7 +69,7 @@ class DiagnosisTestResultRetrievalPeriodicWorker @AssistedInject constructor(
                     stopWorker()
                 }
             }
-        } catch (e: Exception) {
+        } catch (ex: Exception) {
```
d4rken, line 72: Can we add a test for the error case, or did I miss it?

d4rken (commented):
Looks good :coffee: , a few small tweaks to discuss, but no blockers.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
chiljamgossow: Yes, will do

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
chiljamgossow: done

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/worker/DiagnosisTestResultRetrievalPeriodicWorker.kt
chiljamgossow: done

d4rken (approved):
lgtm :+1:

d4rken:
There is currently still an issue with our CI instrumentation tests, if only that check fails, it can still be merged. Will be fixed in #2222

ralfgehrer (approved):

PR #2215, opened 2021-01-26
Run instrumentation tests on Firebase Test Lab (Dev)

mtwalli:
Run Instrumentation tests on Firebase TL 🔥 

Result:
🟢 Tests + Enjoy 👨🏼‍💻


AlexanderAlferov (dismissed):
Great! 🚀

.circleci/config.yml
```diff
@@ -437,35 +450,54 @@ jobs:
...
-          desc: Run instrumentation tests on device
-          cmd: >
-            :Corona-Warn-App:connectedDeviceDebugAndroidTest
-            --info --continue
-            -Pandroid.testInstrumentationRunnerArguments.notAnnotation=testhelpers.Screenshot
-            -Pandroid.testInstrumentationRunnerArguments.clearPackageData=true
-      - kill-all-emulators
+      - restore-android-build-cache
+      - run:
+          name: "Build app APK "
```
d4rken, line 464: Extra space, only remove it if you change the file for any other reason.

.circleci/config.yml
```diff
@@ -437,35 +450,54 @@ jobs:
...
+          command: ./gradlew -PdisablePreDex :Corona-Warn-App:assembleDeviceForTestersDebug
+      - run:
```
d4rken, line 466: Could we use our `run-gradle-cmd` which automatically adds a few optimization flags, or did you try that and there were downsides?

.circleci/config.yml
```diff
@@ -437,35 +450,54 @@ jobs:
...
+              --no-record-video
+          no_output_timeout: 30m
+      - run:
+          name: Create directory to store test results
+          command: mkdir ~/project/firebase
+      - run:
+          name: Install gsutil dependency and copy test results data
+          command: |
+            sudo pip install -U crcmod
+            sudo gsutil -m cp -R -U "`sudo gsutil ls gs://${GOOGLE_PROJECT_ID}-circleci-android/${BUCKETDIR} | tail -1`*"  ~/project/firebase | true
```
d4rken, line 496: I think we could make the path simpler.
The generated path in the zip is currently `/home/circleci/project/firebase/flame-29-de_DE-portrait/`

We could try

```yaml
          sudo gsutil -m cp -R -U "`sudo gsutil ls gs://${GOOGLE_PROJECT_ID}-circleci-android/${BUCKETDIR} | tail -1`*" firebase-results | true

(...)
      - store_test_results:
          path: ./firebase-results/flame-29-de_DE-portrait
      - compress-path:
          input: ./firebase-results/flame-29-de_DE-portrait/*.xml
(...)
```

What do you think?

d4rken (commented):
Looks good, few comments.

.circleci/config.yml
```diff
@@ -437,35 +450,54 @@ jobs:
             --form "file=@${fileName}" \
 
   instrumentation_tests_device:
-    macos:
-      xcode: "12.3.0"
-    resource_class: large
+    resource_class: xlarge
```
d4rken, line 453: Instrumentation tests are now faster than the overall longest job-chain (unit tests + sonar), so we could save some resources by preventing firebase from spinning up if trivial checks fails, i.e.

```yml
      - instrumentation_tests_device:
          requires:
            - ktlint_device_release_check
            - detekt
```

Thoughts?

mtwalli: Good point

mtwalli: Good point

.circleci/config.yml
mtwalli: will try it

mtwalli:
Comments are done, Please have another look

d4rken (approved):
lgtm, thank you for optimizing this

ralfgehrer (approved):

PR #2216, opened 2021-01-27
7 day r value decimal place issue (EXPOSUREAPP-4751)

LukasLechnerDev:
Number formatting issue. "0.00" was cut off.


jurajkusnier (changes requested):
We would have the same issue for other decimal places. For example, if you add this line 
`Arguments.of(1.0, 3, Locale.GERMANY, "1,000")`
to `StatisticsNumberValueFormatterTest` test will fail too

LukasLechnerDev:
@jurajkusnier Thanks for your input. I made a change that fixes this issue. However, we have the same issue when the decimal value is >3 but I don't think that the API will return such decimals.

jurajkusnier:
> @jurajkusnier Thanks for your input. I made a change that fixes this issue. However, we have the same issue when the decimal value is >3 but I don't think that the API will return such decimals.

ok, let is like this for now

jurajkusnier (dismissed):

chiljamgossow (changes requested):
The number leading the separator is gone now, just displaying ,95

LukasLechnerDev:
@chiljamgossow Great that you spotted that! I fixed it.

jurajkusnier (approved):

ralfgehrer (approved):

PR #2225, opened 2021-01-27
Skeleton classes for privacy preserving analytics (EXPOSUREAPP-4753)

d4rken:
Skeleton classes for privacy preserving analytics to parallelize developer effort.

Has to be merged after #4754. Once that is merged I will branch a feature-main branch of 1.13.x, and rebase this PR against it.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/DonorModule.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/DonorModule.kt
@@ -2,7 +2,7 @@ package de.rki.coronawarnapp.datadonation.analytics.modules
 
 interface DonorModule {
 
-    suspend fun startDonation(request: Request): Contribution
+    suspend fun beginDonation(request: Request): Contribution
 
     /**
      * Data that the modules may need to fullfil the request
@@ -27,6 +27,6 @@ interface DonorModule {
          * If the submission failed, i.e. network issues, the donor module may collect their data for the next attempt.
          * If the submission was sucessful, the module may discard the data as it was submitted.
          */
-        suspend fun finishContribution(successful: Boolean)
+        suspend fun finishDonation(successful: Boolean)
     }
 }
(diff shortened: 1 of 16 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/DonorModule.kt
```diff
@@ -0,0 +1,32 @@
+package de.rki.coronawarnapp.datadonation.analytics.modules
+
+interface DonorModule {
```
ralfgehrer, line 3: I very much like that approach: clean, good separation of concerns, and easily extendable

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/DonorModule.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    suspend fun startDonation(request: Request): Contribution
```
ralfgehrer, line 5: very minor: maybe name it `initiateDonation` instead, to clarify it's the first thing being called.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/DonorModule.kt
```diff
@@ -0,0 +1,32 @@
...
+         */
+        suspend fun injectData(protobufContainer: Any)
+
+        /**
+         * This will be called with the submission result.
+         * For modules that track a data delta between each submission time:
+         * If the submission failed, i.e. network issues, the donor module may collect their data for the next attempt.
+         * If the submission was sucessful, the module may discard the data as it was submitted.
+         */
+        suspend fun finishContribution(successful: Boolean)
```
ralfgehrer, line 30: My gut feeling would be that this should be moved outside of the `Contribution` interface - time /actual implementation will tell ...

ralfgehrer (approved):
Thx. Great work.

kolyaopahle (approved):
i like the donor pattern a lot, but i'm not 100% sure if we can cover all the analytics cases with a single call issued when starting the submission, it seems to me that some stuff needs to be prepared beforehand and i would like to see a good pattern to support that as well as an extension.

Existing code looks great though 👍

PR #2230, opened 2021-01-28
End of life behaviour UI (EXPOSUREAPP-4641)

SamuraiKek:
This PR adds the new card for re-enabling the risk calculation after a user was found positive and has submitted their keys.

Testing steps:
- Register a positive test.
- Submit your keys.
- On the home screen, under the "Vielen Dank" card, the new card that lets you re-enable risk calculation is displayed. 
Note: nothing will happen after pressing the new card. The logic will be implemented in a separate PR.


Corona-Warn-App/src/main/res/drawable/ic_submission_home.xml
```diff
@@ -1,12 +1,10 @@
...
-        android:strokeWidth="1"
-        android:strokeColor="#00000000" />
+    android:width="22dp"
+    android:height="20dp"
+    android:viewportWidth="22"
+    android:viewportHeight="20">
+  <path
+      android:pathData="M11,0.5C10.7329,0.5 10.4769,0.6069 10.2891,0.7969L0.2031,9.5977C0.0754,9.6919 -0,9.8412 0,10C0,10.2761 0.2239,10.5 0.5,10.5H3V18.5C3,19.052 3.448,19.5 4,19.5H8C8.552,19.5 9,19.052 9,18.5V12.5H13V18.5C13,19.052 13.448,19.5 14,19.5H18C18.552,19.5 19,19.052 19,18.5V10.5H21.5C21.7761,10.5 22,10.2761 22,10C22,9.8412 21.9246,9.6919 21.7969,9.5977L11.7168,0.8027C11.7149,0.8008 11.7129,0.7988 11.7109,0.7969C11.5231,0.6069 11.2671,0.5 11,0.5Z"
+      android:fillColor="#ffffff"
+      android:fillType="evenOdd"/>
```
jurajkusnier, line 9: We already have a home icon `ic_risk_details_home`. Cannot we reuse it?

SamuraiKek: The one from the designs looks different from this one. Also, the current one that we have is too small. Looked out of place.

Corona-Warn-App/src/deviceForTesters/res/layout/home_reenable_risk_card_layout.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jurajkusnier, line 1: This layout seems to be duplicated.

jurajkusnier:
image
Date implementation is missing.

jurajkusnier (changes requested):
Please check my comments

Corona-Warn-App/src/deviceForTesters/res/layout/home_reenable_risk_card_layout.xml
SamuraiKek: Yep, I forgot to remove it.

SamuraiKek:
> Date implementation is missing.

Yep, I'm working on it.

Corona-Warn-App/src/main/res/drawable/ic_submission_home.xml
SamuraiKek: Oh no, wait. This is another home icon :D I think that will work.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/SubmissionState.kt
```diff
@@ -10,4 +15,9 @@ object TestNegative : SubmissionState()
 object TestError : SubmissionState()
 object TestInvalid : SubmissionState()
 object TestPending : SubmissionState()
-object SubmissionDone : SubmissionState()
+data class SubmissionDone(val testRegisteredOn: Date?) : SubmissionState() {
+
+    fun formatTestRegistrationText(context: Context): String =
```
jurajkusnier, line 20: I wouldn't put formatting methods into data objects, but as I saw this also in other parts of the project I guess it's not a big deal.

jurajkusnier (dismissed):

chiljamgossow (dismissed):
The UI looks fine. I'm just not sure if it's a good idea to merge in case this turns out to be a blocker. My concerns are about the setting risk calculation, which is not aligned with the proposed end-of-life behavior.

harambasicluka:
> My concerns are about the setting risk calculation, which is not aligned with the proposed end-of-life behavior.

What do you mean exactly?

jurajkusnier (approved):

PR #2231, opened 2021-01-28
Risk tracing stopped correct info (EXPOSUREAPP-4414)

AlexanderAlferov:
✅   Correct info text (same as for failed one, "Infektionsrisiko")
✅   No Investigation period section ("Ermittlungszeitraum")

Screenshot 2021-01-28 at 19 20 14

How to test:
1) start the app
2) turn off tracing
3) go to risk details
4) check correct text and no investigation period section ("Ermittlungszeitraum")
5) turn on tracing
6) if low or high risk - there should be another info text on the bottom and investigation period section in place


harambasicluka:
Maybe have to change the base to `release/1.13.x`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/details/TracingDetailsItemProvider.kt
```diff
@@ -53,23 +54,27 @@ class TracingDetailsItemProvider @Inject constructor(
                 )
             }.also { add(it) }
 
-            if (latestCalc.riskState != RiskState.CALCULATION_FAILED) {
+            if (latestCalc.riskState != RiskState.CALCULATION_FAILED && status != Status.TRACING_INACTIVE) {
```
ralfgehrer, line 57: Quite some nesting. Would be nice to have some tests for this.

ralfgehrer (commented):
Otherwise, LGTM. Tested on Pixel(Android 10)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/details/TracingDetailsItemProvider.kt
```diff
@@ -53,23 +54,27 @@ class TracingDetailsItemProvider @Inject constructor(
...
-                    riskState = latestCalc.riskState,
-                    matchedKeyCount = latestCalc.matchedKeyCount
-                )
-                RiskState.INCREASED_RISK -> DetailsIncreasedRiskBox.Item(
-                    riskState = latestCalc.riskState,
-                    lastEncounteredAt = latestCalc.lastRiskEncounterAt ?: Instant.EPOCH
-                )
-                RiskState.CALCULATION_FAILED -> DetailsFailedCalculationBox.Item
-            }.also { add(it) }
+            if (status == Status.TRACING_INACTIVE) {
```
d4rken, line 63: What about

```kotlin
    when {
        status == Status.TRACING_INACTIVE || latestCalc.riskState == RiskState.CALCULATION_FAILED -> {
            DetailsFailedCalculationBox.Item
        }
        latestCalc.riskState == RiskState.LOW_RISK -> DetailsLowRiskBox.Item(
            riskState = latestCalc.riskState,
            matchedKeyCount = latestCalc.matchedKeyCount
        )
        latestCalc.riskState == RiskState.INCREASED_RISK -> DetailsIncreasedRiskBox.Item(
            riskState = latestCalc.riskState,
            lastEncounteredAt = latestCalc.lastRiskEncounterAt ?: Instant.EPOCH
        )
        else -> null
    }?.let { add(it) }
```

d4rken (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/details/TracingDetailsItemProvider.kt
AlexanderAlferov: Was thinking of this variant at first (as it looks less nested), thought that it would be explicitly defined for inactive tracing. But nvm - already switched to it.

ralfgehrer (dismissed):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/tracing/ui/details/TracingDetailsItemProviderTest.kt
```diff
@@ -0,0 +1,349 @@
...
+            status = GeneralTracingStatus.Status.TRACING_ACTIVE,
+            activeTracingDaysInRetentionPeriod = 0,
+            riskLevel = ProtoRiskLevel.LOW,
+            matchedKeyCount = 1
+        )
+
+        val instance = createInstance()
+        val testCollector = instance.state.test(startOnScope = this)
+
+        testCollector.latestValue?.size?.shouldBeGreaterThan(0)
```
d4rken, line 102: Do not use `?.` in test assertions, it will silently fail without anyone knowing.

There needs to be a latest value, but if there is none, this will still return true.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/tracing/ui/details/TracingDetailsItemProviderTest.kt
```diff
@@ -0,0 +1,349 @@
...
+            activeTracingDaysInRetentionPeriod = 0,
+            riskLevel = ProtoRiskLevel.LOW,
+            matchedKeyCount = 0
+        )
+
+        val instance = createInstance()
+        val testCollector = instance.state.test(startOnScope = this)
+
+        testCollector.latestValue?.size?.shouldBeGreaterThan(0)
+        testCollector.latestValue?.any { it is AdditionalInfoLowRiskBox.Item } shouldBe false
```
d4rken, line 120: Please remove any use of `?.` in test assertions, if it should not be null, then it should crash if is null in a test.

d4rken (changes requested):

d4rken (approved):
lgtm :+1:

ralfgehrer (approved):

PR #2232, opened 2021-01-28
Repository for One Time Passwords(EXPOSUREAPP-4857)

chris-cwa:


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/datadonation/OneTimePasswordTest.kt
```diff
@@ -0,0 +1,14 @@
+package de.rki.coronawarnapp.datadonation
+
+import io.kotest.matchers.shouldBe
+import org.junit.Test
+import java.util.UUID
+
+class OneTimePasswordTest {
+
+    @Test
+    fun test() {
```
d4rken, line 10: Why is this an instrumentation test?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OneTimePassword.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import android.util.Base64
+import android.util.Base64.encode
+import java.nio.ByteBuffer
+import java.util.UUID
+
+data class OneTimePassword(val uuid: UUID = UUID.randomUUID()) {
+
+    val payloadForRequest: ByteArray
+        get() = ByteBuffer.wrap(ByteArray(16)).apply {
```
d4rken, line 11: Is this different from `UUID.toString().toByteArray()` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/storage/DataDonationPreferences.kt
```diff
@@ -0,0 +1,28 @@
...
+import android.content.Context
+import de.rki.coronawarnapp.datadonation.OneTimePassword
+import de.rki.coronawarnapp.util.di.AppContext
+import de.rki.coronawarnapp.util.preferences.clearAndNotify
+import de.rki.coronawarnapp.util.preferences.createFlowPreference
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class DataDonationPreferences @Inject constructor(
```
d4rken, line 12: The overall concept is called data donation, with the sub features "event survey" and "privacy preserving analytics".

Maybe we should rename this class as it's "event survey" specific?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -1,7 +1,10 @@
 package de.rki.coronawarnapp.datadonation.survey
 
+import de.rki.coronawarnapp.datadonation.storage.DataDonationPreferences
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class SurveySettings @Inject constructor()
+class SurveySettings @Inject constructor(
```
d4rken, line 8: Do we need the extra level? Could we just use `SurveySetting` or delete it and just use the new preference class you created?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OneTimePassword.kt
chris-cwa: technically it's not the same, but since we're just in need of a random byte array, and don't want to convert back to uuid, i guess your suggestion will suffice. will change it

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/datadonation/OneTimePasswordTest.kt
chris-cwa: because bytebuffer is only stub in android studio. obsolete with https://github.com/corona-warn-app/cwa-app-android/pull/2232#discussion_r566806424

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/storage/DataDonationPreferences.kt
chris-cwa: i thought we could you 1 instance of preferences for data donation. what do you think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
chris-cwa: see https://github.com/corona-warn-app/cwa-app-android/pull/2232#discussion_r566807152 for multi-use of this preferences class, if you are against is, i could skip this layer, but if it were up to me it would stay, because every layer improves finding a way through project structure

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/storage/DataDonationPreferences.kt
d4rken: As the user gives two different types of consent, one for analytics, and one for surveys, we may run into the situation where the user revokes either one, and we want to clear all data associated with that feature, having the data in different preferences makes that easier then if we have to reset certain key-values individually.

I'm not sure yet though what data we need to safe for analytics, the consent, and? Not sure if @kolyaopahle has started on that yet.

chris-cwa: good thinking! i'll make 1 instance of preferences for survey only - with according name

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
chris-cwa: obsolete as of https://github.com/corona-warn-app/cwa-app-android/pull/2232#discussion_r566847377

d4rken: I currently only see the `DataReset` class where we'd need access to both Analytics and Survey data structures, an abstraction layer just for that, seems a bit much. This wouldn't be a blocker for me though, maybe align with @kolyaopahle?

If the extra abstraction is desired, please reverse the order though.
* DataDonation (Parent/TopLevel)
   * Survey (Child)
   * Analytics (Child)

BMItter:
looking good so far 👌 
ps: don't forget to add timestamp (at generation) as discussed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -1,7 +1,27 @@
...
+    @AppContext val context: Context
+) {
+
+    private val preferences by lazy {
+        context.getSharedPreferences("survey_localdata", Context.MODE_PRIVATE)
+    }
+
+    val oneTimePassword =
+        preferences.createFlowPreference<OneTimePassword?>(
```
d4rken, line 21: I don't think this works as `OneTimePassword` is not a primitive type.
This will probably crash as there will be no default mapper for this in `FlowPreference.kt`.

If you want to save the whole data class you can use a gson writer, see `gsonReader` and `gsonWriter`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/OneTimePasswordTest.kt
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.Test
+import java.util.UUID
+
+class OneTimePasswordTest {
+
+    @Test
+    fun test() {
+        "923c977a-617b-11eb-ae93-0242ac130002".apply {
+            OneTimePassword(UUID.fromString(this))
+                .payloadForRequest shouldBe encodeToByteArray()
```
d4rken, line 13: What is `encodeToByteArray`? Are we testing it too somewhere?

I'd prefer a hardcoded comparison to remove any error sources, e.g.
Find out what that UUID looks like in base64/hex encoded form and use it for comparison.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/storage/OTPRepositoryTest.kt
```diff
@@ -0,0 +1,44 @@
...
+
+class OTPRepositoryTest : BaseTest() {
+
+    @MockK lateinit var surveySettings: SurveySettings
+    @MockK lateinit var flowPreference: FlowPreference<OneTimePassword?>
+
+    @BeforeEach
+    fun setup() {
+        MockKAnnotations.init(this)
+        every { surveySettings.oneTimePassword } returns flowPreference
```
d4rken, line 25: There is a helper method if you want: `mockFlowPreference(defaultValue)`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -1,7 +1,27 @@
...
-class SurveySettings @Inject constructor()
+class SurveySettings @Inject constructor(
+    @AppContext val context: Context
+) {
+
+    private val preferences by lazy {
+        context.getSharedPreferences("survey_localdata", Context.MODE_PRIVATE)
+    }
+
+    val oneTimePassword =
```
d4rken, line 20: Please add a unit test that shows that storage is working on both OTP and timestamp are saved&loaded.

d4rken (changes requested):

d4rken (changes requested):
Looks good, would like to increase resilience with a few tweaks and test cases.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -1,7 +1,41 @@
...
 
 @Singleton
-class SurveySettings @Inject constructor()
+class SurveySettings @Inject constructor(
+    @AppContext val context: Context,
+    @BaseGson val gson: Gson
+) {
+
+    companion object {
+        private const val KEY_OTP = "one_time_password"
```
BMItter, line 19: minor:
Constants in Kotlin dont need to be in companion object and preferred to be placed top level

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/storage/OTPRepositoryTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    @AfterEach
+    fun teardown() {
+        clearAllMocks()
+    }
+
+    @Test
+    fun `last otp is read from preferences`() {
+        val uuid = UUID.fromString("e103c755-0975-4588-a639-d0cd1ba421a0")
+        val time = Instant.now()
```
BMItter, line 39: minor: Instants should be hardcoded in tests to avoid flaky tests (also below)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -1,7 +1,41 @@
...
+
+    private val preferences by lazy {
+        context.getSharedPreferences("survey_localdata", Context.MODE_PRIVATE)
+    }
+
+    var oneTimePassword: OneTimePassword?
+        get() {
+            val json = preferences.getString(KEY_OTP, null)
+            if (json != null) {
+                return gson.fromJson(json, OneTimePassword::class.java)
```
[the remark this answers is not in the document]
chris-cwa: We would catch this pretty quick if it would throw a throwable - first time you run the code. So no try-catch-block needed here.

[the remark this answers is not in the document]
d4rken: Who uses this call downstream? Would they handle the exception, or should we just log an error and treat it as `null`?

SamuraiKek (dismissed):
Looks good. Code is readable and easy to understand.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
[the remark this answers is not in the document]
chris-cwa: added try catch block and unit tested parsing of invalid data

d4rken (dismissed):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -1,7 +1,47 @@
...
+                }
+            } catch (t: Throwable) {
+                Timber.e(t, "failed to parse OTP from preferences")
+            }
+            return null
+        }
+        set(value) =
+            preferences
+                .edit()
+                .putString(KEY_OTP, if (value == null) null else gson.toJson(value))
```
LukasLechnerDev, line 41: I am curious: Why are you saving it as `JSON` instead of 2 simple String values?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OneTimePassword.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.google.gson.annotations.SerializedName
+import org.joda.time.Instant
+import java.util.UUID
+
+@Keep
+data class OneTimePassword(
+    @SerializedName("uuid")
+    val uuid: UUID = UUID.randomUUID(),
+    @SerializedName("time")
+    val time: Instant = Instant.now()
```
LukasLechnerDev, line 13: Don't we also have to store the `expirationDate` (that we get after the `otp` authentication) and the time when the `otp` authentication was performed?

See point 1 and 2 in the tech-spec [here](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/068f948fdadfda84712e6f23f79d792e6103e946/docs/spec/data-donation-client-edus.md#authorization-request-of-otp)

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/OneTimePasswordTest.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import io.kotest.matchers.shouldBe
+import okio.ByteString.Companion.decodeBase64
+import org.junit.Test
+import java.util.UUID
+
+class OneTimePasswordTest {
+
+    @Test
+    fun test() {
```
LukasLechnerDev, line 11: A more descriptive test name would be appreciated 🙂

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
chris-cwa: because this way you don't need to change anything here if the data class gets additional fields

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OneTimePassword.kt
chris-cwa: I would rather dynamically calculate the expiration date and not set it at creation time.

chris-cwa: to me the success and time of auth would be part of a request entity, not the otp itself

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/OneTimePasswordTest.kt
chris-cwa: renamed it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OneTimePassword.kt
LukasLechnerDev: Thanks!

LukasLechnerDev: > I would rather dynamically calculate the expiration date and not set it at creation time.

We don't calculate the `expirationDate`, we get it from the backend on a successful authentication, or?

> to me the success and time of auth would be part of a request entity, not the otp itself

But we have to persist those values, right?

chris-cwa: you're right! :)

BMItr (approved):
all right for me 
next steps in upcoming PR as discussed 👌

SamuraiKek (approved):

PR #2235, opened 2021-01-29
Add Tab bar (EXPOSUREAPP-4639)

mtwalli:
1- Add Tab in Home Screen
2- Move Contact diary nav graph into main graph as nested graph
3- Remove contact diary Activity
4- Adjust AppTheme to work with AppCompat and MDTheme
5- Remove Diary Card from Home Screen
6- Remove Overview diary screen <- button, Onboarding  x button.

Tested Dark and Light Themes on Samsung S8 Android 9

> A:Hi, right. We have discussed that we will have two behaviours there. If you open it from the upper menu, it will be shown as a modal dialog and the Tab-Bar will be hidden. In case you enter the contact diary the first time we will display it as a normal view behind the tabs (so that we avoid that I press on the tab and the Tab will be covered by the modal dialog).

PLEASE test it thoroughly and report any difference you spot


harambasicluka:
Tested on a Pixel 3, looks really good and feels smooth :) 
I'm not a big fan of the "jumping" between the logo and the title, but that's not in our hands.

axelherbstreith:
Tested the behavior on a Huawei P30 and a Nokia 7.1 (both Android 10), looking really good! 👍 
One thing which appears strange to me is the "Start page" title in the Contact Journal Tab. Doesn't seem to make sense anymore since the journal was moved to a different tab...

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1632,5 +1632,9 @@
     <string name="statistics_trend_decreasing">"Trend: Downwards"</string>
     <!-- XTXT: Statistics trend stable (Accessibilty) -->
     <string name="statistics_trend_stable">"Trend: Steady"</string>
+    <!-- XHED: Title for BottomNav main screen title -->
+    <string name="bottom_nav_home_title">Home</string>
+    <!-- XHED: Title for BottomNav diary screen title -->
+    <string name="bottom_nav_diary_title">Diary</string>
```
axelherbstreith, line 1638: Not sure whether this needs to be "Diary" or "Journal"

SabineLoss, line 1638:
```suggestion
    <string name="bottom_nav_diary_title">Journal</string>
```
Needs to be "Journal"

mtwalli:
> Tested the behavior on a Huawei P30 and a Nokia 7.1 (both Android 10), looking really good! 👍
> 
> One thing which appears strange to me is the "Start page" title in the Contact Journal Tab. Doesn't seem to make sense anymore since the journal was moved to a different tab...

You are right!, I adjusted the title now

Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -65,7 +65,8 @@
             android:name=".ui.main.MainActivity"
             android:exported="false"
             android:screenOrientation="portrait"
-            android:theme="@style/AppTheme.Main" />
+            android:theme="@style/AppTheme.Main"
+            android:windowSoftInputMode="adjustResize" />
```
d4rken, line 69: Why the new flag?

d4rken: Ok I see, it comes from the ContactDiary Activity. How does effect input in the "non-diary" app?

Where do we have keyboard input that needs to be tested due to this

* TAN input
* :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/UIExtensions.kt
```diff
@@ -13,3 +23,45 @@ fun NavController.doNavigate(direction: NavDirections) {
...
+fun BottomNavigationView.setupWithNavController2(
+    navController: NavController,
+    onItemSelected: () -> Unit
+) {
+    setupWithNavController(navController)
+    setOnNavigationItemSelectedListener { item ->
+        onItemSelected()
+        NavigationUI.onNavDestinationSelected(item, navController)
+    }
+    val weakBottomNavView = WeakReference(this)
```
d4rken, line 40: Why does this need to be a `WeakReference`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -58,11 +62,14 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
 
     @Inject lateinit var deadmanScheduler: DeadmanNotificationScheduler
     @Inject lateinit var contactDiaryWorkScheduler: ContactDiaryWorkScheduler
+    @Inject lateinit var settings: ContactDiarySettings
```
d4rken, line 65: Should this be part of a viewmodel?

Corona-Warn-App/src/main/res/layout/activity_main.xml
```diff
@@ -6,12 +6,33 @@
     android:layout_height="match_parent"
     tools:context="de.rki.coronawarnapp.ui.main.MainActivity">
 
-    <androidx.fragment.app.FragmentContainerView
-        android:id="@+id/nav_host_fragment"
-        android:name="androidx.navigation.fragment.NavHostFragment"
+    <androidx.constraintlayout.widget.ConstraintLayout
```
d4rken, line 9: We could also achieve this with a `LinearLayout`, do you think performance will be better?

Corona-Warn-App/src/main/res/values/colors.xml
```diff
@@ -84,4 +84,8 @@
     <color name="colorStatisticsTrendPositive">#2E854B</color>
     <color name="colorStatisticsTrendNegative">#C11633</color>
     <color name="colorStatisticsTrendNeutral">#5D6F80</color>
+
+    <!-- Bottom Nav bar-->
+    <color name="navItemColorSelected">#017FAD</color>
```
d4rken, line 89: Why is this different from our apps `accent` color?

Corona-Warn-App/src/main/res/values/styles.xml
```diff
@@ -1,37 +1,26 @@
 <resources xmlns:android="http://schemas.android.com/apk/res/android">
 
-    <style name="AppTheme" parent="Theme.AppCompat.DayNight">
+    <style name="AppTheme" parent="Theme.MaterialComponents.DayNight.Bridge">
```
d4rken, line 3: I like this, this was on our refactoring TODO, but this will really require us to check every screen!

@harambasicluka @ralfgehrer This PR needs to be reviewed by more than 2 people.

d4rken (commented):
Nice, code looks good overall, needs more thorough on device testing.

Corona-Warn-App/src/main/res/values/styles.xml
ralfgehrer: @mtwalli  I added the label "author merge" so you have full control and can merge it once several ppl reviewed it.

harambasicluka:
Only did a on device testing (Pixel 3), due to https://github.com/corona-warn-app/cwa-app-android/pull/2235#discussion_r567923879 I visited every screen, I wasn't able to spot a difference.

But I spotted a "jumping", I think it isn't caused by this change.
When I go from home to diary everything works as expected but if I navigate to test and go back it seems like the risk card is redrawn, in this video it isn't very visible but it should be easy to reproduce.

https://user-images.githubusercontent.com/64483219/106487270-c8c14e00-64b2-11eb-9504-2ad828f5decb.mp4

Corona-Warn-App/src/main/AndroidManifest.xml
ralfgehrer: TAN works nicely, also symptoms screens, as well as onboarding.

ralfgehrer:
Did extensive device testing with QR-codes of all types, TAN, App reset, went through all settings, tested all features of the contact journal, ... on Pixel (Android 10). Worked as expected and could not detect any UI glitches. Nice job!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/UIExtensions.kt
mtwalli: Actually, I tried here to have the same behaviour as the original setupWithNavController extension for nav-component.
I assume they do this as for the listener it might live after the view is gone

Corona-Warn-App/src/main/res/layout/activity_main.xml
mtwalli: I don't think it will have a big impact on the performance.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
mtwalli: I will check if possible, just took it as is from the contact diary activity

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/UIExtensions.kt
mtwalli: Original code from the navigation components 
[
Screenshot 2021-02-01 at 18 34 37
](url)

mtwalli:
> Only did a on device testing (Pixel 3), due to [#2235 (comment)](https://github.com/corona-warn-app/cwa-app-android/pull/2235#discussion_r567923879) I visited every screen, I wasn't able to spot a difference.
> 
> But I spotted a "jumping", I think it isn't caused by this change.
> When I go from home to diary everything works as expected but if I navigate to test and go back it seems like the risk card is redrawn, in this video it isn't very visible but it should be easy to reproduce.
> 
>  device-2021-02-01-172504.mp4

The jumping is happening because we hide the bottom bar for second level screens. it takes some space back once it visible

mtwalli:
> Did extensive device testing with QR-codes of all types, TAN, App reset, went through all settings, tested all features of the contact journal, ... on Pixel (Android 10). Worked as expected and could not detect any UI glitches. Nice job!
> 
> The code looks good too. Happy to approve after the VM comment and weak reference comment have been addressed.

I moved the settings to the view model. I will keep the weak reference to have the same behaviour look at my comment ⬆️

ralfgehrer (dismissed):

LukasLechnerDev:
Tested on Samsung S20+, Android 11. Looks good, basically. 

The only thing I found is that the `EditText`s in the contact diary now look different and lack the display of remaining characters. 

image

SamuraiKek:
Tested on Huawei Mate 8 (Android 6). This feels almost perfect. I have also found the issue mentioned by @LukasLechnerDev. Other than that, while I'm on the home screen, if I open the three dot menu, and choose one of the options there, tab bar disappears before the screen is switched. 
Also, can we save the state of the activities when switching between tabs? If I scroll down on the home screen and then switch to the diary and then switch back to the home screen, it looks like the whole activity is redrawn and I'm starting from the to top, not from where I left off.

mtwalli:
> Tested on Samsung S20+, Android 11. Looks good, basically.
> 
> The only thing I found is that the `EditText`s in the contact diary now look different and lack the display of remaining characters.
> 
> image

Thanks for reporting! , this should be addressed by https://github.com/corona-warn-app/cwa-app-android/pull/2235/commits/c82cc1f62853f2b8ed8d17c6559b68e8c1c31863

mtwalli:
> Tested on Huawei Mate 8 (Android 6). This feels almost perfect. I have also found the issue mentioned by @LukasLechnerDev. Other than that, while I'm on the home screen, if I open the three dot menu, and choose one of the options there, the tab bar disappears before the screen is switched (looks kinda weird).
> Also, can we save the state of the activities when switching between tabs? If I scroll down on the home screen and then switch to the diary and then switch back to the home screen, it looks like the whole activity is redrawn and I'm starting from the to top, not from where I left off.

Thx, I will try to improve or (delay) the visibility change part. Regarding resetting the state , I personally don't like too. You be surprised to know that this the recommended MD behaviour for BottomView on Android :https://material.io/components/bottom-navigation#behavior

jurajkusnier:
Teste on Pixel 4 (Android 10) and didn't notice any issues

LukasLechnerDev:
👍 Checked the `EditText`s again and they look good now:

image

ralfgehrer (approved):

chiljamgossow (approved):
tested on Xiaomi
LGTM

PhilippNowak96:
@mtwalli sorry for pulling out this old one but I stumbled over it when testing App Shortcuts. Did you guys discuss about suppressing reselection? The current behaviour is the fragment is created every time you click the icon, even if it is already the active one. Despite being the default behaviour I'm not sure if it is useful here.

dsarkar:
@PhilippNowak96 forwarded your comment to the devs ticket.

mtwalli:
> @mtwalli sorry for pulling out this old one but I stumbled over it when testing App Shortcuts. Did you guys discuss about suppressing reselection of bottom navigation items? The current behaviour is the fragment is created every time you click the icon, even if it is already the active one. Despite being the default behaviour I'm not sure if it is useful here.

@PhilippNowak96, Thanks for the feedback, as you already mentioned this is the default behaviour in the navigation components and the material design guidelines: https://material.io/components/bottom-navigation#behavior.

If you feel that the default behaviour is not providing a good user experience, I suggest that you open a feature request and explain what should be the expected behaviour. So we can better asses your request and discuss it with the UX team.

PR #2236, opened 2021-01-29
Adjust high risk card (EXPOSUREAPP-4835)

LukasLechnerDev:
Screenshot 2021-01-29 at 11 29 38

Talkback should read "An einem Tag" instead of "An eins Tag"

Unfortunately, there is a typo in the ticket number in the branch.


Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -161,15 +161,28 @@
 
     <!-- XTXT: risk card - High risk state - Days with high risk encounters -->
     <plurals name="risk_card_high_risk_encounter_days_body">
+        <item quantity="one">"Begegnungen an %1$d Tag mit erhöhtem Risiko"</item>
+        <item quantity="other">"Begegnungen an %1$d Tagen mit erhöhtem Risiko"</item>
+        <item quantity="zero">"Begegnungen an %1$d Tagen mit erhöhtem Risiko"</item>
```
SabineLoss, line 166: Do we need this? If there are no encounters with high risk, no red risk card should be displayed.

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -161,15 +161,28 @@
 
     <!-- XTXT: risk card - High risk state - Days with high risk encounters -->
     <plurals name="risk_card_high_risk_encounter_days_body">
+        <item quantity="one">"Begegnungen an %1$d Tag mit erhöhtem Risiko"</item>
+        <item quantity="other">"Begegnungen an %1$d Tagen mit erhöhtem Risiko"</item>
```
SabineLoss, line 165: What exactly would be a possible value for "other"?

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -171,15 +171,27 @@
...
         <item quantity="few">"Exposures on %1$d days with increased risk"</item>
         <item quantity="many">"Exposures on %1$d days with increased risk"</item>
     </plurals>
-    <!-- XTXT: risk card - High risk state - Most recent date with high risk -->
-    <string name="risk_card_high_risk_most_recent_body">"Most recently on %1$s"</string>
+    <!-- XACT: risk card - High risk state - Days with high risk encounters description -->
+    <plurals name="risk_card_high_risk_encounter_days_body_description">
+        <item quantity="one">"Exposures on %1$d day with increased risk"</item>
+        <item quantity="other">"Exposures on %1$d days with increased risk"</item>
+        <item quantity="zero">"Exposures on %1$d days with increased risk"</item>
```
SabineLoss, line 185: See comment for German version.

Corona-Warn-App/src/main/res/values-de/strings.xml
LukasLechnerDev: from the [Android docs](https://developer.android.com/guide/topics/resources/string-resource#Plurals)

> When the language does not require special treatment of the given quantity (as with all numbers in Chinese, or 42 in English).

We have all possible quantities in every language. This is actually not necessary - e.g. German only needs `one` and `other` and the app would also work if you only provide these two quantities. 

Android Studio shows me which quantities are relevant for each language, so I could remove the unnecessary ones if you like ...

Corona-Warn-App/src/main/res/values-de/strings.xml
LukasLechnerDev: This is true. 

But as written in https://github.com/corona-warn-app/cwa-app-android/pull/2236#discussion_r566878687 if we would get rid of the unnecessary quantities, `zero` would be gone anyway.

Corona-Warn-App/src/main/res/values-de/strings.xml
SabineLoss: No, at least not for me ;-)

Corona-Warn-App/src/main/res/values-de/strings.xml
SabineLoss: Thx

SabineLoss (dismissed):
UA approved

Corona-Warn-App/src/main/res/values-de/strings.xml
harambasicluka: That is from the translation system.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
```diff
@@ -60,6 +60,16 @@ data class IncreasedRisk(
         )
     }
 
+    fun getRiskContactBodyDescription(c: Context): String = if (daysWithEncounters == 0) {
```
mtwalli, line 63: Please rename c to context ?

LukasLechnerDev: Renamed all parameters in this class from `c` to `context`

SamuraiKek (approved):
LGTM. Tested on Huawei Mate 8 (Android 6) + Nokia 3 (Android 9)

mtwalli (approved):

PR #2242, opened 2021-01-29
SafetyNet Implementation (EXPOSUREAPP-4754)

d4rken:
* Completes the skeletons for SafetyNet related code
* Adds data donation debug menu

### In details
* I added logic the the app config's device time mechanism that allows us to know when we have had a correct device time for at least 24h otherwise Safetynet fails
* The implementation of the `DeviceAttestation` interface `CWASafetyNet` takes a payload in `ByteArray` form and will return an attestation result if there are no errors.
    * Wrong device time, incorrect nonces, packagename missmatches, corrupted response data or plain network errors will throw a `SafetyNetException` upon calling `attest(request)`.
    * If there was no error and we got a result, a `DeviceAttestation.Result` is returned. This result has a method `requirePass(SafetyNetRequirements)` to which an implementation of the class `SafetyNetRequirements` can be passed. This interface is implemented by both the analytics and the survey config and describes what "level" of safetynet attestation is required. If the attestation does not meet that level an exception is thrown.
    * `DeviceAttestation.Result.accessControlProtoBuf` contains the protobuf class that both data donation features need when communicating with the CWA server. You should call `requirePass` before doing anything with the protobuf class. 
 * All `SafetyNetException` return a specific error type `enum class SafetyNetException.Type`
 
### Testing
* Compare against tech spec


tomjschwanke:
So rooted users will be locked out? Is this really required?

d4rken:
@tomjschwanke 

> So rooted users will be locked out? Is this really necessary?

There is no intention of preventing rooted users from using the app. My own device is rooted :wink:.

This only affects the features for surveys and analytics.
The idea is that there needs to be some way to add credibility to the gathered data, otherwise it's useless.
A way to prevent malicious (or just bored) actors from skewing the data by firing up a few emulators.

As only few devices are rooted, excluding them from the data donation features, will have little impact on the overall trends the donated data will show.

I did not envision this solution myself but I didn't have a better suggestion either. I'd be happy to brainstorm an alternative solution. If you have an idea on how to achieve this without using SafetyNet, please open an issue ticket and let's discuss it!

tomjschwanke:
Thanks for the clarification. I thought that Safety-Net would now be required to pass to use the application. 

As for root detection, Safety-Net is only reliable when powered by hardware backed attestation. I have no idea how to do it otherwise though.

fynngodau:
I have some more questions on the architecture of these new "data donation" and survey features:

* When sending data to the server, is it going to verify the SafteyNet attestation server-side?
* Is there an iOS-equivalent to SafteyNet that is going to be used in the iOS version? Is it also going to be validated server-side?

mar-v-in:
> As only few devices are rooted, excluding them from the data donation features, will have little impact on the overall trends the donated data will show.

Just a reminder that by doing this you **once again** exclude users (even if just from some features) that either restrict themselves to only use free software (being it on a rooted device or not) or that run devices which don't have Google services (Huawei).

Of course I can put effort into making the SafetyNet service running on devices that don't have Google services and thereby again make sure users can use this "open-source" app on open-source operating systems. But I'd rather suggest you use some open source solution for spam protection instead of some proprietary technology.

tomjschwanke:
@mar-v-in 
https://github.com/corona-warn-app/cwa-app-android/pull/2242#issuecomment-770069361

> I did not envision this solution myself but I didn't have a better suggestion either. I'd be happy to brainstorm an alternative solution. If you have an idea on how to achieve this without using SafetyNet, please open an issue ticket and let's discuss it!

thomasaugsten:
@tomjschwanke @mar-v-in @fynngodau 
Thanks for your feedback. But we really thought this through and we see no other possibility to prevent wrong data via data donation. This feature is an opt-in feature to send for example the information when you receive an exposure risk. Because of decentralized approach we cannot check if the reported exposure risk is valid. 

For this we use SafetyNet to ensure the integrity of our via app digest. Which is send via SafetyNet JWT to the data donation server and the data donation server is checking the certificate chain of the signature and the app digest.

We don't check for rooted device and this is also not possible with basic SafetyNet.

We aware that we exclude non Google Service devices from the data donation service but there is at the moment no other way to prevent false information without identify the user.
 
We will use DeviceCheck for iOS.

thomasaugsten:
@tomjschwanke @mar-v-in @fynngodau 
I added a issue to discuss alternative solution
https://github.com/corona-warn-app/cwa-wishlist/issues/356

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
```diff
@@ -0,0 +1,113 @@
...
+        val signature: ByteArray
+    ) {
+        val nonce: String? = body.get("nonce")?.asString
+
+        val apkPackageName: String? = body.get("apkPackageName")?.asString
+
+        val basicIntegrity: Boolean = body.get("basicIntegrity")?.asBoolean == true
+        val ctsProfileMatch = body.get("ctsProfileMatch")?.asBoolean == true
+
+        val evaluationTypes = body.get("evaluationType")?.asString
```
mlenkeit, line 96: This also handles the scenario where `evaluationType` is not set at all, right?

d4rken: Correct, it would fallback to `emptyList()`.
I have added additional test cases to check that.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
```diff
@@ -0,0 +1,116 @@
...
+        )
+    }
+
+    private fun String.decodeBase64Json(): JsonObject {
+        val rawJson = decodeBase64()!!.string(Charsets.UTF_8)
+        return JsonParser.parseString(rawJson).asJsonObject
+    }
+
+    private suspend fun callClient(nonce: ByteArray): SafetyNetApi.AttestationResponse = suspendCoroutine { cont ->
+        safetyNetClient.attest(nonce, environmentSetup.safetyNetApiKey)
```
LukasLechnerDev, line 68: A `NullPointerException` is thrown here when there is no `safetyNetApiKey` defined in the environment. The error message that the user/developer then sees is not really helpful. Maybe we could tell him something like "safety net API key missing, please add it in test_environments.json"?

LukasLechnerDev: I updated the`test_environments` in our wiki so that it now includes these new keys.

d4rken: > A NullPointerException is thrown here when there is no safetyNetApiKey defined in the environment. The error message that the user/developer then sees is not really helpful. Maybe we could tell him something like "safety net API key missing, please add it in test_environments.json"?

I will add a more descriptive error to this and the other environment variables that can be accessed.

> I updated thetest_environments in our wiki so that it now includes these new keys.

I was waiting for the server environments to be final.

d4rken: As this has nothing to do with safetynet directly but our environments, I'd like to do this in an extra PR, would that be okay with you?

LukasLechnerDev (commented):
Some additional comments. I will have another look on it tomorrow to finish my Review

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/AttestationContainer.kt
```diff
@@ -0,0 +1,56 @@
...
+) : DeviceAttestation.Result {
+    override val accessControlProtoBuf: PpacAndroid.PPACAndroid
+        get() = PpacAndroid.PPACAndroid.newBuilder().apply {
+            salt = ourSalt.toByteString().base64()
+            safetyNetJws = report.jwsResult
+        }.build()
+
+    override fun requirePass(reqs: SafetyNetRequirements) {
+        if (reqs.requireBasicIntegrity && !report.basicIntegrity) {
+            throw SafetyNetException(Type.BASIC_INTEGRITY_REQUIRED, "Requirement 'basicIntegrity' not met.")
```
LukasLechnerDev, line 20: I read in the docs, that the attestation response optionally returns:
> `advice`: A suggestion for how to get a device back into a good state.

What do you think about adding the `advice` property to the `Report` and add it to the exception messages here, if available?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
LukasLechnerDev: Yes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
```diff
@@ -0,0 +1,116 @@
...
+            .addOnSuccessListener {
+                Timber.tag(TAG).v("Attestation finished with %s", it)
+                cont.resume(it)
+            }
+            .addOnFailureListener {
+                Timber.tag(TAG).w(it, "Attestation failed.")
```
LukasLechnerDev, line 74: This [checklist](https://developer.android.com/training/safetynet/attestation#use-response-server) contains: 

> Your app handles transient errors by retrying the request with an increasing amount of time between retries (exponential backoff).

Should we implement a retry mechanism?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetException.kt
```diff
@@ -1,6 +1,32 @@
...
 
 class SafetyNetException constructor(
-    message: String?,
+    val type: Type,
+    message: String,
     cause: Throwable? = null
-) : Exception(message, cause)
+) : Exception("$type: $message", cause) {
+
+    enum class Type {
```
BMItter, line 9: would it make sense, to make the enum available for the whole data donation, e.g. in an own class? ->could be reused for the survey e.g. (not only in the SafetyNetException)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
```diff
@@ -0,0 +1,116 @@
...
+            signature = signature
+        )
+    }
+
+    private fun String.decodeBase64Json(): JsonObject {
+        val rawJson = decodeBase64()!!.string(Charsets.UTF_8)
+        return JsonParser.parseString(rawJson).asJsonObject
+    }
+
+    private suspend fun callClient(nonce: ByteArray): SafetyNetApi.AttestationResponse = suspendCoroutine { cont ->
```
BMItter, line 67: possibly a timeout would be helpful here...

LukasLechnerDev: According to the [documentation](https://developer.android.com/training/safetynet/attestation#verify-attestation-response), only for the blocking version of the request (we are using the non-blocking version), a timeout is necessary:

> Note: The preceding code snippet uses a non-blocking version of the attest() method. If you're making a blocking call instead, always set a response timeout. That way, your app remains responsive, even if a slow network connection causes a delayed response from the SafetyNet Attestation API.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapperTest.kt
```diff
@@ -0,0 +1,216 @@
...
+                "nonce": "AAAAAAAAAAAAAAAAAAAAAA==",
+                "timestampMs": 1608558363702,
+                "apkPackageName": "de.rki.coronawarnapp.test",
+                "apkDigestSha256": "9oiqOMQAZfBgCnI0jyN7TgPAQNSSxWrjh14f0eXpB3U=",
+                "ctsProfileMatch": false,
+                "apkCertificateDigestSha256": [
+                    "9VLvUGV0Gkx24etruEBYikvAtqSQ9iY6rYuKhG+xwKE="
+                ],
+                "basicIntegrity": false,
+                "nobodyExpects":"TheSpanishInquisition",
```
LukasLechnerDev, line 176: 😂

LukasLechnerDev:
Also from the [checklist](https://developer.android.com/training/safetynet/attestation-checklist)

> You can configure your app to function normally when the SafetyNet Attestation API experiences a large-scale outage.

In this case our data donation features simply don't work and that's okay for us, because otherwise we would be vulnerable for misuse, right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/CWASafetyNet.kt
```diff
@@ -1,19 +1,93 @@
...
+            )
+        }
+
+        if (context.packageName != report.apkPackageName) {
+            throw SafetyNetException(
+                Type.APK_PACKAGE_NAME_MISMATCH,
+                "Our APK name doesn't match response (${context.packageName} != ${report.apkPackageName})"
+            )
+        }
+
```
LukasLechnerDev, line 86: Quote from this [article](https://android-developers.googleblog.com/2017/11/10-things-you-might-be-doing-wrong-when.html)

> ... and use timestampMs to check how much time has passed since you made the request and you got the response. A delayed response that arrives several hours or days after the request may suggest suspicious activity.

Should we check the timestamp too?

LukasLechnerDev: Okay, the TechSpec says that this is done on the server.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/safetynet/CWASafetyNetTest.kt
```diff
@@ -0,0 +1,195 @@
...
+        appConfigProvider = appConfigProvider,
+        googleApiVersion = googleApiVersion,
+        timeStamper = timeStamper,
+        cwaSettings = cwaSettings
+    )
+
+    @Test
+    fun `salt generation returns new value each call`() {
+        createInstance().apply {
+            generateSalt() shouldBe "LMK0jFCu/lOzl07ZHmtOqQ==".decodeBase64()!!.toByteArray()
```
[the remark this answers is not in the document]
d4rken: I've adjusted the function name.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/safetynet/AttestationContainerTest.kt
```diff
@@ -0,0 +1,237 @@
...
+            }
+       """.trimIndent().let { create(it) }
+
+        shouldNotThrowAny {
+            attestation.requirePass(SafetyNetRequirementsContainer())
+        }
+    }
+
+    @Test
+    fun `basic integrity required`() {
```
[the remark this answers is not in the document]
d4rken: I agree, for now I would like to keep it this way. If we add additional test cases we can re-evaluate it. Would that be okay with you?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/AttestationContainer.kt
d4rken: That's a nice thought. Will do.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
```diff
@@ -15,7 +15,14 @@ data class ConfigDataContainer(
 ) : ConfigData, ConfigMapping by mappedConfig {
 
     override val isDeviceTimeCorrect: Boolean
```
LukasLechnerDev, line 17: What do you think about renaming this property to `isDeviceTimeCorrectOrAssumedCorrect` and add another property `isDeviceTimeCorrect` that simply checks `if(deviceTimeState ==ConfigData.DeviceTimeState.CORRECT)`. 

We need to check `isDeviceTimeCorrect` before performing data donation features.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/safetynet/AttestationContainerTest.kt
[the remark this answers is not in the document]
LukasLechnerDev: Is fine for me. I just want to raise awareness for writing `ParemeterizedTests` in the future as they make tests much more readable IMO.

LukasLechnerDev (dismissed):
Good Work 👍.

My comments are optional and no blockers.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
d4rken: I'm not sure we can make the decision at this level. For analytics we can retry a lot and also delay for longer, it's in the background. For the survey stuff the user is active on screen. Both cases likely need different retry behavior.

I'd say this is up to the classes consuming the API?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetException.kt
d4rken: > would it make sense, to make the enum available for the whole data donation, e.g. in an own class?

This is available for the whole data donation, isn't it?
Both survey and analytics use safetynet, and if they receive an exception they can check the error type if they want to perform specific behavior depending on what error it is.
As this is an error type specific to SafetyNet and only available via `SafetyNetException`s being thrown it makes sense to keep it here imho.

> could be reused for the survey e.g. (not only in the SafetyNetException)

According to the tech spec, these error codes are specifically related to safetynet behavior. Did I missunderstand the spec or are you suggestion to put all error types (not just safetynet) into the same enum?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
d4rken: A timeout is useful, but as survey and analytics are very different use-cases I'd prefer if the timeout is handled downstream by the caller. What do you think?

d4rken:
> > Also from the checklist
> > You can configure your app to function normally when the SafetyNet Attestation API experiences a large-scale outage.

> In this case our data donation features simply don't work and that's okay for us, because otherwise we would be vulnerable for misuse, right?

If the SafetyNet API is down there will be an exception or you'd run into a timeout, the caller survey or analytics can then decide what to do, I'd expect them to currently treat any exception the same was as "attestation" failed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
d4rken: I've added a timeout of 30 seconds for the attestation call as "maximum", lower timeout values can be enforced by the caller.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
d4rken: > What do you think about renaming this property to isDeviceTimeCorrectOrAssumedCorrect and add another property isDeviceTimeCorrect that simply checks if(deviceTimeState ==ConfigData.DeviceTimeState.CORRECT).

Maybe, changing the name or what it returns would be out of scope for this PR as it would affect many other features.

> We need to check isDeviceTimeCorrect before performing data donation features.

You don't need to do that manually. According to the tech spec the attestation should (and my implementation does?) check the device time and return a `SafetyNetException` if the device time is not sufficiently accurate.

API consumers (survey+analytics) should just need to call the classes in this PR, and only proceed if there was no exception.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
LukasLechnerDev: True, there are different circumstances in which our API is called...
Then it makes sense to leave it to the consumers of the API.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainer.kt
LukasLechnerDev: You are right, I missed that `CWASafetyNet` is doing the correct `deviceTime` checks. I thought it would call this method here.

LukasLechnerDev (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetException.kt
BMItter: "_are you suggestion to put all error types (not just safetynet) into the same enum?_"
yes, that was the idea. Just one allover enum class for all error codes regarding dataDonation...

d4rken: Hm, I'm not sure when this would be an advantage. Which class would have to handle all types of errors? By scoping it we have more robust error handling, e.g. SafetyNet errors will only popup in specific locations in survey and analytics core logic, as the enum error types are specific to SafetyNet and need to be handled in the survey and analytics class, and not further downstream.

The current typing would allow `ImaginarySurveyRepo` to just use a `when(error.type) { }` and let the compiler enforce that all error types that can happen during a SafetyNet check are handled as desired.

Any additional exceptions e.g. `SurveyLogicSaysNoException` that occur in the survey related logic could have their own exception typing if necessary, and could also include the `SafetyNetException` as cause, but I'd expect any classes downstream of `ImaginarySurveyRepo` to not need to be aware of the ways that SafetyNet can fail. This should be a hidden implementation detail of the class that uses the SafetyNet code.

:thinking:  a bit like "per feature not per layer", i.e. group errors by feature/module, not by layer, i.e. "all errors in one package".

BMItter: I think, both ways are possible.
It won't stop the train..

BMItr (approved):
lgtm,
thx for addressing some comments

chris-cwa (approved):

PR #2255, opened 2021-02-01
Collect Exposure Risk Metadata (EXPOSUREAPP-4921)

kolyaopahle:
### Description
This PR implements the first metadata donator for the privacy preserving analytics feature,
it also contains a few small changes to the dontator interface, mainly using the now provided proto classes where applicable 

### Steps to reproduce
1. check if the metadata generating code is up to spec with https://github.com/corona-warn-app/cwa-app-tech-spec/blob/e88e3c2b6e9ed76d34ac744974e79d912b3b078f/docs/spec/data-donation-client-ppa.md#exposure-risk-metadata


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsSettings.kt
```diff
@@ -1,7 +1,35 @@
...
+        key = PREVIOUS_EXPOSURE_RISK_METADATA,
+        reader = { key ->
+            getString(key, null)?.let {
+                ExposureRiskMetadata.parseFrom(ByteString.copyFromUtf8(it))
+            }
+        },
+        writer = { key, value ->
+            putString(key, value?.toByteString()?.toStringUtf8())
+        }
+    )
```
d4rken, line 30: Be cautious here, we are using two different implementations for data conversion.

* On read: we use `ByteString` from `com.google.protobuf`
* On write: we use `ByteString` conversion functions from Kotlin

Can we use read/write from the same implementation source?
e.g. for a Kotlin only solution:

```kotlin
        ByteArray(1).toByteString().base64()
        "".decodeBase64()!!.toByteArray()
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsSettings.kt
```diff
@@ -1,7 +1,35 @@
...
 @Singleton
-class AnalyticsSettings @Inject constructor()
+class AnalyticsSettings @Inject constructor(
+    @AppContext private val context: Context
+) {
+    private val prefs by lazy {
+        context.getSharedPreferences("analytics_localdata", Context.MODE_PRIVATE)
+    }
+
+    val previousExposureRiskMetadata = FlowPreference(
```
d4rken, line 19: Please add a unit test for the conversion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposureriskmetadata/ExposureRiskMetadataDonor.kt
```diff
@@ -0,0 +1,75 @@
...
+            .setRiskLevelChangedComparedToPreviousSubmission(previousMetadata?.riskLevel != riskLevelForMetadata)
+            .setMostRecentDateAtRiskLevel(lastRiskResult.calculatedAt.millis)
+            .setDateChangedComparedToPreviousSubmission(
+                previousMetadata?.mostRecentDateAtRiskLevel != lastRiskResult.calculatedAt.millis
+            )
+            .build()
+
+        return ExposureRiskMetadataContribution(
+            contributionProto = newMetadata,
+            onContributionFinished = {
```
d4rken, line 42: I'm wondering if we need additional guards on `previousExposureRiskMetadata` to prevent threading issues.

d4rken (commented):
Looks good. Some comments on LOCs where we might need to code a bit more defensively.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsSettings.kt
kolyaopahle: done 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsSettings.kt
kolyaopahle: done 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposureriskmetadata/ExposureRiskMetadataDonor.kt
kolyaopahle: hmm, i'm not sure, technically analytics collections only happen every 24 hours right, so we should not be able to call this in succession often enough to be a problem right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsSettings.kt
```diff
@@ -15,6 +19,21 @@ class AnalyticsSettings @Inject constructor(
         context.getSharedPreferences("analytics_localdata", Context.MODE_PRIVATE)
     }
 
+    val previousExposureRiskMetadata = FlowPreference(
+        preferences = prefs,
+        key = PREVIOUS_EXPOSURE_RISK_METADATA,
+        reader = { key ->
+            getString(key, null)?.let { prefString ->
+                prefString.decodeBase64()?.toByteArray()?.let {
+                    ExposureRiskMetadata.parseFrom(it)
```
d4rken, line 28: What happens if there is a change (accidental?) to the protobuf files or the way they are generated. Could this throw an exception? Will we remember that there is data that might throw an exception then?

Would it be okay for logic consuming this value, to receive `null` in that case, so we could just wrap this with a warm & comfy try-catch?

d4rken (commented):
Looks good, can we get someone else to take a look too?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsSettings.kt
kolyaopahle: logic handling with null is fine here, also added a test that validates changes in the proto definition that we need to handle

ralfgehrer (approved):

d4rken (approved):

PR #2257, opened 2021-02-01
Delta-Onboarding interoperability info not shown (EXPOSUREAPP-4910)

LukasLechnerDev:
A description of how to test is in the Jira ticket description. 

The problem was, that in our `HomeFragmentViewModel` => `postValue(ShowInteropDeltaOnboarding)` and `postValue(HomeFragmentEvents.ShowNewReleaseFragment)` were both executed within a short timespan and our `HomeFragment` therefore only observed the later event - `HomeFragmentEvents.ShowNewReleaseFragment`.

My solution here is to not `post()` both `Events` as soon as the user sees our `HomeFragment`, but to at first do `postValue(ShowInteropDeltaOnboarding)` to navigate to the Delta Onboarding Screen and when the user returns to the `HomeFragment` again, `showPopupsOrNavigate()` is executed again and the other event `HomeFragmentEvents.ShowNewReleaseFragment` will be fired to trigger navigation to `NewReleaseInfoFragment`. 

I think this logic needs to be refactored at some point, but since we want to release this week, I wanted to be as minimal invasive as possible to not introduce new bugs. 

The small problem with this solution is that we access 2 values of our shared preferences whenever the user returns to the `HomeFragment`, which costs us a bit of performance.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -488,7 +488,7 @@ jobs:
               --os-version-ids 29 \
               --locales de_DE \
               --orientations portrait \
-              --no-record-video
+              --no-record-video | true
           no_output_timeout: 30m
       - run:
           name: Create directory to store test results
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -488,7 +488,7 @@ jobs:
               --os-version-ids 29 \
               --locales de_DE \
               --orientations portrait \
-              --no-record-video | true
+              --no-record-video
           no_output_timeout: 30m
       - run:
           name: Create directory to store test results
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
@@ -91,28 +91,27 @@ class HomeFragmentViewModel @AssistedInject constructor(
         .map { it.toHeaderState() }
         .asLiveData(dispatcherProvider.Default)
 
-    val popupEvents: SingleLiveEvent<HomeFragmentEvents> by lazy {
-        SingleLiveEvent<HomeFragmentEvents>().apply {
-            if (!LocalData.isInteroperabilityShownAtLeastOnce) {
-                postValue(ShowInteropDeltaOnboarding)
-            } else {
-                launch {
-                    if (!LocalData.tracingExplanationDialogWasShown()) {
-                        postValue(
-                            ShowTracingExplanation(
-                                TimeVariables.getActiveTracingDaysInRetentionPeriod()
-                            )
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -488,7 +488,7 @@ jobs:
               --os-version-ids 29 \
               --locales de_DE \
               --orientations portrait \
-              --no-record-video
+              --no-record-video | true
```
mtwalli, line 491: Please revert this, I have updated my PR, this change should achieve the right behaviour https://github.com/corona-warn-app/cwa-app-android/pull/2258/commits/e53c6762d6c41f540cf761620944601c29f099da

LukasLechnerDev: Reverted.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -91,28 +91,27 @@ class HomeFragmentViewModel @AssistedInject constructor(
...
-                            )
+    val popupEvents = SingleLiveEvent<HomeFragmentEvents>()
+
+    fun showPopUpsOrNavigate() {
+        if (!LocalData.isInteroperabilityShownAtLeastOnce) {
+            popupEvents.postValue(ShowInteropDeltaOnboarding)
+        } else if (cwaSettings.lastChangelogVersion.value < BuildConfigWrap.VERSION_CODE) {
+            popupEvents.postValue(HomeFragmentEvents.ShowNewReleaseFragment)
+        } else {
+            launch {
```
BMItter, line 102: why 2 different launch blocks? (not needed)
also using "when" instead of if else blocks is good practice :older_woman:

BMItter:
looking good overall 👌 
left one comment (but just minor)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
LukasLechnerDev: This is not code that I have written. I only moved 

`if (cwaSettings.lastChangelogVersion.value < BuildConfigWrap.VERSION_CODE) {
                postValue(HomeFragmentEvents.ShowNewReleaseFragment)
}`

upwards into the `else if` condition. I am also not sure why there are 2 `launch{}` blocks but I don't want to remove one because they might be there for a reason we don't know.

SamuraiKek (approved):
LGTM. Tested on Nokia 3 (Android 9). First installed version 1.3.1, did the onboarding, then installed v 1.12 from the PR. The delta-onboarding screen first showed up, followed by the new release screen.

BMItr (approved):

PR #2260, opened 2021-02-01
Fix: First and last item of statistic cards was not reachable on some devices (EXPOSUREAPP-4945) (closes #2259)

### LINKED ISSUE #2259: Statistics snaps to second card after swiping right
## Avoid duplicates

## Technical details

- Device name: Samsung Galaxy Tablet A SM-T510
10.1inch 1200 x 1920 pixels, 16:10 ratio (~224 ppi density)
- Android version: 10
- App version: ~~1.11.0~~ 1.12.0

## Describe the bug

After swiping the statistics left, then swiping right it is not possible to leave the card "Confirmed New Infections" showing completely.

The same issue appears when swiping completely left. The card "7-Day R Value" is not completely displayed.

1. Scroll down to the statistics cards starting with "Confirmed New Infections" which looks similar to the following screenshot
Starting statistics
2. Now swipe to the left and back to the right fully. The first card on the left is no longer completely displayed. 
Snapped statistics
3. To restore the view, tap on any of the ℹ️ icon and then tap the back button.

## Expected behaviour

If the cards are swiped fully to the right, then they should stay there and not be snapped slightly to the left again and vice versa.

## Possible Fix

PagerSnapHelper() is responsible. 

## Additional context

I found only my tablet (16:10 ratio), but not my phone (19.5:9 ratio) is affected. I don't know exactly the conditions where it starts to become problematic unfortunately, but it looks like it may only affect relatively wider devices.

---
Internal Tracking ID: [EXPOSUREAPP-4945](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4945)

PhilippNowak96:
### Description

As written in #2259 the first (and the last) card of the statistics cards were not 100% reachable on some devices due to a problem in the snap functionality.

The fix basically skips the snap if the first completely visible item is also the first one of the whole list. The same goes for the last item. The approach is basically taken from here: https://stackoverflow.com/questions/52688974/recyclerview-snaphelper-fails-to-show-first-last-items/58876237#58876237
It's linked in the class comment, i just added two small comments to better understand whats going on.

### Steps to reproduce
1. Create a Nexus 10 (2560x1600) Android 10 Emulator or take a Samsung Galaxy Tablet A SM-T510 10.1inch 1200x1920 device (others might be affected too, but it's more the exception than the regular case)
2. Swipe left on the statistics cards to reach the rightmost one. See how it snaps a little back and the last card is not fully visibile.
3. Swipe right to reach the leftmost card. See how it snaps a little back and the first card is not fully visible.

#### Comment

It feels like the behaviour is a little different than before (was a little smoother with the PagerSnapHelper IMO). But I don't feel like it's a UX problem in the end. It's still working fine for me.

---
Internal Tracking ID: [EXPOSUREAPP-4945](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4945)


mtwalli:
Thank you very much for the PR! while it might work for first and last items, I still can experience this for other items
I would suggest to disable the snapping behaviour for tablets.
Something like in `StatisticsHomeCard`:
```kotlin
val isPhone = resources.configuration.smallestScreenWidthDp < 600
if (isPhone) {
    PagerSnapHelper().attachToRecyclerView(statisticsRecyclerview)
}
```

what do you think about that?

PhilippNowak96:
@mtwalli Were you able to reproduce anything like this? I was not after the fix but 4 items are not that much of course to try to reproduce anything like this

mtwalli:
> @mtwalli Were you able to reproduce anything like this? I was not after the fix but 4 items are not that much of course to try to reproduce anything like this

Yes I was.
1- Please use a Nexus 10 emulator in Portrait mode 
2-Try to scroll the item before the last item while it is out of the screen while it is completely visible leave it it will snap out of the screen again

MikeMcC399:
@mtwalli

Certainly disabling `PagerSnapHelper()` works on the tablet where I first noticed this issue. That tablet has a smallestScreenWidthDp = 800dp. At 600dp, two cards (which are 300dp wide) can be displayed almost completely side-by-side. Padding and card separation puts the two cards just very slightly over 600dp. Manual scrolling on a wide display of 800dp didn't feel difficult and is acceptable. 

On a narrower phone display which I tried with smallestScreenWidthDp = 360dp it was nice to have the PagerSnapHelper() align each tile so it could be completely displayed.

So to me the workaround of disabling `PagerSnapHelper()` for tablets looks good.

PhilippNowak96:
@mtwalli Ah I understood what you mean. I can reproduce it as well. I pushed a commit to apply your suggestion which seems to be okay in this case.

mtwalli (dismissed):
Thank you very much!

MikeMcC399:
@mtwalli 
Will this PR be merged before the 1.12.0 release?

ralfgehrer:
@PhilippNowak96 many thanks for your contribution. May I ask you to please rebase the branch onto `release/1.13.x`

PhilippNowak96:
@ralfgehrer Done 👍🏼

mtwalli (approved):

MikeMcC399 (approved):
I successfully checked that this fix is still working for me after this branch was re-based on release/1.13.x

On tablet - no snapping
On phone - snapping

Both as expected. 👍

ralfgehrer (approved):
Successfully tested as well. Many thanks @PhilippNowak96 and @MikeMcC399.

PR #2262, opened 2021-02-02
Pull firebase results anyways (Dev)

mtwalli:
Run pulling results always regardless of the result of previous steps.
- Suppressing the error only did allow to grab the report ,but it gave false impression that the job succeeded.
- Verified in this build: https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/9145/workflows/736a7a5f-7c26-4978-af97-98f16443d6f5/jobs/50356

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -493,11 +493,13 @@ jobs:
       - run:
           name: Create directory to store test results
           command: mkdir firebase-results
+          when: always
       - run:
           name: Install gsutil dependency and copy test results data
           command: |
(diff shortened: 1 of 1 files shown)


BMItr (approved):
pulling result and test completion is working. (even if the app crashes)

LukasLechnerDev (approved):
Looks good! We can now see the information about the failing test and the pipeline isn't green anymore when a test fails. 

Helped me to fix the pipeline in my bugfix!

PR #2263, opened 2021-02-02
Change escaping of percentage sign (DEV)

ralfgehrer:
All other options did not work well with the internal translation tool.

TEST:
- Check the statistics info screen.
- Scroll to Legend
- See Trend
- Check the description and make sure it says "...1%..."  and "...5%..."

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/StatisticsExplanationFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/StatisticsExplanationFragment.kt
@@ -28,9 +28,11 @@ class StatisticsExplanationFragment : Fragment(R.layout.fragment_statistics_expl
             R.string.statistics_explanation_seven_day_r_link_label,
             R.string.statistics_explanation_faq_url
         )
+
         binding.statisticsExplanationTrendText.apply {
-            text = String.format(getString(R.string.statistics_explanation_trend_text))
-            contentDescription = String.format(getString(R.string.statistics_explanation_trend_text))
+            val label = String.format(getString(R.string.statistics_explanation_trend_text))
+            text = label
+            contentDescription = label
         }
     }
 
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/StatisticsExplanationFragment.kt
```diff
@@ -28,6 +28,12 @@ class StatisticsExplanationFragment : Fragment(R.layout.fragment_statistics_expl
             R.string.statistics_explanation_seven_day_r_link_label,
             R.string.statistics_explanation_faq_url
         )
+
+        binding.statisticsExplanationTrendText.apply {
+            val label = String.format(getString(R.string.statistics_explanation_trend_text))
+            text = label
+            contentDescription = label
```
harambasicluka, line 35: You just adapted what was there in XML, but why do we even need `contentDescription` on a `TextView` without any difference to the text?.

harambasicluka (approved):
Finally! :)
IMO the `contentDescription` can be removed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/StatisticsExplanationFragment.kt
harambasicluka: Is it due to?
> Check the description and make sure it says "...1%..." and "...5%..."

ralfgehrer: In this case, yes.

axelherbstreith:
Looking good on Samsung S8 but somehow the app crashes when opening the screen on my P30... not sure what causes this

BMItr (approved):
tested on Nexus 5x and Samsung S20. looking good 👌 
PS: I also checked on a P20 Device, but was also working as expected (no crash). But I have no P30

axelherbstreith (approved):

ralfgehrer:
> tested on Nexus 5x and Samsung S20. looking good 👌
> PS: I also checked on a P20 Device, but was also working as expected (no crash). But I have no P30

The crash was due to English language settings. This will updated with the upcoming Translation Delivery. @axelherbstreith  could validate it with German language settings.

PR #2264, opened 2021-02-02
Scrolling issues with back navigation from Statistics Info screen to Homescreen (EXPOSUREAPP-4771)

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
```diff
@@ -71,52 +71,49 @@ class SubmissionRepository @Inject constructor(
         }
     }
 
-    // TODO this should be more UI agnostic
     fun refreshDeviceUIState(refreshTestResult: Boolean = true) {
-        var refresh = refreshTestResult
+        if (LocalData.submissionWasSuccessful()) {
```
d4rken, line 75: Why is this necessasry to fix the scrolling issue of the statistics?

I like the refactoring, but it would be better as this is part of an extra PR so it's easier to explicitly test these changes.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/modular/ModularAdapter.kt
```diff
@@ -7,6 +7,7 @@ import de.rki.coronawarnapp.ui.lists.BaseAdapter
 
 abstract class ModularAdapter<VH : ModularAdapter.VH> : BaseAdapter<VH>() {
     val modules = mutableListOf<Module>()
+    val holders = mutableListOf<VH>()
```
d4rken, line 10: Don't we create a memory leak here?

I think the RecyclerView creates the viewholders on demand and we now store a reference to all viewholders that were ever created.

chiljamgossow: I need access to the instance of the viewholder to reset the scrolling state. Do you have a better idea?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
chiljamgossow: Yes, this is part of the bug ticket. It runs through several states quickly on each resume, resulting in quick changes of the list items.

d4rken: Ok :+1:.

Would be nice to have a little bit more PR description :grin:. 
What/Why, and what needs explicit testing, e.g. run a test result through all possible test states and go back and forth between home screen and statistics info.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/modular/ModularAdapter.kt
d4rken: I'll look into it.

chiljamgossow: I added a clear after usage which would be called onPause()

d4rken (approved):

ralfgehrer (approved):
Tested on Pixel (Android 10)

PR #2269, opened 2021-02-02
PPA User Input Strings Only (EXPOSUREAPP-4752)

d4rken:


AlexanderAlferov (dismissed):

ralfgehrer (dismissed):

harambasicluka (dismissed):

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1644,4 +1644,60 @@
...
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_SL -->
+    <string name="analytics_userinput_federalstate_sl">Saarland</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_SN -->
+    <string name="analytics_userinput_federalstate_sn">Sachsen</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_ST -->
+    <string name="analytics_userinput_federalstate_st">Sachsen-Anhalt</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_SH -->
+    <string name="analytics_userinput_federalstate_sh">Schleswig-Holstein</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_TH -->
+    <string name="analytics_userinput_federalstate_th">Thüringen</string>
```
mtwalli, line 1697: I suggest to to have the states as a String array resource

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1666,4 +1666,60 @@
...
+    <!-- XHED: Analytics voluntary user input, age group toolbar title -->
+    <string name="analytics_userinput_agegroup_title">Ihr Alter</string>
+    <!-- XTXT: Analytics voluntary user input, age group: AGE_GROUP_UNSPECIFIED -->
+    <string name="analytics_userinput_agegroup_unspecified">keine Angabe</string>
+    <!-- XTXT: Analytics voluntary user input, age group: AGE_GROUP_0_TO_29 -->
+    <string name="analytics_userinput_agegroup_0_to_29">bis 30 Jahre</string>
+    <!-- XTXT: Analytics voluntary user input, age group: AGE_GROUP_30_TO_59 -->
+    <string name="analytics_userinput_agegroup_30_to_59">30 bis 60</string>
+    <!-- XTXT: Analytics voluntary user input, age group: AGE_GROUP_FROM_60 -->
+    <string name="analytics_userinput_agegroup_from_60">60 oder älter</string>
```
mtwalli, line 1682: here too?

Corona-Warn-App/src/main/res/values-de/strings.xml
d4rken: That was my first idea too. I really liked it when @chiljamgossow  did it for the release notes.

I decided against it here, as `strings.xml` only supports arrays, and not key-value mapping. If we use a `string-array` here, then we are very succeptible to subtle bugs. If someone or something (translation pipeline) changes the order of the array, in any of the languages, then the user selects something, but something else gets send to the server. We will likely never notice.

I think it's better to just add the entries than requiring us to write unit tests for the array order in every locale.

Am I overlooking a solution here that makes the array approach less susceptible to bugs?

mtwalli (dismissed):

janetback (dismissed):
UA approved

PR #2270, opened 2021-02-02
PPA UserInput Selection Screens (EXPOSUREAPP-4752)

d4rken:
A fragment that allows the user to input agegroup, federal state and district.
The selections are stored in `AnalyticsSettings`.

There is a `Playground` entry in the test menu that allows you to check out these screens until @AlexanderAlferov has integrated them into the final settings screens.

Final use will look like:

```kotlin
doNavigiate(Directions.....(type=AnalyticsUserInputFragment.InputType.AGEGROUP)
```

when the user has selects something the fragment will close and if you will get the new value via the `AnalyticsSettings` instance.

### Testing
* Compare with tech spec
* Check sorting
* Selection/Unselection

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputFragment.kt
@@ -36,8 +36,8 @@ class AnalyticsUserInputFragment : Fragment(R.layout.analytics_ppa_userinfo_inpu
         super.onViewCreated(view, savedInstanceState)
 
         val toolbarLabel = when (navArgs.type) {
-            InputType.AGEGROUP -> R.string.analytics_userinput_agegroup_title
-            InputType.FEDERALSTATE -> R.string.analytics_userinput_federalstate_title
+            InputType.AGE_GROUP -> R.string.analytics_userinput_agegroup_title
+            InputType.FEDERAL_STATE -> R.string.analytics_userinput_federalstate_title
             InputType.DISTRICT -> R.string.analytics_userinput_district_title
         }
         binding.toolbar.apply {
@@ -57,8 +57,8 @@ class AnalyticsUserInputFragment : Fragment(R.layout.analytics_ppa_userinfo_inpu
     }
 
     enum class InputType {
-        AGEGROUP,
-        FEDERALSTATE,
+        AGE_GROUP,
+        FEDERAL_STATE,
         DISTRICT
     }
 }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputFragment.kt
@@ -1,0 +1,64 @@
+package de.rki.coronawarnapp.datadonation.analytics.ui.input
+
+import android.os.Bundle
+import android.view.View
+import androidx.fragment.app.Fragment
+import androidx.navigation.fragment.navArgs
+import androidx.recyclerview.widget.LinearLayoutManager
+import de.rki.coronawarnapp.R
+import de.rki.coronawarnapp.databinding.AnalyticsPpaUserinfoInputFragmentBinding
+import de.rki.coronawarnapp.util.di.AutoInject
+import de.rki.coronawarnapp.util.ui.observe2
+import de.rki.coronawarnapp.util.ui.popBackStack
+import de.rki.coronawarnapp.util.ui.viewBindingLazy
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactoryProvider
+import de.rki.coronawarnapp.util.viewmodel.cwaViewModelsAssisted
+import javax.inject.Inject
+
+class AnalyticsUserInputFragment : Fragment(R.layout.analytics_ppa_userinfo_input_fragment), AutoInject {
+
+    val navArgs by navArgs<AnalyticsUserInputFragmentArgs>()
+
+    @Inject lateinit var viewModelFactory: CWAViewModelFactoryProvider.Factory
+    private val vm: AnalyticsUserInputViewModel by cwaViewModelsAssisted(
+        factoryProducer = { viewModelFactory },
+        constructorCall = { factory, _ ->
+            factory as AnalyticsUserInputViewModel.Factory
+            factory.create(navArgs.type)
+        }
+    )
+
+    private val binding: AnalyticsPpaUserinfoInputFragmentBinding by viewBindingLazy()
+
+    @Inject lateinit var itemAdapter: UserInfoItemAdapter
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+
+        val toolbarLabel = when (navArgs.type) {
+            InputType.AGEGROUP -> R.string.analytics_userinput_agegroup_title
+            InputType.FEDERALSTATE -> R.string.analytics_userinput_federalstate_title
+            InputType.DISTRICT -> R.string.analytics_userinput_district_title
+        }
+        binding.toolbar.apply {
+            setTitle(toolbarLabel)
+            setNavigationOnClickListener { popBackStack() }
+        }
+
+        binding.inputList.apply {
+            layoutManager = LinearLayoutManager(requireContext())
+            adapter = itemAdapter
+        }
+        vm.userInfoItems.observe2(this) {
+            itemAdapter.data = it
+        }
+        itemAdapter.onItemClickListener = { vm.selectUserInfoItem(it) }
+        vm.finishEvent.observe2(this) { popBackStack() }
+    }
+
+    enum class InputType {
+        AGEGROUP,
+        FEDERALSTATE,
+        DISTRICT
+    }
+}
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputViewModel.kt
@@ -1,0 +1,121 @@
+package de.rki.coronawarnapp.datadonation.analytics.ui.input
+
+import android.content.Context
+import androidx.annotation.StringRes
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.asLiveData
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import de.rki.coronawarnapp.R
+import de.rki.coronawarnapp.contactdiary.ui.day.ContactDiaryDayViewModel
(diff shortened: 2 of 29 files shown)


d4rken:
These are the same strings as in #2269

d4rken:
The radiobutton checkboxes currently have the default color, while they should have the accent color, I'm wondering whether the easiest way to fix this is to wait for #2235 to be merged (which adjusts the theme to material design) and then use the MD radio button class here.... :thinking: 

Is there an easy way to change the color? `buttonTint` colors both the selected and the unselected state....

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputViewModel.kt
```diff
@@ -0,0 +1,177 @@
...
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactory
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.catch
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.map
+import timber.log.Timber
+import java.util.Locale
+
+class AnalyticsUserInputViewModel @AssistedInject constructor(
```
chris-cwa, line 29: much of what is going on here belongs to domain layer or data access layer. why put all this in 1 file? please consider some investment into app architecture to increase later-on maintainability.

chris-cwa (changes requested):

Corona-Warn-App/src/main/res/layout/analytics_ppa_userinfo_input_adapter_item.xml
```diff
@@ -0,0 +1,29 @@
...
+    <RadioButton
+        android:id="@+id/radiobutton"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content"
+        android:clickable="false"
+        android:focusable="false"
+        android:layout_marginStart="16dp"
+        style="@style/Widget.AppCompat.CompoundButton.RadioButton"
+        android:layout_marginTop="8dp"
+        android:layout_marginBottom="8dp" />
```
mtwalli, line 17: Please reformat the file?

Corona-Warn-App/src/main/res/layout/analytics_ppa_userinfo_input_fragment.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools">
```
mtwalli, line 4: Is it still required to have this tag? if `ViewBinding` is enough we should get rid of it

Corona-Warn-App/src/main/res/layout/analytics_ppa_userinfo_input_fragment.xml
```diff
@@ -0,0 +1,28 @@
...
+            android:background="@color/colorBackground"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent"
+            tools:title="Ihr Alter/Bundesland/Kreis " />
+
+        <androidx.recyclerview.widget.RecyclerView
+            android:id="@+id/input_list"
+            android:layout_width="match_parent"
+            android:layout_height="match_parent" />
```
mtwalli, line 25: Please use `tools:listitem` ? it gives the other overview if later the layout is visited

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1610,4 +1610,60 @@
...
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_SL -->
+    <string name="analytics_userinput_federalstate_sl">Saarland</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_SN -->
+    <string name="analytics_userinput_federalstate_sn">Sachsen</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_ST -->
+    <string name="analytics_userinput_federalstate_st">Sachsen-Anhalt</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_SH -->
+    <string name="analytics_userinput_federalstate_sh">Schleswig-Holstein</string>
+    <!-- XTXT: Analytics voluntary user input, federal state: FEDERAL_STATE_TH -->
+    <string name="analytics_userinput_federalstate_th">Thüringen</string>
```
mtwalli, line 1663: I added a comment regarding this in strings PR

mtwalli:
> The radiobutton checkboxes currently have the default color, while they should have the accent color, I'm wondering whether the easiest way to fix this is to wait for #2235 to be merged (which adjusts the theme to material design) and then use the MD radio button class here.... 🤔
> 
> Is there an easy way to change the color? `buttonTint` colors both the selected and the unselected state....

1- ` android:buttonTint="@color/radio_button_tint"`
2- 
```<?xml version="1.0" encoding="utf-8"?>
<selector xmlns:android="http://schemas.android.com/apk/res/android">
    <item android:color="@color/colorAccent"  android:state_checked="true"/>
    <item android:color="@color/colorStableHairlineLight" />
</selector>
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+        }
+        vm.userInfoItems.observe2(this) {
+            itemAdapter.data = it
+        }
+        itemAdapter.onItemClickListener = { vm.selectUserInfoItem(it) }
+        vm.finishEvent.observe2(this) { popBackStack() }
+    }
+
+    enum class InputType {
+        AGEGROUP,
```
mtwalli, line 60: `AGE_GROUP`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+        FEDERALSTATE,
```
mtwalli, line 61: `FEDERAL_STATE`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputViewModel.kt
```diff
@@ -0,0 +1,177 @@
...
+        PpaData.PPAFederalState.FEDERAL_STATE_MV -> R.string.analytics_userinput_federalstate_mv
+        PpaData.PPAFederalState.FEDERAL_STATE_NI -> R.string.analytics_userinput_federalstate_ni
+        PpaData.PPAFederalState.FEDERAL_STATE_NRW -> R.string.analytics_userinput_federalstate_nrw
+        PpaData.PPAFederalState.FEDERAL_STATE_RP -> R.string.analytics_userinput_federalstate_rp
+        PpaData.PPAFederalState.FEDERAL_STATE_SL -> R.string.analytics_userinput_federalstate_sl
+        PpaData.PPAFederalState.FEDERAL_STATE_SN -> R.string.analytics_userinput_federalstate_sn
+        PpaData.PPAFederalState.FEDERAL_STATE_ST -> R.string.analytics_userinput_federalstate_st
+        PpaData.PPAFederalState.FEDERAL_STATE_SH -> R.string.analytics_userinput_federalstate_sh
+        PpaData.PPAFederalState.FEDERAL_STATE_TH -> R.string.analytics_userinput_federalstate_th
+        PpaData.PPAFederalState.UNRECOGNIZED -> throw IllegalArgumentException("PpaData.PPAFederalState.UNRECOGNIZED")
```
mtwalli, line 161: Can we simply this somehow? I see that this enum is coming from the server. In this case I would just create another one for the UI state. We can have more control over it and the fields that are related. what do you think?
This might also simplify https://github.com/corona-warn-app/cwa-app-android/pull/2270/files#diff-8da4ec62b6a184a1706e14e77b4184ee344079fd3a174efdafcf23c4c1c15be0R76-R91

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/Districts.kt
```diff
@@ -0,0 +1,42 @@
...
+        @SerializedName("districtName") val districtName: String = "",
+        @SerializedName("districtShortName") val districtShortName: String = "",
+        @SerializedName("districtId") val districtId: Int = 0,
+        @SerializedName("federalStateName") val federalStateName: String = "",
+        @SerializedName("federalStateShortName") val federalStateShortName: String = "",
+        @SerializedName("federalStateId") val federalStateId: Int = 0
+    )
+
+    companion object {
+        private const val ASSETNAME = "ppdd-ppa-administrative-unit-set.json"
```
mtwalli, line 39: ASSET_NAME

mtwalli:
Tested on Samsung S8 ,functionality is really good

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputViewModel.kt
d4rken: Creating another enum or data class would just increase the overhead as we then need to start mapping between additional types. I think this is a very good use-case for extension functions. I have extracted them into an extra file so that  @AlexanderAlferov  can reuse them.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/Districts.kt
d4rken: Done, but this is a bit pedantic imho :wink:.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/ui/input/AnalyticsUserInputViewModel.kt
d4rken: I've refactored the code, but at this point would prefer not to introduce any additional abstractions. I've tried a `UserInfoRepo` approach but in the end deleted it, because there is not enough information there to scope it correctly.
I'd prefer to revisit this once the functionality has been integrated with the other screens and logic.

d4rken:
> https://github.com/corona-warn-app/cwa-app-android/pull/2270#issuecomment-772336760

I was hoping for a solution without an extra selector as the accentColor should be set app wide, and the radiobutton should use it by default. I've used your proposal for now though, maybe we can revisit this later once the theming/styles are more aligned in the app.

Thank you for the snippet.

chris-cwa (approved):

mtwalli (approved):
LGTM 👍🏼

PR #2272, opened 2021-02-02
Update klint plugins (DEV)

mtwalli:
Update klint plugins to enable training commas, Please not changes in here are from klint formatting


build.gradle
```diff
@@ -18,8 +18,8 @@ buildscript {
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "com.google.protobuf:protobuf-gradle-plugin:$protobufVersion"
         classpath "androidx.navigation:navigation-safe-args-gradle-plugin:$navVersion"
-        classpath "org.jlleitschuh.gradle:ktlint-gradle:9.2.1"
-        classpath "io.gitlab.arturbosch.detekt:detekt-gradle-plugin:1.9.1"
+        classpath "org.jlleitschuh.gradle:ktlint-gradle:9.4.1"
```
d4rken, line 21: Why does "Allowing trailing commas", require 127 files to be changed?

A lot of the reformatting is unnecessary or worse imho, and the potential for merge conflicts is huge.

d4rken (changes requested):
The PR should be about the linter allowing trailing commas. Just updating for the sake of updating is not desired if it fixes no issues and reformats 127 files. Linting rules change almost monthly....

build.gradle
d4rken: I appreciate the effort :heart:  but would like to reduce this PR to "just allowing trailing commas".

mtwalli: Well I looked for a way to enable only the training commas did not find a rule that was preventing it in our config file.
After some Googling I found out that it is only possible by upgrading the plugin

d4rken: I don't want to decide this. We should check with the rest of the team.

mtwalli: Fair enough

BMItter:
I'd be ok with this change.  also like the formatting

build.gradle
chris-cwa: i also fear a lot of merge conflicts. is this a reasonable price for trailing commas?

mtwalli:
Closing it, till we agree on a way of doing it

PR #2279, opened 2021-02-03
Submission process swapped dialogs (EXPOSUREAPP-4899)

SamuraiKek:
Fixes the dialog displayed when canceling the submission process after getting a positive result. 

Very small change to the condition that sets the dialogs.

Test steps can be found in the ticket


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/resultavailable/SubmissionTestResultAvailableFragment.kt
```diff
@@ -83,7 +83,7 @@ class SubmissionTestResultAvailableFragment : Fragment(R.layout.fragment_submiss
         var dialogTitle = R.string.submission_test_result_available_close_dialog_title_consent_not_given
         var dialogBody = R.string.submission_test_result_available_close_dialog_body_consent_not_given
 
-        if (binding.submissionTestResultAvailableConsentStatus.consent) {
+        if (!binding.submissionTestResultAvailableConsentStatus.consent) {
```
harambasicluka, line 86: I think your fix is valid, but if I read this:
1. consent not given strings are set
2. consent isn't given -> consent given strings are used

Are the strings maybe named wrong?

SamuraiKek: Yes, it looks like they are. The last changes pushed should take care of this.

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1274,15 +1274,15 @@
     <!-- YTXT: Text for consent given -->
     <string name="submission_test_result_available_text_consent_given">"Vielen Dank, dass Sie mithelfen, andere zu warnen und damit einverstanden sind, Ihr Testergebnis zu teilen. \n\n"<b>"Bitte geben Sie Ihr Testergebnis im nächsten Schritt frei, indem Sie „Teilen“ antippen."</b>""</string>
     <!-- XHED: Close screen popup title for consent given -->
-    <string name="submission_test_result_available_close_dialog_title_consent_given">"Wollen Sie wirklich abbrechen?"</string>
+    <string name="submission_test_result_available_close_dialog_title_consent_given">"Achtung:\nSie sind dabei den Vorgang abzubrechen.\nSie haben Ihre Mitmenschen noch nicht gewarnt. Bitte schützen Sie Ihre Mitmenschen und schließen Sie den Vorgang ab."</string>
```
chiljamgossow, line 1277: This seems wrong. The title and body are switched now. Title is too long to be visible.
device-2021-02-04-141213

chiljamgossow (changes requested):

Corona-Warn-App/src/main/res/values-de/strings.xml
SamuraiKek: This ticket only solves the switched dialogs problem. It's known that the texts need to be changed and a new PR with the right string will be added when we get them

chiljamgossow: I can see above and below in the strings file  that text has been switched. The popup before also shows that is was the other way around before.

SamuraiKek: That's true, but check Luka's comment https://github.com/corona-warn-app/cwa-app-android/pull/2279#discussion_r569453901. They were wrongly named since the beginning.

chiljamgossow: The content of submission_test_result_available_close_dialog_body_consent_given and submission_test_result_available_close_dialog_title_consent_given needs to be switched. They are not correct in the UI as shown above.

harambasicluka:
I think the body and title are interchanged.

image

Corona-Warn-App/src/main/res/values-de/strings.xml
harambasicluka: Yes @chiljamgossow is right, the naming of the keys where wrong like `given` vs `not_given`. But now the body and the title are still interchanged.

harambasicluka (changes requested):

SamuraiKek:
Hey folks, after looking further into it, I'm still convinced that the dialogs show up when expected and have the right titles and dialogs:

The consent is on by default and the user must press "Andere warnen" and put it to "Gestoppt" to revoke the consent.

Ticket 4899 - cancel submission with consent (default state):



`submission_test_result_available_close_dialog_title_consent_given` has text 'Achtung:\nSie sind dabei den Vorgang abzubrechen.\nSie haben Ihre Mitmenschen noch nicht gewarnt. Bitte schützen Sie Ihre Mitmenschen und schließen Sie den Vorgang ab.'

'submission_test_result_available_close_dialog_body_consent_given` has text 'Möchten Sie den Vorgang wirklich abbrechen?`

The dialog that shows up: 

4899-dialog


Ticket 4946 - cancel submission without consent:



`submission_test_result_available_close_dialog_title_consent_not_given` has text `Wollen Sie wirklich abbrechen?`

`submission_test_result_available_close_dialog_body_consent_not_given` has text `Ihre bisherigen Angaben werden nicht gespeichert.`

The dialog that shows up:

4946-dialog

chiljamgossow (approved):
LGTM

harambasicluka (approved):

janetback (approved):
UA approved

PR #2280, opened 2021-02-03
Submit Analytics Data (EXPOSUREAPP-4823)

kolyaopahle:
### Description
This PR contains the backend logic for Analytics submission
It should only be merged after all other Analytics PRs are merged as it will be used to combine front and backend interactions.

### Testing
There are quite a few options in the Data Donation tab of the Tester Menu to already test functionality, but there is currently no working testing environment providing the necessary backend.
You could use the cwa-app-cli project which has support for the backend or check logcat for errors


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/datadonation/analytics/storage/DefaultLastAnalyticsSubmissionLogger.kt
```diff
@@ -0,0 +1,87 @@
...
+import java.io.File
+import javax.inject.Inject
+
+class DefaultLastAnalyticsSubmissionLogger @Inject constructor(
+    @AppContext private val context: Context,
+    private val dispatcherProvider: DispatcherProvider,
+    @BaseGson private val baseGson: Gson
+) : LastAnalyticsSubmissionLogger {
+    private val analyticsDir = File(context.cacheDir, "analytics_storage")
+    private val analyticsFile = File(analyticsDir, "analytics.bin")
```
d4rken, line 31: Minor, `analytics.bin` seems very generic, isn't it `last_analytics.bin`?

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/datadonation/analytics/storage/DefaultLastAnalyticsSubmissionLogger.kt
```diff
@@ -0,0 +1,87 @@
...
+            try {
+                gson.toJson(dataObject, analyticsFile)
+            } catch (e: Exception) {
+                Timber.e(e, "Failed to store analytics data.")
+            }
+        }
+
+    override suspend fun getLastAnalyticsData(): LastAnalyticsSubmission? = withContext(dispatcherProvider.IO) {
+        try {
+            gson.fromJson(analyticsFile)
```
d4rken, line 60: Based on issues we had with the exposure detection which also stores json to file:
We should prevent empty files from causing invalid data classes being created by GSON (examples in `ExposureDetectionTrackerStorageTest`).

Despite not being allowed GSON can create data classes with null fields that shouldn't be null (unsafe init), which the JVM will only notice once accessed.

TL;DR:  `.also { requireNotNull(it.timestamp) }`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/AnalyticsConfigMapper.kt
```diff
@@ -4,20 +4,59 @@ import dagger.Reusable
...
 @Reusable
 class AnalyticsConfigMapper @Inject constructor() : AnalyticsConfig.Mapper {
     override fun map(rawConfig: AppConfigAndroid.ApplicationConfigurationAndroid): AnalyticsConfig {
-        // TODO
-        return AnalyticsConfigContainer()
+        if (!rawConfig.hasPrivacyPreservingAnalyticsParameters() &&
+            !rawConfig.privacyPreservingAnalyticsParameters.hasCommon() &&
+            !rawConfig.privacyPreservingAnalyticsParameters.hasPpac()
+        ) {
+            throw ApplicationConfigurationInvalidException(
```
d4rken, line 18: Is the expected behavior to throw here (and discard the whole config), or would it be okay to return a default of "analytics disabled", as analytics being disabled would still allow the rest of the app to function?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/AnalyticsConfigMapper.kt
```diff
@@ -4,20 +4,59 @@ import dagger.Reusable
...
+                    .probabilityToSubmitExposureWindows
+            )
+        }
+
     data class AnalyticsConfigContainer(
         override val safetyNetRequirements: SafetyNetRequirements = SafetyNetRequirementsContainer(),
-        override val probabilityToSubmit: Float = 1f,
-        override val probabilityToSubmitAfterRiskCalculation: Float = 1f,
-        override val probabilityToSubmitNewExposureWindows: Float = 1f
+        override val probabilityToSubmit: Double = 1.0,
```
d4rken, line 57: Should the defaults be 1.0, or would it be better if the defaults were 0.0, i.e. disabled? Does the spec say?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -1,34 +1,156 @@
...
+    private val logger: LastAnalyticsSubmissionLogger
 ) {
+    private suspend fun trySubmission(ppaData: PpaData.PPADataAndroid): Boolean {
+        try {
+            val ppaAttestationRequest = PPADeviceAttestationRequest(
+                ppaData = ppaData
+            )
 
-    val isEnabled: Boolean = true
+            val attestation = deviceAttestation.attest(ppaAttestationRequest)
```
d4rken, line 36: This only means that SafetyNet passed "generally". The analytics config contains specific requirements that you need to check, i.e.

`attestation.requirePass(analyticsConfig.requirements)`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -1,34 +1,156 @@
...
 
+            val ppaContainer = PpaDataRequestAndroid.PPADataRequestAndroid.newBuilder()
+                .setAuthentication(attestation.accessControlProtoBuf)
+                .setPayload(ppaData)
+                .build()
+
+            dataDonationAnalyticsServer.uploadAnalyticsData(ppaContainer)
```
d4rken, line 43: I'd :heart: a log statement before attestation, after attestation pass and after upload.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -1,34 +1,156 @@
...
+        val success = trySubmission(analyticsProto)
+
+        contributions.forEach {
+            Timber.d("Finishing contribution: %s", it::class.simpleName)
+            it.finishDonation(success)
+        }
+
+        if (success) {
+            settings.lastSubmittedTimestamp.update {
+                Instant.now()
```
d4rken, line 87: `Timestamper` would like to be part of the analytics gang.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -1,34 +1,156 @@
...
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun stopDueToNoUserConsent(): Boolean {
+        return !settings.analyticsEnabled.value
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    suspend fun stopDueToProbabilityToSubmit(): Boolean {
+        val submitRoll = Random.nextDouble(0.0, 1.0)
+        return submitRoll > appConfigProvider.getAppConfig().analytics.probabilityToSubmit
```
d4rken, line 104: `>=`?

What if we have chance 1.0 and roll a 1.0?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -1,34 +1,156 @@
...
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun stopDueToLastSubmittedTimestamp(): Boolean {
+        val lastSubmit = settings.lastSubmittedTimestamp.value ?: return false
+        return lastSubmit.plus(Hours.hours(LAST_SUBMISSION_MIN_AGE_HOURS).toStandardDuration()).isAfterNow
```
d4rken, line 110: `isAfterNow` internally calls `Instant.now` right? `TimeStamper` so we can fake this in tests?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -1,34 +1,156 @@
...
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    fun stopDueToTimeSinceOnboarding(): Boolean {
+        val onboarding = LocalData.onboardingCompletedTimestamp()?.let { Instant.ofEpochMilli(it) } ?: return true
+        return onboarding.plus(Hours.hours(ONBOARDING_DELAY_HOURS).toStandardDuration()).isAfterNow
+    }
+
+    suspend fun submitIfWanted() {
```
d4rken, line 119: I guess this is the entry point to the whole thing, right?
Can we sleep better if we guard this with a `Mutex` or are we absolutely sure there will be no overlaps? If yes why are we so sure?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsModule.kt
```diff
@@ -5,26 +5,61 @@ import dagger.Provides
...
+import javax.inject.Singleton
 
 @Module
 class AnalyticsModule {
 
-    @IntoSet
+    @Singleton
     @Provides
-    fun newExposureWindows(module: NewExposureWindowsDonor): DonorModule = module
+    fun provideAnalyticsSubmissionApi(
```
d4rken, line 25: Please align with @chris-cwa regarding the http client for the datadonation server, there are some differences in your PRs.

https://github.com/corona-warn-app/cwa-app-android/pull/2311#discussion_r573655003

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsModule.kt
```diff
@@ -5,26 +5,61 @@ import dagger.Provides
...
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+import retrofit2.converter.protobuf.ProtoConverterFactory
+import javax.inject.Singleton
 
 @Module
 class AnalyticsModule {
 
-    @IntoSet
+    @Singleton
```
d4rken, line 23: `Singleton` or `Reusable`, or neither?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/server/DataDonationAnalyticsServer.kt
```diff
@@ -0,0 +1,30 @@
...
+
+@Singleton
+class DataDonationAnalyticsServer @Inject constructor(
+    private val api: Lazy<DataDonationAnalyticsApiV1>
+) {
+
+    suspend fun uploadAnalyticsData(ppaDataRequestAndroid: PpaDataRequestAndroid.PPADataRequestAndroid) {
+        val response = api.get().submitAndroidAnalytics(ppaDataRequestAndroid)
+
+        if (response.code() == 204) {
```
d4rken, line 17: Could I have some logging for each response type/code :relaxed: ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsOneTimeWorker.kt
```diff
@@ -0,0 +1,51 @@
...
+import de.rki.coronawarnapp.util.worker.InjectedWorkerFactory
+import de.rki.coronawarnapp.worker.BackgroundConstants
+import timber.log.Timber
+
+/**
+ * One time background Analytics worker
+ *
+ * @see DataDonationAnalyticsScheduler
+ */
+class DataDonationAnalyticsOneTimeWorker @AssistedInject constructor(
```
d4rken, line 19: Do we need a onetime worker? Can win with just the periodic worker?

d4rken: Why do they both look the same?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsScheduler.kt
```diff
@@ -0,0 +1,59 @@
...
+ */
+@Reusable
+class DataDonationAnalyticsScheduler @Inject constructor(
+    private val workManager: WorkManager,
+    private val workBuilder: DataDonationAnalyticsWorkBuilder,
+    private val timeCalculation: DataDonationAnalyticsTimeCalculation
+) {
+
+    /**
+     * Enqueue background deadman notification periodic work
```
d4rken, line 23: deadman?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsScheduler.kt
```diff
@@ -0,0 +1,59 @@
+package de.rki.coronawarnapp.datadonation.analytics.worker
+
+import androidx.work.ExistingPeriodicWorkPolicy
+import androidx.work.ExistingWorkPolicy
+import androidx.work.WorkManager
+import dagger.Reusable
+import javax.inject.Inject
+
+/**
+ * We schedule two workers to allow for randomness in the execution times,
```
d4rken, line 10: We just need randomness across the userbase (to be nice to the server), not check on the users device, couldn't we achieve this by setting a random initial delay for the periodic worker?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingAnalyticsViewModel.kt
```diff
@@ -32,15 +33,24 @@ class OnboardingAnalyticsViewModel @AssistedInject constructor(
...
 
     fun onDisableClick() {
-        settings.userInfoAgeGroup.update { PpaData.PPAAgeGroup.AGE_GROUP_UNSPECIFIED }
-        settings.userInfoFederalState.update { PpaData.PPAFederalState.FEDERAL_STATE_UNSPECIFIED }
-        settings.userInfoDistrict.update { 0 }
+        settings.analyticsEnabled.update {
+            false
+        }
+
+        launch {
```
d4rken, line 48: If the user smashes that like button, err disable button.... multiple presses can launch this multiple times, right? So donor modules would have to be thread safe on `deleteData`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    @AfterEach
+    fun tearDown() {
+        clearAllMocks()
+    }
+
+    private fun createDonorModules(): Set<DonorModule> = setOf(exposureRiskMetadataDonor)
+
+    private fun createInstance() = spyk(
```
d4rken, line 64: Why do we need  `spy` here?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
```diff
@@ -0,0 +1,191 @@
...
+        every { settings.analyticsEnabled } returns mockFlowPreference(false)
+        every { analyticsConfig.probabilityToSubmit } returns 0.0
+
+        val analytics = createInstance()
+
+        runBlockingTest2 {
+            analytics.submitIfWanted()
+        }
+
+        // TODO: is there a way to make sure a spy function returned x?
```
d4rken, line 101: You could inject the randomness source and hardcode it to a specific value.

We currently have `SecureRandom` in the DI graph for SafetyNet.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
```diff
@@ -0,0 +1,191 @@
...
+            analytics.submitAnalyticsData()
+        }
+    }
+
+    @Test
+    fun `abort due to time since onboarding`() {
+        every { settings.analyticsEnabled } returns mockFlowPreference(false)
+        every { analyticsConfig.probabilityToSubmit } returns 1.0
+        every { settings.lastSubmittedTimestamp } returns mockFlowPreference(
+            Instant.now().minus(Days.TWO.toStandardDuration())
```
d4rken, line 133: Dynamic time in tests is :ghost:

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
```diff
@@ -0,0 +1,191 @@
...
+
+        val analytics = createInstance()
+
+        runBlockingTest2 {
+            analytics.submitIfWanted()
+        }
+
+        coVerify(exactly = 1) {
+            analytics.submitAnalyticsData()
+            dataDonationAnalyticsServer.uploadAnalyticsData(any())
```
d4rken, line 188: Do we have access to the data that would be uploaded such that we can be specific than `any()`?

e.g.

Can we construct `PpaDataRequestAndroid` with the same value? Or would the ProtoBuf equals/hash still not be equal and allow this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/server/DataDonationAnalyticsApiV1.kt
```diff
@@ -0,0 +1,13 @@
+package de.rki.coronawarnapp.datadonation.analytics.server
+
+import de.rki.coronawarnapp.server.protocols.internal.ppdd.PpaDataRequestAndroid
+import retrofit2.Response
+import retrofit2.http.Body
+import retrofit2.http.POST
+
+interface DataDonationAnalyticsApiV1 {
+    @POST("/version/v1/android/dat")
+    suspend fun submitAndroidAnalytics(
```
d4rken, line 10: Can I have a test for this stuff too? :cat2: 

Should be straight forward, check the other API tests: Making sure serialization does what it should, bad data leads to an error, and the API path is what we expect.

d4rken (commented):
I like what you got

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
kolyaopahle: we can't roll a 1.0 afaik `nextDouble` is (inclusive, exclusive)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/AnalyticsConfigMapper.kt
kolyaopahle: i talked to @mlenkeit about this, he mentioned that the analytics config part of the app config should not be optional so we can assume a broken config if no analytics config is provided

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/AnalyticsConfigMapper.kt
kolyaopahle: not really, i'll remove the defaults for now as there is no way to instantiate the defaults anyways at the moment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
kolyaopahle: i am relatively sure because of the delay implied in the scheduled background work, but i will add a mutex to be extra safe :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsOneTimeWorker.kt
kolyaopahle: currently the randomness was achieved by scheduling one time work every 24 hours which would then schedule periodic work at a random time in the next 23 hours, but i like the idea of just randomly distributing the initial start delay of the periodic work, will remove the one time worker

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsModule.kt
BMItter: I'd prefer @Reuseable

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_datadonation.xml
```diff
@@ -115,6 +115,98 @@
...
+                    android:layout_width="match_parent"
+                    android:layout_height="wrap_content"
+                    android:layout_marginTop="@dimen/spacing_tiny"
+                    android:textIsSelectable="true"
+                    app:layout_constraintEnd_toEndOf="parent"
+                    app:layout_constraintStart_toStartOf="parent"
+                    app:layout_constraintTop_toBottomOf="@+id/analytics_last_submit_title"
+                    tools:text="Body" />
+
+            </androidx.constraintlayout.widget.ConstraintLayout>
```
BMItter, line 147: please check for layout Stacking (constraint->constraint->...)
I'm not sure here

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/datadonation/analytics/storage/DefaultLastAnalyticsSubmissionLogger.kt
```diff
@@ -0,0 +1,87 @@
...
+            .create()
+    }
+
+    override suspend fun storeAnalyticsData(analyticsProto: PpaData.PPADataAndroid) =
+        withContext(dispatcherProvider.IO) {
+            if (!analyticsDir.exists()) {
+                analyticsDir.mkdirs()
+            }
+
+            val dataObject = LastAnalyticsSubmission(
```
BMItter, line 46: Inject and use TimeStamper 👵

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingAnalyticsViewModel.kt
kolyaopahle: hmm true, i am also not 100% happy with the positioning here, as we would need to duplicate the code in the AnalyticsSettings view model later on, maybe also some kind of mutex lockout and we move this to a more generic place where it listens to the analyticsEnabled flow preference?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
kolyaopahle: i was trying to use a spy to check for the return value of the internal check functions in Analytics i.e. `stopDueToProbabilityToSubmit` to not only check that submission was aborted as it is done now, but make sure that it is aborted for the right reason, maybe throwing a more specific exception in the stop functions and raising the exception handling one level further up in the analytics submission could also help with testability

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
kolyaopahle: this comment was more in reference to https://github.com/corona-warn-app/cwa-app-android/pull/2280#discussion_r573931976

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingAnalyticsViewModel.kt
```diff
@@ -32,15 +33,24 @@ class OnboardingAnalyticsViewModel @AssistedInject constructor(
     }.asLiveData(dispatcherProvider.IO)
 
     fun onNextButtonClick() {
+        settings.analyticsEnabled.update {
+            true
+        }
+
         completedOnboardingEvent.postValue(Unit)
-        // TODO: ♫ ♪ Hello Kolya, my old friend ♫ ♪
```
BMItter, line 36: 👌 😄

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/server/DataDonationAnalyticsServer.kt
```diff
@@ -0,0 +1,30 @@
...
+import de.rki.coronawarnapp.server.protocols.internal.ppdd.PpaDataRequestAndroid
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class DataDonationAnalyticsServer @Inject constructor(
+    private val api: Lazy<DataDonationAnalyticsApiV1>
+) {
+
+    suspend fun uploadAnalyticsData(ppaDataRequestAndroid: PpaDataRequestAndroid.PPADataRequestAndroid) {
```
BMItter, line 14: consider to use 'when' instead of various if statements if or when you want (optional)

Corona-Warn-App/src/deviceForTesters/res/layout/fragment_test_datadonation.xml
kolyaopahle: i'm too stupid to do it without as the stacked layout also contains a card border

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/server/DataDonationAnalyticsServer.kt
kolyaopahle: done :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingAnalyticsViewModel.kt
d4rken: Note that FlowPreference is only safe when accessed from the same thread. It's not safe for concurrent access from multiple threads.

I think it makes sense if all deletions, just like starting the analytics submission goes the through `Analytics` main class, and we add a single `Mutex' that guards the deletion and the submission call (as those should not overlap either).

What do you think?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
d4rken: Maybe set up the `setup()` that the default state is equal to a successful submission, then in each test case (e.g. probability) change a single parameter and check that the process fails appropriately, maybe by checking that no submission happened.

Or if `stopDueToProbabilityToSubmit` causes an abort, then check that `stopDueToNoUserConsent` was called but `stopDueToLastSubmittedTimestamp`, maybe by verifying calls on the mocks used by those functions.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingAnalyticsViewModel.kt
kolyaopahle: sounds good, i've wrapped changes to the property in calls on `Analytics` which should allow us to execute delete on every settings change

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsTest.kt
kolyaopahle: I like it :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/AnalyticsModule.kt
kolyaopahle: i agree :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/AnalyticsConfigMapper.kt
```diff
@@ -5,19 +5,66 @@ import de.rki.coronawarnapp.appconfig.AnalyticsConfig
 import de.rki.coronawarnapp.appconfig.SafetyNetRequirements
 import de.rki.coronawarnapp.appconfig.SafetyNetRequirementsContainer
 import de.rki.coronawarnapp.server.protocols.internal.v2.AppConfigAndroid
+import timber.log.Timber
 import javax.inject.Inject
 
 @Reusable
 class AnalyticsConfigMapper @Inject constructor() : AnalyticsConfig.Mapper {
     override fun map(rawConfig: AppConfigAndroid.ApplicationConfigurationAndroid): AnalyticsConfig {
```
BMItter, line 13: using return if... below would be nice

BMItr (approved):
one optional comment.
lgtm overall 👌

harambasicluka:
Did on device testing (Pixel 3), logs looked good.

d4rken (approved):
lgtm

PR #2281, opened 2021-02-03
Accessibility: Contact Journal Edit Screens not announced (EXPOSUREAPP-4456)

axelherbstreith:
This PR addresses following issues:
- Announcement of the edit screens (both places and person) on Samsung devices.
- Announcement of the Contact Journal Onboarding Screen on Samsung devices.

To test this changes: Activate TalkBack on a **Samsung Device**, go to the Contact Journal Tab and open the edit screens. Titles should now be announced.


chiljamgossow:
I tried on a Xiaomi as do not have a Samsung device. There it is announced twice, once from the constraint layout and once from the toolbar. I suppose it is only announced once on Samsung.

axelherbstreith:
@chiljamgossow yes, on my Samsung device it is announced only once. But I just tested the changes on my P30 and it is also announced twice... I will test this again, since the double announcement also occurs in other parts of the app which wasn't affected by my changes

chiljamgossow:
@axelherbstreith How about removing the part that is not announced on Samsung?

axelherbstreith:
@chiljamgossow Already tried that, but somehow then the screen is not announced on Samsung again (at least on my phone), really weird... there must be an easier way, need to invest some more time there :)

chiljamgossow (approved):

LukasLechnerDev (approved):
Tested On Samsung S20+ and Google Pixel 3A, both have Android 11. 
On these phones, it was already correctly announced without this fix. 
With this fix, it is now announced twice ...

However, since this probably is a general issue and not correlated to the contact diary edit screen, I also think it makes sense to open another Ticket for it.

PR #2296, opened 2021-02-05
Appconfig survey handling (EXPOSUREAPP-4847)

BMItr:
EXPOSUREAPP-4847
This PR handles the Survey by app-config (e.g. Enable/Disable)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SurveyConfigMapper.kt
```diff
@@ -5,17 +5,46 @@ import de.rki.coronawarnapp.appconfig.SafetyNetRequirements
...
+                    surveyConfigContainer = surveyConfigContainer.copy(
+                        safetyNetRequirements = SafetyNetRequirementsContainer(
+                            requireBasicIntegrity = it.requireBasicIntegrity,
+                            requireCTSProfileMatch = it.requireCTSProfileMatch,
+                            requireEvaluationTypeBasic = it.requireEvaluationTypeBasic,
+                            requireEvaluationTypeHardwareBacked = it.requireEvaluationTypeHardwareBacked
+                        )
+                    )
+                }
+            }
```
d4rken, line 38: Is it possible that we have only one of the parameters ( `common` / `ppac`)  and not the other?
Should we throw an error or use a fallback if both are not available?

d4rken: If e.g. `ppac` is optional what should the default values be for survey related ppac?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SurveyConfigMapper.kt
```diff
@@ -5,17 +5,46 @@ import de.rki.coronawarnapp.appconfig.SafetyNetRequirements
...
-        return SurveyConfigContainer()
+        var surveyConfigContainer = SurveyConfigContainer()
+
+        if (rawConfig.hasEventDrivenUserSurveyParameters()) {
+            if (rawConfig.eventDrivenUserSurveyParameters.hasCommon()) {
+                rawConfig.eventDrivenUserSurveyParameters.common.also {
+                    surveyConfigContainer = SurveyConfigContainer(
+                        otpQueryParameterName = it.otpQueryParameterName,
+                        surveyOnHighRiskEnabled = it.surveyOnHighRiskEnabled,
+                        surveyOnHighRiskUrl = it.surveyOnHighRiskUrl
```
d4rken, line 22: Should we validate these? e.g. "isNotEmpty()"?

LukasLechnerDev: And should we also validate if the url-is well-formed here? Currently, I am doing this in the `SurveyUrlProvider` component of my subtask, but if we validate it here and I can be sure to have a valid URL, I don't need to do that anymore. 

What about exposing `surveyOnHighRiskUrl` as a `okhttp3.HttpUrl` instead of `String` in `SurveyConfigContainer`?

BMItter: @d4rken I already use the default values as fallback (see skelton)
@LukasLechnerDev will adjust to validate url here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SurveyConfigMapper.kt
BMItter: would it make sense to have an own mapper for PPAC too? because it is also needed at analytics

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SurveyConfigMapper.kt
```diff
@@ -4,18 +4,69 @@ import dagger.Reusable
...
+    } catch (e: Exception) {
+        throw ApplicationConfigurationInvalidException(cause = e, message = "Survey on high risk url is invalid")
     }
 
     data class SurveyConfigContainer(
-        override val safetyNetRequirements: SafetyNetRequirements = SafetyNetRequirementsContainer(),
-        override val enableSurveyLinkOnHighRiskCard: Boolean = true
+        override val otpQueryParameterName: String = "",
+        override val surveyOnHighRiskEnabled: Boolean = false,
+        override val surveyOnHighRiskUrl: String = "",
```
LukasLechnerDev, line 69: Why not use `HttpUrl` here instead of `String`? We are already doing a conversion in the URL validation above...

BMItter: interfaces should be kept as generic as possible and you have to rebuild the url anyay to append the OTP as param.
But anyway, you shall get what you requested 😄 
adjusted.

LukasLechnerDev (approved):
LGTM

d4rken (approved):
lgtm!

PR #2297, opened 2021-02-05
Fix exposure logging progress circle inconsistencies. (EXPOSUREAPP-1835)

SamuraiKek:
This PR makes the circle progress in line with the iOS one. Adding the x/14 days inside the circle and changing the progress color depending on the state of the exposure logging.

How to test:
- Go to exposure logging and look at the number of days inside the progress circle. It should look like x/14 days now.
- Try playing with the exposure logging states. Stop it, turn it back on again. Make sure that the circle is blue while logging is on and grey while it's off.
- I've removed two layers of includes in the fragment as they were only being used in this screen. Make sure the layout hasn't changed. Try disabling bluetooth and make sure the right cards appear on the screen.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CircleProgress.kt
```diff
@@ -69,6 +69,19 @@ class CircleProgress @JvmOverloads constructor(
             invalidate()
         }
 
+    /**
+     * Setter for the progress circle color.
+     * The progress bar also needs to be repainted when the
+     * color changes (ex: when the risk calculation gets turned on/off)
+      */
+    var progressColor: Int = 0
```
mtwalli, line 77: Please use `Color.TRANSPARENT` instead of 0

mtwalli:
### LGTM, tested on S8

device-2021-02-05-154326
device-2021-02-05-154649

mtwalli (dismissed):
👍🏼

ralfgehrer:
Screenshot 2021-02-08 at 13 35 51
Screenshot 2021-02-08 at 13 36 18


The progress indicator on the risk details is not shown - the blue bar is missing.

SamuraiKek:
> The progress indicator on the risk details is not shown - the blue bar is missing.

Forgot about that. I'll add them as well.

mtwalli:
@SamuraiKek  after your changes:
Low risk details screen does not show progress in case bluetooth is disabled, while it shows 2/14
device-2021-02-08-174056

SamuraiKek:
> @SamuraiKek after your changes:
> Low risk details screen does not show progress in case bluetooth is disabled, while it shows 2/14

Yes, I've intentionally done this because when bluetooth is turned off, the risk calculation becomes restricted. There isn't really a design guidline for this case, maybe we should ask somebody about it? 
iOS doesn't even have the progress circle in the low risk details screen ¯\_(ツ)_/¯

ralfgehrer:
> > @SamuraiKek after your changes:
> > Low risk details screen does not show progress in case bluetooth is disabled, while it shows 2/14
> 
> Yes, I've intentionally done this because when bluetooth is turned off, the risk calculation becomes restricted. There isn't really a design guidline for this case, maybe we should ask somebody about it? I can quickly change it to the blue tint.
> iOS doesn't even have the progress circle in the low risk details screen ¯_(ツ)_/¯
> 
> @ralfgehrer, @harambasicluka, what do you think?

IMHO, the progress should be shown. Otherwise, the circle does not make much sense, does it?

SamuraiKek:
> IMHO, the progress should be shown. Otherwise, the circle does not make much sense, does it?

Alright, the same behaviour was happening when the location is disabled so I'll take care of that too.

ralfgehrer:
Screenshot 2021-02-09 at 11 05 42


When exposure logging is turned off, the progress indicator shall be dark grey - according to the ticket. In the current implementation, it is gone.

mtwalli (approved):
LGTM , tested latest changes on Samsung S8 while location and bluetooth on/off.

ralfgehrer (approved):

PR #2298, opened 2021-02-05
Clean up environment types (DEV)

d4rken:
`INT-FED` no longer exists, and we can instead name it `LOCAL` to conveniently setup local environment addresses.


ralfgehrer (approved):
Thanks! Tested all environments in the debug menu.

BMItr (approved):

PR #2300, opened 2021-02-05
Implement Donor for User Metadata (EXPOSUREAPP-4823)

kolyaopahle:
### Description
This PR adds an analytics donor for the user metadata collected from the screens implemented by @d4rken
it also changes the Protobuf instances in the DonorModule interface to their respective builder which is also part of the 
Submit Analytics PR #2280


d4rken (approved):
lgtm :+1:

LukasLechnerDev (approved):
LGTM

PR #2303, opened 2021-02-08
Remove redundant district prefixes (EXPOSUREAPP-4815)

mlenkeit:
This PR updates the JSON file with the districts with new names where redundant prefixes have been removed and relevant prefixes move to the end of the district. (Discussed with UA)


chiljamgossow (approved):
LGTM

mlenkeit:
Set it back to _draft_, one more change needs to be done...

d4rken (approved):

Issue #2013, opened 2020-12-30
Wrong font in risk explanation and transnational screen


Bananeweizen:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: One Plus One
- Android version: 10
- App version: 1.10.1

## Describe the bug

The app uses the same font everywhere, but in 2 places it uses much more letter spacing, making the text harder to read.

## Steps to reproduce the issue

Open the transnational screen or the risk explanation screen and compare fonts in different areas. If you don't see a difference immediately, then please compare identical words like the highlighted "Ermittlung". In the wide letter-spacing variant, the space between the characters is much wider than the character "lowercase L", while in the normal font the space is roughly as wide as "lowercase L".

Screenshot 1
![Screenshot_20201230-191359_Corona-Warn](https://user-images.githubusercontent.com/406876/103373477-5d601700-4ad5-11eb-965d-4dae881804de.png)

Screenshot 2
![Screenshot_20201230-191330_Corona-Warn](https://user-images.githubusercontent.com/406876/103373355-fb071680-4ad4-11eb-8487-f878dcf7fd53.png)

## Expected behaviour

Same font used everywhere.

---
Internal Tracking ID: [EXPOSUREAPP-4489](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4489)
PR #2304

dsarkar:
Hi @Bananeweizen, 

Thanks for reporting.  We will have a look at this.

  
Best wishes,
DS

---
Corona-Warn-App Open Source Team

dsarkar:
- Reproduced on Samsung Galaxy A21s / Android 10 / CWA 1.10.0
- NOT reproduced on iPhone6s / iOS 14.3 / CWA 1.10.1

---
Corona-Warn-App Open Source Team

dsarkar:
Dear @Bananeweizen,

Thanks again. We have created a ticket for the bug in the internal Jira system. Any further developments will be notified here in this Github issue.
  
Best wishes,
DS

---
Corona-Warn-App Open Source Team
[...]

PR #2304, opened 2021-02-08
Change text style (EXPOSUREAPP-4489)

chiljamgossow:
change text style from body1 to subtitle as requested in ticket
some refactoring of the layout to better be able to identify the screen and text elements in preview


Corona-Warn-App/src/main/res/layout/include_settings_switch_row.xml
```diff
@@ -74,11 +74,11 @@
 
             </androidx.constraintlayout.widget.ConstraintLayout>
 
-            <Switch
+            <androidx.appcompat.widget.SwitchCompat
```
mtwalli, line 77: could you please use `SwitchMaterial` instead?

chiljamgossow: Does it make sense if the style is set to Widget.AppCompat.CompoundButton.Switch? I replaced it because AS was complaining.

mtwalli: `SwitchMaterial` is actually a `SwitchCompat`, it will still work with `CompatTheme`. When we move to MD completely ,we will have to change the theme only

Corona-Warn-App/src/main/res/layout/fragment_interoperability_configuration.xml
```diff
@@ -23,35 +23,24 @@
             app:layout_constraintStart_toStartOf="parent"
             app:layout_constraintTop_toTopOf="parent" />
 
-        <androidx.constraintlayout.widget.ConstraintLayout
-            android:id="@+id/interoperability_configuration_country_list_container"
+        <include
```
d4rken, line 26: The TextView now touches the bottom screen end.
I think you might have forgot to readd `android:layout_marginBottom` when refactoring?

d4rken (commented):
See open comments, but otherwise :+1:

Corona-Warn-App/src/main/res/layout/include_settings_switch_row.xml
chiljamgossow: I replaced it as requested

Corona-Warn-App/src/main/res/layout/fragment_interoperability_configuration.xml
chiljamgossow: Actually, I did not forget but as it was reused adding a padding to the end of the layout messed up the onboarding. I refactored to solve this issue. The preview looks almost like the actual screen now.

chiljamgossow:
device-2021-02-09-163619
device-2021-02-09-163556

mtwalli (approved):

d4rken (approved):

MikeMcC399:
Link back to original issue https://github.com/corona-warn-app/cwa-app-android/issues/2013.

PR #2306, opened 2021-02-08
Show normal behavior box when tracing disabled (EXPOSUREAPP-4768)

chiljamgossow:
Always show normal behavior box when tracing disabled independent of prior risk status
Use toolbar


d4rken (approved):
lgtm

ralfgehrer (approved):
LGTM. Back navigation works, too.

PR #2307, opened 2021-02-08
Provide survey URL (EXPOSUREAPP-4858)

LukasLechnerDev:
The component that provides the URL for Surveys. 

For our current survey (high-risk cards), the base URL is loaded from our `AppConfig` and then the one-time-password is added as a query parameter. The name of this query parameter is also contained in the `AppConfig`


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveyUrlProvider.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val otpRepo: OTPRepository
+) {
+
+    /**
+     * Provides an Url for the Data Donation Surveys.
+     *
+     * @throws IllegalStateException If no one time password (otp) can be loaded from the OTPRepository of if the
+     * AppConfig doesn't contain a link to a survey
+     */
+    suspend fun provideUrl(type: Surveys.Type): String {
```
BMItter, line 20: please consider to put this logic/method to _Surveys.kt_
we could then validate if the OTP has been used already this month.
If no OTP is there, we can just generate one...

LukasLechnerDev: My plan was that after the OTP generation/validation/authentication (Part of Chris' subtask) ,`Surveys.kt` can simply call `provideUrl()`. At this point, a generated, validated and authenticated OTP should be stored in our `OtpRepository` that is then loaded by `SurveyUrlProvider` to assemble the URL.

I pushed a new commit where the `UrlProvider` is called in `Surveys.kt`. 

Do you think the `UrlProvider` shouldn't be a separate component but a method in `Surveys.kt`? What would be the benefit?

BMItter: The SurveyUrlProvider could be irritating if one calls it from another class than Surveys.
For e.g. one tries to get an URL for a High_Risk_Survey. He could use the url provider instead of surveys and so accidentally skip the generation/validation/authentication of the OTP.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/survey/SurveyHighRiskUrlProviderTest.kt
```diff
@@ -0,0 +1,71 @@
...
+        every { otpRepo.lastOTP } returns OneTimePassword(uuid)
+        coEvery { appConfigProvider.getAppConfig().survey.surveyOnHighRiskUrl } returns null
+
+        shouldThrow<IllegalStateException> { createInstance().provideUrl(Surveys.Type.HIGH_RISK_ENCOUNTER) }.also {
+            it.message shouldBe "AppConfig doesn't contain a link to the high-risk card survey"
+        }
+    }
+
+    @BeforeEach
+    fun setUp() {
```
d4rken, line 56: Test "architecture" should be at the top of a test imho, as you read top->down so it would make sense to first learn about  mocks etc.
But not a blocker, just subjective preference.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveyUrlProvider.kt
d4rken: Is the OTP (re)generated on-demand?

What is the calling classes expected to do when `provideUrl` throws `IllegalStateException`?

LukasLechnerDev: @BMItter 
I added a new commit. Now you need to pass the `OTP` to the `UrlProvider` and since only the Surveys class is going to generate `OTP`s, it shouldn't be called from anywhere else.

@d4rken 
I changed the UrlProvider so that we now already have to pass an authenticated `OTP` to it. 
I guess the calling class should show a "Something went wrong...." error message to the user in this case.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveyUrlProvider.kt
```diff
@@ -0,0 +1,42 @@
+package de.rki.coronawarnapp.datadonation.survey
+
+import de.rki.coronawarnapp.appconfig.AppConfigProvider
+import de.rki.coronawarnapp.datadonation.storage.OTPRepository
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
```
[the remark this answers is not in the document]
LukasLechnerDev: Added `@Reusable`

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/survey/SurveyHighRiskUrlProviderTest.kt
LukasLechnerDev: I see code in `setUp` and `tearDown` as mostly boilerplate, which is often not necessary to read in order to understand the tests, so I often move it to the bottom, so one can dig into it if required. However, since your suggestion seems to be the "standard" in this Project, I changed it.

jurajkusnier (approved):

BMItr (approved):
lgtm 👌

PR #2308, opened 2021-02-08
Fix SafetyNet nonce comparison (DEV)

d4rken:
* Fix SafetyNet nonce comparison. Apparently SafetyNet base64 encodes the nonce in the returned JSON. 
* Add additional test menu options to test requirement validation.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
@@ -95,7 +95,7 @@ class SafetyNetClientWrapper @Inject constructor(
         val body: JsonObject,
         val signature: ByteArray
     ) {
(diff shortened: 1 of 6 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/safetynet/SafetyNetClientWrapper.kt
```diff
@@ -95,15 +95,17 @@ class SafetyNetClientWrapper @Inject constructor(
         val body: JsonObject,
         val signature: ByteArray
     ) {
-        val nonce: String? = body.get("nonce")?.asString
+        val nonce: String? = body.get("nonce")?.asString?.decodeBase64()?.utf8()
```
d4rken, line 98: This was the culprit.

kolyaopahle (approved):
Nice catch 👌🏻

LukasLechnerDev (approved):
LGTM!

Great that we can now check the validation in the Test Menu 👍

PR #2314, opened 2021-02-10
Ignore latest risk calculation status when it is turned off (EXPOSUREAPP-4414)

chiljamgossow:
Whenever risk calculation status is turned off, the risk details screen should show the same items independent of prior risk status.

Testing: trigger high risk card and low risk card with exposures via test menu, then turn off exposure logging. In all cases, the screen should show the same items as seen here https://www.figma.com/file/c35ugXL5qZufiVFBLRITyG/COVID19_Master_1.13?node-id=770%3A14368


BMItter:
tested on pixel 2 device. lgtm 👌 
thx for minding the survey card

BMItr (approved):

d4rken (approved):

PR #2316, opened 2021-02-10
Adjust device time test setting (DEV)

d4rken:
Instead of faking the "disable-device-time-check" which won't work for SafetyNet,
we now fake the time offset to be within the acceptable delta.

### Testing
* Check that this can't become active in production, and that the tests cover this.
* Check that you can now time travel and do and can perform a "casual" verification via the data donation test menu


ralfgehrer:
When activating fake time and traveling one day into the future, I receive the following exception on  the "Casual" SafetyNet Report
`SafetyNetException: TIME_SINCE_ONBOARDING_UNVERIFIED: Time since first reliable timestamp <24h`

ralfgehrer (approved):
Clarified with @d4rken. 
- Launch app 
- check test menu --> Remote config - isDeviceTimeCorrec should be true
- enable fake time
- travel > 24h
- check SafetyNet Report "Casually" --> Requirements passed

LukasLechnerDev (approved):
- Device has correct time
- Reset App
- Do Onboarding
- TestMenu: Create SafetyNet Report
=> casual verification fails with "Time since first reliable timestamp < 24h"

=====

- Enable "Fake correct device time"
- time travel 24h into the future
- TestMenu: Create SafetyNet Report
=> casual verification: "Requirements passed"

=====

- time travel only 23h in the future
- TestMenu: Create SafetyNet Report
=> casual verification fails with "Time since first reliable timestamp < 24h"


**=> Works as expected.** 

- Tests also look good and verify that the time can't be faked in production builds. 
- One Suggestion would be to describe the new switch a bit better in the Developer Menu. e.g. by adding "This allows time travelling while still getting positive device time checks" or something like that

PR #2317, opened 2021-02-10
Fix coroutine scope canceling log spam (DEV)

d4rken:
:point_up:


ralfgehrer (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
```diff
@@ -80,8 +81,11 @@ class HotDataFlow<T : Any>(
             }
         }
         .onCompletion { err ->
-            if (err != null) Timber.tag(tag).w(err, "internal onCompletion due to error")
-            else Timber.tag(tag).v("internal onCompletion")
+            when {
+                err is CancellationException -> Timber.tag(tag).d("internal onCompletion due to cancelation")
+                err != null -> Timber.tag(tag).w(err, "internal onCompletion due to error")
+                else -> Timber.tag(tag).v("internal onCompletion")
```
mtwalli, line 87: Please include the `err.message` in the log? is not better to use `Timber.e` in this cases?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/BackgroundModeStatus.kt
```diff
@@ -90,7 +90,10 @@ class BackgroundModeStatus @Inject constructor(
         .onStart { emit(pollisIgnoringBatteryOptimizations()) }
         .distinctUntilChanged()
         .onCompletion {
-            if (it != null) Timber.w(it, "isIgnoringBatteryOptimizations failed.")
+            when {
+                it is CancellationException -> Timber.d("isIgnoringBatteryOptimizations canceled.")
+                it != null -> Timber.w(it, "isIgnoringBatteryOptimizations failed.")
```
mtwalli, line 95: is `it`  an error ? If so could you please use `Timber.e`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/flow/HotDataFlow.kt
d4rken: The signature is `it: Throwable?`

* If it's a cancellation exception we don't care about the specific error message
* If it is an error we log it which would include it's error message automatically ( `.w(err,...)` ) 
* If it is neither of the above, then it was null and we have completion not caused by an exception.
* 
I'll adjust the log level to `e`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/BackgroundModeStatus.kt
d4rken: `it: Throwable?`

I'll adjust the log level to `e`

mtwalli (approved):
thanks

BMItr (approved):

PR #2320, opened 2021-02-10
Adapt 'fake correct device time' feature (DEV)

LukasLechnerDev:
This adjusts and extends the test menu feature from @d4rken => https://github.com/corona-warn-app/cwa-app-android/pull/2316

- Change the hardcoded offset from 90 to 0
- Add a description of the "Fake correct device time" in the Test Menu

image


d4rken (approved):
lgtm

ralfgehrer (approved):

PR #2324, opened 2021-02-11
Persist result from otp authorization server (EXPOSUREAPP-4637)

chris-cwa:
OTP repository is extended to store the result of an authorization request.

Instead of changing the fields of the locally generated otp here the option was chosen to have 2 separate entities. And here is why:

Both are used in different points of time in the OTP validation process and keeping them separate seemed cleaner.
otp repo
The OTP Repository also internally makes sure that at any point in time there can be only one of both persisted.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.google.gson.annotations.SerializedName
+import org.joda.time.Instant
+import java.util.UUID
+
+@Keep
+data class OTPAuthorizationResult(
+    @SerializedName("uuid")
+    val uuid: UUID,
+    @SerializedName("expirationDate")
+    val expirationDate: Instant,
```
LukasLechnerDev, line 13: This needs to be `Instant?` because unsuccessful authentications won't have an `expireDate`, or?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
```diff
@@ -0,0 +1,18 @@
...
+    @SerializedName("authorized")
+    val authorized: Boolean,
+    @SerializedName("redeemedAt")
```
LukasLechnerDev, line 16: Is this the time when we received the authentication result?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
```diff
@@ -42,7 +43,32 @@ class SurveySettings @Inject constructor(
                 .putString(KEY_OTP, if (value == null) null else gson.toJson(value))
                 .apply()
 
+    var otpAuthorizationResult: OTPAuthorizationResult?
```
LukasLechnerDev, line 46: Because we now store the `AuthorizationResult` I think there is no reason to store the `otp` in the preferences anymore?

LukasLechnerDev (changes requested):
I really like the diagram and the new `OTPAuthResult` entity 👍

I have some questions:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
LukasLechnerDev: According to the change from Max, we might don't even need the `expirationDate`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
chris-cwa: yes :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
chris-cwa: yes, if you want to have a look at OTPReposititory.otpAuthorizationResult.set: the otp is removed from preferences.

LukasLechnerDev: I mean we actually **never** have to store the otp in the preferences? Since the flow is always Generate OTP => Authorize OTP, the OTP would only be in the `preferences` for a couple of milliseconds or seconds. Or am I wrong?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
chris-cwa: removed the field

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/SurveySettings.kt
chris-cwa: what if the connection breaks when trying to authorize?

chris-cwa: can we leave it for now? i see your point, but i'm afraid to delete already.

LukasLechnerDev (approved):
As discussed, LGTM

kolyaopahle (approved):
Nice Drawings 👍 LGTM

PR #2326, opened 2021-02-11
Missing tracing duration on low risk detail screen (EXPOSUREAPP-5092)

LukasLechnerDev:
Issue: The active tracing history ("Exposure logging was active for x of the past 14 days.) was missing in the low-risk detail screen. It was only missing when there were encounters. 

Fix: I made it visible on the detail screen via editing the binding expression in the layout. I created a new binding adapter to simplify the expression.

To get a low-risk card with encounters do: Test Menu => Select LOW_RISK_DEFAULT => press "Calculate Risk Level".

Before: 
image

After:
image


harambasicluka (dismissed):
Tested on a Pixel3, works as expected! Thanks :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/ViewExtensions.kt
```diff
@@ -4,6 +4,11 @@ import android.view.View
 import androidx.databinding.BindingAdapter
 import de.rki.coronawarnapp.util.recyclerview.ThrottledClickListener
 
+@BindingAdapter("visible")
+fun View.setVisible(visible: Boolean) {
```
ralfgehrer, line 8: This is exactly  the same but negated as the `setGone` below

Corona-Warn-App/src/main/res/layout/tracing_content_low_view.xml
```diff
@@ -74,7 +74,7 @@
 
         <androidx.constraintlayout.widget.ConstraintLayout
             android:id="@+id/row_saved_days"
-            gone="@{state.getRiskContactLast(context) != null}"
+            visible="@{state.getRiskContactLast(context) == null || state.inDetailsMode}"
```
ralfgehrer, line 77: Why not use `gone` instead. No point in maintaining both adapters IMHO

LukasLechnerDev: I implemented it with `gone` in the beginning and it looked like this: 
`gone="@{state.getRiskContactLast(context) != null &amp;&amp; !state.inDetailsMode}"`
which is really hard to understand (`&amp;&amp;` is the logical AND operator (&&) in data binding )

The version here with `visible` is much easier to understand. 

I don't see any "costs" of maintaining this additional adapter, but I see some value in simplifying the expression.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/ViewExtensions.kt
LukasLechnerDev: See other comment.

Corona-Warn-App/src/main/res/layout/tracing_content_low_view.xml
ralfgehrer: How about moving the logic to the state object --> `gone=@{state.isGoneOnContentLowView(context)}` for  instance

LukasLechnerDev: Good idea. Moved the logic to the state object and wrote tests for it.

harambasicluka (approved):

ralfgehrer (approved):

PR #2329, opened 2021-02-11
Moving legal strings to strings (DEV)

harambasicluka:
As discussed with @janetback the following strings should be part of our normal translatable text and aren't treated as legal texts:

* `datadonation_survey_consent_details_title_below`
* `datadonation_survey_consent_details_text_below`


janetback (approved):
UA reviewed

PR #2341, opened 2021-02-12
Survey debug menu (EXPOSUREAPP-4855)

BMItr:
This Pr addresses the testmenu regarding the survey.
Details: EXPOSUREAPP-4637

Features:
- show OTP is available
- show Survey Config if available
- simulate several Exception for SaftyNet & Survey and show dialog

How To Test:
- testerDebugBuild
- Testmenu
- DataDonation (SafetyNet, Analytics, Surveys, etc)
- Check Cards

Screen:
image


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/datadonation/ui/DataDonationTestFragmentViewModel.kt
```diff
@@ -32,6 +41,21 @@ class DataDonationTestFragmentViewModel @AssistedInject constructor(
     val errorEvents = SingleLiveEvent<Throwable>()
     val copyJWSEvent = SingleLiveEvent<String>()
 
+    val otp: String = otpRepository.otp?.toString() ?: "No OTP received yet"
```
LukasLechnerDev, line 44: We should show the `OTPAuthorizationResult` here instead. The otp will be deleted after it got authenticated. And since the OTP is always authenticated after it generation, it will always be null when this screen is opened.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/datadonation/ui/DataDonationTestFragmentViewModel.kt
```diff
@@ -85,6 +109,42 @@ class DataDonationTestFragmentViewModel @AssistedInject constructor(
         }
     }
 
+    fun selectSafetyNetExceptionType(type: SafetyNetException.Type) {
+        currentSafetyNetExceptionTypeInternal.value = type
+    }
+
+    fun showSafetyNetErrorDialog() {
+        when (currentSafetyNetExceptionTypeInternal.value) {
+            SafetyNetException.Type.APK_PACKAGE_NAME_MISMATCH,
```
LukasLechnerDev, line 118: We also have these same mappings in `SurveyConsentViewModel`. I think we should re-use them. Because if they change in the `ViewModel` and we forget to change them here, the TestMenu shows a different error message than in the one in the real usage.

LukasLechnerDev (changes requested):
I left 2 comments.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/datadonation/ui/DataDonationTestFragmentViewModel.kt
BMItter: was not there when I made it yesterday

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/datadonation/ui/DataDonationTestFragmentViewModel.kt
BMItter: How?

BMItter: adjusted

LukasLechnerDev (dismissed):
LGTM

chris-cwa (approved):
lgtm - thanks!

ralfgehrer (approved):

LukasLechnerDev (approved):

PR #2348, opened 2021-02-15
Strings for contact diary encounter/visit button groups and extensions (EXPOSUREAPP-5108)

d4rken:
`contact_diary_location_visit_circumstances_hint` and `contact_diary_person_encounter_circumstances_hint` still need to be clarified to due differences within the mockups between Android and iOS.


janetback (approved):
UA approved

SamuraiKek (approved):

PR #2353, opened 2021-02-15
Discard OTP when high risk card disappears (EXPOSUREAPP-5144)

LukasLechnerDev:
Fulfills the requirement from the Survey story about: 

> When the high-risk card disappears (status change red --> green), a previously generated OTP is discarded.

There are already some unit tests in `RiskLevelChangeDetectorTest`

How to test it manually: 
- Trigger high-risk card (Test Menu => ENF v2 Calculation => INCREASED_RISK_DEFAULT => Calculate Risk Level)
- See high-risk card in HomeMenu; Enter the Risk details screen to see the survey (should currently work on INT)
- Trigger low-risk card (Test Menu => ENF v2 Calculation => LOW_RISK_DEFAULT => Calculate Risk Level)
- Check LogCat => "Discarding one time password for survey about previous high-risk state." should be printed out


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
```diff
@@ -49,7 +49,9 @@ class Surveys @Inject constructor(
     }
 
     suspend fun resetSurvey(type: Type) {
-        // TODO
+        if (type == Type.HIGH_RISK_ENCOUNTER) {
+            settings.clear()
```
BMItter, line 53: Shouldn't the flow be: OTPRepos clear method calls SurveySettings clear method?

LukasLechnerDev: Yes! Thank you. Changed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
```diff
@@ -49,7 +50,9 @@ class Surveys @Inject constructor(
     }
 
     suspend fun resetSurvey(type: Type) {
-        // TODO
+        if (type == Type.HIGH_RISK_ENCOUNTER) {
+            oneTimePasswordRepo.clear()
+        }
```
ralfgehrer, line 55: Minor: Would it be useful to add a Timber to the else case?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
```diff
@@ -49,7 +50,9 @@ class Surveys @Inject constructor(
     }
 
     suspend fun resetSurvey(type: Type) {
```
ralfgehrer, line 52: It's not a suspend function

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
```diff
@@ -49,7 +50,9 @@ class Surveys @Inject constructor(
...
-        // TODO
+        if (type == Type.HIGH_RISK_ENCOUNTER) {
```
ralfgehrer, line 53: Minor: please add a Timber to the `if` and `else` case

ralfgehrer (commented):
Successfully tested on Pixel (Android 10)

BMItr (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
LukasLechnerDev: Thanks, removed!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
LukasLechnerDev: Moved the log statement from `RiskLevelChangeDetector` to this if.

BMItr (approved):

ralfgehrer (approved):

PR #2362, opened 2021-02-15
Contact journal day screen: New attributes (EXPOSUREAPP-5108)

d4rken:
Adds new attributes to person and location entry for each day.

Adjusting the style to 100% match the mockups might take some time, we could evaluate merging this earlier so other tickets tha depend on data entry can better be worked on. The functionality itself is already functional.

### TODO
- [x] A handful of tests (Extra PR)
- [x] Adjust style to fit mockups (Extra PR)

### Testing
* Compare with mockups
* Play with data entry, leaving the app, restoring it, updating the entries etc.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/common/DiaryCircumstancesTextView.kt
```diff
@@ -0,0 +1,66 @@
...
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : ConstraintLayout(context, attrs, defStyleAttr) {
+
+    private val input: EditText
+    private val infoButton: ImageView
+
+    private var afterTextChangedListener: ((String) -> Unit)? = null
+    private val circumStanceTextWatcher = object : TextWatcher {
```
LukasLechnerDev, line 23: Instead of `TextWatcher` with empty method implementations, you could use `TextView.doAfterTextChanged` from KTX: https://developer.android.com/kotlin/ktx/extensions-list#extension_functions_18

d4rken: Using the focus approach for now, but thanks will use next time!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/common/DiaryCircumstancesTextView.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    init {
+        LayoutInflater.from(context).inflate(R.layout.view_diary_circumstances_textview, this, true)
+
+        input = findViewById<EditText>(R.id.input).apply {
+            setOnFocusChangeListener { v, hasFocus ->
+                if (hasFocus) {
+                    Timber.v("Focused on %s", v)
+                } else {
+                    Timber.v("Lost focus on %s", v)
```
LukasLechnerDev, line 31: On my device the focus isn't lost when putting the app in background (and then close it) 🤔

LukasLechnerDev (approved):
This was a triumph. 'Huge success'. It's hard to overstate my satisfaction.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/day/tabs/common/DiaryCircumstancesTextView.kt
d4rken: Will offer a solution to this in a follow up PR.

axelherbstreith:
LGTM 👍  Tested on Samsung S8 and Huawei P30

axelherbstreith (approved):

PR #2371, opened 2021-02-16
Your Risk Status Announcements (EXPOSUREAPP-5135)

axelherbstreith:
Currently, the Guidelines in the Risk Detail Screen are announced as one block when using TalkBack.

This PR fixes this issue by splitting up the entries and announcing them separately.

To test this, activate TalkBack and navigate to the Risk Detail Screen and go through the guideline items.
Make sure you test it with both `low risk` and `increased risk`.


d4rken (approved):
sgtm

ralfgehrer (approved):
Tested on Pixel (Android 10)

PR #2372, opened 2021-02-16
Collect test result data (EXPOSUREAPP-4820)

mtwalli:
- Implements `TestResultDonor`
- I don't persist initial data as it will be overwritten later before submission.

### The idea is the following:
- if the test result turns POSITIVE or NEGATIVE, it is included in the next submission. Afterwards, the collected metric data is removed.

- More than `hoursSinceTestRegistration` hours have passed since the test was registered, it is included in the next submission and removed afterwards. That means if the test result turns POS or NEG afterwards, this will not submitted

## Testing 
- From test menu you can create and submit data to the server.
- Once data submitted if you hit create again `Test result metadata` won't be created anymore.

## TODO

- [x] Finalise logic 
- [x] Add unit tests


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/storage/AnalyticsSettings.kt
```diff
@@ -89,12 +89,35 @@ class AnalyticsSettings @Inject constructor(
         defaultValue = false
     )
 
+    val testScannedAfterConsent = prefs.createFlowPreference(
+        key = PREFS_KEY_TEST_SCANNED_AFTER_CONSENT,
+        reader = { key ->
```
d4rken, line 94: I think for primitives `FlowPreference` has direct support, you can just specify the default value.

kolyaopahle (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/registeredtest/TestResultDonor.kt
```diff
@@ -0,0 +1,214 @@
...
+) : DonorModule {
+
+    override suspend fun beginDonation(request: DonorModule.Request): DonorModule.Contribution {
+        val scannedAfterConsent = analyticsSettings.testScannedAfterConsent.value
+        if (!scannedAfterConsent) {
+            Timber.d("Skipping TestResultMetadata donation (testScannedAfterConsent=%s)", scannedAfterConsent)
+            return TestResultMetadataNoContribution
+        }
+
+        val timestampAtRegistration = LocalData.initialTestResultReceivedTimestamp()
```
chiljamgossow, line 35: As we want to get rid of local data, we should avoid further usage and store the needed data in a dedicated and separate space.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
```diff
@@ -74,6 +88,25 @@ class SubmissionQRCodeScanViewModel @AssistedInject constructor(
         }
     }
 
+    // Collect Test result registration only after user has given a consent.
+    // To exclude any registered test result before giving a consent
+    private suspend fun saveTestResultAnalyticsSettings(testResult: TestResult) = with(analyticsSettings) {
```
chiljamgossow, line 93: Can you move this logic to dedicated class as it has nothing to do with the screen logic?

chiljamgossow (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/registeredtest/TestResultDonor.kt
mtwalli: I used this based on a discussion with Max, refactoring is a different topic. When is refactored it will replaced here too. Additionally I don't see any reason to create a new one saving the same value

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
mtwalli: I did not get it. Where do you want me to put? it is in the ViewModel , Does the ViewModel handles only screen logic ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/storage/AnalyticsSettings.kt
```diff
@@ -89,12 +90,73 @@ class AnalyticsSettings @Inject constructor(
         defaultValue = false
     )
 
+    val testScannedAfterConsent = prefs.createFlowPreference(
```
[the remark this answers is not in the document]
mtwalli: Do you suggest to create a new settings class  for these values only?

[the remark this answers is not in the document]
chiljamgossow: yes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/registeredtest/TestResultDonor.kt
chiljamgossow: You could just add it to the other data you collect. It makes it easy to wipe data without side effects.

mtwalli: It is not related to analytics data, I don't really like we should create a new one for every feature that uses it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
mtwalli: Improved ,Moved the logic of saving to `TestResultDonorSettings` ,but it still entry point is here. As per tech specs this should happen in QR scan, if you know a better handling let me know.

chiljamgossow: You could just move the whole method to a dedicated class and inject that one instead.

mtwalli: Done ✅, new class , new method , new test coverage

kolyaopahle (approved):

chiljamgossow (commented):
When I tested the data appeared repeatedly in the analytics

mtwalli:
> When I tested the data appeared repeatedly in the analytics

The data is not repeated, As I mentioned in the chat. It can be created multiple times as long as the submission is not done

chiljamgossow (approved):

PR #2373, opened 2021-02-16
PPA: Collect Client Metadata (EXPOSUREAPP-5147)

kolyaopahle:
### Description

This PR implements the Client Metadata Donor for the Privacy Preserving Analytics

### Steps to reproduce
1. Do an analytics collection in the test menu
2. Check if submitted data matches app / device specs


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientMetadataDonor.kt
```diff
@@ -0,0 +1,52 @@
...
+@Singleton
+class ClientMetadataDonor @Inject constructor(
+    private val clientVersionParser: ClientVersionParser,
+    private val clientVersionWrapper: ClientVersionWrapper,
+    private val appConfigProvider: AppConfigProvider,
+    private val enfClient: ENFClient
+) : DonorModule {
+
+    override suspend fun beginDonation(request: DonorModule.Request): DonorModule.Contribution {
+        val config = appConfigProvider.getAppConfig()
```
d4rken, line 19: We should use `configProvider.latestConfig` here. I don't think a 5 hour old config would be a problem here and `Analytics` will already attempt a refresh when checking for conditions.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientVersionWrapper.kt
```diff
@@ -0,0 +1,12 @@
+package de.rki.coronawarnapp.datadonation.analytics.modules.clientmetadata
+
+import android.os.Build
+import dagger.Reusable
+import de.rki.coronawarnapp.BuildConfig
+import javax.inject.Inject
+
+@Reusable
+class ClientVersionWrapper @Inject constructor() {
```
d4rken, line 9: Do we need an extra class for this?

We also already have `ApiLevel`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientVersionParser.kt
```diff
@@ -0,0 +1,37 @@
+package de.rki.coronawarnapp.datadonation.analytics.modules.clientmetadata
+
+import dagger.Reusable
+import de.rki.coronawarnapp.server.protocols.internal.ppdd.PpaData
+import javax.inject.Inject
+
+@Reusable
+class ClientVersionParser @Inject constructor() {
+    fun parseClientVersion(buildNumber: Int): ClientVersion {
+        val major = buildNumber / MAJOR_OFFSET
```
d4rken, line 10: Nice

d4rken: If we change the way the version is calculated this needs to change too. Please add a test case that fails if the buildNumber pattern changes, e.g. gets larger for different offsets.

d4rken (changes requested):
Code works good, would like a few small additions :hugs:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientVersionWrapper.kt
kolyaopahle: ah nice, didn't know about ApiLevel 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientMetadataDonor.kt
kolyaopahle: Sounds good 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientVersionParser.kt
kolyaopahle: makes sense, added a test that compares a generated version string with the original version string 👍

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/modules/clientmetadata/ClientVersionParserTest.kt
```diff
@@ -0,0 +1,35 @@
...
+    private fun createInstance() = ClientVersionParser()
+
+    @Test
+    fun `client version is properly parsed`() {
+        val currentVersionName = BuildConfig.VERSION_NAME
+        val currentVersionNumber = BuildConfig.VERSION_CODE
+
+        val parsedVersion = createInstance().parseClientVersion(currentVersionNumber)
+
+        parsedVersion.toVersionString() shouldBe currentVersionName
```
d4rken, line 33: Careful. Only `deviceForTesters` Variant has `-RCx`, would this unit test fail  if executed for a different build flavor?

kolyaopahle: noted, tossed out the whole parsing of version codes and switched to major minor and patch version being provided by the buildconfig

d4rken (approved):
lgtm

chiljamgossow (approved):
Tested, LGTM

PR #2387, opened 2021-02-17
PPA Collect exposure windows (EXPOSUREAPP-4819)

chiljamgossow:
Tech spec: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/40e00dfd0e4544bdf05a2ae83fa7123af86b9438/docs/spec/data-donation-client-ppa.md#new-exposure-windows

New exposure windows are stored on a new database. Primary key is a sha256Hash calculated on the string generated by the corresponding data class. For reported exposure windows only the hash value is stored.

A db rollback is performed in case the submission fails.

Stale data is being deleted on each donation.

How to test with test menu:
- fake exposure windows via ENF v2 calculation
- create current analysis via Data Donation


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowRepository.kt
```diff
@@ -0,0 +1,90 @@
...
+        val new = dao.getAllNew()
+        dao.deleteExposureWindows(new.map { it.exposureWindowEntity })
+        dao.deleteScanInstances(new.flatMap { it.scanInstanceEntities })
+        dao.deleteReported(dao.getAllReported())
+    }
+}
+
+private fun AnalyticsExposureWindow.sha256Hash() = toString().getSha256Hash()
+
+private fun String.getSha256Hash(): String? {
```
d4rken, line 57: Maybe `HashExtensions.kt` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/Database.kt
```diff
@@ -0,0 +1,164 @@
...
+)
+abstract class AnalyticsExposureWindowDatabase : RoomDatabase() {
+    abstract fun exposureWindowContributionDao(): AnalyticsExposureWindowDao
+
+    class Factory @Inject constructor(@AppContext private val context: Context) {
+        fun create(): AnalyticsExposureWindowDatabase = Room
+            .databaseBuilder(
+                context,
+                AnalyticsExposureWindowDatabase::class.java,
+                "AnalyticsExposureWindowDatabase"
```
d4rken, line 37: Maybe add an extension .db -db like the other database names in the app.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/Model.kt
```diff
@@ -0,0 +1,40 @@
+package de.rki.coronawarnapp.datadonation.analytics.modules.exposurewindows
+
+import com.google.android.gms.nearby.exposurenotification.ExposureWindow
+import com.google.android.gms.nearby.exposurenotification.ScanInstance
+import de.rki.coronawarnapp.risk.result.RiskResult
+
```
d4rken, line 6: `Model.kt` is a bit too generic.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelTask.kt
```diff
@@ -169,6 +177,34 @@ class RiskLevelTask @Inject constructor(
...
+            riskResultsPerWindow[window]?.let { result ->
+                val contribution = createAnalyticsExposureWindow(
+                    window,
+                    result
+                )
+                analyticsExposureWindowRepository.addNew(contribution)
+            }
+        }
+    }
+
```
d4rken, line 207: I would prefer it if we lower the analytics specific code in `RiskLevelTask`, e.g. inject a single analytics specific class on which we call a single function  e.g. `donorModule.maybeTrack(riskResultsPerWindow)` with the data it needs checking `analyticsEnabled` and using `analyticsExposureWindowRepository` would happen inside that class.

chiljamgossow: I extracted the code to AnalyticsExposureWindowCollector

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowCollector.kt
```diff
@@ -0,0 +1,50 @@
...
+    private suspend fun collectAnalyticsData(riskResultsPerWindow: Map<ExposureWindow, RiskResult>) {
+        riskResultsPerWindow.keys.forEach { window ->
+            riskResultsPerWindow[window]?.let { result ->
+                val analyticsExposureWindow = createAnalyticsExposureWindow(
+                    window,
+                    result
+                )
+                analyticsExposureWindowRepository.addNew(analyticsExposureWindow)
+            }
+        }
```
kolyaopahle, line 29: why not like this? can result actually be null?
```suggestion
        riskResultsPerWindow.forEach {
            val analyticsExposureWindow = createAnalyticsExposureWindow(
                it.key,
                it.value
            )
            analyticsExposureWindowRepository.addNew(analyticsExposureWindow)
         }
```

chiljamgossow: Both work. I like that I can name it so there is no confusion which one is the key and which one the value.

chiljamgossow: I replaced it

kolyaopahle (dismissed):
Works on my machine, LGTM 👍

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowDonorTest.kt
```diff
@@ -0,0 +1,147 @@
...
+            analyticsExposureWindowRepository
+        )
+
+
+    companion object {
+        @AfterAll
+        @JvmStatic
+        fun cleanup() {
+            unmockkAll()
+        }
```
mtwalli, line 145: This should happen in `BaseTest` already. is not it ?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowsRepositoryTest.kt
```diff
@@ -0,0 +1,127 @@
...
+    fun setup() {
+        MockKAnnotations.init(this)
+        every { timeStamper.nowUTC } returns now
+        coEvery { analyticsExposureWindowDao.deleteReportedOlderThan(any()) } just Runs
+    }
+
+    @AfterEach
+    fun teardown() {
+        clearAllMocks()
+    }
```
mtwalli, line 57: if you use @BeforeEach, this is not needed as they already initialised again

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowDonor.kt
```diff
@@ -0,0 +1,99 @@
...
+        Timber.w("Random number is $random. probabilityToSubmitNewExposureWindows is $probability.")
+        return random > probability
+    }
+
+    @VisibleForTesting
+    internal val emptyContribution by lazy {
+        Contribution(
+            data = emptyList(),
+            onDonationFailed = {}
+        )
```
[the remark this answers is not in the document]
mtwalli: Q: why is wrapped in lazy ?  I don't think it a heavy resource.
and I think you should improve it, in the logs it looks like this:
```2021-02-24 13:45:09.731 13186-13364/de.rki.coronawarnapp.test I/AnalyticsExposureWindowDonor: Contribution(data=[], onDonationFailed=Function1<kotlin.coroutines.Continuation<? super kotlin.Unit>, java.lang.Object>)
```
if it is an object it will just print the name `EmptyExposureWindowDonation` for example

mtwalli:
Tested it on Samsung S8 , data is collected and submitted

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowDonor.kt
```diff
@@ -0,0 +1,99 @@
...
+    data class Contribution(
+        val data: List<PpaData.PPANewExposureWindow>,
+        val onDonationFailed: suspend () -> Unit
+    ) : DonorModule.Contribution {
+        override suspend fun injectData(protobufContainer: PpaData.PPADataAndroid.Builder) {
+            protobufContainer.addAllNewExposureWindows(data)
+        }
+
+        override suspend fun finishDonation(successful: Boolean) {
+            if (!successful) onDonationFailed()
```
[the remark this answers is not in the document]
chiljamgossow: I collect data in a database, where it accumulates. I discussed with @d4rken if the default should be to move the data to reported, and rollback only in case of failure. We agreed on this approach.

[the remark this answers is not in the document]
mtwalli: Alirght

mtwalli (dismissed):
👍🏼

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowsDatabaseTest.kt
```diff
@@ -0,0 +1,76 @@
...
+        val allNew = dao.getAllNew()
+        allNew.size shouldBe 1
+        allNew[0].exposureWindowEntity.sha256Hash shouldBe "hash"
+
+        // move to reported
+        dao.moveToReported(listOf(wrapper), 999999)
+        dao.getAllNew() shouldBe listOf()
+        val reported = dao.getReported("hash")
+        assert(reported != null)
+        reported?.sha256Hash shouldBe "hash"
```
d4rken, line 46: Subjective opinion, please don't use `?.` for values under test, just let it fail with `!!.`, while `assert` also ensures this, it doesn't add much to the readability of the test.

d4rken (dismissed):
lgtm :+1:

kolyaopahle (dismissed):

mtwalli (approved):

ralfgehrer (approved):

PR #2388, opened 2021-02-17
PPA User Metadata not deleted on In-App Reset (EXPOSUREAPP-5203)

kolyaopahle:
### Description

The current version of the PPA feature does not delete all analytics related data on an app reset,
this PR introduces calls to the respective deletion functions on app reset

### Steps to reproduce
1. do an in-app reset
2. follow the onboarding until the Share Data screen
3. No user data should be populated
4. Do a Analytics Submission
5. No analytics data from before should be submitted


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/storage/AnalyticsSettings.kt
```diff
@@ -89,6 +90,8 @@ class AnalyticsSettings @Inject constructor(
         defaultValue = false
     )
 
+    fun clear() = prefs.clearAndNotify()
```
ralfgehrer, line 93: A unit test would be nice.

chiljamgossow (approved):
LGTM

ralfgehrer (approved):
Successfully tested on a Pixel (Android 10).

PR #2392, opened 2021-02-17
Reduce timeout to  5 sec (EXPOSUREAPP-5196)

mtwalli:
- This PR addresses an issue when Broadcast receiver is not triggered after requesting TEK keys using new Google APIs
- Before 20 sec timeout it gives the user the ability to click the button many times ,while DefaultTEKProvider is waiting to timeout after 20 sec if receiver did not deliver the keys

- Reducing timeout to 5 sec will timeout quickly with an error which then will fallback to old APIs to get the keys
- Show Progress dialog while retrieving the keys to give the user indication that there's something going in the background and prevent multiple many clicks


harambasicluka:
LGTM, tested on a Pixel3. Won't approve it until we have a decision how to proceed.

ralfgehrer (approved):
Successfully tested on Pixel (Android 10) with and w/o giving consent.

harambasicluka (approved):

PR #2393, opened 2021-02-17
Adjust PPDD error dialog messages (EXPOSUREAPP-5201)

d4rken:
Refactor SafetyNetException.kt and SurveyException.kt to use `HasHumanReadableError` interface and fix error codes not being added to the localized strings.

### Testing
* Check the code
* Check the dialogs in the test menu


ralfgehrer (approved):

BMItr (approved):

PR #2395, opened 2021-02-17
Fix text on bottom of new release info screen (DEV)

chiljamgossow:
text on bottom of new release info screen (preview was correct, but not the actual screen)


chris-cwa (approved):

jurajkusnier (approved):
LGTM

PR #2399, opened 2021-02-18
Add styles for diary circumstances text field (EXPOSUREAPP-5161)

LukasLechnerDev:
This PR styles the "circumstances" text field of the contact journal extension. 

Remarks: 
- Although not in the design, we (@SamuraiKek has also implemented them this way in the screen to enter a person/location) decided to show the label when text is entered. This is also according to the material design guidelines ("label should always be visible")

image


d4rken:
Hm it doesn't look "nice" :fearful: 
Maybe we should drop the TextInputLayout? So the white box less tall.
Is the InfoButton aligned correctly?

LukasLechnerDev:
@d4rken I removed the label of the text field so now it should look better hopefully. 

I also aligned the info button. 

What do you think, is it nicer?

image

d4rken (approved):
Marvelous

PR #2408, opened 2021-02-18
Person Button Styling(EXPOSUREAPP-5160)

axelherbstreith:
This PR adds styling to the newly added buttons in the contact journal day screen (person).

button

Having four rounded corners as well as a divider between the elements is not possible with our current implementation using `toggleGroups`.


Corona-Warn-App/src/main/res/color/button_contact_diary_person.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<selector xmlns:android="http://schemas.android.com/apk/res/android">
```
jurajkusnier, line 2: Comments are misleading. Also why there are 3 states when you change only one boolean variable `state_checked` ?

SamuraiKek: Agree, should just be default and state_checked true for this to work.

SamuraiKek:
This looks pretty good. I think the text in the design is not bold though. I've also tried to play with the styling of the buttons but you're right, toggle groups don't have an option for a divider and also the inner corners of the buttons can't be rounded as far as I can tell (design has all corners rounded). We could get closer to the current design by using normal Material buttons but I think we should stick to using the toggle group since it's more in line with the functionality that we want here.

Corona-Warn-App/src/main/res/color/button_contact_diary_person.xml
axelherbstreith: Good point, is changed 👍

axelherbstreith:
@SamuraiKek yes I know, the text looks bold... however, when using material buttons this seems to be the default appearance. I also set the fontFamily to `sans-serif-medium` in order to be aligned with the other parts of the app. Further tried to set it to `sans-serif-light` or `sans-serif-thin` to align it to the design but this resulted in an text appearance which was way to thin and didn't look nice at all. Therefore I decided to stay with this appearance for now

jurajkusnier (approved):

SamuraiKek (approved):
Looks good. Tested on Nokia 3 (Android 9).

PR #2410, opened 2021-02-18
Don't additionally base64 encode the nonce (DEV)

d4rken:


BMItr (dismissed):
tested on pixel 2. lgtm

mlenkeit (approved):
LGTM, tested it against the mock server and INT and the nonce check passed on both!

BMItr (approved):

PR #2413, opened 2021-02-18
Improve analytics error handling (DEV)

d4rken:
If there is a bug in a single data donation module, we don't want to abort the process and we do want the modules to have a chance to cleanup their state instead of just throwing the exception and leaving everyone hanging.

This change catches the exceptions logs them and lets the process continue, talked with @mlenkeit  and continuing with the data we have is the desired behavior.

For tester builds we log this into the error tool such that testers can easily check if something went wrong without having to parse the log.

### Testing
* Review code
* Check analytics submission via test menu


ralfgehrer (approved):

jurajkusnier (approved):
LGTM. Tested on Pixel 4 (Android 11)

PR #2414, opened 2021-02-18
Analytics: Improve AppConfig handling (DEV)

d4rken:
Analytics.kt should provide the current config to all DonorModule.kt in the donation request.
This prevents unnecessary config refreshes and race conditions where the config may change mid process.


chiljamgossow (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
```diff
@@ -67,8 +68,13 @@ class Analytics @Inject constructor(
...
 
-    suspend fun collectContributions(ppaDataBuilder: PpaData.PPADataAndroid.Builder): List<DonorModule.Contribution> {
-        val request: DonorModule.Request = object : DonorModule.Request {}
+    suspend fun collectContributions(
+        configData: ConfigData,
+        ppaDataBuilder: PpaData.PPADataAndroid.Builder
+    ): List<DonorModule.Contribution> {
+        val request: DonorModule.Request = object : DonorModule.Request {
+            override val currentConfig: ConfigData = configData
+        }
```
mtwalli, line 78: With SAM fun in `Request` we can avoid this boilerplate code, this would simply it to:
```kotlin
       val contributions = donorModules.map {
            Timber.d("Beginning donation for module: %s", it::class.simpleName)
            it.beginDonation { configData }
       }
```

by doing:
```kotlin 
    fun interface Request {
        fun currentConfig(): ConfigData
    }
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/modules/ExposureRiskMetadataDonorTest.kt
```diff
@@ -88,7 +90,11 @@ class ExposureRiskMetadataDonorTest : BaseTest() {
         val parentBuilder = PpaData.PPADataAndroid.newBuilder()
 
         runBlockingTest2 {
-            val contribution = createInstance().beginDonation(object : DonorModule.Request {})
+            val contribution = createInstance().beginDonation(
+                object : DonorModule.Request {
+                    override val currentConfig: ConfigData = mockk()
+                }
+            )
```
mtwalli, line 97: With the suggestion this would be:
```kotlin
createInstance().beginDonation { mockk() }
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/modules/ExposureRiskMetadataDonorTest.kt
```diff
@@ -136,7 +142,11 @@ class ExposureRiskMetadataDonorTest : BaseTest() {
...
```
mtwalli, line 149: Here too:
`createInstance().beginDonation { mockk() }`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/Analytics.kt
d4rken: Analytics is still a work in progress and we may have to extend the `Request` interface. `SAM fun` only allows a single function, right?

mtwalli: Yes, if you expect to add more properties, it is better not to have it.

mtwalli (dismissed):

PR #2424, opened 2021-02-19
Fix comment field not saving on app moving into the background (EXPOSUREAPP-5108)

d4rken:
In the new comment/circumstances field in the contact journal:
Previously we only saved using `EditText.onFocusLost` which does not trigger if the user just leaves the app and swipes the app away. Using `viewTreeObserver.addOnWindowFocusChangeListener` we can catch that.

This PR contains an additional refactoring for `View.addLifecycleEventCallback`, which I tried first, but this does not work in this case as we are not able to find the parent Fragment from within the viewholder layout. It's a nice refactoring nonetheless, so I'd like to keep it.

### Testing
* Enter something in the contact-journal day screen, swipe the app away, check that it was saved.
* Check that scroll position stays the same for statistics cards if you view the info screen

### TODO
* [x] After #2399 is merged, adjust callback to use `notifyTextChanged` method


kolyaopahle (approved):
LGTM 👍

ralfgehrer (approved):
Tested on Pixel (Android 10)

PR #2426, opened 2021-02-20
Enhance screenshots-FTL(EXPOSUREAPP-5250)

mtwalli:
- Run screenshots 📸 tests on Firebase Test Lab
- Refactor screenshots logic to work with both fastlane (locally) and Firebase Test Lab
- We can always switch at anytime with less effort

## Outcome
- Quick and reliable screenshots job
- High quality screenshots (No blank anymore)

## Changes
- No html file to display the results 🧾 
- Tests tab shows number of tests combined for both locales (DE, EN) = `96` Tests (In case. something went wrong, you would still know)
- Screenshots 🖼  and other artifacts (Logs + test_result) are zipped together.

[Build Link](https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/10132/workflows/a6bdbfc1-1250-4776-80ed-1c8b54131c49/jobs/61908/steps)

Enjoy 😉 release builds 🏭  🚀

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -601,9 +601,7 @@ jobs:
             sudo gsutil -m cp -R -U gs://${GOOGLE_PROJECT_ID}-circleci-android/${BUCKETDIR}/flame* firebase-screenshots
           when: always
       - store_test_results:
-          path: ./firebase-screenshots/flame-29-en_US-portrait/*.xml
-      - store_test_results:
-          path: ./firebase-screenshots/flame-29-de_DE-portrait/*.xml
+          path: ./firebase-screenshots
       - compress-path:
           input: ./firebase-screenshots
           output: /tmp/screenshots.zip
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -550,6 +550,65 @@ jobs:
           path: /tmp/device_screenshots.zip
           destination: zips/device_screenshots.zip
 
+  firebase_screenshots:
+    resource_class: xlarge
+      docker:
+          - image: circleci/android:api-28
+    steps:
+      - skip-for-external-pull-requests
+        - setup-google-cloud
+        - checkout
+        - restore-gradle-cache
+        - restore-android-build-cache
+      - run-gradle-cmd:
+          desc: Build APKs for screenshots
+          cmd: >
+            :Corona-Warn-App:assembleDebug
+            :Corona-Warn-App:assembleAndroidTest
+      - run:
+          name: Setup Testlab environment
+            command: |
+              echo "export BUCKETDIR=\"`date "+%Y-%m-%d-%H:%M:%S:%3N"`-${RANDOM}\"" >> $BASH_ENV
+              source $BASH_ENV
+              echo "$BUCKETDIR is setup."
+      - run:
+          name: Test with Firebase Test Lab
+          command: |
+            echo "Using bucketdir $BUCKETDIR"
+            sudo gcloud firebase test android run \
+              --type instrumentation \
+              --app Corona-Warn-App/build/outputs/apk/deviceForTesters/debug/*.apk \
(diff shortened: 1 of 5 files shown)


Corona-Warn-App/src/debug/AndroidManifest.xml
```diff
@@ -15,4 +15,8 @@
         android:name="android.permission.CHANGE_CONFIGURATION"
         tools:ignore="ProtectedPermissions" />
 
+    <application
+        android:requestLegacyExternalStorage="true"
```
harambasicluka, line 19: Why is this needed? Is it visible to the user as permission in the playstore?

mtwalli: It is needed to avoid new storage api restrictions in Android 29, So I can save screenshots on `sdcard/..`. it is the path where firebase can pull screenshots from according to their docs. [or the only path that really worked for me  :)]
https://cloud.google.com/sdk/gcloud/reference/firebase/test/android/run#--directories-to-pull.
AND it is not affecting end user at all. it is just debug builds as you can see the link of the file ⬆️  `/debug/AndroidManifest.xml`

ralfgehrer (approved):
Bummer that we lose fastlane's HTML overview, but well.

Corona-Warn-App/src/androidTest/java/testhelpers/ScreenShotter.kt
```diff
@@ -36,3 +57,35 @@ inline fun <reified F : Fragment> captureScreenshot(
     launchFragmentInContainer2<F>(fragmentArgs, themeResId, factory)
     takeScreenshot<F>(suffix)
 }
+
+/**
+ * Saves screenshots on the device's sdcard
+ */
+object SDCardCallback : ScreenshotCallback {
```
d4rken, line 64: The `screenshots.zip` seems to contain duplicate images:

* In `firebase-screenshots/flame-29-de_DE-portrait/artifacts/sdcard/screenshots`
* In `firebase-screenshots/flame-29-de_DE-portrait/artifacts`

.circleci/config.yml
```diff
@@ -550,6 +550,65 @@ jobs:
...
+          name: Install gsutil dependency and copy test results data
+          command: |
+            sudo pip install -U crcmod
+            sudo gsutil -m cp -R -U gs://${GOOGLE_PROJECT_ID}-circleci-android/${BUCKETDIR}/flame* firebase-screenshots
+          when: always
+      - store_test_results:
+          path: ./firebase-screenshots
+      - compress-path:
+          input: ./firebase-screenshots
+          output: /tmp/screenshots.zip
```
d4rken, line 607: Could we improve the naming so when others download the screenshots, there is no confusion about which version it is from? I think we have access to the commit hash in CircleCI, maybe `CWA-Android-Screenshots-$GitSha.zip`, can we get the build version easily?

d4rken (approved):
This can be merged for 1.13, but maybe we  do a follow up for 1.14?

Corona-Warn-App/src/androidTest/java/testhelpers/ScreenShotter.kt
mtwalli: Yes, I did not find a nice way to avoid the duplicated images. When I filter them gutil saved the result from both locales into one folder. i.e it does not keep the folder structure from Firebase test results. Which leads to an exception in the command when pulling two files with the same into the same source folder. 
/en_US/Onboarding.png , /de_DE/Onboarding.png. ->/ screenshots/Onboarding.png. if there is away to avoid this easily please let me know?

.circleci/config.yml
mtwalli: I think was late to change, I will address this together with the comment above (if possible) in another PR

PR #2429, opened 2021-02-22
Use last OTP auth result to check month on client side (EXPOSUREAPP-5246)

chris-cwa:
how to test:
1. take a survey (in month m)
2. make the card green, then red again
3. try to take the survey again (still in month m)
expected: app shows info (not: error) that you already took part this month

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
@@ -12,5 +12,10 @@ data class OTPAuthorizationResult(
     @SerializedName("authorized")
     val authorized: Boolean,
     @SerializedName("redeemedAt")
-    val redeemedAt: Instant
-)
+    val redeemedAt: Instant,
+    @SerializedName("invalidated")
+    val invalidated: Boolean
+) {
+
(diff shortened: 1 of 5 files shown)


LukasLechnerDev (approved):
LGTM! Test as described is successful. If I fake the time to next month, I get a new otp and can access the survey again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
```diff
@@ -12,5 +12,10 @@ data class OTPAuthorizationResult(
     @SerializedName("authorized")
     val authorized: Boolean,
     @SerializedName("redeemedAt")
-    val redeemedAt: Instant
-)
+    val redeemedAt: Instant,
+    @SerializedName("invalidated")
+    val invalidated: Boolean
```
BMItter, line 17: could be false by default

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/OTPAuthorizationResult.kt
```diff
@@ -12,5 +12,10 @@ data class OTPAuthorizationResult(
...
+) {
+
+    fun toInvalidatedInstance() = OTPAuthorizationResult(uuid, authorized, redeemedAt, true)
```
BMItter, line 20: just use copy

BMItr (approved):

PR #2430, opened 2021-02-22
FIX: No PPA submission in background over 36+ hours (EXPOSUREAPP-5286)

kolyaopahle:
### Description

This PR changes a few values in the PPA Worker Scheduling logic to combat huge initial execution delays and provide a more testable experience

Note that this does not influence the repeat of submissions every 24Hours afterwards, these should still happen as usual.

### Steps to reproduce
1. Install App and consent to analytics donation
2. Wait 24 Hours (or check logcat for actual analytics scheduling delay)
3. An Analytics donation should have been triggered in the background

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
@@ -17,6 +17,7 @@ import dagger.android.HasAndroidInjector
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.contactdiary.retention.ContactDiaryWorkScheduler
 import de.rki.coronawarnapp.databinding.ActivityMainBinding
+import de.rki.coronawarnapp.datadonation.analytics.worker.DataDonationAnalyticsScheduler
 import de.rki.coronawarnapp.deadman.DeadmanNotificationScheduler
 import de.rki.coronawarnapp.storage.LocalData
 import de.rki.coronawarnapp.ui.base.startActivitySafely
@@ -61,6 +62,7 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
     @Inject lateinit var powerManagement: PowerManagement
     @Inject lateinit var deadmanScheduler: DeadmanNotificationScheduler
     @Inject lateinit var contactDiaryWorkScheduler: ContactDiaryWorkScheduler
+    @Inject lateinit var dataDonationAnalyticsScheduler: DataDonationAnalyticsScheduler
 
     override fun onCreate(savedInstanceState: Bundle?) {
         AppInjector.setup(this)
@@ -108,6 +110,7 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
         scheduleWork()
         vm.doBackgroundNoiseCheck()
         contactDiaryWorkScheduler.schedulePeriodic()
+        dataDonationAnalyticsScheduler.schedulePeriodic()
         if (!LocalData.isAllowedToSubmitDiagnosisKeys()) {
             deadmanScheduler.schedulePeriodic()
         }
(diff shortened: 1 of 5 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -108,6 +110,7 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
         scheduleWork()
         vm.doBackgroundNoiseCheck()
         contactDiaryWorkScheduler.schedulePeriodic()
+        dataDonationAnalyticsScheduler.schedulePeriodic()
```
mtwalli, line 113: Adding this here will not schedule new worker does it?
The logs are misleading:
```
2021-02-22 15:15:12.896 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 8 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:18:40.306 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 8 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:18:57.219 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 18 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:19:15.002 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 1 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:19:21.499 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 21 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:19:27.070 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 6 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:19:55.359 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 1 Hours in the future, will repeat every 24 Hours from there
2021-02-24 15:20:29.996 25009-25009/de.rki.coronawarnapp.test D/DataDonationAnalyticsScheduler: scheduling Analytics job for 0 Hours in the future, will repeat every 24 Hours from there
```

While actually the first worker was scheduled and kept check the timestamps for the first log and the start time in the screenshot:
Screenshot 2021-02-22 at 15 24 31


my question here why it is moved from Application into here ?

kolyaopahle: It is moved here because on application init the Initial onboarding might not be completed and there is no need to schedule a job that will never be able to execute, also onboarding might complete after app initialisation (first launch) which would not have scheduled a job before

The logging is quite misleading yes, since we enqueue the work using `ExistingPeriodicWorkPolicy.KEEP` only the first ever call to `schedulePeriodic` will actually enqueue a task, sadly there is (afaik) no way for us to know if a task has been queued before so there is no way to make the logging more usefull

kolyaopahle: maybe removing the log completely would be a solution, but i don't really like the idea

mtwalli: Ok, I checked `enqueueUniquePeriodicWork` which return Operation status, but it does not say whether it was enqueued before or not. Maybe just rephrase the log message ,so it gives the reader a hint.
such as : "Scheduling Analytics job with KEEP policy for %s Hours in the future, will repeat every 24 Hours from there. Won't schedule if is already enqueued" or whatever you like 😃

kolyaopahle: As @d4rken has also added logging here to his pr i will remove mine

harambasicluka: see: https://github.com/corona-warn-app/cwa-app-android/pull/2436

mtwalli (approved):

d4rken (approved):
lgtm

ralfgehrer (approved):

PR #2431, opened 2021-02-22
Add network connected constraint to DataDonationAnalyticsWorker (EXPOSUREAPP-5286)

kolyaopahle:
### Description
This PR adds a network connected constraint to the work builder of the DataDonationAnalyticsWorker as we cannot proceed with analytics donation without a network connection there is no need to enqueue work in that case

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsWorkBuilder.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsWorkBuilder.kt
@@ -1,6 +1,8 @@
 package de.rki.coronawarnapp.datadonation.analytics.worker
 
 import androidx.work.BackoffPolicy
+import androidx.work.Constraints
+import androidx.work.NetworkType
 import androidx.work.PeriodicWorkRequest
 import androidx.work.PeriodicWorkRequestBuilder
 import dagger.Reusable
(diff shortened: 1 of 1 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsWorkBuilder.kt
```diff
@@ -25,5 +27,11 @@ class DataDonationAnalyticsWorkBuilder @Inject constructor() {
                 BackgroundConstants.BACKOFF_INITIAL_DELAY,
                 TimeUnit.MINUTES
             )
+            .setConstraints(buildConstraints())
```
harambasicluka, line 30: Would it make sense to add a log if the worker wasn't started due to the network constraint?

chiljamgossow (approved):
LGTM

mtwalli (approved):
I confirm it is added :)
Screenshot 2021-02-22 at 15 34 49

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/worker/DataDonationAnalyticsWorkBuilder.kt
kolyaopahle: I'm not sure if we can, the work is scheduled by the WorkManager api and if there is no network connection the worker is simply not spawned, i don't think we get any kind of callback

kolyaopahle: we could maybe remove the constraint and do the check inside of the worker ourselves but i think this kind of defeats the purpose of the work manager in the first place

PR #2435, opened 2021-02-22
Exposure Checks are continued in End-of-Life State (EXPOSUREAPP-5298)

jurajkusnier:
**Fixing:** Exposure Checks are continued in End-of-Life State

**Testing:**
1. Add a positive test to the app
2. There shouldn't be any new exposure check from the time you scanned the QR code until you remove the test from the app. Check in `os settings > google > covid 19 > three dots > exposure checks`
3. Remove QR code exposure checks should continue


harambasicluka (approved):
Tested it via time traveling on a Pixel 3 (Android 11):

`2021-02-23 23:25:15.004 11861-11900/de.rki.coronawarnapp.test I/DownloadDiagnosisKeysTask: task aborted, positive test result`

ralfgehrer (approved):
Checked with time travel on Pixel (Android 10)

PR #2439, opened 2021-02-23
Updating Scrolling Behavior of the Contact Journal Overview Screen (DEV)

axelherbstreith:
This PR changes the scrolling behavior of the contact journal overview screen so that the information text is not static anymore but acts as a part of the list and scrolls with it.

**Seems like there is some delay when switching tabs using this solution. Please do not merge until we clarified whether this is acceptable or not!**


Corona-Warn-App/src/main/res/layout/contact_diary_overview_fragment.xml
```diff
@@ -20,32 +20,51 @@
...
-            android:text="@string/contact_diary_overview_subtitle"
-            app:layout_constraintEnd_toEndOf="@id/guideline_end"
-            app:layout_constraintStart_toStartOf="@id/guideline_start"
-            app:layout_constraintTop_toBottomOf="@id/toolbar"
-            tools:text="Tragen Sie ein, mit wem Sie sich getroffen haben und wo Sie gewesen sind." />
 
-        <androidx.recyclerview.widget.RecyclerView
-            android:id="@+id/contact_diary_overview_recyclerview"
+        <androidx.core.widget.NestedScrollView
+            android:id="@+id/scroll"
```
kolyaopahle, line 25: maybe use a more specific id or none at all if you don't need to reference it

axelherbstreith: The id is needed to keep the current scrolling state when navigating between the overview and day screen. Without having an id, the scrolling state wouldn't be remembered and it would jump to the top when navigation back to the overview screen. But you're right, naming is kind of meaningless, I will think about a better name :)

jurajkusnier:
I would suggest setting margin/padding inside `LinearLayout` and let `NestedScrollView` have full width for a nicer scrolling effect.

kolyaopahle (dismissed):

SamuraiKek (dismissed):

mtwalli:
LGTM, I tested in on Samsung S8, Android 9 with https://github.com/corona-warn-app/cwa-app-android/pull/2439/commits/573d8f7f80374d71f221dc596210458f4abcf4b4

jurajkusnier (approved):
LGTM tested on Pixel (Android 11)

mtwalli (approved):
lgtm

d4rken:
Superseeded by #2482

PR #2441, opened 2021-02-23
Resize diary home screen attributes when empty (DEV)

SamuraiKek:
Small fix to hide the attributes textview when there aren't any attributes selected as it was causing unnecessary vertical space.


axelherbstreith (approved):

harambasicluka (approved):
LGTM, tested on a Pixel 3 (Android 11).

PR #2443, opened 2021-02-23
Hide progress dialog on missing consent from both Google and App (EXPOSUREAPP-5320)

mtwalli:
Before the fix:
1- Open the App and proceed  with QR code scanning 
2- Disallow Google consent 
3- Scan a positive test
4- Proceed to sharing
5- Click on Consent button and disallow our consent flag
6- return back and try to share the keys 
7- the progress dialog will be shown forever

After:
Dialog will hide if consent has not been given


jurajkusnier (approved):
LGTM. Tested on Pixel (Android 11)

ralfgehrer (approved):
LGTM. Tested on Pixel (Android 10).

PR #2449, opened 2021-02-23
Add Base32 extensions and unit tests for it (EXPOSUREAPP-5329)

mtwalli:
- Create Base32 wrapper and add unit tests for it to verify its correctness 
- Ticket: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5329 
- Sample test data : https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration/docs/spec/event-registration-client.md#base32-examples


harambasicluka:
This has to go in the feature branch for the event registration. 
ping @d4rken

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/Base32Encoder.kt
```diff
@@ -0,0 +1,35 @@
+package de.rki.coronawarnapp.eventregistration
+
+import com.google.common.io.BaseEncoding
+
+/**
+ * Base32 Encoding wrapper around [BaseEncoding.base32]
+ */
+class Base32Encoder {
```
d4rken, line 8: What do you think about making this an extension function and using the same pattern als `decodeBase64()` / `base64()`, see `ByteString`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/Base32EncoderTest.kt
```diff
@@ -0,0 +1,46 @@
+package de.rki.coronawarnapp.eventregistration
+
+import io.kotest.matchers.shouldBe
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ArgumentsSource
+
+class Base32EncoderTest {
```
d4rken, line 7: Please extend `BaseTest`.

mtwalli:
Tested on Samsung S8 Android 9, it flickers sometimes.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/Base32.kt
```diff
@@ -0,0 +1,25 @@
+package de.rki.coronawarnapp.eventregistration
+
+import com.google.common.io.BaseEncoding
+
+/**
+ * Decodes input String into byte array using Base32 decoder
+ * @return [ByteArray]
+ */
+fun String.decodeBase32ToArray(): ByteArray = BaseEncoding.base32().decode(this)
```
d4rken, line 9: What do you think about decoding to `ByteString` (`decodeBase64()` also uses it), less custom code from our end, and the API consumer of these extensions has more options.

d4rken: `decodeBase32(): ByteString`

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/Base32Test.kt
```diff
@@ -0,0 +1,45 @@
+package de.rki.coronawarnapp.eventregistration
+
+import io.kotest.matchers.shouldBe
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.ArgumentsSource
+import testhelpers.BaseTest
+
+class Base32Test : BaseTest() {
```
d4rken, line 8: Nice tests :heart:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/Base32.kt
mtwalli: Good idea, Added a new variant for it. we can use what suits the use case.
`decodeBase32ToByteString`

d4rken (approved):
:heart_eyes:  so nice!

mlenkeit:
Looks promising 👍

ralfgehrer (approved):

PR #2458, opened 2021-02-24
Update contact journal screenshots (DEV)

jurajkusnier:
Update mock data for contact journal screenshots

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -618,6 +618,7 @@ workflows:
   check_buildtype_device:
     jobs:
       - detekt
+      - firebase_screenshots
       - lint_device_release_check
       - ktlint_device_release_check
       - quick_build_device_release_no_tests
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -618,7 +618,6 @@ workflows:
   check_buildtype_device:
     jobs:
       - detekt
-      - firebase_screenshots
       - lint_device_release_check
       - ktlint_device_release_check
       - quick_build_device_release_no_tests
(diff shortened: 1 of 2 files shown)


ralfgehrer:
Thx. If we were to enable the `firebase_screenshots` on this branch, we could easily validate the fix and won't need to build an RC.

ralfgehrer (dismissed):
Many thanks. LGTM

harambasicluka:
[Screenhots](https://65263-268027139-gh.circle-artifacts.com/0/zips/screenshots.zip)

.circleci/config.yml
```diff
@@ -618,6 +618,7 @@ workflows:
   check_buildtype_device:
     jobs:
       - detekt
+      - firebase_screenshots
```
mtwalli, line 621: Please remove it?

harambasicluka: yes, as the screenshots are taken we can remove it :)

LukasLechnerDev: But the option to run the screenshots on firebase without modifying the config.yml file for certain branches/PRs would be nice. Any ideas about how we could accomplish this?

ralfgehrer: done

ralfgehrer (approved):

LukasLechnerDev (approved):
The correct text "Diese müssen nicht in Zusammenhang mit den von Ihnen erfassten Personen und Orten stehen." is now added to a high or low risk in our contact journal, so LGTM!

.circleci/config.yml
mtwalli: It is a nice idea, I don't have any on top of my head. if anyone has ideas let me know?

mtwalli (approved):

PR #2461, opened 2021-02-25
Improve screenshots testing(DEV)

mtwalli:
It looks like Screenshots is becoming an important feature for our project. 
A question was raised by @LukasLechnerDev  [here](https://github.com/corona-warn-app/cwa-app-android/pull/2458/files/7c19cb6dade4b529035150cc94796a36b3f52142#r582047283)

on how to  test that without adding `firebase_screenshots` to the flow?

I thought 💭 that might we need a solution that we can use during the implementation phase before even pushing to GitHub.
Here it is:
- Added a flag so you can you execute the test locally from Android Studio or command line 

### How?
1- From Android Studio edit the configuration and your test parameters:
Screenshot 2021-02-25 at 15 22 28

2- From the command line:
```
$ adb shell am instrument -w -m -e androidStudioMode true -e annotation testhelper.Screenshot   -e debug false -e class 'de.rki.coronawarnapp.ui.main.MainActivityTest' de.rki.coronawarnapp.test.test/testhelpers.TestApplicationUIRunner
```


LukasLechnerDev:
Hi! Thanks for this! This is really helpful when extending / adapting our screenshots. 

I have an idea how we could further improve this so that we don't need to change actual code during development:
We could add an instrumentation param like "testMode" in the Run Configuration of a test or TestClass in Android Studio whenever we want to store a screenshot on the sdcard: 

image

Then, in our function that takes the screenshot, we could have something like this: 

image

So without this Instrumentation param, our screenshots are taken as usual, but when a developer sets the param, they are stored on the sdcard. 

What do you think?

mtwalli:
> Hi! Thanks for this! This is really helpful when extending / adapting our screenshots.
> 
> I have an idea how we could further improve this so that we don't need to change actual code during development:
> We could add an instrumentation param like "testMode" in the Run Configuration of a test or TestClass in Android Studio whenever we want to store a screenshot on the sdcard:
> 
> image
> 
> Then, in our function that takes the screenshot, we could have something like this:
> 
> image
> 
> So without this Instrumentation param, our screenshots are taken as usual, but when a developer sets the param, they are stored on the sdcard.
> 
> What do you think?

This was my first thought, if you like this way, I can simplify this even further. Wait for my commit below ⬇️

LukasLechnerDev:
Yes, I like it very much 👍 ! We also have to write it in our Wiki so the other Devs know this "trick" too 😉

mtwalli:
> Yes, I like it very much 👍 ! We also have to write it in our Wiki so the other Devs know this "trick" too 😉

Done: https://github.com/corona-warn-app/cwa-app-android/pull/2461#issuecomment-785929008

LukasLechnerDev (approved):
LGTM! Tested it for the screenshot adaptions for the contact diary extension and works really nice. Great PR :)

ralfgehrer (approved):

PR #2463, opened 2021-02-25
Diary exporter improvements (DEV)

LukasLechnerDev:
This PR now includes some good suggestions that were given by @BMItter in a review yesterday. Due to time pressure yesterday, we decided to implement them in a separate PR. 

The suggestions were (all in `ContactDiaryExporter`):
- Use the Reusable scope instead of Singleton
- Use `getDateAndNameString` as extension function on `Duration`
- Use `getReadableDuration` as extension function on `Duration`

Since I found an already existing extension function that is similar to `getReadableDuration()` in `DurationExtensions`, I re-used this one and wrote some unit tests for it.


BMItr (approved):
lgtm 👌 thx

axelherbstreith (approved):

PR #2467, opened 2021-02-25
Duration Picker Minute State (DEV)

axelherbstreith:
This PR fixes an issue where the previously picked minutes are not set when opening the duration picker again, instead 00 was set for minutes. Now, the minutes for the duration picker are set to the current duration.

To test this PR: Go to the contact diary day screen (Places) and pick a duration with Minutes ≠ 00 e.g. 04:15. Save the duration and press on the duration picker again, check if the minutes are displayed correctly.


AlexanderAlferov (approved):
Good catch!

d4rken (approved):

Issue #2049, opened 2021-01-07
CWA 1.10.1 crashes/closes/not starting/hanging


dsarkar:
***This bug was mirrored from feedback in the Play Store, so it can be discussed by the Github community.***

## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: Several models
- Android version: 6, 9, 10, 11
- App version: CWA 1.10.1

## Describe the bug

From the Play store, multiple reports on CWA 1.10.1 crashing/not starting/hanging/closing immediately: 

- v1.10.1 (1100100) Android 10.0, Galaxy s10+ (beyond2)

- v1.10.1 (1100100) Android 6.0, P8 (hwgra)

- v1.10.1 (1100100) Android 9.0, G8s thinq (betalm)

- v1.10.1 (1100100) Android 10.0, Galaxy a6 (a6lte) 

- v1.10.1 (1100100) Android 6.0, P8 青春版 (hwale-h) 

- v1.10.1 (1100100) Android 11.0, Pixel 2 (walleye) 

- v1.10.1 (1100100) Android 11.0, Galaxy s20+ 5g (y2s)

- v1.10.1 (1100100) Android 6.0, Honor 7 (hwplk) 

---

EDIT 210108 similar (old) issue https://github.com/corona-warn-app/cwa-app-android/issues/1456: 
- Huawei P8 Lite (ALE-L21) Android 6.0 CWA 1.5 EMUI 4.0.3 - requested update to 1.10.1 and waiting feedback

- v1.10.1 (1100100) Android 6.0, Honor 7 (hwplk) (crash already since 1.9.1)

---
Internal Tracking ID: [EXPOSUREAPP-4471](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4471)
Related tickets would be EXPOSUREAPP-2850 and EXPOSUREAPP-4524

PhilippNowak96:
Can you provide any stacktrace from the playstore for further information or are the devs already on it?

dsarkar:
@PhilippNowak96 will check. Thanks.

d4rken:
[...]

PR #2468, opened 2021-02-25
Shared preferences migration: SETTINGS DATA (EXPOSUREAPP-5378)

jurajkusnier:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/preferences/EncryptedPreferencesMigration.kt
```diff
@@ -0,0 +1,22 @@
...
+
+class EncryptedPreferencesMigration @Inject constructor(
+    @Preferences private val settingsPreferences: SettingsPreferences,
+    @EncryptedPreferences private val encryptedSettingsPreferences: SettingsPreferences
+) {
+
+    fun migrate() {
+        Timber.d("EncryptedPreferencesMigration START")
+        if (encryptedSettingsPreferences.containsData()) {
+            settingsPreferences.isNotificationsRiskEnabled = encryptedSettingsPreferences.isNotificationsRiskEnabled
```
d4rken, line 16: Do we need to do this one by one? Have you evaluated just taking the whole preference key/value map and moving  it from one encrypted to non encrypted preferences?

@kolyaopahle  @jurajkusnier

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/launcher/LauncherActivityViewModel.kt
```diff
@@ -15,12 +16,14 @@ class LauncherActivityViewModel @AssistedInject constructor(
...
     dispatcherProvider: DispatcherProvider,
     private val cwaSettings: CWASettings,
+    private val encryptedPreferencesMigration: EncryptedPreferencesMigration
 ) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
 
     val events = SingleLiveEvent<LauncherEvent>()
 
     init {
         launch {
+            encryptedPreferencesMigration.migrate()
```
d4rken, line 26: Can we migrate earlier? Can we migrate async?

Note that `launch` will be canceled when the viewmodel is destroyed. That seems risky.

fynngodau:
Did I understood correctly that you are going to migrate some preferences from encrypted shared preferences to (unencrypted) "normal" shared preferences? Which data are you going to move – all of them, including the test registration data?

d4rken:
@fynngodau Evaluating it. At least temporary. 

99% of crashes in the Google Play Console are from encryptedprefs/tink. While the keystore-retry mechanism workaround from  October reduced the crashes significantly, there are still people unable to use the app.

I posted some details here: https://github.com/corona-warn-app/cwa-app-android/issues/2049#issuecomment-756254926

The majority of effected devices are low-budget older Android devices, where it likely comes down to ROM issues, and is permanent, but there are also newer high-end devices where the issues are sporadic.

The additional protection through EncryptedPrefs on-top of the sandbox + system dm/fscrypt, is not significant IMHO.
Not significant enough to warrant depriving our fellow citizens of the option to use the app to protect themselves, just because they don't have or can't afford a different Android device.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
```diff
@@ -42,6 +42,20 @@ class CWASettings @Inject constructor(
         ).let { raw -> ConfigData.DeviceTimeState.values().single { it.key == raw } }
         set(value) = prefs.edit { putString(PKEY_DEVICE_TIME_LAST_STATE_CHANGE_STATE, value.key) }
 
+    var numberOfRemainingSharePositiveTestResultReminders: Int
+        get() = prefs.getInt(PKEY_POSITIVE_TEST_RESULT_REMINDER_COUNT, Int.MIN_VALUE)
+        set(value) = prefs.edit { putInt(PKEY_POSITIVE_TEST_RESULT_REMINDER_COUNT, value) }
+
+    val isNotificationsRiskEnabledFlow = prefs.createFlowPreference(
```
kolyaopahle, line 49: should we use the Flow postfix? I feel like this could confuse people expecting to use this var as a flow instead of usings its .flow property

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
```diff
@@ -42,6 +42,20 @@ class CWASettings @Inject constructor(
...
+        key = PKEY_NOTIFICATIONS_RISK_ENABLED,
+        defaultValue = false
+    )
+
+    val isNotificationsTestEnabledFlow = prefs.createFlowPreference(
```
kolyaopahle, line 54: same here

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/settings/notification/NotificationsSettingsTest.kt
```diff
@@ -25,27 +25,27 @@ class NotificationsSettingsTest : BaseTest() {
...
         mockkObject(LocalData)
 
-        every { LocalData.isNotificationsRiskEnabledFlow } returns flow { emit(true) }
-        every { LocalData.isNotificationsRiskEnabled = any() } just Runs
-        every { LocalData.isNotificationsRiskEnabled } returns true
-
-        every { LocalData.isNotificationsTestEnabledFlow } returns flow { emit(true) }
-        every { LocalData.isNotificationsTestEnabled = any() } just Runs
-        every { LocalData.isNotificationsTestEnabled } returns true
+        every { cwaSettings.isNotificationsRiskEnabledFlow.flow } returns flow { emit(true) }
```
kolyaopahle, line 35: you could also use
```suggestion
        every { cwaSettings.isNotificationsRiskEnabledFlow } returns mockFlowPreference(true)
```
note that this needs to be imported and im not 100% sure of the function name

kolyaopahle (approved):
Looks good to me, a few small ideas but nothing merge blocking

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
jurajkusnier: good point, updated

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/settings/notification/NotificationsSettingsTest.kt
jurajkusnier: didn't notice we have this `mockFlowPreference()`, updated

PR #2476, opened 2021-02-26
Contact diary person and place with more than 250 letters can be saved (EXPOSUREAPP-5400)

LukasLechnerDev:
Testers reported that it was possible to store more than 250 characters for names of persons and locations in our contact diary. 
This fix now sets a maxLength to the input fields so that users can't even add more than 250 characters and also trims the string again before writing to the database. 

I also applied this 250 character restriction to the email + phone fields of persons and locations.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/entity/ContactDiaryLocationEntity.kt
```diff
@@ -19,10 +20,13 @@ data class ContactDiaryLocationEntity(
         get() = locationId
 }
 
+private const val MAX_CHARACTERS = 250
+private fun String.trimMaxCharacters(): String = this.trim().trimToLength(MAX_CHARACTERS)
```
ralfgehrer, line 24: Afaik, we shall not trim the input values. Leading and trailing whitespaces for instance shall be possible (no input validation + apparently this also allowed on most phone contact apps)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/entity/ContactDiaryPersonEntity.kt
```diff
@@ -19,10 +20,13 @@ data class ContactDiaryPersonEntity(
...
+private const val MAX_CHARACTERS = 250
+private fun String.trimMaxCharacters(): String = this.trim().trimToLength(MAX_CHARACTERS)
+
 fun ContactDiaryPerson.toContactDiaryPersonEntity(): ContactDiaryPersonEntity =
     ContactDiaryPersonEntity(
         personId = this.personId,
-        fullName = this.fullName,
-        phoneNumber = this.phoneNumber,
-        emailAddress = this.emailAddress
+        fullName = this.fullName.trimMaxCharacters(),
```
ralfgehrer, line 29: maxLength should be sufficient

LukasLechnerDev: That's true. It's not necessarily needed. However, I think the data layer shouldn't rely on the UI layer to make correct sanitization and therefore should do it too. We also have that for the circumstances field that way.

LukasLechnerDev:
@SamuraiKek asked me to remove some uneccessary style items, see 👆

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/entity/ContactDiaryLocationEntity.kt
LukasLechnerDev: removed it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/entity/ContactDiaryLocationEntity.kt
```diff
@@ -19,10 +20,13 @@ data class ContactDiaryLocationEntity(
         get() = locationId
 }
 
+private const val MAX_CHARACTERS = 250
+private fun String.trimMaxCharacters(): String = this.trimToLength(MAX_CHARACTERS)
```
ralfgehrer, line 24: `trimToLength` also removes leading and trailing white spaces

LukasLechnerDev: I checked it, it uses substring internally, so it doesn't remove them ...

BMItr (dismissed):
lgtm.
tested on pixel 2 device

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/entity/ContactDiaryPersonEntity.kt
```diff
@@ -19,10 +20,13 @@ data class ContactDiaryPersonEntity(
         get() = personId
 }
 
+private const val MAX_CHARACTERS = 250
+private fun String.trimMaxCharacters(): String = this.trimToLength(MAX_CHARACTERS)
```
d4rken, line 24: This does not automatically remove whitespaces at start or end, and we want that right?
So `this.trim().trimToLength(....)`

LukasLechnerDev: Ralf made a comment earlier that we shouldn't doo that: https://github.com/corona-warn-app/cwa-app-android/pull/2476#discussion_r583497360

SamuraiKek (dismissed):
LGTM. Tested on Nokia 3 (Android 9). Tried by pasting randomly generated strings of 250 characters and then typing in more from the keyboard but couldn't.

LukasLechnerDev:
Clarified the `trim()` topic with @ralfgehrer: All input values (name, phone and email) should allow leading and trailing whitespace.  

I added another commit for that: https://github.com/corona-warn-app/cwa-app-android/pull/2476/commits/3f429aa62685ea144252bd9742e3c4a83f018219

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/entity/ContactDiaryLocationEntity.kt
LukasLechnerDev: The ViewModel and the Fragment did some trimming too, that's why it was still trimmed 🙃

SamuraiKek (approved):

ralfgehrer (approved):
LGTM. Thanks for the fix. Tested on Pixel (Android 10)

Issue #780, opened 2020-06-28
No beacons sent on Fairphone 2 (Fairphone OS 19.11.2 / Android 7.1.2)


gollub:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Describe the bug

The Fairphone 2, currently running Android 7.1.2 with access to GMS, is not sending any beacon / exposure notification.

(Note: there are two firmware flavors, Fairphone Open OS and Fairphone OS. This is about the later one, which has GMS access)

This was noticed while playing with https://github.com/marq24/UUID0xFD6FTracer (@marq24 kudos!)

There is absolutely no indication to the user that the CWA App is not fully functional.

This behavior could be verified on two Fairphone 2 devices running the same Android / firmware. Also verified with other Bluetooth Beacon scanners that the Fairphone 2 OS is is never  sending a beacon. Did also a functional cross check with Samsung device.

Followed also the instructions from https://github.com/corona-warn-app/cwa-app-android/issues/674#issuecomment-646949156 to verify that the Fairphone 2 has no "Peripheral mode support" (and the Samsung has).

From the adb logs of the Fairphone 2:

`06-28 17:06:54.186  3370 22809 I ExposureNotification: Utils#isSupported enabled=true, isDeviceSupported=true, isBluetoothSupported=true, BluetoothAdapter.isMultipleAdvertisementSupported=false [CONTEXT service_id=236 ]`

## Expected behaviour

Give the users warning/notice/hint that this device is not able to send beacons, notify other devices.

Or even better would be to get exposure notification working. But unclear if that's possible. Will reach out to Fairphone team in parallel to clarify if they are aware of that issue.

## Steps to reproduce the issue

1. Buy a smartphone certified with the Blaue Engel by the Umweltbundesamt ;)
2. Install the CWA App as recommended by the RKI / Bundesministerium für Gesundheit ;)
3. Follow the instruction of the CWA App and grant the App access for "Use location" and enable Bluetooth

... the users get the impression that they actively participating, since no warnings or notice is given that the device is actually not capable sending beacons.

## Technical details

- Mobile device: Fairphone 2 / Fairphone OS 19.11.2
- Android version: Android 7.1.2

[...]

Issue #688, opened 2020-06-20
User generated list of incompatible devices lacking BLE "Peripheral mode"


gizmo21:
CWA app has requirement of Android 6 and Bluetooth Low Energy, but that alone is not enough to get the full functionality.
To **send** the "Exposure Notification" beacon your device not only needs BLE but also a special feature called "Peripheral mode". That way you can never warn others if you are infected, cause your device never sends IDs to other phones. But you can get warned, cause your device is able the record others phones IDs. 
Even with lack of that "Peripheral mode" on your device the CWA seems to work flawlessly from user standpoint cause neither google play service nor CWA shows the limitation that it is only working one way.

Users can test their devices with this app https://play.google.com/store/apps/details?id=no.nordicsemi.android.mcp

@JProof provided screenshots here https://github.com/corona-warn-app/cwa-app-android/issues/674#issuecomment-646949156
**Selecting device information and look for Peripheral mode supported - NO**
![](https://user-images.githubusercontent.com/12245927/85194325-abe25c80-b2cd-11ea-8c05-eb899c8f8f61.png)

---
---

### Samsung A7 (SM-A700F)
![](https://user-images.githubusercontent.com/12245927/85194329-ac7af300-b2cd-11ea-81e4-b40831adf9d3.png)

----
Internal Tracking ID: EXPOSUREAPP-1667

gizmo21:
### LG Google Nexus 5

![8f617cd9-93f9-49f8-bc2c-f6a49cf98cc1](https://user-images.githubusercontent.com/2779732/85200801-920b3e80-b2fa-11ea-99c5-7543c707dfb2.png)

sqrt-1764:
Instead of generating a compatibility list (that would have to be maintained) Ii would prefer to disable the option to send the infection report to the server in that case as it is useless (as already commented in #674 ) and instead replace it with a note that explains this.
The avaliablility of the "peripheral mode" can be detected programmatically and the replaced note informs the user of anything that needs to be known ...

gizmo21:
As I understood, the devs atm had no direct access to BT but only the GAProtocol, so perhaps no quick solution available as they (correctly) say it's Google job to warn or even disable protocol for those devices.
That's why I thought users soon should at least be able to check and inform others that there are older devices out there that SEEM to be fully compatible but aren't.  And a separate thread so the other threads aren't spammed with "this device too" .

sqrt-1764:
[...]

PR #2481, opened 2021-02-26
Incompatibility warning card on main screen (COMMUNITY)

fynngodau:
As discussed in https://github.com/corona-warn-app/cwa-app-android/issues/688#issuecomment-785622753, this PR introduces a warning card on the main screen that informs users when their device is incompatible or only partially compatible with Exposure Notifications.

### Screenshots

| 1: BLE broadcasts not supported | 2: BLE scans not supported |
|---|---|
|Screenshot_1614350339|Screenshot_1614350116|

#### 

### To test

* On most devices: open app's main screen. Nothing changes with this PR.
* Use a device that doesn't support broadcasting beacons. Open app's main screen. New card like in screenshot 1. (For example, some operating systems for Fairphone 2 have a Bluetooth driver that doesn't support BLE peripheral mode, see also #780. Note that I don't own such a device)
* Use Android emulator. Open app's main screen. New card like in screenshot 2. (Android emulator doesn't support Bluetooth at all)

### What this PR doesn't do, but should still be done

* Add some further information to the app or to the FAQ or another website that can be displayed once the warning card is clicked.
* Block sharing test results if BLE broadcasts are not supported.

---
Internal Tracking ID: [EXPOSUREAPP-5464](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5464)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeAdapter.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeAdapter.kt
@@ -20,6 +20,7 @@ import de.rki.coronawarnapp.tracing.ui.homecards.TracingFailedCard
 import de.rki.coronawarnapp.tracing.ui.homecards.TracingProgressCard
 import de.rki.coronawarnapp.ui.main.home.items.FAQCard
 import de.rki.coronawarnapp.ui.main.home.items.HomeItem
+import de.rki.coronawarnapp.ui.main.home.items.IncompatibleCard
 import de.rki.coronawarnapp.ui.main.home.items.ReenableRiskCard
 import de.rki.coronawarnapp.util.lists.BindableVH
 import de.rki.coronawarnapp.util.lists.diffutil.AsyncDiffUtilAdapter
@@ -42,6 +43,7 @@ class HomeAdapter :
                 StableIdMod(data),
                 DataBinderMod<HomeItem, HomeItemVH<HomeItem, ViewBinding>>(data),
                 TypedVHCreatorMod({ data[it] is FAQCard.Item }) { FAQCard(it) },
+                TypedVHCreatorMod({ data[it] is IncompatibleCard.Item}) { IncompatibleCard(it) },
                 TypedVHCreatorMod({ data[it] is ReenableRiskCard.Item }) { ReenableRiskCard(it) },
                 TypedVHCreatorMod({ data[it] is IncreasedRiskCard.Item }) { IncreasedRiskCard(it) },
                 TypedVHCreatorMod({ data[it] is LowRiskCard.Item }) { LowRiskCard(it) },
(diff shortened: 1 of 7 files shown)


Diapolo:
I own a FP2 and use it with CWA (and microG). As far as I can tell I can send and receive beacons, microG is showing me collected IDs and a changing own ID.

fynngodau:
If microg is not showing a very similar warning text to one of the screenshots within its EN settings, this PR would not any as well, because the code for detecting incompatible devices is actually the same.

@Diapolo Is microg not displaying such a warning? Which operating system have you installed on your Fairphone 2?

Diapolo:
I'm using LineageOS for microG 17.1 for FP2, while my girlfriend is still using LineageOS 16 with GAPPS and both versions are able to communicate with eachother via EN beacons.

thomasaugsten:
@Diapolo is the microgG settings giving some warnings?
@fynngodau can you double check if the check is valid for the FP2 phones?

fynngodau:
To our knowledge there is a driver issue with Fairphone 2 that prevents BLE from working, it is not a hardware limitation. For this reason, it is possible that an OS for this device exists which ships with a driver that implements BLE broadcasting appropriately. 

The LineageOS that @Diapolo installed on their device could contain such a driver. If it does, and the device is broadcasting IDs, and if microG is (hence correctly) not showing a warning, then this PR will also not display a warning.

Our latest information is that the stock ROM (though this information could be outdated and there could be a fix in newer versions) doesn't have such a driver, and we heard from users who are correctly seeing the warning and would also see them in CWA with this PR.

Diapolo:
> @Diapolo is the microgG settings giving some warnings?

No there is NO warnings shown and everything EN-related seems to work fine...

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-5463](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5463)

doak:
Don't know it it is eventually the same code, but you can also use the _nrf Connect_ application to detect if "peripheral mode" is working.

More important, you can check if you actually receive BLE beacons from the other device (which proofs that that device is able to send).

Diapolo:
Screenshot_20210305-092109_nRF_Connect

This is my device and it shows "Peripheral mode supported" YES.

fynngodau:
Thanks for the interesting information @Diapolo, I'll remember that LineageOS for Fairphone 2 supports BLE broadcasting and forward this information to users in case it is needed. I'll also change the PR's description accordingly to avoid confusion.

vaubaehn:
Just for additional information/reference, and to support @fynngodau findings: No BLE beacons sent on FP2 with Fairphone OS 19.11.2 / Android 7.1.2, as reported here: https://github.com/corona-warn-app/cwa-app-android/issues/780#issue-646971831

vaubaehn:
Great to see that you consider to implement the proposed solution for release 2.1! 🎉

harambasicluka:
@fynngodau also haven't forgotten that one :) But the same applies as here: https://github.com/corona-warn-app/cwa-app-android/pull/2702#issuecomment-811269646

d4rken:
As this PR didn't make it into 1.15.x, it fell a bit behind, I rebased your commits and got it up to speed so that it should make the schedule for 2.2.x.

Closing in favor of #2873

PR #2482, opened 2021-02-26
Contact Journal Overview Screen Scrolling (DEV)

d4rken:
@axelherbstreith raised the issue in #2439 that the subheader text in the diary consumes valuable screen estate.

This initial attempt of making the whole screen scrolling was problematic due to the performance of nested layout scrolling.

* This PR makes the subheader a list item like the other entries, by reusing the same modular adapter we have for the homescreen.
* The overall layout nesting was slightly improved.

* Check that the screen still looks correct and scrolls nicely.


ralfgehrer (approved):
LGTM. Scrolls smoothly and works as expected on Pixel (Android 10) and Huawei P10 (Android 9).

BMItr (approved):

PR #2489, opened 2021-03-01
Fix race condition in test result donation (DEV)

mtwalli:
- Fix race condition when for ex:`navigationEventLiveData.postValue()` is called before calling `saveTestResultData`
- The coroutine will get cancelled as it is launched from `viewModelScope`
- For more information read Matthias post in `[PRIVATE] Android Team` chat

### Testing
- Collect test result data and check if is already collected in Test-Menu


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
```diff
@@ -56,10 +56,11 @@ class SubmissionQRCodeScanViewModel @AssistedInject constructor(
         try {
             registrationState.postValue(RegistrationState(ApiRequestState.STARTED))
             val testResult = submissionRepository.asyncRegisterDeviceViaGUID(scanResult.guid!!)
+            // Order here is important.to void race condition the coroutine will get cancelled
+            // when registrationState.postValue() is called before
+            testResultDataCollector.saveTestResultAnalyticsSettings(testResult)
```
chiljamgossow, line 61: What is the reason that you did not choose to grab the test result in the SubmissionRepository? No issues with race conditions there.

mtwalli: Yes, it should not `SubmissionRepository` is singleton. As TestResult metadata should be collected only for QR Code I did not want to add a new dependency that will live in the repo object as long as it is there. `SubmissionRepository` is used in other places too.

mtwalli: Adding it to the ViewModel, it will serve its purpose for the use case and it is gone once the user is not on this screen

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
```diff
@@ -56,10 +56,11 @@ class SubmissionQRCodeScanViewModel @AssistedInject constructor(
         try {
             registrationState.postValue(RegistrationState(ApiRequestState.STARTED))
             val testResult = submissionRepository.asyncRegisterDeviceViaGUID(scanResult.guid!!)
+            // Order here is important.to void race condition the coroutine will get cancelled
+            // when registrationState.postValue() is called before
```
d4rken, line 60: Typos and not really a sentence?

d4rken (dismissed):
lgtm, comment could be improved, but not a blocker for me.

chiljamgossow (dismissed):

d4rken (approved):

chiljamgossow (approved):

PR #2490, opened 2021-03-01
QR code and  pdf creation (EXPOSUREAPP-5430)

mtwalli:
- Create QR Code and PDF locally on the device
- Tech Specs : https://github.com/corona-warn-app/cwa-app-tech-spec/pull/25/commits/e91bbad4b54d7717afb0b4425be71223f2999f5f
- TextField is editable(see video below), can be replaced by another one by pasting and then hit QR Code button to generate a new one
- So far I was able to specify `Error Correction Level` , Input mode in [Zxing](https://github.com/zxing/zxing) is  chosen by examining the content https://github.com/zxing/zxing/blob/44267115986e7fdcb2a9629fd7986f65ab9a2670/core/src/main/java/com/google/zxing/qrcode/encoder/Encoder.java#L226
- QR Code creation can be checked also on Zxing website https://zxing.appspot.com/generator (Choose URL from `Contents` menu)

## How to test?

https://user-images.githubusercontent.com/25054729/109480869-5d03de00-7a7c-11eb-8544-a0dcd89396bd.mp4


LukasLechnerDev:
Tested it on Samsung S20 and Google Pixel 3A and worked really well. 

As far as I understand the code, we simply create a PDF out of the ConstraintLayout, right? So could there be bad resolution PDFs on bad resolution phones?

Have you thought about this point in the spec:
> - evaluate the possibility to describe the position of the QR code in the PDF via configuration

=> Should be possible by positioning the view in which the QR code is loaded according to the position data we get from the backend, or?

mtwalli:
@LukasLechnerDev  Regarding the positioning, Max is aware of that, AppConfig will probably hold x,y offsets. I also mentioned that width and height of Qrcode image can be specified too.

I have not tested it on different resolutions , this was just a POC to check whether it is possible or not.
For final implementation, there will be more specifications

LukasLechnerDev (approved):
LGTM

AlexanderAlferov:
Samsung S8 (Accessibility -> Visibility enhancements -> Font size and style -> 2 of 8) - does not look good 😢 

Vuzix View_20210304_182741_655

Scanning test:

- iPhone 11 - redirected to site
- Samsung S8, built-in scanner - redirected to Galaxy Notes
- Samsung S8, external app - redirected to site

mtwalli:
@AlexanderAlferov , thanks for the feedback , it is just for testing at the moment. this is not what the end user will see.
it is just to verify the idea of the POC.

Yeah, I think if you don't have an App (iOS or Android ) that handles the link, it will open the browser

AlexanderAlferov (approved):

PR #2493, opened 2021-03-01
Remove app shortcut during submission process (TAN) (EXPOSUREAPP-5425)

jurajkusnier:
The shortcut menu should not be visible if the submission process was started (from entering the positive test result screen until returning back to the home screen).
### Testing
Install CWA, finish onboarding, check if the app shortcut is available. 
Add positive TAN and immediately close the app before finishing the submission process. App shortcut shouldn't be available now. Finish the submission process and go back to the home screen. App shortcut should be available again.


ralfgehrer (approved):
LGTM. Tested on Pixel (Android 10).

LukasLechnerDev (approved):
Also works on Samsung S20+.

PR #2494, opened 2021-03-01
Debug log refactoring (storage limit & state information) (EXPOSUREAPP-5031)

d4rken:
* Add low storage check
* Exposure state information via flow (isRunning, isLowStorage etc.) to make it easier for the UI
* Improve testing
    * `DebugLogger` is no longer a singleton, but instead is instantiated once by the `CWADebug` singleton
    * Split up logic into extra classes, i.e. `LogWriter`, as the new requirements made the original `DebugLogger` a bit unwiedly.


janetback (approved):
UA reviewed

axelherbstreith (approved):

PR #2495, opened 2021-03-01
Pre-Authorise TEK history in QR - Code Scan [Only] (EXPOSUREAPP-5254) 

mtwalli:
- This PR removes `preAuthorizeExposureKeyHistory ` check when requesting TEKs later in the submission flow
- `preAuthorizeExposureKeyHistory ` is ONLY called before QR-Code Scan
- In the TAN flow you should not see the new consent at all ,but you might see new TEKs notification in one scenario:
       1- Go to QR-Code 
       2- Allow the consent
       3- Go back 
       4- Choose TAN and proceed to the submission flow 
       5- Notification of new Google API will be shown
- Otherwise in TAN flow old consent should be shown

Please note that if you give the consent and then changed device time to 2,3 hours in the future and went to QR-Code scanning again , the consent will be shown again in this case. this is not covered by this PR and I don't see any solution for it.

Issues addressed by this PR:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5254
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5427


ralfgehrer (approved):

LukasLechnerDev (approved):
LGTM. Tested on Samsumg S20+, Android 11 the QR-code (content dialog not shown again after the pending -> positive transition) and the 2 Tan flows (with and without pre-auth => notification was shown with pre-auth)

PR #2502, opened 2021-03-02
Deep linking for Event Registration (EXPOSUREAPP-5305)

mtwalli:
This PR unifies shortcuts and deeplinking handling in different cases such as:
1- Shortcuts scenarios MUST function as before
2- Opening a deep-link while user is not onboarded yet
3- Opening a deep-link when user is onboarded 
4- Opening a deep-link when user is onboarded ,but delta onboarding should be shown 
5- Opening a deep-link when App is already open

### Testing
- Test shortcuts 
- Test App reset 
- Test deep-linking by scanning  the QR Code in the PDF attached to test the deep-linking
[CoronaWarnApp-Event.pdf](https://github.com/corona-warn-app/cwa-app-android/files/6067823/CoronaWarnApp-Event.pdf)


jurajkusnier (approved):
- LGTM
- App-shortcuts work like before

mtwalli:
Thanks @jurajkusnier

mtwalli:
Merged into its parent branch, better to test all together

PR #2507, opened 2021-03-02
QR Code scanning to check-in (EXPOSUREAPP-5305)

mtwalli:
- Adds new scan fragment for Events - accessible from Test menu
- Handle deep-linking from Google Lens or third party Scanner
- Dummy confirmation screen
- Unifies shortcuts and deeplinking handling in different cases such as:
      1- Shortcuts scenarios MUST function as before
      2- Opening a deep-link while user is not onboarded yet
      3- Opening a deep-link when user is onboarded
      4- Opening a deep-link when user is onboarded ,but delta onboarding should be shown
      5- Opening a deep-link when App is already open
      6- Shortcuts

### Testing 
- Test shortcuts
- Test App reset (Just it launches as before)
- Scan Qr-Code attached
   - from `Scan Fragment` -> navigates to Confirmation screen
   - from `Qr-code Scanner` -> click on the link -> choose CWA to handle it -> navigates to Confirmation screen
[CoronaWarnApp-Event.pdf](https://github.com/corona-warn-app/cwa-app-android/files/6070765/CoronaWarnApp-Event.pdf)


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/launcher/LauncherActivityTest.kt
```diff
@@ -0,0 +1,105 @@
...
+        }
+    }
+
+    @After
+    fun teardown() {
+        clearAllViewModels()
+    }
+
+    @Test
+    fun testDeepLinkLowercase() {
```
LukasLechnerDev, line 62: What are you testing here? There is no assert? 🤔

mtwalli: Testing that the activity is launching when an intent with `data=https://coronawarn.app/E1/SOME_PATH_GOES_HERE` is fired

as you can see for cases below it will throw a `RuntimeException`

LukasLechnerDev (approved):
LGTM! Tested on Samsung S20+ and Google Pixel 3A. 

Shortcuts work as before. 

Scanning the QR code from an external app also doesn't skip onboarding, if not yet done, and then navigates to the Confirmation Screen.

AlexanderAlferov (approved):
Deeplinking works nice.
On Samsung S8 I had an issue with scanning QR code with built-in scanner: after scanning redirected to Galaxy Notes application, meanwhile other QR-code links works fine. After rebuilding QR code built-in scanner worked fine too. Probably some issues with test code.
With other QR-scanning apps no issues - everything works great 🚀

PR #2508, opened 2021-03-03
Dialog runs forever (EXPOSUREAPP-5493)

mtwalli:
From the logs I see the callback when TEK is received is called ,but despite that the dialog was not hidden.
My best guess here it is a race condition where `routeToScreen.postValue ` is navigating to the next screen and
`showKeysRetrievalProgress.postValue` never got the chance to have an effect.

Fix: `showKeysRetrievalProgress.postValue` is called before `routeToScreen.postValue` to guarantee it is called.

Logs section:
```
2021-03-02T19:16:45.202Z  D/TEKHistoryUpdater$updateHistoryAndTriggerCallback: onTEKAvailable(teks.size=1)
2021-03-02T19:16:45.203Z  I/AutoSubmission: updateMode(mode=MONITOR)
2021-03-02T19:16:45.207Z  V/SharedPreferencesImpl$EditorImpl: android.app.SharedPreferencesImpl@496ca1a:submission.user.activity.last changed to 2021-03-02T19:16:45.203Z
2021-03-02T19:16:45.211Z  V/SharedPreferencesImpl$EditorImpl: android.app.SharedPreferencesImpl@496ca1a:submission.auto.enabled changed to true
2021-03-02T19:16:45.212Z  V/AutoSubmission: scheduleWorker(): Creating periodic worker request for submission.
2021-03-02T19:16:45.229Z  D/AppInjector: Injecting SubmissionTestResultConsentGivenFragment{d21503f} (29a58653-0455-4dc5-814e-03cf0bd44616) id=0x7f090329}
2021-03-02T19:16:45.241Z  D/DefaultTEKHistoryProvider$getPreAuthTEKHistoryOrRequestPermissionOnV18: onTEKHistoryAvailable() -> permission were already available
2021-03-02T19:16:45.288Z  V/SubmissionTestResultConsentGivenViewModel: Initialized
```

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/resultavailable/SubmissionTestResultAvailableViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/resultavailable/SubmissionTestResultAvailableViewModel.kt
@@ -42,43 +42,41 @@ class SubmissionTestResultAvailableViewModel @AssistedInject constructor(
             override fun onTEKAvailable(teks: List<TemporaryExposureKey>) {
                 Timber.d("onTEKAvailable(teks.size=%d)", teks.size)
                 autoSubmission.updateMode(AutoSubmission.Mode.MONITOR)
-
+                showKeysRetrievalProgress.postValue(false)
                 routeToScreen.postValue(
                     SubmissionTestResultAvailableFragmentDirections
                         .actionSubmissionTestResultAvailableFragmentToSubmissionTestResultConsentGivenFragment()
                 )
-
-                showKeysRetrievalProgress.postValue(false)
             }
 
             override fun onTEKPermissionDeclined() {
                 Timber.d("onTEKPermissionDeclined")
+                showKeysRetrievalProgress.postValue(false)
                 routeToScreen.postValue(
                     SubmissionTestResultAvailableFragmentDirections
                         .actionSubmissionTestResultAvailableFragmentToSubmissionTestResultNoConsentFragment()
                 )
-                showKeysRetrievalProgress.postValue(false)
             }
 
             override fun onTracingConsentRequired(onConsentResult: (given: Boolean) -> Unit) {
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/resultavailable/SubmissionTestResultAvailableViewModel.kt
```diff
@@ -42,43 +42,41 @@ class SubmissionTestResultAvailableViewModel @AssistedInject constructor(
...
                     SubmissionTestResultAvailableFragmentDirections
                         .actionSubmissionTestResultAvailableFragmentToSubmissionTestResultConsentGivenFragment()
                 )
-
-                showKeysRetrievalProgress.postValue(false)
             }
 
             override fun onTEKPermissionDeclined() {
                 Timber.d("onTEKPermissionDeclined")
+                showKeysRetrievalProgress.postValue(false)
```
ralfgehrer, line 54: I'm wondering if it made sense to implement some sort of time-out for the dialog (as there is a submission time-out as well) or make the progress dialog dismissable. If we did, the user will not get stuck in that scenario.

mtwalli: For submission dialog I did not see any timeout.
There are many questions for introducing a timeout , we can discuss it and create a task for it ,but it is out of scope here. I don't like the whole thing ,it was not introduced because the tester is just using a tool that is not supposed to reflect what is happening on an end user device

ralfgehrer: thx. I agree with it being another task

ralfgehrer (approved):

BMItr (approved):
code lgtm.
We should keep an eye on it checking whether the implementation is sufficent for the issue

PR #2512, opened 2021-03-03
V2 Configuration Options for Risk Calculation (EXPOSUREAPP-5217)

BMItr:
This PR addresses ExposureApp-5217 (Update Configuration Options for Risk Calculation)

Story:
- changes to the app config may not break older client releases
- the configuration options shall allow to define a mapping of Normalized Time per Day to a Risk Level (already the case)
the risk calculation shall ignored Exposure Windows whose Normalized Time per Day cannot be mapped to a Risk Level 
- the configuration options shall allow to specify the Transmission Risk Value per Transmission Risk Level (one-to-one mapping) instead of using a generic multiplier
-updated protobufs

How to test:
- check if you receive the the new appconfig (v2) inside logcat
- check riskcalculation tests
- check card behavior triggered from testmenu (increased risk ->red, low risk ->green, etc.)

ToDo:

- [x] Replace default app config


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -264,19 +272,19 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
             throw NormalizedTimePerDayToRiskLevelMappingMissingException()
```
mlenkeit, line 272: @BMItter I think this should also be gone... is this the one failing test case you're still seeing? Otherwise, we should have a test case that's failing because of this (and it should be part of the JSON thingy)

mlenkeit (changes requested):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/sources/fallback/DefaultAppConfigSourceTest.kt
```diff
@@ -81,6 +82,7 @@ class DefaultAppConfigSourceTest : BaseIOTest() {
 
     @Test
     fun `exceptions when getting the default config are rethrown`() = runBlockingTest {
+        every { assetManager.open("default_app_config_android.bin") } throws FileNotFoundException("default_app_config_android.bin does not exist")
```
d4rken, line 85: Why did you add this? The original mock specifies a file to be read, which doesn't exist without `configFile.writeBytes(APPCONFIG_RAW)` and should thus throw an exception here without requiring us to mock an exception being thrown.

BMItter: But it doesn't and even fails on release 1.15.
Just tried to fix this test for test pipe.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
BMItter: outdated Exceptions have been removed.
(failing test cases this morning had other issues (v2<->v1)), also solved.

mlenkeit (approved):
Changes to the risk calculation itself look good to me.

Where/how do you make sure that the app config cache is invalidated?

BMItter:
> Changes to the risk calculation itself look good to me.
> 
> Where/how do you make sure that the app config cache is invalidated?
Invalidation will be added in a separte PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
BMItter: yes right, you can extend the testcase.json if you want

d4rken (approved):

PR #2516, opened 2021-03-03
Small refactoring, a review of used scopes for coroutines (DEV)

d4rken:
Use AppScope instead of ViewModelScope, if there is a reasonable expectation that the a screen could be closed prematurely, and we still want an operation scoped to it's viewmodel to complete.


ralfgehrer (approved):
LGTM. Took the liberty of fixing two broken tests.

LukasLechnerDev (approved):
LGTM

PR #2518, opened 2021-03-04
Fix timeformat. API expects exposureWindow timestamp to be in seconds. (EXPOSUREAPP-5529)

d4rken:
Milliseconds to seconds + a unit test.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposurewindows/AnalyticsExposureWindowDonor.kt
```diff
@@ -76,7 +76,7 @@ internal fun List<AnalyticsExposureWindowEntityWrapper>.asPpaData() = map {
     }
 
     val exposureWindow = PpaData.PPAExposureWindow.newBuilder()
-        .setDate(it.exposureWindowEntity.dateMillis)
+        .setDate(it.exposureWindowEntity.dateMillis / 1000)
```
ralfgehrer, line 79: very minor: didn't we have an extension function for this?

ralfgehrer (approved):

chiljamgossow (approved):

PR #2416, opened 2021-02-18
Fix endless uploading progress (EXPOSUREAPP-5237)

mtwalli:
- Hide Progress dialog while leaving the last screen in submission flow.
-  To avoid having endless progress

PR #2527, opened 2021-03-05
Submission dialog runs forever (EXPOSUREAPP-5547)

mtwalli:
Similar to the fix we had in #2416.
There are two other places where we end up having dialog running forever if the state was not synced before leaving the screen


ralfgehrer (approved):

BMItr (approved):

BMItter:
tested on pixel 2 device. lgtm

PR #2529, opened 2021-03-05
Submission process wrong cancelation dialog text (EXPOSUREAPP-4946)

SamuraiKek:
Replacing the text in the dialog that shows up when cancelling the submission process with new strings. 

How to test:
- Register a positive test
- In the "Ihr Testergebnis liegt vor" screen, press X
- The dialog with new strings should show up.
- Pressing "Anzeigen" closes the dialog and keeps you on the screen
- Pressing "Nicht Anzeigen" returns you to the main fragment
- The same behaviour happens whether consent is given or not

P.S. Change your phone's language to German, as it's the only language in which we have the right strings so far.


LukasLechnerDev (approved):
LGTM! Correct Strings are shown (see screenshot below) and the button navigation is also correct.

image

janetback (approved):
UA reviewed

harambasicluka (approved):

PR #2530, opened 2021-03-05
Refactored VerificationKeys.kt: Made the class more general purpose (DEV)

d4rken:
@mtwalli  and me refactored the verification keys class (now `SignatureValidation`), to be more general purpose.
This allows it to be reused for event registration.

###
* Check changed unit tests (invertation `hasInvalid` -> `hasValid`)
* Check on device, that statistics and app config are downloaded (because they use the same signature checks)


mtwalli (approved):
Tested on different environments and looks fine

chiljamgossow (approved):
Seems fine, statistics and countries appear in different environments

PR #2536, opened 2021-03-05
Create Check-in confirmation screen (EXPOSUREAPP-5424)

mtwalli:
- Move screens under `ui.eventregistration.attendee` package 
- Create two navigation graphs for attendee and organizer
- Add tab item for checkins and empty screen for it 
- Handle deep-linking in Check-Ins fragment
- Validate deeplink uri

### Testing 
[Invalid.pdf](https://github.com/corona-warn-app/cwa-app-android/files/6103541/Invalid.pdf)
[Valid.pdf](https://github.com/corona-warn-app/cwa-app-android/files/6103542/Valid.pdf)

 -  Scan QR Code from CWA scan screen or using third party scanner 
 -  For valid QRCode:
     -  A third party App will trigger CWA and opens Confirmation screen
     -  Scan screen will open  Confirmation screen
 -  For Invalid QRCode:
     -  A third party App will `NOT` trigger CWA -> browser will open this link
     -  Scan screen will open  an error dialog

### Update
After merging `Verification PR` , scanning Valid qr-code PDF will show an error too, since it is signed by a public that does not exist in any of our testing env


AlexanderAlferov (approved):
Tested on S8. Everything looks good.
Still have issues with navigation from built-in application on Samsung S8, but this is probably test pdf issue.

mtwalli:
> Tested on S8. Everything looks good.
> Still have issues with navigation from built-in application on Samsung S8, but this is probably test pdf issue.

Thanks @AlexanderAlferov  for the feedback. I think I know where the issue is. 
Camera Samsung App is triggering intent:

```xml
            <intent-filter>
                <action android:name="android.intent.action.SEND" />
                <category android:name="android.intent.category.DEFAULT" />
                <data
                    android:mimeType="text/plain" />
            </intent-filter>
```

while the App should respond to :
```xml
           <intent-filter>
                <action android:name="android.intent.action.VIEW" />

                <category android:name="android.intent.category.DEFAULT" />
                <category android:name="android.intent.category.BROWSABLE" />

                <data
                    android:pathPrefix="/"
                    android:scheme="https" />
            </intent-filter>
```
According to Android docs :https://developer.android.com/training/app-links/deep-linking
the intent action should be `VIEW` not `SEND`

I noticed also that Chrome is showing up in the sharing menu options .
I can add, but this will make CWA respond to sharing messages too like WhatsApp 😄

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkin/CheckInsFragment.kt
```diff
@@ -0,0 +1,65 @@
...
+import de.rki.coronawarnapp.databinding.FragmentCheckInsBinding
+import de.rki.coronawarnapp.util.di.AutoInject
+import de.rki.coronawarnapp.util.ui.doNavigate
+import de.rki.coronawarnapp.util.ui.observe2
+import de.rki.coronawarnapp.util.ui.viewBindingLazy
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactoryProvider
+import de.rki.coronawarnapp.util.viewmodel.cwaViewModels
+import javax.inject.Inject
+
+class CheckInsFragment : Fragment(R.layout.fragment_check_ins), AutoInject {
```
chiljamgossow, line 22: Can we call it CheckInFragment? The plural would make me expect a list of check ins. Same for the view model. Unless it is a list of course.

Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -69,12 +69,7 @@
                 <category android:name="android.intent.category.BROWSABLE" />
 
                 <data
-                    android:host="coronawarn.app"
-                    android:pathPrefix="/"
-                    android:scheme="https" />
-
-                <data
-                    android:host="www.coronawarn.app"
+                    android:host="e.coronawarn.app"
```
chiljamgossow, line 72: This is not case sensitive? I'm asking because the Validator checks both

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkin/CheckInsViewModel.kt
```diff
@@ -0,0 +1,45 @@
...
+            val encodedEvent = uri.substringAfterLast("/")
+            val traceLocationQRCode = qrCodeVerifier.verify(encodedEvent)
+            Timber.i("traceLocationQRCode: $traceLocationQRCode")
+            navigationRoutes.postValue(
+                TraceLocationQRCode(
+                    traceLocationQRCode.singedTraceLocation.event
+                )
+            )
+        } catch (e: Exception) {
+            Timber.d(e, "TraceLocation verification failed")
```
chiljamgossow, line 38: Are we not showing the error to the user? Or is this a separate task?

Corona-Warn-App/src/main/AndroidManifest.xml
mtwalli: It is not allowed to set a HOST in the manifest ,but it works with UPPER case uri

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkin/CheckInsViewModel.kt
mtwalli: We do as normal , `e.report()` display an error dialog

chiljamgossow: Ok, I expected e.report to do internal error logging

chiljamgossow (approved):

PR #2537, opened 2021-03-05
v2 config - invalidate appconfig (EXPOSUREAPP-5217)

BMItr:
ExposureApp-5217
FollowUp PR for https://github.com/corona-warn-app/cwa-app-android/pull/2512

Feature: Invalidate Appconfig due transmissionRiskValueMappingList is empty.


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/mapping/ExposureWindowRiskCalculationConfigMapperTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun `throws when riskCalculationParameters are missing`() {
+        val rawConfig = AppConfigAndroid.ApplicationConfigurationAndroid.newBuilder()
+            .build()
+
+        rawConfig.run {
+            hasRiskCalculationParameters() shouldBe false
+            hasDiagnosisKeysDataMapping() shouldBe false
+        }
+
+        shouldThrowAny {
```
d4rken, line 33: Would prefer to be more specific here, otherwise even something like a `NotMockedException` could accidentally cause the test to pass.

Maybe something like 

```kotlin
shouldThrow<ApplicationConfigurationInvalidException> {
    // Blub
}
```

d4rken (commented):
Looks good from my POV, if the underlying assumption (list empty = bad) is always the case (ping @mlenkeit ).

Thanks for adding the missing test cases.
Few comments, but no blockers.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/mapping/ExposureWindowRiskCalculationConfigMapperTest.kt
d4rken: If desired, to check for a specific abort condition, maybe something like:

```kotlin
val error = shouldThrow<ApplicationConfigurationInvalidException> {
    // Blub
}
error.msg shouldContain "Transmission Risk Value"
```

BMItter: 👌 
adjusted

harambasicluka:
@BMItter is there a good way to test this?

mlenkeit:
> Looks good from my POV, if the underlying assumption (list empty = bad) is always the case (ping @mlenkeit ).
> 
> Thanks for adding the missing test cases.
> Few comments, but no blockers.

@d4rken  yes, the assumption is fine

ralfgehrer (approved):

BMItter:
> @BMItter is there a good way to test this?
not really, at least not by UI.
but you can run the new tests or watch out inside the logcat for: 
"Transmission Risk Value Mapping List is empty which indicates an outdated app config"

chris-cwa (approved):

chris-cwa:
LGTM!

d4rken (approved):
:ok_hand:

PR #2544, opened 2021-03-08
Technical Support Functions UI Polishing (EXPOSUREAPP-5520)

axelherbstreith:
Refactoring and adapting of the UI according to the target design. 

Changed by now:
- id referenced changed from `@+id` to `@id `
- renaming of buttons
- removed NavigationEvents class
- removed greyish background from legal screen in dark mode


Corona-Warn-App/src/main/res/values/styles.xml
```diff
@@ -112,8 +112,8 @@
         <item name="android:textColor">@color/button_text_color_emphasized</item>
     </style>
 
-    <style name="debugLogButton" parent="button">
-        <item name="android:backgroundTint">@color/button_debug_log</item>
+    <style name="buttonBarAlertRed" parent="button">
```
harambasicluka, line 115: I'm a little late to the party, but don't we already have a red button? Settings > Reset

harambasicluka (approved):
One small comment, checked out this feature for the first time, looks nice. I'm just not happy about the sticky/floating bottom rectangle with a title, status box and buttons.

image

ralfgehrer (approved):

PR #2546, opened 2021-03-08
Generic duration picker (DEV)

mtwalli:
Duration is picker is required for event registration too. This PR make the current implementation more generic and customisable using a builder pattern like Material way.

- Rename the current picker to just `DurationPicker`,moved it a `durationpicker` package
- Add Builder pattern to easy add more options for the dialog if wanted (such as title for example)
- Hide args and dialog in creation in Builder

### Testing

- Check in the contact-diary and from Test menu


harambasicluka:
@mtwalli Would you mind changing the base to 1.16.x? It's not required for 1.15.x and we still have a lot of PRs to go :)

AlexanderAlferov (approved):
LGTM. Tested on Samsung S8.

jurajkusnier (approved):
LGTM
Duration Picker will be updated for the purpose of Event Registration in another PR

PR #2550, opened 2021-03-08
Initialize test results with timestamp from LocalData (EXPOSUREAPP-5595)

jurajkusnier:
Fixing an issue where after update from older app version to 1.13 user see EOL screen with old test result but current day

## Testing
You can find more details about how to reproduce the bug here: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5595


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
```diff
@@ -37,7 +37,8 @@ class SubmissionRepository @Inject constructor(
     private val deadmanNotificationScheduler: DeadmanNotificationScheduler,
     private val analyticsKeySubmissionCollector: AnalyticsKeySubmissionCollector
 ) {
-    private val testResultReceivedDateFlowInternal = MutableStateFlow(Date())
+    private val testResultReceivedDateFlowInternal =
+        MutableStateFlow(Date(LocalData.initialTestResultReceivedTimestamp() ?: System.currentTimeMillis()))
```
harambasicluka, line 41: Why should we ever show the current system time? In this state we always have a time when the user received the test result otherwise you won't be in EOL. Or did I understood something wrong?

harambasicluka (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
jurajkusnier: I don't know the reason why it was implemented this way in the beginning.
Another solution would be to change the flow to be nullable, but it could influence more modules.

ralfgehrer (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
ralfgehrer: We agreed to this change in a follow-up PR on 1.15 - which goes through full regression testing.

harambasicluka (approved):
LGTM, tested on a Pixel 3.

PR #2562, opened 2021-03-09
Custom View for Detailed Information (EXPOSUREAPP-5488)

axelherbstreith:
This PR introduces a custom view for detailed information

Custom View contains:

- **Mandatory title** -> if left empty, the view is still visible
- **Optional subtitle** -> if left empty, the view is collapsed so that the spacing between the title and divider fits
- **a divider** below the texts -> no Divider above the texts should be added to be aligned with material style guidelines

Sample (without subtitle): 

image


harambasicluka:
@axelherbstreith can this go in `release/2.0.x`?

d4rken (approved):

PR #2565, opened 2021-03-09
Debug Log Confirmation Dialog (DEV)

axelherbstreith:
This PR introduces a new dialog that ask the user for confirmation when pressing "Stoppen und Löschen". Only if the user pressed "Anayse stoppen" the recording will be stopped and deleted. If the user presses "Analyse fortführen" the alert will be dismissed and the recording continues. 

image


janetback (approved):
UA approved

d4rken (approved):
:+1:

PR #2571, opened 2021-03-10
Adjust server upload & screens (DEV)

d4rken:
* This makes the expiration date in the servers response optional as the mock server does not provide it, and the app doesn't use it
* The scrollview with all the text will scroll and highlight the ID history entry after successful upload (because it's hidden by all the legal text :frowning_face:)

Make sure it doesn't crash and use the mock server to upload a logfile.


axelherbstreith (approved):
Looks good 👍

BMItter:
> Looks good 👍

did the upload really succeed during your testing? ;-)

BMItr (approved):
testinstance for log uploads instable currently. 
but code lgtm.
thx for fixing npe.

PR #2575, opened 2021-03-11
Protocol Buffer update for release 2.0.x at Thu Mar 11 10:00:42 UTC 2021

Corona-Warn-App-Technical-User:


mtwalli (dismissed):
lgtm

Server-Protocol-Buffer/src/main/proto/internal/evreg/event.proto
```diff
@@ -4,9 +4,22 @@ package de.rki.coronawarnapp.server.protocols.internal.evreg;
...
 
 message Event {
+  // 16 random bytes
   bytes guid = 1;
-  string description = 2;
-  uint32 start = 3;
-  uint32 end = 4;
-  uint32 defaultCheckInLengthInMinutes = 5;
+  EventType type = 2;
+  // max. 150 characters
```
LukasLechnerDev, line 10: Figma designs say that we should only allow 100 characters

Server-Protocol-Buffer/src/main/proto/internal/pt/trace_location.proto
```diff
@@ -0,0 +1,31 @@
...
+syntax = "proto3";
+package de.rki.coronawarnapp.server.protocols.internal.pt;
+
+
+message TraceLocation {
+  // uuid
+  string guid = 1;
+  uint32 version = 2;
+  TraceLocationType type = 3;
+  // max. 150 characters
```
LukasLechnerDev, line 11: Figma designs say that we should only allow 100 characters

LukasLechnerDev (dismissed):
One small issue in the comments, but otherwise LGTM

Server-Protocol-Buffer/src/main/proto/internal/evreg/event.proto
```diff
@@ -4,9 +4,22 @@ package de.rki.coronawarnapp.server.protocols.internal.evreg;
 
 
 message Event {
+  // 16 random bytes
   bytes guid = 1;
```
LukasLechnerDev, line 8: I don't know why the guid is there in the Event. It is generated by the server, so we don't need to provide it here.

mtwalli: GUID is generate on the client , server is verify that it is unique. https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-concept.md#concept-for-event-creation

LukasLechnerDev: There was this change recently, that says that its generated on the server: https://github.com/corona-warn-app/cwa-app-tech-spec/pull/30/commits/15a0d0e5c8329d2a295de87881dbf119e11088bd

Server-Protocol-Buffer/src/main/proto/internal/evreg/event.proto
```diff
@@ -4,9 +4,22 @@ package de.rki.coronawarnapp.server.protocols.internal.evreg;
 
 
 message Event {
```
LukasLechnerDev, line 6: I think we don't need `Event` anymore since we now have `TraceLocation`?

mtwalli:
Can we clean up old models also here ? I think we don't need `SignedEvent` and `Event` anymore right?

Server-Protocol-Buffer/src/main/proto/internal/evreg/event.proto
mtwalli: If it has to be generated from the server, I suggest to provide no way of setting it in the client as currently it can be set from the client accidentally:
```kotlin
            TraceLocationOuterClass.TraceLocation
                .newBuilder()
                .setGuid(UUID.randomUUID().toString())
                .build()
```

LukasLechnerDev: Since (according to the tech spec) the client sends this same object as the one in the Server Response, it needs to be there.

mlenkeit:
@LukasLechnerDev  @mtwalli  the entire `evreg` subdir and its messages will be deleted. No need to comment on them. I only left them in for now so that the functionality from the Spike doesn't break.
/cc @harambasicluka

LukasLechnerDev (approved):

mtwalli (approved):

PR #2578, opened 2021-03-11
Fix logging related OOM issue (EXPOSUREAPP-5670/5691)

d4rken:
The issue was that we didn't filter out `""` (empty string) phone numbers, emails etc. so when it came to logging we attempted to replace `""` in every log line, with `fold` this lead to an OOM if more than one person/location existed due to recursion.

To reproduce without this PR, create 3 persons or locations that just have a name and no phone or email.

You can also check out this early commit in this PR, which has unit tests that reproduce this issue:
https://github.com/corona-warn-app/cwa-app-android/commit/dc5631421c614fbfe3ce776ee1d853552e89f914


BMItr (approved):
lgtm

ralfgehrer (approved):

PR #2580, opened 2021-03-11
Disable log-upload in `device` (prod) builds again (DEV)

d4rken:
Build type checks prevent the logger from initializing, and the OTP for log upload from being obtainable.
Some refactoring was necessary, tests were added.

😩


ralfgehrer (approved):
LGTM although I don't like it. 
Tested on deviceForTesters and device debug builds.

LukasLechnerDev (approved):
LGTM - Logging accessible in variant `deviceForTesters`, but not in `device`

¯\_(ツ)_/¯

PR #2595, opened 2021-03-12
Add check in risk to RiskLevelResult (DEV)

chiljamgossow:
extend interface to allow parallel work


BMItter:
@chiljamgossow 
PR looking good so far.
But one request after a little thinking:
could you make _TraceLocationCheckInRisk_ an own interface?
We can then create different objects e.g. for DB later..

chiljamgossow:
> @chiljamgossow
> PR looking good so far.
> But one request after a little thinking:
> Could you create _TraceLocationCheckInRisk_ as an Interface?
> So we can create different classes like an Entity for TraceLocationCheckInRisk with db related stuff

Sure, moved it to separate file as well

BMItr (approved):

d4rken (approved):

PR #2596, opened 2021-03-12
Encryption migration edge case handling (DEV)

d4rken:
@jurajkusnier  and me handled a few more edge cases after discussions on #2569 

* Handle errors during copy of data and still show the error reset dialog
* Initialize migration early, before all other components are injected to prevent possible edge cases where a component may initialize and try to access the preference before migration.

### Testing
Switch between 1.14 branch and this PR and read the start of logcat on app launch.


ralfgehrer (approved):
LGTM

jurajkusnier (approved):
- LGTM
- Migration works as expected

PR #2597, opened 2021-03-12
Trace location icon custom view (DEV)

AlexanderAlferov:
Trace location custom view for icon with date:

Screenshot from 2021-03-12 20-01-46Screenshot from 2021-03-12 20-07-55


Corona-Warn-App/src/main/res/layout/trace_location_organiser_qr_codes_list_item.xml
```diff
@@ -0,0 +1,72 @@
...
+            android:layout_width="42dp"
+            android:layout_height="42dp"
+            android:importantForAccessibility="no"
+            android:src="@drawable/ic_qr_code_list_item_icon"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintStart_toStartOf="parent"
+            app:layout_constraintTop_toTopOf="parent" />
+
+    </de.rki.coronawarnapp.ui.eventregistration.common.TraceLocationCardHighlightView>
```
d4rken, line 30: ### Usage

Set text as caption by using `android:text` on `TraceLocationCardHighlightView`, and any qr codes or other text views that should be displayed on top of the gradient, can be added as child of `TraceLocationCardHighlightView`.

The container with the gradient as background has a fixed size, while the caption text has a fixed width, but will grow in height to wrap it's content (i.e. line breaks possible).

d4rken (approved):

ralfgehrer (approved):

PR #2600, opened 2021-03-12
Log uncaught exceptions through the bugreporter (DEV)

d4rken:
Help testers to better catch exceptions when the app is executing through a worker in the background without UI.


ralfgehrer (approved):

jurajkusnier (approved):
LGTM

PR #2602, opened 2021-03-12
Add check-Ins to submission payload (EXPOSUREAPP-5656)

mtwalli:
- Provide Check-Ins to Submission payload
- Create a transformer to transform Database check-ins to port-buff check-ins (Currently just mapping, it will do more than that )

### Testing
- Check if all fields are mapped according to Tech-Specs
- Register a Test and submit your keys and check that is not affected by the change


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/EventRegistrationModule.kt
```diff
@@ -6,10 +6,15 @@ import de.rki.coronawarnapp.eventregistration.checkins.qrcode.DefaultQRCodeVerif
...
 import de.rki.coronawarnapp.eventregistration.storage.repo.DefaultTraceLocationRepository
 import de.rki.coronawarnapp.eventregistration.storage.repo.TraceLocationRepository
+import de.rki.coronawarnapp.eventregistration.checkins.CheckInsTransformer
+import de.rki.coronawarnapp.eventregistration.checkins.DefaultCheckInsTransformer
 
 @Module
 abstract class EventRegistrationModule {
 
+    @Binds
+    abstract fun checkInsTransformer(transformer: DefaultCheckInsTransformer): CheckInsTransformer
```
d4rken, line 16: I think in the future we can omit the interface if it is not necessary from the start (for features, or parallelizing work).

d4rken (dismissed):
lgtm

LukasLechnerDev (approved):
LGTM. Scanned positive code and submitted my keys - everything continues to work fine.

d4rken (approved):

PR #2606, opened 2021-03-12
Fix contact journal data visit/encounter updates (EXPOSUREAPP-5751)

### LINKED ISSUE #2594: Contact Journal: When entering place information, the sequence (duration first, then note) is relevant
## Avoid duplicates
* [ ] Enhancement request is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)
* [ ] If you are proposing a new feature, please do so in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)

## Current Implementation

When entering place information in the contact journal, currently the sequence is relevant (duration first, then note). If entered the other way around (note first, then duration) the note is lost.

## Suggested Enhancement

Either sequence should be feasible without data loss.

## Expected Benefits

No data loss.

---
Internal Tracking ID: [EXPOSUREAPP-5751](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5751)

d4rken:
This fixes #2594 where depending on the input order, the UI uses an older data object, changes one value, but due to working with an outdated data object, another value that was also changed is lost when the database is updated.

Let the diary repo offer an update method for changing visits and encounters, so we can do delta updates, and don't accidentally overwrite information. So now we pass the ID, the repo gives us the newest object, we change the value we want, and that's then written to the DB.


BMItter:
code lgtm for now 👌
Engage!

BMItr (approved):

chris-cwa (approved):

PR #2607, opened 2021-03-12
Test menu additions for download, matching, risk calculation

chiljamgossow:
Adds buttons to test menu
Some dummy implementations


d4rken (approved):
lgtm

mtwalli (approved):

PR #2611, opened 2021-03-15
Presence tracing app config (DEV)

mtwalli:
- Add configuration parameters for Presence Tracing (Event Registration)
- Check the specs https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#configuration-parameters

### Testing
- It is used in QR Code creation test to provide the correction error


axelherbstreith (approved):
Looks good 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/PresenceTracingConfigMapper.kt
```diff
@@ -0,0 +1,93 @@
...
+
+@Reusable
+class PresenceTracingConfigMapper @Inject constructor() : PresenceTracingConfig.Mapper {
+    override fun map(rawConfig: AppConfigAndroid.ApplicationConfigurationAndroid): PresenceTracingConfig {
+        if (!rawConfig.hasPresenceTracingParameters()) {
+            Timber.w("AppConfig does not have PresenceTracingParameters")
+            return PresenceTracingConfigContainer(
+                qrCodeErrorCorrectionLevel = ErrorCorrectionLevel.H,
+                revokedTraceLocationVersions = emptyList(),
+                riskCalculationParameters = emptyRiskCalculationParameters(),
```
d4rken, line 27: Might be nicer to use data class default parameters for these things?

d4rken (approved):
lgtm

PR #2612, opened 2021-03-15
Technical-Support-Function Screenshots (EXPOSUREAPP-5033)

axelherbstreith:
Added Screenshots for `DebugLogFragment` and `DebugLogUploadFragment`


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
```diff
@@ -0,0 +1,134 @@
...
+    @Rule
+    @JvmField
+    val localeTestRule = LocaleTestRule()
+
+    @get:Rule
+    val systemUIDemoModeRule = SystemUIDemoModeRule()
+
+    @Before
+    fun setup() {
+
```
d4rken, line 54: Why do you add an empty line at the start of functions?

d4rken (dismissed):

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
axelherbstreith: Good point, will be removed ;)

d4rken (approved):

chiljamgossow (approved):
LGTM

PR #2613, opened 2021-03-15
Fix notification settings (EXPOSUREAPP-5776)

jurajkusnier:
**Fixing**
- After Onboarding or Reset, Notifications is always Off


harambasicluka:
@jurajkusnier I think this part from the bug is still missing:
* Expected: During the Reset procedure, the user is asked wether to allow notifications or not
* Actual: During the Reset procedure, the user was NOT asked wether to allow notifications or not

In the past there where two switches on the onboarding screen.

jurajkusnier:
> @jurajkusnier I think this part from the bug is still missing:
> 
>     * Expected: During the Reset procedure, the user is asked wether to allow notifications or not
> 
>     * Actual: During the Reset procedure, the user was NOT asked wether to allow notifications or not
> 
> 
> In the past there where two switches on the onboarding screen.

Actual behaviour is correct according to the current Figma design.

harambasicluka (approved):
🚀

ralfgehrer (approved):

PR #2619, opened 2021-03-15
"My Check-Ins" skeleton UI & prep work (EXPOSUREAPP-5410)

d4rken:
* Wiring up the UI a bit more
* Some refactoring.
* Graphics, layouts
* Strings
* QR Code tests that are not yet running, I'll follow with another PR for that

### Testing
* Look at the "My Check-Ins"  screen, make sure everything you need to debug still works
* Strings and graphics should be correct, other functionality will be part of a different PR.


janetback (dismissed):
UA reviewed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
```diff
@@ -0,0 +1,68 @@
...
+    @Assisted private val deepLink: String?,
+    dispatcherProvider: DispatcherProvider,
+    private val qrCodeVerifier: QRCodeVerifier,
+    private val qrCodeUriParser: QRCodeUriParser
+) : CWAViewModel(dispatcherProvider) {
+
+    private val verifyResultData = MutableLiveData<QRCodeVerifyResult>()
+    val verifyResult: LiveData<QRCodeVerifyResult> = verifyResultData
+
+    init {
```
mtwalli, line 30: Will this still work in case the activity is open on this Screen? and when it is coming from Scanning fragment?
1- Case 1: there could be an uri coming from `newIntent` in main activity 
2- Case 2: when navigating to scan fragment and then back here again
is the view model created newly that is why is added in `init` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
```diff
@@ -0,0 +1,68 @@
...
+        deepLink?.let {
+            if (deepLink != savedState.get(SKEY_LAST_DEEPLINK)) {
+                Timber.i("New deeplink: %s", deepLink)
+                verifyUri(it)
+            } else {
+                Timber.d("Already consumed this deeplink: %s", deepLink)
+            }
+        }
+        savedState.set(SKEY_LAST_DEEPLINK, deepLink)
```
mtwalli, line 39: Do we need to save it? or it is better to clear it ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
d4rken: It worked in my tests :thinking:, we are further a long we will test this again too.

> 1- Case 1: there could be an uri coming from newIntent in main activity

Causes navigation via navgraph. Navargs use fragment arguments under the hood, fragment arguments can only be set on new fragments. So a new fragment is created, and this did create a new viewmodel in my tests as the fragment viewmodels are scoped to the fragment.
This would would become an issue if we scope the viewmodel to an activity.

> 2- Case 2: when navigating to scan fragment and then back here again
is the view model created newly that is why is added in init ?

It would depend on if the checkins fragment is destroyed in that time. If it is not destroyed, nothing happens because `init` is not called again. If it is destroyed (low memory / don't keep activies), but because it was still on the backstack, we will get a `SavedStateHandle` with `SKEY_LAST_DEEPLINK` and then know not to execute the deepLink action again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
d4rken: We need to save it in case the system restores the app after "don't keep activities state". The previously solution you wrote using "clear" on the arguments bundle works too, in most cases I guess, but this is not documented behavior, it just happens that on most devices the same bundle is reused, but theoretically the original bundle (not the one we called  "clear" on, could be returned to us.

BMItter:
code lgtm..
tested on pixel 2 & nexus 5 device (no issues found)
please resolve conflict & resolve comments when you are done.
Approve from my side (will reapprove after upcoming commit)

BMItr (dismissed):

d4rken:
Conflict was due to changes in protobuf, my solution for this PR was to comment it out, it will be overwritten in #2622

BMItr (approved):

mtwalli (approved):
LGTM

PR #2621, opened 2021-03-15
Protocol Buffer update for release 2.0.x at Mon Mar 15 14:42:34 UTC 2021

Corona-Warn-App-Technical-User:

PR #2622, opened 2021-03-15
QR Code parsing (DEV/EXPOSUREAPP-5841)

d4rken:
To be merged after #2619, replaces #2621.

Adjusts to the new protobuf format for signedTraceLocations.


mtwalli (dismissed):
LGTM

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
```diff
@@ -11,7 +11,7 @@
         tools:layout="@layout/fragment_confirm_check_in">
         <argument
             android:name="traceLocation"
-            app:argType="de.rki.coronawarnapp.ui.eventregistration.attendee.checkins.VerifiedTraceLocation" />
+            app:argType="de.rki.coronawarnapp.eventregistration.checkins.qrcode.VerifiedTraceLocation" />
```
chiljamgossow, line 14: Thanks for moving

chiljamgossow, line 14: Thanks for moving it

chiljamgossow (dismissed):

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
mtwalli: Why it is not in `ui` package though? it is parcelable entity for UI navigation purpose is not it?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/QRCodeVerifyResult.kt
```diff
@@ -2,18 +2,40 @@ package de.rki.coronawarnapp.eventregistration.checkins.qrcode
 
 import de.rki.coronawarnapp.server.protocols.internal.pt.TraceLocationOuterClass
 import de.rki.coronawarnapp.util.TimeAndDateExtensions.seconds
+import okio.ByteString.Companion.toByteString
 import org.joda.time.Instant
+import java.util.concurrent.TimeUnit
 
 data class QRCodeVerifyResult(
-    val singedTraceLocation: TraceLocationOuterClass.SignedTraceLocation
+    val singedTraceLocation: TraceLocationOuterClass.SignedTraceLocation,
```
LukasLechnerDev, line 10: typo: `singed` => `signed`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/VerifiedTraceLocation.kt
```diff
@@ -0,0 +1,18 @@
+package de.rki.coronawarnapp.eventregistration.checkins.qrcode
+
+import android.os.Parcelable
+import de.rki.coronawarnapp.server.protocols.internal.pt.TraceLocationOuterClass
+import kotlinx.parcelize.Parcelize
+import org.joda.time.Instant
+
+@Parcelize
+data class VerifiedTraceLocation(
```
LukasLechnerDev, line 9: We have `DefaultTraceLocation` that you can use here instead of creating a new model. It has the same properties.

LukasLechnerDev: We also need the `signature`, because we need to store it in the db after verifying the qr code at event creation / checkin

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/LocationQRCodeVerifier.kt
```diff
@@ -2,15 +2,14 @@ package de.rki.coronawarnapp.eventregistration.checkins.qrcode
 
 import de.rki.coronawarnapp.server.protocols.internal.pt.TraceLocationOuterClass
 import de.rki.coronawarnapp.util.security.SignatureValidation
-import okio.ByteString.Companion.decodeBase64
 import timber.log.Timber
 import javax.inject.Inject
 
-class DefaultQRCodeVerifier @Inject constructor(
+class LocationQRCodeVerifier @Inject constructor(
```
LukasLechnerDev, line 8: I would re-use this Verifier to verify the traceloation that is returned from the server. Does it make sense to rename this class to something like `TraceLocationVerifier`?

LukasLechnerDev: Wait... I don't need to use this class for my task. I can simply use `SignatureValidation`, so the comment above is obsolete.

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
d4rken: Kinda subjektive, but I'd have put it as a "core class" and not necessarily specific to UI, parcelable can also be passed to services. The whole package setup is currently a bit chaotic we will have to refactor some things once there are less moving parts and fewer chances for conflicts.

Based on @LukasLechnerDev  comments I've removed the whole class though.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/TraceLocation.kt
```diff
@@ -1,9 +1,11 @@
 package de.rki.coronawarnapp.eventregistration.events
 
+import android.os.Parcelable
 import de.rki.coronawarnapp.server.protocols.internal.pt.TraceLocationOuterClass
+import okio.ByteString
 import org.joda.time.Instant
 
-interface TraceLocation {
+interface TraceLocation : Parcelable {
```
mtwalli, line 8: I don't think `Parcelable` should be here. More important that we keep things related to the context where it is used. Parcelable can be used in a service ,but we don't have a service and we will not. So I would like to keep things where they only belong

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
```diff
@@ -11,7 +11,7 @@
         tools:layout="@layout/fragment_confirm_check_in">
         <argument
             android:name="traceLocation"
-            app:argType="de.rki.coronawarnapp.ui.eventregistration.attendee.checkins.VerifiedTraceLocation" />
+            app:argType="de.rki.coronawarnapp.eventregistration.events.TraceLocation" />
```
mtwalli, line 14: This it is used in UI I would keep it in UI` package `. this is the main reason why it is Praceable in the first place is not it ?

chiljamgossow (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/TraceLocation.kt
d4rken: > If you want to have DefaultTraceLocation as Parcelable, Please remove the interface I think it is not needed and don't really know why we create interface to be implemented by data class.

I don't think it makes sense two have two data classes just to have one of them parcelable, if both are the same like they currently are. If this becomes an issue we can split them. The interface/data class separation was created by @LukasLechnerDev, Interfaces were likely created during initial setup so everyone can start working on it. I will ask him.

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
d4rken: TraceLocation is not a "UI only" class, it's being used by the repository and I think also when verifying the server signed QR Code.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/TraceLocation.kt
LukasLechnerDev: Yes, I think we can remove the interface. We created it initially and I created the data class for it to stay compatible. 

Maybe we can create already data classes instead of interfaces next time at feature start?

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
LukasLechnerDev: I also see it as a "core" class.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/DefaultTraceLocation.kt
```diff
@@ -14,7 +18,7 @@ data class DefaultTraceLocation(
     override val startDate: Instant?,
     override val endDate: Instant?,
     override val defaultCheckInLengthInMinutes: Int?,
-    override val signature: String,
+    override val signature: ByteString,
```
LukasLechnerDev, line 21: `CheckIn` also contains the signature, so we should also change it to `ByteString` there. However, these refactoring are getting out of scope for this PR, so I can do this in my "Store binary representation of SignedTraceLocation in database" Task if you like?

Corona-Warn-App/src/main/res/navigation/trace_location_attendee_nav_graph.xml
mtwalli: Ok, I think discussion regarding that can go forever 😄. I think using it for the server does not require to have it `Pracelable` does it ?.  if you both agree why not. democracy wins

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/TraceLocation.kt
d4rken: I've refactored this and removed the interface.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/DefaultTraceLocation.kt
d4rken: > CheckIn also contains the signature, so we should also change it to ByteString there. 

Yes

> However, these refactoring are getting out of scope for this PR, so I can do this in my "Store binary representation of SignedTraceLocation in database" Task if you like?

Yes please.

mtwalli (dismissed):
thank you!

LukasLechnerDev (dismissed):
👍

mtwalli (approved):

LukasLechnerDev (approved):

chiljamgossow (approved):

PR #2624, opened 2021-03-15
Use TraceLocationType enums from protobuf (DEV)

LukasLechnerDev:
Before this PR, we had our own enum for TraceLocationTypes. 

This PR changes this to use the enum from the protobuf `TraceLocationOuterClass.TraceLocationType`.


mtwalli (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/entity/TraceLocationConverters.kt
```diff
@@ -1,13 +1,14 @@
...
 import androidx.room.TypeConverter
-import de.rki.coronawarnapp.eventregistration.events.TraceLocation
+import de.rki.coronawarnapp.server.protocols.internal.pt.TraceLocationOuterClass
 
 class TraceLocationConverters {
 
     @TypeConverter
-    fun toTraceLocationType(value: Int) = enumValues<TraceLocation.Type>().single { it.value == value }
+    fun toTraceLocationType(value: Int): TraceLocationOuterClass.TraceLocationType =
+        TraceLocationOuterClass.TraceLocationType.forNumber(value)!!
```
chiljamgossow, line 10: Would this crash here if the mapping is missing?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/events/DefaultTraceLocationKtTest.kt
```diff
@@ -23,7 +25,7 @@ internal class DefaultTraceLocationKtTest : BaseTest() {
         ).toTraceLocation() shouldBe DefaultTraceLocation(
             guid = "TestGuid",
             version = 1,
-            type = TraceLocation.Type.TEMPORARY_OTHER,
+            type = LOCATION_TYPE_PERMANENT_OTHER,
```
chiljamgossow, line 28: Before it was temporary. Was this change on purpose?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/events/DefaultTraceLocationKtTest.kt
```diff
@@ -13,7 +15,7 @@ internal class DefaultTraceLocationKtTest : BaseTest() {
         TraceLocationEntity(
             guid = "TestGuid",
             version = 1,
-            type = TraceLocation.Type.TEMPORARY_OTHER,
+            type = LOCATION_TYPE_PERMANENT_OTHER,
```
chiljamgossow, line 18: Before it was temporary. Was this change on purpose?

chiljamgossow (commented):
Looks good over all

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/entity/TraceLocationConverters.kt
LukasLechnerDev: Yes, it would crash if the user has a value in his database that we can't map to an enum value. 

However, this can only happen if we change the enum values for a new app version. In this case, we also would need to write a database migration in which we replace the invalid values with new values. 

I discussed this with @d4rken and he also recommended the crash (instead of e.g. returning a fallback value) as this situation is unexpected. This way we also notice this issue faster.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/events/DefaultTraceLocationKtTest.kt
LukasLechnerDev: No, not really on purpose 😅

However, since the only important thing is that the mapper maps the type to the same type, this test is still ok.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/events/DefaultTraceLocationKtTest.kt
LukasLechnerDev: See other comment.

chiljamgossow (approved):
LGTM

PR #2627, opened 2021-03-15
Update Generic Duration Picker (EXPOSUREAPP-5746)

jurajkusnier:
Update generic duration look and feel.
image


mtwalli (dismissed):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/durationpicker/DurationExtension.kt
```diff
@@ -3,18 +3,8 @@ package de.rki.coronawarnapp.ui.durationpicker
 import org.joda.time.Duration
 
 fun Duration.toContactDiaryFormat(): String {
```
LukasLechnerDev, line 5: Maybe add tests to verify if this extension still works as expected?

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1921,6 +1921,14 @@
     <string name="duration_dialog_ok_button">OK</string>
     <!-- NOTR -->
     <string name="duration_dialog_default_value">00:00</string>
+    <!-- XTXT: Duration hours text -->
+    <string name="duration_hours">"Stunden"</string>
+    <!-- XTXT: Duration minutes text -->
+    <string name="duration_minutes">"Minuten"</string>
+    <!-- XTXT: Duration Organize title -->
+    <string name="duration_typical_lenght_of_stay">"Typische Aufenthaltsdauer"</string>
```
LukasLechnerDev, line 1929: These last two Strings don't belong in this PR I guess.

LukasLechnerDev (dismissed):
LGTM! 2 small remarks

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/durationpicker/DurationExtension.kt
jurajkusnier: One more test is never a bad idea 😃

Corona-Warn-App/src/main/res/values-de/strings.xml
jurajkusnier: yeah, they don't need to be here

mtwalli (approved):

LukasLechnerDev (approved):

janetback (approved):
UA reviewed

PR #2628, opened 2021-03-16
Bring TracingExplanationDialog back (EXPOSUREAPP-5813)

jurajkusnier:
**Fixing**
Tracing explanation dialog got removed from the onboarding process
**Testing**
Reset app & go through onboarding. You should see the Tracing explanation dialog with new strings.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/main/CWASettings.kt
@@ -25,6 +25,10 @@ class CWASettings @Inject constructor(
         get() = prefs.getBoolean(PKEY_DEVICE_TIME_INCORRECT_ACK, false)
         set(value) = prefs.edit { putBoolean(PKEY_DEVICE_TIME_INCORRECT_ACK, value) }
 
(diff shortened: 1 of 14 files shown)


harambasicluka (commented):
image
looks good :)

@ralfgehrer should we keep the string changes in to not trigger a new translation? But this will cause merge problems with the translation delivery tomorrow...

harambasicluka:
I also tested an update to a newer version (`VERSION_MAJOR=2`) and it isn't shown again 👍

jurajkusnier:
> @ralfgehrer should we keep the string changes in to not trigger a new translation? But this will cause merge problems with the translation delivery tomorrow...

Everything in the PR is already translated and some texts were removed in the previous commits so we need to update string files anyway to display something in the popup.

ralfgehrer:
> @ralfgehrer should we keep the string changes in to not trigger a new translation? But this will cause merge problems with the translation delivery tomorrow...

Let's keep the changes here and request another translation upload once merged.

ralfgehrer (approved):
LGTM. Tested on Pixel (Andoid 10)

harambasicluka (approved):

PR #2630, opened 2021-03-16
Fix test result donor calculation (EXPOSUREAPP-5814)

mtwalli:
- Introduced new time specifically for test result donor calculation saved in RiskResultDetector
- Skip donation if the timestamps required for calculation is missing

### Testing
- Register a test and try donation from the menu

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
@@ -82,7 +82,7 @@ class RiskLevelChangeDetector @Inject constructor(
     private fun saveTestDonorRiskLevelAnalytics(
         newRiskState: RiskLevelResult
     ) {
-        // Save if not already set before for increased risk
+        // Save riskLevelTurnedRedTime if not already set before for high risk detection
         Timber.i("riskLevelTurnedRedTime:%s", testResultDonorSettings.riskLevelTurnedRedTime.value)
         if (testResultDonorSettings.riskLevelTurnedRedTime.value == null) {
             if (newRiskState.isIncreasedRisk) {
@@ -97,7 +97,7 @@ class RiskLevelChangeDetector @Inject constructor(
             }
         }
 
-        // Save most recent date high or low risks
+        // Save most recent date of high or low risks
         if (newRiskState.riskState in listOf(RiskState.INCREASED_RISK, RiskState.LOW_RISK)) {
             Timber.i(
                 "newRiskState:%s, mostRecentDateWithHighOrLowRiskLevel:%s",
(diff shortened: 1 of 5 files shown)


jurajkusnier (dismissed):
LGTM data in the test menu looks good

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
```diff
@@ -65,9 +67,53 @@ class RiskLevelChangeDetector @Inject constructor(
...
+
+        // Save most recent date of high or low risks
+        if (newRiskState.riskState in listOf(RiskState.INCREASED_RISK, RiskState.LOW_RISK)) {
+            Timber.i(
+                "newRiskState:%s, mostRecentDateWithHighOrLowRiskLevel:%s",
+                newRiskState.riskState,
+                newRiskState.calculatedAt
+            )
+            testResultDonorSettings.mostRecentDateWithHighOrLowRiskLevel.update {
+                newRiskState.calculatedAt
```
d4rken, line 108: This is the time when this result was calculated, which happens 5-6 times per day, but I think the value we want here is the day (e.g. 3 days ago) when we had an encounter.

For `riskLevelTurnedRedTime` (above) this works because it's only read and written once, the first time it happens.

But for `mostRecentDateWithHighOrLowRiskLevel` it will happen every day (hopefully), as `newRiskState.riskState` will be `FAILED/LOW/HIGH`.

We probably want this from `RiskLevelResult`:

```kotlin
    val lastRiskEncounterAt: Instant?
        get() = if (isIncreasedRisk) {
            aggregatedRiskResult?.mostRecentDateWithHighRisk
        } else {
            aggregatedRiskResult?.mostRecentDateWithLowRisk
        }
```

d4rken (changes requested):
Possible bug, can you confirm?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelChangeDetector.kt
mtwalli: which date do we rely on for  `the Home Screen Card ` ? I think this should be the one we use

d4rken: The homecards have multiple dates for calculation and encounters. I don't think the time this was calculated is interesting to analytics though, but rather (similar to the first value) it would be interesting for analytics when the user last had contact with someone that tested positive.

jurajkusnier: I didn't notice this, but I agree it makes sense to use the first value when the increased risk was calculated for analytics.

mtwalli: 👍🏼

d4rken (approved):

jurajkusnier (approved):

PR #2633, opened 2021-03-17
Split check ins midnight (EXPOSUREAPP-5706)

mtwalli:
- Split CheckIn across multiple days by midnight UTC time
- Add Unit tests for different scenarios

### Testing 
- Check the [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#split-checkin-by-midnight-utc) 
- Check test cases


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
```diff
@@ -0,0 +1,94 @@
...
+    val startTimeSeconds = checkInStart.seconds
+    val endTimeSeconds = checkInEnd?.seconds ?: 0L
+    val durationSeconds = endTimeSeconds - startTimeSeconds
+
+    Timber.i("durationSeconds=$durationSeconds")
+
+    val durationDays = max(1L, ceil(durationSeconds.toDouble() / DAY_IN_SECONDS).toLong())
+    Timber.i("durationDays=$durationDays")
+
+    return Array(durationDays.toInt()) { day ->
```
BMItter, line 26: (0 until durationDays).map{  } would be a little bit simpler

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
```diff
@@ -0,0 +1,94 @@
...
+            ),
+            checkInEnd = toInstant(
+                seconds = toMidnightUTC(startTimeSeconds) + daysInSeconds(day + 1)
+            )
+        )
+
+    else -> copy()
+}
+
+private fun toMidnightUTC(timestampSeconds: Long): Long =
```
BMItter, line 67: could be extension function.
also below for **daysInnSeconds**, **toInstant**, **isFirstDay** etc

BMItr (commented):
lgtm, two comments

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
mtwalli: For `isFirstDay` ,` isLastDay`, `toMidnightUTC`  it is nice to have them as extensions. For 
`toInstant` and `daysInSeconds` , it does not look simpler. I keep as it is  

```kotlin
// Extensions 
(day + 1).daysInSeconds()
(startTimeSeconds.toMidnightUTC() + daysInSeconds(day + 1)).toInstant()

// Normal functions
 daysInSeconds(day + 1)
toInstant(
  seconds = startTimeSeconds.toMidnightUTC() + daysInSeconds(day)
)
```

BMItr (dismissed):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
```diff
@@ -0,0 +1,94 @@
...
+import kotlin.math.max
+
+/**
+ * Splits a [CheckIn] by midnight UTC time into multiple [CheckIn]s across multiple days.
+ *
+ * @return [List] of [CheckIn]s
+ */
+fun CheckIn.splitByMidnightUTC(): List<CheckIn> {
+    val startTimeSeconds = checkInStart.seconds
+    val endTimeSeconds = checkInEnd?.seconds ?: 0L
```
chiljamgossow, line 18: How about an early return if the checkInEnd is null or before the checkInStart?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitterTest.kt
```diff
@@ -0,0 +1,247 @@
...
+
+        checkIn.splitByMidnightUTC().apply {
+            size shouldBe 1 // No splitting
+            get(0) shouldBe checkIn
+        }
+    }
+
+    @Test
+    fun `Scenario 4`() {
+        /*
```
chiljamgossow, line 89: Can you add a test for this scenario:      val checkIn = defaultCheckIn.copy(
            checkInStart = Instant.parse("2021-03-04T09:30:00Z"),
            checkInEnd = Instant.parse("2021-03-05T09:15:00Z")
        )
That fails for me. What happens when startTime < endTime?

chiljamgossow (changes requested):
Looks fine, can you add more tests ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
mtwalli: No,  I would like to keep that way.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitterTest.kt
mtwalli: what do you mean by `startTime < endTime` ? is not that case in all test cases?

chiljamgossow: Yes, I mean startTime > endTime, we should handle that possibility correctly.

Please test the example above, where only the time of the day is reversed, not the full date.

mtwalli: I expect that there is no splitting should happen since the duration still less than one day, but I will align with Max regarding those cases? my understanding that the check-ins are split across multiple duration days. It does mean that they have to split for different dates

chiljamgossow: @mlenkeit Can you confirm this?

mtwalli: @chiljamgossow  I already aligned with Max regarding that. pushing a fix soon and a test cases for that

BMItr (dismissed):

chiljamgossow (approved):

BMItr (approved):

PR #2637, opened 2021-03-17
Test registration date remains from deleted test after adding a new test (EXPOSUREAPP-4484)

jurajkusnier:
I wasn't able to reproduce the bug so this is just an attempt to fix it or mitigate the effects

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
@@ -40,8 +40,8 @@ class SubmissionRepository @Inject constructor(
     private val tracingSettings: TracingSettings
 ) {
     private val testResultReceivedDateFlowInternal =
-        MutableStateFlow((submissionSettings.initialTestResultReceivedAt ?: timeStamper.nowUTC).toDate())
-    val testResultReceivedDateFlow: Flow<Date> = testResultReceivedDateFlowInternal
+        MutableStateFlow(submissionSettings.initialTestResultReceivedAt?.toDate())
+    val testResultReceivedDateFlow: Flow<Date?> = testResultReceivedDateFlowInternal
 
     private val deviceUIStateFlowInternal =
         MutableStateFlow<NetworkRequestWrapper<DeviceUIState, Throwable>>(NetworkRequestWrapper.RequestIdle)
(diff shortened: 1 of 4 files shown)


d4rken:
Can you further elaborate on where you think the issue is and how this mitigates it?

d4rken:
Hm, that would mean the bug happens if the user registers a second test but the CWA has never left RAM (no low memory, no reboots) since the first test?

Then the bug would be with the second test never updating the timestamp though :thinking: 

I don't like that we make the date nullable further done the line in UI related classes. It's not good to allow this value to be nullable if the UI elements make no sense if it is null.

d4rken:
This doesn't fix the underlying cause: The second test didn't update the timestamp
It would hide the symptoms: The first removed test sets the date to null, the second never updates it, the test date would be displayed as "today" and most users wouldn't notice it unless they see that the date keeps showing "today" and the keep looking at it after 24h again.

I would prefer to not hide the symptoms as it's not a fix and will make it more difficult to spot and debug this issue, and eventually come up with an actual fix. In case this was somehow related to encrypted preferences, this might already be fixed in 1.15.x too.

jurajkusnier:
> Hm, that would mean the bug happens if the user registers a second test but the CWA has never left RAM (no low memory, no reboots) since the first test?
> 
> Then the bug would be with the second test never updating the timestamp though 🤔
> 
> I don't like that we make the date nullable further done the line in UI related classes. It's not good to allow this value to be nullable if the UI elements make no sense if it is null.

From the bug description, it could be possible that the date is just in memory as the user deleted the test just before entering another test and the test result came in a few hours.
Date was actually always nullable, was changed to not nullable just recently. Also clearing nullable dates doesn't seem right when it always must have some value.

jurajkusnier:
> This doesn't fix the underlying cause: The second test didn't update the timestamp
> It would hide the symptoms: The first removed test sets the date to null, the second never updates it, the test date would be displayed as "today" and most users wouldn't notice it unless they see that the date keeps showing "today" and the keep looking at it after 24h again.
> 
> I would prefer to not hide the symptoms as it's not a fix and will make it more difficult to spot and debug this issue, and eventually come up with an actual fix. In case this was somehow related to encrypted preferences, this might already be fixed in 1.15.x too.

Yes, it was another assumption that it could be related to encrypted preferences, but without logs, it's hard to predict where the issue could be.

harambasicluka:
As we don't exactly know where the problem is it won't go into 1.1.-0, if we are able to implement a reliable fix we'll ship it with 2.0.x.

PR #2638, opened 2021-03-17
If there is an `initialTestResultReceivedTimestamp`  it may be (EXPOSUREAPP-4484)

d4rken:
If a user removed a test prior to 1.11, they may have an old timestamp that should have been deleted, this tries to detect that and delete the timestamp.

If there is also a registration value, but no `devicePairingSuccessfulAt` timestamp, then we are being called from `asyncRegisterDeviceViaGUID` or `asyncRegisterDeviceViaTAN`, and it's a new test and `initialTestResultReceivedTimestamp` should be null.

Also see #2637

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
@@ -165,17 +165,17 @@ class SubmissionRepository @Inject constructor(
             deadmanNotificationScheduler.cancelScheduledWork()
         }
 
-        val initialTestResultReceivedTimestamp = submissionSettings.initialTestResultReceivedAt
-
         // https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4484
         // User removed a test before 1.11 where due to a bug the timestamp was not removed.
-        if (initialTestResultReceivedTimestamp != null &&
+        if (submissionSettings.initialTestResultReceivedAt != null &&
             submissionSettings.registrationToken.value != null &&
             submissionSettings.devicePairingSuccessfulAt == null
         ) {
             Timber.tag(TAG).w("User has stale initialTestResultReceivedAt, fixing EXPOSUREAPP-4484.")
             submissionSettings.initialTestResultReceivedAt = null
         }
+
+        val initialTestResultReceivedTimestamp = submissionSettings.initialTestResultReceivedAt
 
         if (initialTestResultReceivedTimestamp == null) {
             val currentTime = timeStamper.nowUTC
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
@@ -167,6 +167,16 @@ class SubmissionRepository @Inject constructor(
 
         val initialTestResultReceivedTimestamp = submissionSettings.initialTestResultReceivedAt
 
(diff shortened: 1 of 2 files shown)


d4rken:
Called from 3 places:

```kotlin
    private suspend fun fetchTestResult(registrationToken: String): DeviceUIState = try {
        val testResult = submissionService.asyncRequestTestResult(registrationToken)
        updateTestResult(testResult)
        deriveUiState(testResult)
    } catch (err: NoRegistrationTokenSetException) {
        DeviceUIState.UNPAIRED
    }
```

```kotlin
    suspend fun asyncRegisterDeviceViaGUID(guid: String): TestResult {
        analyticsKeySubmissionCollector.reset()
        val registrationData = submissionService.asyncRegisterDeviceViaGUID(guid)
        submissionSettings.registrationToken.update {
            registrationData.registrationToken
        }
        updateTestResult(registrationData.testResult)
        submissionSettings.devicePairingSuccessfulAt = timeStamper.nowUTC
        backgroundNoise.scheduleDummyPattern()
        analyticsKeySubmissionCollector.reportTestRegistered()
        return registrationData.testResult
    }
```

```kotlin
    suspend fun asyncRegisterDeviceViaTAN(tan: String) {
        analyticsKeySubmissionCollector.reset()
        val registrationData = submissionService.asyncRegisterDeviceViaTAN(tan)
        submissionSettings.registrationToken.update {
            registrationData.registrationToken
        }
        updateTestResult(registrationData.testResult)
        submissionSettings.devicePairingSuccessfulAt = timeStamper.nowUTC
        backgroundNoise.scheduleDummyPattern()
        analyticsKeySubmissionCollector.reportTestRegistered()
        analyticsKeySubmissionCollector.reportRegisteredWithTeleTAN()
    }
```

d4rken:
```kotlin
        if (
            // We have a timestamp, but is it the bug or a legitimate timestamp
            initialTestResultReceivedTimestamp != null && 
            // This get's set directly before we are being called from  asyncRegisterDeviceViaGUID or asyncRegisterDeviceViaTAN
            submissionSettings.registrationToken.value != null &&
            // Now we need to make sure that we are not being called as part of a fetchTestResult
            // and this value gets set after we are called
            submissionSettings.devicePairingSuccessfulAt == null
            // All of these values are nulled when a test is removed on 1.15
        ) {
            Timber.tag(TAG).w("User has stale initialTestResultReceivedAt, fixing EXPOSUREAPP-4484.")
            submissionSettings.initialTestResultReceivedAt = null
        }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
```diff
@@ -167,6 +167,16 @@ class SubmissionRepository @Inject constructor(
 
         val initialTestResultReceivedTimestamp = submissionSettings.initialTestResultReceivedAt
 
+        // https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4484
+        // User removed a test before 1.11 where due to a bug the timestamp was not removed.
+        if (initialTestResultReceivedTimestamp != null &&
+            submissionSettings.registrationToken.value != null &&
+            submissionSettings.devicePairingSuccessfulAt == null
+        ) {
+            Timber.tag(TAG).w("User has stale initialTestResultReceivedAt, fixing EXPOSUREAPP-4484.")
```
BMItter, line 176: don't know if this reference make sense ->4484  
but doesnt matter too much

BMItr (dismissed):
lgtm
ps: please fix tests

BMItr (approved):

ralfgehrer (approved):

PR #2641, opened 2021-03-18
Check-In risk level transmission (EXPOSUREAPP-5770)

mtwalli:
- Finalise check-ins submission steps and map to server check-ins
- Put all Check-in submission in order as per specs 
- Adjusting plausible deniability and (server request changes) will follow up in another PR
### Testing 
- Please check the [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#attendee-check-in-submission) and see if there anything missing 
- Register a test and submit your keys (should not be affected by this change)


d4rken (commented):
I like it :+1:

d4rken (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/derivetime/TimeIntervalDeriver.kt
```diff
@@ -16,11 +16,14 @@ private fun alignToInterval(timestamp: Long) =
  * Derive CheckIn start and end times
  * @param startTimestampInSeconds [Long] timestamp in seconds
  * @param endTimestampInSeconds [Long] timestamp in seconds
+ *
+ * @return [Pair] of [Long] including derived startTime and endTime in seconds
```
LukasLechnerDev, line 20: Not really related to this PR actually, but I think a data class with 2 properties as a return type is always better than a pair because the code then gets more self-explanatory and less error-prone.

mtwalli: what would be the difference? `Pair` is data class actually. Since it is from Kotlin, I assume that usability is well described

LukasLechnerDev: The main benefit is that in `CheckInsTransformer`the readability improves -  instead of: 

```
timesPair.first
timesPair.second
```
we would have something like
```
derivedTime.startTime
derivedTime.endTime
```

So its more self-explanatory and we also wouldn't really need the one line of documentation

LukasLechnerDev (dismissed):
LGTM! 

Submitted keys without noticeable side effects.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/derivetime/TimeIntervalDeriver.kt
mtwalli: You convinced me , addressed below

d4rken (approved):

LukasLechnerDev (approved):

PR #2644, opened 2021-03-18
CheckIn cards & interaction (EXPOSUREAPP-5410)

d4rken:
This PR aims to add:
* CheckIn list behavior
* Cards for active CheckIns + interaction
* Cards for old CheckIns + interaction

The list displays dummy data and the menus/clicks don't do anything yet, this will come in a different PR.

The card for old check-ins currently says `21.01.2021, 18:01 - 18:26`, without the `Uhr`, I will add this in a different PR after it is clear how we handle check-ins spanning multiple days.


AlexanderAlferov (approved):
LGTM 🚀 
Looks nice on Samsung S8:
screenshot-1616080455904

janetback (approved):
UA reviewed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
```diff
@@ -24,8 +31,28 @@ class CheckInsViewModel @AssistedInject constructor(
     private val qrCodeUriParser: QRCodeUriParser
 ) : CWAViewModel(dispatcherProvider) {
 
-    private val verifyResultData = MutableLiveData<TraceLocationVerifyResult>()
-    val verifyResult: LiveData<TraceLocationVerifyResult> = verifyResultData
+    val confirmationEvent = SingleLiveEvent<TraceLocationVerifyResult>()
+
+    val checkins = FAKE_CHECKIN_SOURCE
+        .map { checkins -> checkins.sortedBy { it.checkInEnd } }
```
ralfgehrer, line 37: Do we need 2nd level sort order - e.g. `checkInStart` to have a stable sorting for active events?

ralfgehrer:
It seems that the padding of the floating button on the right side slightly too big.

Screenshot_20210319-104127

Screenshot_20210319-104140

ralfgehrer (approved):
LGTM. Minor comments - also fine to be fixed with upcoming PRs.
Tested on Pixel (Android 10). Also Talkback works nicely IMHO.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
d4rken: Good point, by title would make sense, will add in follow up PRs.

PR #2647, opened 2021-03-18
Add tracelocation byterepresentation to TraceLocation database table (DEV)

LukasLechnerDev:
TechSpec says to store the byte representation of the TraceLocation into the [TraceLocation](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#database-table-for-tracelocations) database table.

This PR does exactly that.

We decided to store the byte representation as base64 encoded string in the database.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/entity/TraceLocationEntity.kt
```diff
@@ -18,8 +18,8 @@ data class TraceLocationEntity(
     @ColumnInfo(name = "startDate") val startDate: Instant?,
     @ColumnInfo(name = "endDate") val endDate: Instant?,
     @ColumnInfo(name = "defaultCheckInLengthInMinutes") val defaultCheckInLengthInMinutes: Int?,
+    @ColumnInfo(name = "byteRepresentation") val byteRepresentationBase64: String,
```
d4rken, line 21: Should the columnName then also be `byteRepresentationBase64` ?

LukasLechnerDev: Yes makes sense. However we are also not suffixing the base64 encoded values in the `CheckInEntity`. I think a PR in which we refactor the 2 entities would be best. I can do this after this one is merged, if that's fine for you?

chiljamgossow (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/entity/TraceLocationEntity.kt
d4rken: Sounds good.

d4rken (approved):

PR #2650, opened 2021-03-19
Adjust margins for lists with card like items (DEV)

d4rken:
Follow up PR to our previous UX/UI meeting.

We have lists of cards in 3 places, the margins on each where different.
* HomeFragment: 16,16,16,16
* DiaryOverview: 24,8,24,8
* DiaryPerson/Location: 24,16,24,16
* CheckIns: 24,16,24,16

I've adjusted the margin to match the home fragment, so when moving from home fragment to diary/checkins the transitions looks nicer because the card sizes match. We also get more estate for texts and titles, which were a bit crunched.

### DIFF
--- a/Corona-Warn-App/schemas/de.rki.coronawarnapp.eventregistration.storage.TraceLocationDatabase/1.json
+++ b/Corona-Warn-App/schemas/de.rki.coronawarnapp.eventregistration.storage.TraceLocationDatabase/1.json
@@ -2,11 +2,11 @@
   "formatVersion": 1,
   "database": {
     "version": 1,
-    "identityHash": "48d71b0a00d0c7fe01ffe05d5fecb512",
+    "identityHash": "3130771db80d789cd0ef50b8d24e0354",
     "entities": [
       {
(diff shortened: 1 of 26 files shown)


d4rken:
Screenshot from 2021-03-19 12-23-32
Screenshot from 2021-03-19 12-23-24
Screenshot from 2021-03-19 12-23-17
Screenshot from 2021-03-19 12-23-11
Screenshot from 2021-03-19 12-23-01

harambasicluka (approved):
Looks good! Tested on a OnePlus 8.

Is there a reason why you only considered this for lists? E.g. Contact Diary -> Add place: the input field or also indented. Could be part of another refactoring round.

chiljamgossow (approved):

d4rken:
> Is there a reason why you only considered this for lists? E.g. Contact Diary -> Add place: the input field or also indented. Could be part of another refactoring round.

Only the lists came up so far, when I find time, I can check the other margings too, but AFAICT, the add person/location margin is correct at `24dp`.

PR #2654, opened 2021-03-19
CheckIn deletion (EXPOSUREAPP-5410) 

d4rken:
* Wire up UI elements with delete confirmation and repository
* Add delete method to respository
* Refactore repository to offer blocking calls to the UI

### DIFF
--- a/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/eventregistration/storage/TraceLocationCheckInDaoTest.kt
+++ b/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/eventregistration/storage/TraceLocationCheckInDaoTest.kt
@@ -4,8 +4,10 @@ import androidx.room.Room
 import androidx.test.core.app.ApplicationProvider
 import de.rki.coronawarnapp.eventregistration.storage.CheckInDatabaseData.testCheckIn
 import de.rki.coronawarnapp.eventregistration.storage.CheckInDatabaseData.testCheckInWithoutCheckOutTime
+import io.kotest.assertions.throwables.shouldThrow
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNotBe
+import io.mockk.mockk
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.runBlocking
 import org.joda.time.Instant
@@ -101,5 +103,24 @@ class TraceLocationCheckInDaoTest : BaseTestInstrumentation() {
         checkInDao.deleteByIds(listOf(generatedId1))
         checkInDao.entryForId(generatedId1) shouldBe null
         checkInDao.entryForId(generatedId2) shouldNotBe null
+    }
+
+    @Test
+    fun traceLocationCheckInDaoUpdateById() = runBlocking {
+        val generatedId1 = checkInDao.insert(testCheckIn)
+
+        checkInDao.updateEntityById(generatedId1) {
+            it.copy(address = "test")
+        }
+        checkInDao.entryForId(generatedId1)!!.address shouldBe "test"
+    }
+
+    @Test
+    fun traceLocationCheckInDaoUpdateById_raceCondition1(): Unit = runBlocking {
+        shouldThrow<IllegalStateException> {
+            checkInDao.updateEntityById(123) {
+                mockk()
+            }
+        }
     }
 }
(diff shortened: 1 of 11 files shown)


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/eventregistration/storage/TraceLocationCheckInDaoTest.kt
```diff
@@ -82,4 +85,42 @@ class TraceLocationCheckInDaoTest : BaseTestInstrumentation() {
...
+        val generatedId1 = checkInDao.insert(testCheckIn)
+
+        checkInDao.updateEntityById(generatedId1) {
+            it.copy(address = "test")
+        }
+        checkInDao.entryForId(generatedId1)!!.address shouldBe "test"
+    }
+
+    @Test
+    fun traceLocationCheckInDaoUpdateById_raceCondition1(): Unit = runBlocking {
```
LukasLechnerDev, line 119: Why did you call the test raceCondition? Isn't it just asserting that an `IllegalStateException` is thrown when we want to  update a non-existing checkIn?

chris-cwa (approved):
Looks good!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
```diff
@@ -68,6 +76,22 @@ class CheckInsViewModel @AssistedInject constructor(
         savedState.set(SKEY_LAST_DEEPLINK, deepLink)
     }
 
+    fun onRemoveCheckInConfirmed(checkIn: CheckIn?) {
+        Timber.d("removeCheckin(checkIn=%s)", checkIn)
+        launch(scope = appScope) {
```
LukasLechnerDev, line 81: Because of the `withContext(NonCancellable)`, I think we don't need to use the `appScope` and can simply use the `viewModelScope`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
```diff
@@ -29,59 +27,25 @@ class CheckInRepository @Inject constructor(
...
-        }
+        checkInDao.insert(checkIn.toEntity())
     }
 
-    fun clear() {
-        appScope.launch {
-            checkInDao.deleteAll()
-        }
+    suspend fun updateCheckIn(checkInId: Long, update: (CheckIn) -> CheckIn) = withContext(NonCancellable) {
+        Timber.d("updateCheckIn(checkInId=%d, update=%s)", checkInId, update)
```
LukasLechnerDev, line 38: Aren't String templates nicer in Kotlin? 😀

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/eventregistration/storage/TraceLocationCheckInDaoTest.kt
d4rken: The name was from a previous test case that I adapted :grin:. But I think it still fits and would just keep it as is. Isn't that a racecondition? We try to update something that should exist, but when we try it, it actually doesn't.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
d4rken: If we print plain values like `checkInId`, yes, because there won't be a huge impact, but if we print complex parameters like data classes or callbacks in this case, using formatting args is better for performance, if logging is disabled, as they will not be evaluated `toString()`.

We could now put `checkInId` as template, and update as formatting arg, but.... would this be a blocker for you :slightly_smiling_face: ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsViewModel.kt
d4rken: I think there might still be a racecondition.
Maybe, I'll think about this a bit more and fix it in a follow up PR if necessary.

```kotlin
    launch(scope = appScope) {
            if (checkIn == null) { // <------ If we cancel the scope here, is `clear` still called? 
                checkInsRepository.clear()
            } else {
                checkInsRepository.deleteCheckIns(listOf(checkIn))
            }
     }
```

LukasLechnerDev: Oh, good point. I am pretty sure `clear()` wouldn't be called in the situation you described, because the suspend function `withContext()` would immediately throw a `CancellationException` and the "NonCancellable" code block wouldn't be executed. 

=> Leave it as it is :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
LukasLechnerDev: Thanks for the explanation. Then its fine for me as it is.

LukasLechnerDev (approved):
LGTM!

PR #2659, opened 2021-03-21
Check-in matching (EXPOSUREAPP-5694)

chiljamgossow:
Calculates time overlap of check-ins and trace time warnings

Tech spec https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#calculate-overlap-of-checkin-and-tracetimeintervalwarning

This is how the algorithm works:
1. split check-ins by midnight GMT (reason for this is that risk is calculated per day)
2. check for all check-ins and relevant warnings if there is a time overlap
3. keep the overlaps for further processing -> risk calculation follows in a different PR 

Deletes the guidHashBase64 from TraceLocationCheckInEntity because this db field's only purpose is facilitating check-in matching, but proved to be unhelpful. Replaced by an attribute in the CheckIn class.

cannot be tested yet with the test menu, due to missing download/check-ins
but has plenty unit tests

so please check the code


Corona-Warn-App/schemas/de.rki.coronawarnapp.eventregistration.storage.TraceLocationDatabase/1.json
```diff
@@ -116,7 +110,7 @@
       },
       {
         "tableName": "traceLocations",
-        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`guid` TEXT NOT NULL, `version` INTEGER NOT NULL, `type` INTEGER NOT NULL, `description` TEXT NOT NULL, `address` TEXT NOT NULL, `startDate` TEXT, `endDate` TEXT, `defaultCheckInLengthInMinutes` INTEGER, `byteRepresentation` TEXT NOT NULL, `signature` TEXT NOT NULL, PRIMARY KEY(`guid`))",
```
chiljamgossow, line 119: This deletion shouldn't be in here, was not intended. Will fix it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
```diff
@@ -14,12 +15,15 @@ import kotlin.math.ceil
...
 fun CheckIn.splitByMidnightUTC(): List<CheckIn> {
     Timber.d("Starting splitByMidnightUTC ...")
+    if (checkInStart.toLocalDate() == checkInEnd.toLocalDate()) return listOf(this)
     val startTimeSeconds = checkInStart.seconds
     val endTimeSeconds = checkInEnd.seconds
     val durationSecondsUTC = endTimeSeconds - startTimeSeconds.toMidnightUTC()
 
     Timber.i("durationSecondsUTC=$durationSecondsUTC")
 
+    if (durationSecondsUTC <= 0) return listOf(this)
```
d4rken, line 25: Can you add a unit test for this case?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
```diff
@@ -0,0 +1,75 @@
...
+): List<CheckInWarningOverlap> {
+    return warningPackage
+        .extractTraceTimeIntervalWarnings()
+        .flatMap { warning ->
+            checkIns
+                .mapNotNull {
+                    it.calculateOverlap(warning)
+                }
+        }
+}
```
d4rken, line 49: Please add some logging here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val dispatcherProvider: DispatcherProvider
+) {
+    suspend fun execute(): List<CheckInWarningOverlap> {
+        val checkIns = checkInsRepository.allCheckIns.firstOrNull() ?: return emptyList()
+
+        val warningPackages = traceTimeIntervalWarningRepository.allWarningPackages.firstOrNull() ?: return emptyList()
+
+        val splitCheckIns = checkIns.flatMap { it.splitByMidnightUTC() }
+
+        return launchMatching(splitCheckIns, warningPackages, dispatcherProvider.IO)
```
d4rken, line 23: Love the parallelization effort!

We might have to cap the parallel executions though.
Currently we launch one coroutine (possibly one thread due to `IO` dispatcher) per `warningPackage`, which may be 24*15, though capped at 64 by kotlin (afaik).  Parallel I/O (reading the packages) is usually the bottle neck on most devices, especially low end ones, so 64 parallel attempts to read the files may starve the system of I/O resources and could cause the device to lag during our background execution.

This is all theoretical though, we'd need to test this with an actual set of warning packages (@mlenkeit is something like that available?).

Maybe we should cap parallel execution at 2-4 for now? What do you think? Anything above 4 might have diminishing returns anyways :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
```diff
@@ -0,0 +1,75 @@
...
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toLocalDate
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.async
+import kotlinx.coroutines.withContext
+import org.joda.time.Duration
+import org.joda.time.Instant
+import kotlin.coroutines.CoroutineContext
+
+suspend fun launchMatching(
```
d4rken, line 15: I'd prefer a different name to prevent missunderstandings as this does not "launch" until `await` is called on it, devs unexperienced with coroutines may be caught off-guard.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
```diff
@@ -0,0 +1,75 @@
+package de.rki.coronawarnapp.presencetracing.risk
```
d4rken, line 1: What about moving these global functions into a class so they can be mocked better when writing unit tests for `CheckInWarningMatcher`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
```diff
@@ -0,0 +1,75 @@
...
+        List<CheckIn>,
+        TraceTimeIntervalWarningPackage
+    ) -> Deferred<List<CheckInWarningOverlap>> =
+        { list, warningPackage ->
+            async {
+                findMatches(list, warningPackage)
+            }
+        }
+
+    return warningPackages.map { warningPackage ->
```
d4rken, line 30: Please add a unit test that shows how this is expected to behave if there is an error during execution.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
```diff
@@ -0,0 +1,75 @@
...
+                    it.calculateOverlap(warning)
+                }
+        }
+}
+
+fun CheckIn.calculateOverlap(
+    warning: TraceWarning.TraceTimeIntervalWarning
+): CheckInWarningOverlap? {
+
+    if (warning.locationGuidHash != locationGuidHash) return null
```
d4rken, line 55: I'd love some logging for the case where there is a match, does not have to be right here, but somewhere?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class CheckInWarningMatcher @Inject constructor(
+    private val checkInsRepository: CheckInRepository,
+    private val traceTimeIntervalWarningRepository: TraceTimeIntervalWarningRepository,
+    private val dispatcherProvider: DispatcherProvider
+) {
+    suspend fun execute(): List<CheckInWarningOverlap> {
+        val checkIns = checkInsRepository.allCheckIns.firstOrNull() ?: return emptyList()
+
+        val warningPackages = traceTimeIntervalWarningRepository.allWarningPackages.firstOrNull() ?: return emptyList()
```
d4rken, line 19: Some logging for the early aborts please :smile:

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/risk/OverlapTest.kt
```diff
@@ -0,0 +1,257 @@
...
+import de.rki.coronawarnapp.eventregistration.checkins.CheckIn
+import de.rki.coronawarnapp.server.protocols.internal.pt.TraceWarning
+import de.rki.coronawarnapp.util.HashExtensions.toSHA256
+import io.kotest.matchers.shouldBe
+import okio.ByteString
+import org.joda.time.Duration
+import org.joda.time.Instant
+import org.junit.Test
+
+class OverlapTest {
```
d4rken, line 13: `: BaseTest()`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
```diff
@@ -0,0 +1,75 @@
...
+
+internal suspend fun findMatches(
+    checkIns: List<CheckIn>,
+    warningPackage: TraceTimeIntervalWarningPackage
+): List<CheckInWarningOverlap> {
+    return warningPackage
+        .extractTraceTimeIntervalWarnings()
+        .flatMap { warning ->
+            checkIns
+                .mapNotNull {
```
d4rken, line 45: The `mapNotNull` is a crucial part of the matching calculation, maybe a test for this?

d4rken (commented):
Nice! Some comments, no real blockers.

I think this empty file is a leftover from refactoring:
* `Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/checkins/DefaultCheckInsTransformerTest.kt`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/split/CheckInSplitter.kt
chiljamgossow: There was actually a bug here, thanks

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
chiljamgossow: Most likely, it would be a file reading issue. No idea what should happen there. @mlenkeit What should happen if a package cannot be unpacked/evaluated? Omit the package or fail the whole calculation?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
chiljamgossow: I will add logging in the risk calculation PR as the logic here changes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
chiljamgossow: I actually do not want them to be mocked in the CheckInWarningMatcherTest to have a test that covers the whole unit of work

d4rken: Should we make them part of that class? Why are they global functions in an extra file?

chiljamgossow: I love Kotlin for offering global functions 😃 I can easily write unit tests for all kinds of edge cases without having to mock anything. And I can call them easily from everywhere, e.g the test menu. I can of course move them into the same file.

d4rken: I agree that they are very useful, but we run the risk of littering the Android Studio auto completion with functions that are only relevant in specific circumstances.

Scoping functions so only those that are interesting are available at specific places in the code make it easier for new developers to work on something, find relevant functions.

It's may also have a negative impact on IDE performance if there many global functions, similar to overuse of unscoped extension functions.

It would not be a blocker for me if you feel strongly about this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
mlenkeit: Abort the calculation, ideally show a white card on the home screen.

chiljamgossow: I will fix it in the other PR

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInMatching.kt
chiljamgossow: I marked them a internal, only visible for testing (for now)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
```diff
@@ -0,0 +1,117 @@
...
+                }
+        }
+}
+
+@VisibleForTesting(otherwise = PRIVATE)
+internal fun CheckIn.calculateOverlap(
+    warning: TraceWarning.TraceTimeIntervalWarning
+): CheckInWarningOverlap? {
+
+    if (warning.locationGuidHash != locationGuidHash) return null
```
harambasicluka, line 97: The techspec mentions to return `0`, not null. So is this a wanted behaviour here?

harambasicluka: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#calculate-overlap-of-checkin-and-tracetimeintervalwarning

harambasicluka (dismissed):
LGTM 🚀 

Just left one question.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
chiljamgossow: Yes, nulls are skipped for further processing

d4rken (dismissed):
lgtm

harambasicluka (approved):

d4rken (approved):

PR #2660, opened 2021-03-21
Risk calculation (EXPOSUREAPP-5698)

chiljamgossow:
Can be tested in test menu with mock server (necessary for the risk calculation parameters in the config)
1. run matcher, wait for result
2. calculate risk

Tech spec https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#risk-calculation

How it works:
1. Result of matching is stored on DB table
2. For all entries, the time overlap is mapped to a normalised time by multiplying with a risk factor
3. Normalised time is summed up over date, then mapped to a risk state
4. Risk states per day and check-in are published in the RiskLevelStorage (for contact journal)
5. Risk states per day of exposure windows and presence tracing are merged (for contact journal)

builds on top of matching PR https://github.com/corona-warn-app/cwa-app-android/pull/2659


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
```diff
@@ -11,17 +11,25 @@ import javax.inject.Inject
 class CheckInWarningMatcher @Inject constructor(
     private val checkInsRepository: CheckInRepository,
     private val traceTimeIntervalWarningRepository: TraceTimeIntervalWarningRepository,
+    private val presenceTracingRiskRepository: PresenceTracingRiskRepository,
     private val dispatcherProvider: DispatcherProvider
 ) {
-    suspend fun execute(): List<CheckInWarningOverlap> {
-        val checkIns = checkInsRepository.allCheckIns.firstOrNull() ?: return emptyList()
+    suspend fun execute() {
```
d4rken, line 17: I only see this being called from a test, shouldn't this be called from somewhere?

d4rken: Which PR will add code that calls this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/Conversions.kt
```diff
@@ -0,0 +1,17 @@
+package de.rki.coronawarnapp.presencetracing.risk
```
d4rken, line 1: Please name the class more specific.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskCalculator.kt
```diff
@@ -0,0 +1,52 @@
+package de.rki.coronawarnapp.presencetracing.risk
+
+import javax.inject.Inject
+
+class PresenceTracingRiskCalculator @Inject constructor(
```
d4rken, line 5: Could be `@Reusable`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskMapper.kt
```diff
@@ -0,0 +1,63 @@
...
+        if (presenceTracingRiskCalculationParamContainer == null) {
+            presenceTracingRiskCalculationParamContainer =
+                configProvider.currentConfig.first().presenceTracing.riskCalculationParameters
+            Timber.d(presenceTracingRiskCalculationParamContainer.toString())
+        }
+        return presenceTracingRiskCalculationParamContainer
+    }
+}
+
+private fun RiskCalculationParametersOuterClass.Range.isInRange(value: Double): Boolean {
```
d4rken, line 57: We also have this in `DefaultRiskLevels`, can it be re-used?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskMapper.kt
```diff
@@ -0,0 +1,63 @@
...
+
+import de.rki.coronawarnapp.appconfig.AppConfigProvider
+import de.rki.coronawarnapp.appconfig.PresenceTracingRiskCalculationParamContainer
+import de.rki.coronawarnapp.risk.RiskState
+import de.rki.coronawarnapp.server.protocols.internal.v2.RiskCalculationParametersOuterClass
+import kotlinx.coroutines.flow.first
+import timber.log.Timber
+import javax.inject.Inject
+
+class PresenceTracingRiskMapper @Inject constructor(
```
d4rken, line 11: Could be `@Reusable`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskMapper.kt
```diff
@@ -0,0 +1,63 @@
...
+    private suspend fun getTransmissionRiskValueMapping() =
+        getRiskCalculationParameters()?.transmissionRiskValueMapping
+
+    private suspend fun getNormalizedTimePerDayToRiskLevelMapping() =
+        getRiskCalculationParameters()?.normalizedTimePerDayToRiskLevelMapping
+
+    private suspend fun getNormalizedTimePerCheckInToRiskLevelMapping() =
+        getRiskCalculationParameters()?.normalizedTimePerCheckInToRiskLevelMapping
+
+    private suspend fun getRiskCalculationParameters(): PresenceTracingRiskCalculationParamContainer? {
```
d4rken, line 47: Should we maybe pass the config as parameter, so we have control over which config is used? When the calculation is called as part of some other code which might also use an app config for other purposes, so that at one instance in time, all the operations use the same config and we are not at the mercy of the caching mechanism always providing the same config. Could be the source of some difficult to reproduce bugs.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
```diff
@@ -0,0 +1,118 @@
...
+import de.rki.coronawarnapp.util.TimeStamper
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.map
+import org.joda.time.Days
+import org.joda.time.Instant
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class PresenceTracingRiskRepository @Inject constructor(
```
d4rken, line 21: This calculates the risk when we subscribe to a flow, shouldn't we calculate the risk as part of a task and store the the tasks result and expose that as a flow? I'd have imagined this to be similar to or re-using the current `RiskLevelTask`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelResult.kt
```diff
@@ -49,8 +49,6 @@ interface RiskLevelResult {
             aggregatedRiskResult?.mostRecentDateWithLowRisk
         }
 
-    val traceLocationCheckInRiskStates: List<TraceLocationCheckInRisk>
```
d4rken, line 52: Check-In high risk can also cause a red card on the home-screen, as well as influence other parts of the app. All of these consume, the RiskLevelResult, don't we need something like that here?

d4rken (commented):
I'm not sure about the approach using `PresenceTracingRiskRepository` for calculation instead of making it similar to the task based approach we are already using for the other risk calculation. Please elaborate.

janetback (dismissed):
UA

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/eventregistration/ui/EventRegistrationTestFragmentViewModel.kt
```diff
@@ -3,28 +3,131 @@ package de.rki.coronawarnapp.test.eventregistration.ui
...
-    dispatcherProvider: DispatcherProvider,
-    private val checkInWarningMatcher: CheckInWarningMatcher
+    private val dispatcherProvider: DispatcherProvider,
+    private val presenceTracingRiskCalculator: PresenceTracingRiskCalculator
 ) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
 
-    val checkInOverlaps = MutableLiveData<List<CheckInWarningOverlap>>()
-    val checkInRiskPerDayList = MutableLiveData<List<CheckInRiskPerDay>>()
+    private val checkInWarningOverlaps = mutableListOf<CheckInWarningOverlap>()
+    val checkInOverlapsText = MutableLiveData<String>()
```
BMItter, line 27: don't expose mutableLiveData..

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
```diff
@@ -0,0 +1,118 @@
...
+    suspend fun deleteAll()
+
+    @Query("DELETE FROM TraceTimeIntervalMatchEntity WHERE endTimeMillis < :endTimeMillis")
+    suspend fun deleteOlderThan(endTimeMillis: Long)
+
+    @Insert
+    suspend fun insert(entities: List<TraceTimeIntervalMatchEntity>)
+}
+
+@Entity
```
BMItter, line 88: Use @ColumnInfo to specify the name

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/TraceLocationCheckInRisk.kt
```diff
@@ -4,6 +4,6 @@ import org.joda.time.LocalDate
 
 interface TraceLocationCheckInRisk {
     val checkInId: Long
-    val localDate: LocalDate
+    val localDateUtc: LocalDate
```
BMItter, line 7: A LocalDate doesn't have timezone information, if I'm not mistaken..

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/CheckInWarningMatcher.kt
```diff
@@ -11,17 +11,25 @@ import javax.inject.Inject
...
+
+        if (checkIns.isNullOrEmpty() || warningPackages.isNullOrEmpty()) {
+            presenceTracingRiskRepository.deleteAllMatches()
+            return
+        }
 
         val splitCheckIns = checkIns.flatMap { it.splitByMidnightUTC() }
 
-        return launchMatching(splitCheckIns, warningPackages, dispatcherProvider.IO)
+        val matches = launchMatching(splitCheckIns, warningPackages, dispatcherProvider.IO)
```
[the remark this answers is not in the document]
chiljamgossow: @mlenkeit What should happen in case of failure?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskCalculator.kt
chiljamgossow: That would also mean the same config is used. Do you think we should use the same config?

d4rken: > That would also mean the same config is used. 

Why :thinking: ?

> Do you think we should use the same config?

We should use the same config "per operation", i.e. if we are currently doing a risk calculation, we should use the same config for all parts of the operation. Once the calculation is done, the next calculation may use a new config, but again it should not have the chance to change during calculation within the same operation.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskMapper.kt
chiljamgossow: Only if we want to reuse the same config, do we?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskMapper.kt
chiljamgossow: You mean add a setter for the config? The config is only requested once per instance (unless there is none)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
chiljamgossow: leaving it as it is for now as discussed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/RiskLevelResult.kt
chiljamgossow: Yes, will come with a new PR

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskCalculator.kt
chiljamgossow: The config is requested once per instance, so the whole calculation uses the same config.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/TraceLocationCheckInRisk.kt
chiljamgossow: The local data does not have a timezone anymore, that's correct. The conversion to local date uses a timezone, so I wanted to make it clear that the date here is derived using utc, as I would otherwise expect a local date to be derived using the local timezone.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
chiljamgossow: What would I needed for? The attribute names are fine to be used as column names.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/eventregistration/ui/EventRegistrationTestFragmentViewModel.kt
chiljamgossow: What is your concern? I cannot post values if is is not mutable

BMItter: ofc, you can call postValue on a MutableLiveData outside of the VM.
Just expose it as LiveData

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
BMItter: If one just wants to adjust the name of a variable, you don't necessarily need to update the schema of the db or all handwritten sql queries. Same for r8 obfuscation

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskCalculator.kt
d4rken: You are right, it would use same one, and upwards towards `CheckInWarningMatcher` no extra config is used.

d4rken (commented):
lgtm, would approve after the ColumnInfo comment is addressed:

https://github.com/corona-warn-app/cwa-app-android/pull/2660#discussion_r603257565

d4rken (approved):
lgtm

BMItr (approved):

PR #2668, opened 2021-03-22
Run ktlint for whole project (including test classes)

LukasLechnerDev:
I removed 2 ktlint checks from our pipeline: 
- ktlint_device_release_check (`:Corona-Warn-App:ktlintDeviceReleaseCheck`)
- ktlint_device_for_testers_release_check (`:Corona-Warn-App:ktlintDeviceForTestersReleaseCheck`)

Instead, only this ktlint will now run: 
- ktlint_check (`:Corona-Warn-App:ktlintCheck`)

This now also checks if our test classes are formatted correctly. 

❗️ There is a problem: The 2 deleted checks are still shown (see below) and the pipeline seems to be stuck. Does anybody have an idea why this is the case? 🤔


harambasicluka:
>❗️ There is a problem: The 2 deleted checks are still shown (see below) and the pipeline seems to be stuck. Does anybody have an idea why this is the case? 🤔

I think @d4rken has to adapt it in the settings.

d4rken:
We can adjust the restrictions once this merged, if we adjust them now, all other PRs are blocked from merging.

LukasLechnerDev:
@d4rken I talked with @AlexanderAlferov : We can only set these restrictions globally, right? So all releases need to have these config.yml changes. So does it make sense to make this PR against 1.15.2 after it got released and then merge 1.15.2 into 2.0.x ?

d4rken:
> We can only set these restrictions globally, right? 

The restrictions are enforced via regex on the branch names.

> So does it make sense to make this PR against 1.15.2 after it got released and then merge 1.15.2 into 2.0.x ?

I'd say that depends on how many changes we have to make on 1.15.x to get the adjusted check to pass, i.e. do we have to reformat a lot?
If we have to reformat a lot, keep it against `2.0.x`, for a few (if any) hotfix PRs against `1.15.x` I can override the restriction manually.
If there is not a lot of formatting necessary, you can target `1.15.x`.

d4rken (approved):

mtwalli (approved):
LGTM, Checked it locally works for both Test and App

PR #2671, opened 2021-03-22
Pass more information the confirmation screen to allow CheckIn creation (EXPOSUREAPP-5410)

d4rken:
Refactoring so that we pass a `VerifiedTraceLocation` to the confirmation screen's viewmodel. In a follow up PR, @kolyaopahle  then has all the information necessary to create and add a `CheckIn` to the repository.

+Some random fixes to the repo.

You should be able to long the `QR-CODE SCANNEN` button then confirm the test QR code and see a check-in in the list.


kolyaopahle (dismissed):
beautiful as always, will make my life a lot easier ❤️

mtwalli (dismissed):

Issue #2658, opened 2021-03-20
Contact Journal: Keyboard not visible when adding multiple entries in a Row


Paul-Lukas:
## Avoid duplicates
* [ x ] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [ x ] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ x ] Bug is not already reported in another issue

## Technical details

- Device name: Motorola one Vision / Motorola one Macro
- Android version: 10
- App version: 1.14.3

## Describe the bug

When adding multiple entries in a row the keyboard for the name field won't show up automatically in every entry after the first one, but the cursor will. Because of the name field being selected, as shown by the blinking cursor. 
Also, you can't just tap on the name field, you have to first select a different field.

## Steps to reproduce the issue

1. Open CWA and tap Journal
2. If using Journal for first time, then tap "OPEN JOURNAL"
3. Tap date today, tap + PERSON
4. Note how the Keyboard shows up
5. Type "Test" for example
6. Tap Save
7. Tap + PERSON
8. Note how the Keyboard won't show up but the Cursor in the Name field blinks.
9. Try tapping the Name field and notice how it won't change anything

## Expected behaviour

The keyboard just showing up, similar to the first time

---
Internal Tracking ID: [EXPOSUREAPP-5914](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5914)
Internal Tracking ID: [EXPOSUREAPP-5869](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5869) related issue.

DerVogel2020:
I discovered a similar bug in 1.14.3 when editing existing persons or locations but wanted to wait for 1.15 and first check whether it has been already fixed before reporting it.

MikeMcC399:
@Paul-Lukas 
I can't reproduce this on Samsung Galaxy or on Google Pixel so I'm guessing it is device specific.

heinezen:
@DerVogel2020 Can you tell us what device you are using?

----
Corona-Warn-App Open Source Team

heinezen:
Hey @Paul-Lukas ,

Thanks for the hint. We have created a ticket for this issue in our internal Jira system. The development team will take a look at your issue.

----
Corona-Warn-App Open Source Team

[...]

PR #2674, opened 2021-03-23
Fix keyboard issue when editing contact journal entries (EXPOSUREAPP-5869)

ralfgehrer:
Also addresses: EXPOSUREAPP-5749, EXPOSUREAPP-5914

Tested on Pixel (Android 10) and Samsung A3 (Android 7.0)

**Steps:**
- Create a person in contact journal
- Edit Name of that person
- go back to "Edit People" screen
- Try to edit name again of the same person
- keyboard should appear on first attribute (Name)

**Expected result:**
- Keybord should appear and cursor should be at the end of the  name field entry of a person  (same for location) so one can continue writing


MikeMcC399:
I successfully tested this fix on:
- Pixel 3a Android 11 emulator
- Samsung Galaxy A5 (2017) Android 8.0.0

With 1.14.3 on the Pixel device the keyboard failed to display the second time.
On the A5 with Android 8 the keyboard did not even appear the first time.

Both these issues are no longer occurring with this fix. 👍🏻

This corresponds to issue https://github.com/corona-warn-app/cwa-app-android/issues/2658#issuecomment-803650687.

harambasicluka (commented):
Works as expected. Tested on a OnePlus 8/Android 11.

But I fear that this will cause some trouble with Google Talkback. Will test it on a Pixel 3. -> @axelherbstreith could you also have a look?

harambasicluka (changes requested):
I think we need to check this with our accesibility colleagues, if i go to the screen I only get an announcement for the keyboard but nothing else. IMHO we should get an announcement for the screen than for the focused input field and then for the keyboard.

I currently have some trouble with my Pixel 3, so I also tested it on my OnePlus.

axelherbstreith:
@harambasicluka @ralfgehrer tested it on my Samsung Galaxy S8. For me, the screen announcement is 100% identical to version without the bug fix. 

The issue that the screen itself is not announced is due to the fact that the focus changes to the text box "Name" immediately after opening the the screen. Therefore Talkback stops announcing the screen and starts announcing the text box itself. I personally would clarify this with the accessibility colleagues since as far as I can tell, this is not directly a bug.

axelherbstreith (approved):

harambasicluka (approved):
@axelherbstreith thanks for testing it also on an older version. So the problem wasn't introduced here 👍

PR #2680, opened 2021-03-23
Manual checkout, Check-In navigation (EXPOSUREAPP-5410)

d4rken:
* Add navigation for editing events (cc/ @kolyaopahle )
* Add navigation for onboarding (cc/ @chris-cwa )
* Manual checkout, journal creation will be added after @BMItter's PR
* Adjusted list sorting to techspec

For testing tap on-going and finished check-ins, you should navigate to the event confirmation/edit screen.
On-going check-ins pressing the manual check-out button should work.
You can long press the SCAN button to create test check-ins.
Deletion should also work.


kolyaopahle (approved):
lgtm 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/confirm/ConfirmCheckInFragment.kt
```diff
@@ -24,7 +25,8 @@ class ConfirmCheckInFragment : Fragment(R.layout.fragment_confirm_check_in), Aut
         factoryProducer = { viewModelFactory },
         constructorCall = { factory, savedState ->
             factory as ConfirmCheckInViewModel.Factory
-            factory.create(navArgs.verifiedTraceLocation)
+            val editId = if (navArgs.editCheckInId == 0L) null else navArgs.editCheckInId
```
mtwalli, line 28: should `editCheckInId ` be nullable ? if `0` represents a nullable value

mtwalli (approved):
LGTM, just one question

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/confirm/ConfirmCheckInFragment.kt
d4rken: NavArguments didn't allow primitives like `long` to be `null`, the option was greyed out.

PR #2688, opened 2021-03-24
Fix risk card detail screenshots and device time (EXPOSUREAPP-5970)

ralfgehrer:
- The device time is now fixed to 10:00 on all screens.
- On the risk detail card, with low risk and no exposure, the days since installation shall be visible.
- On all risk cards, the time of the last successful calculation is now before the device time (at 9:55).

TracingDetailsFragment_tracing_low_risk

TODO:

- [x] remove the firebase task from circle-ci config after a successful test run

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -628,7 +628,6 @@ workflows:
   check_buildtype_device:
     jobs:
       - detekt
-      - firebase_screenshots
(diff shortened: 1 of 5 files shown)


mtwalli (approved):
LGTM

harambasicluka (approved):

PR #2689, opened 2021-03-25
QR code Generation Component (EXPOSUREAPP-5961)

axelherbstreith:
Small PR that creates a component for the QR code generation which can be reused in the different screens 

To see the generation in action, go to: Test Menu -> Event Registration -> QR code Generation -> QR code


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/QrCodeGenerator.kt
```diff
@@ -0,0 +1,65 @@
...
+import timber.log.Timber
+import javax.inject.Inject
+
+@Reusable
+class QrCodeGenerator @Inject constructor(
+    private val appConfigProvider: AppConfigProvider,
+    @AppContext private val context: Context,
+) {
+
+    private val errorMessage = SingleLiveEvent<String>()
```
mtwalli, line 23: Since this is not  attached to any fragment and ideally should be used from ViewModel , maybe you change how you report errors such as throw an exception if the bitmap could not be generated.

```kotlin
@Throws(ExceptionTypeHere)
suspend fun createQRCode(input: String, size: Int = 1000){
   // Same code without try catch block , so the caller  of this function can handle the exception
}

axelherbstreith: Valid point, will change it

mtwalli (approved):

LukasLechnerDev (approved):
LGTM

PR #2690, opened 2021-03-25
Implement Error Handling for Event Creation (EXPOSUREAPP-5864)

jurajkusnier:
**Testing**

1. Go to `Test Menu > Event Registration > Events > Start Create Event Flow`
2. Create Event
3. Check if the event was created ( in `Test Menu > Show Stored Events`) or an error message was displayed


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/ApiRequestState.kt
```diff
@@ -0,0 +1,5 @@
+package de.rki.coronawarnapp.ui.eventregistration.organizer.create
+
+enum class ApiRequestState {
```
LukasLechnerDev, line 3: What do you think about using a sealed class hierarchy here instead? Then we could add e.g. the exception to the FAILED and the uiState to SUCCESS and don't need additional LiveData properties to send this data to the Fragment.

jurajkusnier: yeah, but the problem is that error is displayed only as a message so I'm using `SingleLiveEvent` not `MutableLiveData`. Also, the state of the network request is updated from another thread so in case I would want to implement it as part of `uiState` I would need to keep it as a separate variable out of `uiState` anyway to prevent accidental rewriting, so I'm not sure if another approach would save us anything.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateViewModel.kt
```diff
@@ -46,30 +57,55 @@ class TraceLocationCreateViewModel @AssistedInject constructor(
...
+                    }
+                    TraceLocationUIType.EVENT -> {
+                        description?.trim()?.length in 1..100 &&
+                            address?.trim()?.length in 0..100 &&
+                            begin != null &&
+                            end != null &&
+                            end?.isAfter(begin) == true
+                    }
+                } && (ignoreRequestState || requestState.value != ApiRequestState.STARTED)
+            )
```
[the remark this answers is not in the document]
d4rken: You can keep the states seperate, but this looks a bit fragile:


`} && (ignoreRequestState || requestState.value != ApiRequestState.STARTED)`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/ApiRequestState.kt
d4rken: Keeping it separate is also okay.
Currently state management here looks a bit chaotic, could we make it nicer somehow?
For threadsafe updates from multiple sources you could look into our `HotDataFlow` class.

d4rken (commented):
Nice! Some thoughts:

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateViewModelTest.kt
```diff
@@ -18,10 +22,17 @@ import testhelpers.extensions.observeForTesting
...
+    @BeforeEach
+    fun setup() {
+        MockKAnnotations.init(this)
+    }
+
     @ParameterizedTest
     @MethodSource("provideArguments")
     fun `send should not be enabled for empty form`(category: TraceLocationCategory) {
-        val viewModel = TraceLocationCreateViewModel(category = category)
+        val viewModel = TraceLocationCreateViewModel(category = category, traceLocationCreator = traceLocationCreator)
```
LukasLechnerDev, line 35: In other tests, we usually have a `createInstance()` function that creates the class we want to test and then call it from every test.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateViewModelTest.kt
```diff
@@ -130,7 +145,8 @@ internal class TraceLocationCreateViewModelTest : BaseTest() {
 
     @Test
     fun `send should not be enabled when end is before begin and category is EVENT`() {
-        val viewModel = TraceLocationCreateViewModel(category = categoryEvent)
+        val viewModel =
+            TraceLocationCreateViewModel(category = categoryEvent, traceLocationCreator = traceLocationCreator)
 
         viewModel.address = "Address"
         viewModel.description = "Description"
```
LukasLechnerDev, line 152: Maybe add some more tests to assert the right ViewModel State after `send()` is executed and the server responds with success / error ?

jurajkusnier: add more tests

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateViewModelTest.kt
jurajkusnier: updated

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateViewModel.kt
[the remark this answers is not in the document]
jurajkusnier: I refactored the states so I have API response as `SingleLiveEvent` and UI state as `LiveData` object and I moved loading state as part of UI state. Let me know if it's more clear now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/ApiRequestState.kt
jurajkusnier: Code is updated. I have API response as `SingleLiveEvent` and UI state as `LiveData` object and I moved loading state as part of UI state. Let me know if it's more clear now.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInsTransformerTest.kt
```diff
@@ -6,8 +6,10 @@ import de.rki.coronawarnapp.appconfig.ConfigData
 import de.rki.coronawarnapp.appconfig.PresenceTracingConfigContainer
 import de.rki.coronawarnapp.appconfig.PresenceTracingSubmissionParamContainer
 import de.rki.coronawarnapp.server.protocols.internal.pt.TraceLocationOuterClass
-import de.rki.coronawarnapp.server.protocols.internal.v2.PresenceTracingParametersOuterClass.PresenceTracingSubmissionParameters.AerosoleDecayFunctionLinear
-import de.rki.coronawarnapp.server.protocols.internal.v2.PresenceTracingParametersOuterClass.PresenceTracingSubmissionParameters.DurationFilter
+import de.rki.coronawarnapp.server.protocols.internal.v2
+    .PresenceTracingParametersOuterClass.PresenceTracingSubmissionParameters.DurationFilter
+import de.rki.coronawarnapp.server.protocols.internal.v2
+    .PresenceTracingParametersOuterClass.PresenceTracingSubmissionParameters.AerosoleDecayFunctionLinear
```
d4rken, line 12: Is this enforced by detekt or klint? I think import statements exceeding the line width is okay.

Can we? Should we? change the linter config here? /cc @LukasLechnerDev

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateViewModel.kt
```diff
@@ -114,6 +158,24 @@ class TraceLocationCreateViewModel @AssistedInject constructor(
         }
     }
 
+    private class UpdateDelegateWithDefaultValue<T>(defaultValue: T) :
```
d4rken, line 161: The `UpdateDelegate` patterns do not seem easy to read or debug, but it's not a blocker for me here. Maybe we can come up with an even better solution to then also use in other screens :thinking:

d4rken (approved):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInsTransformerTest.kt
jurajkusnier: I guess `:ktlintFormat` changed it because of the exceeded max line length

LukasLechnerDev: @d4rken I am not really an expert in this regard. Will look it up once I have a bit more time ...

LukasLechnerDev (approved):
👍

janetback (approved):
UA reivewed.

PR #2692, opened 2021-03-25
Event Checkin and Event Edit Fragments (EXPOSUREAPP-5743/5953)

kolyaopahle:
### Description
This PR contains the checkin and edit fragments for the event registration feature

PR #2694, opened 2021-03-26
Automatic checkout for check-ins (EXPOSUREAPP-5410)

d4rken:
* Adds automatic check-out creation for check-ins.
    * Uses alarm manager to schedule an alarm for the next check-out that is due. Alarm triggers a worker that processes the check-in and reschedules the alarm for the next check-out.
    * Each alarm execution (i.e. check-out) we also check for overdue check-outs and also process them to clean up.
    * The `CheckOutHandler` observes the database and will automatically adjust (or cancel) alarms if check-out times change.
* I refactored the notification system because this PR introduces a new channel for presence tracing related notifications.
    * We have two "Notification Helpers" now, the previous `NotificationHelper` is now `GeneralNotifications` and we have a new `TraceLocationNotifications`.
    * Each represents a distinct notification channel. They handle channel setup and offer a `fun newBaseBuilder(): NotificationCompat.Builder` that provides sane defaults.
    * There are a bit of chaos in `NotificationHelper` due to using two methods to show notifications and providing inconsistent defaults.
        * All notifications are privat by default now. The deadman notification is public and the automatic check-out notification is public. "Public" here means that it shows the specific notification text on the lock screen.
        * As part of the defaults we provide a non-descriptive CoronaWarnApp notification, e.g "hey maybe open the CWA".
* This PR adds the permission `RECEIVE_BOOT_COMPLETED`, because we need to restore the check-out alarm on reboots. We also listen for `intent.action.MY_PACKAGE_REPLACED` after which we may also need to restore our alarms.
* The Check-Ins list now refreshes every 5 seconds, so any automatic check-out countdown is visibility decreasing the time, and increasing the duration, without leaving the screen.

### Testing hints
* If you switch to the mock server environment, you can long press the scan button to add check-ins which expire after 9min.
    * By altering the `checkInEnd` parameter in the `ConfirmCheckInViewModel` you can further decrease the time until automatic checkout, you can make it only a few seconds long. 
* You can view scheduled alarms via ADB: `adb shell dumpsys alarm | grep rki`
* Filtering the log for `AutoCheckOut` may be interesting


Corona-Warn-App/src/main/res/values-de/event_registration_strings.xml
```diff
@@ -132,4 +132,12 @@
     <!-- XTXT: Text for trace location creation card -->
     <string name="create_trace_location_card_button">"QR-Code erstellen"</string>
 
+    <!-- XHED: Title of the notification channel for event registration related notifications. -->
+    <string name="tracelocation_notification_channel_title">"Check-ins"</string>
+    <!-- XTXT: Description of the notification channel for event registration related notifications. -->
+    <string name="tracelocation_notification_channel_description">"Benachrichtungen zu ihren Check-ins, z.B. automatischer Check-out."</string>
```
janetback, line 138: ihren müsste groß geschrieben werden ->Ihren Check-Ins.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
```diff
@@ -28,6 +28,11 @@ class CheckInRepository @Inject constructor(
         .allEntries()
         .map { list -> list.map { it.toCheckIn() } }
 
+    suspend fun getCheckInById(checkInId: Long): CheckIn? {
```
kolyaopahle, line 31: i also implemented this method in my checkin pr here: https://github.com/corona-warn-app/cwa-app-android/pull/2692/files#diff-c66309f5c643d2c824d912080bb056dadca4875c0347335c7885dfd7c5e139b6R53

we should probably agree on one or the other concept

kolyaopahle (dismissed):
LGTM 👍  one small comment, also why are we using work manager and alarm manager cant work manager handle all the tasks?

jurajkusnier (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
d4rken: Should looking up an ID throw an exception? :thinking:

d4rken:
> also why are we using work manager and alarm manager cant work manager handle all the tasks?

I would have prefered workers too... but...
Specs said we should perform automatic checkout as accurate as possible, this isn't possible via WorkManager, is it?
Worker tasks can also only run once every 15min, while automatic checkouts could happen more often.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
kolyaopahle: looking up no, for getting something by id that does not exist i feel it could be valid, but im not saying you should change your implementation to mine, we should just agree on one 😃

d4rken:
> makes perfect sense just wanted to make sure i understand the reason :)

Didn't mean to sound snarky, :heart:.

kolyaopahle (dismissed):

d4rken:
To fix this:
https://stackoverflow.com/questions/61463006/java-lang-illegalstateexception-cancel-call-cannot-happen-without-a-mayberun

I had to upgrade lifecycle libraries to next stable `2.3.1` (`androidx.lifecycle:lifecycle-*`)

kolyaopahle (approved):

jurajkusnier (approved):

mtwalli (approved):

PR #2695, opened 2021-03-26
Filter check-ins based on TRV (EXPOSUREAPP-5992)

mtwalli:
- Filter check-ins that its TRV is 0.0 for not found 
- Update Tests to cover that
- Check steps 5, 6 from the specs :https://github.com/corona-warn-app/cwa-app-tech-spec/blob/2b1f0fcfe25e74b82949231dcaec88caf368972c/docs/spec/event-registration-client.md#attendee-check-in-submission


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInsTransformer.kt
```diff
@@ -61,16 +62,38 @@ class CheckInsTransformer @Inject constructor(
...
         }
     }
 
-    private fun CheckIn.toOuterCheckIn(
+    private suspend fun CheckIn.toOuterCheckIn(
         transmissionVector: TransmissionRiskVector
-    ): CheckInOuterClass.CheckIn {
+    ): CheckInOuterClass.CheckIn? {
+        val trvMappings: List<TransmissionRiskValueMapping> =
+            appConfigProvider.getAppConfig()
```
d4rken, line 76: For performance reasons, and to prevent the config from changing during calculation, please re-use the retrieved one from line 42.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInsTransformer.kt
```diff
@@ -61,16 +62,38 @@ class CheckInsTransformer @Inject constructor(
...
+                .presenceTracing
+                .riskCalculationParameters
+                .transmissionRiskValueMapping
+
+        val transmissionRiskLevel = determineRiskTransmission(
+            timeStamper.nowUTC,
```
d4rken, line 82: Should we maybe get the timestamp once, like the config, and then use it for every calculation, so that it doesn't change during calculation?

d4rken (commented):
Looks good overall, some concerns.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInsTransformer.kt
mtwalli: `timeStamper.nowUTC`  is used only here

d4rken: But "here" is in `toOuterCheckIn` which is called from `mapNotNull` which is called from `flatMap`, so this is called many times, and depending on how fast the whole operation is, will very likely deliver different time values.

Using a timestamp that changes during calculation on the same dataset might at be problematic? At best difficult to compare results, at worst unexpected results if this happens around 23:59:59?

d4rken (approved):
lgtm

chris-cwa (approved):

PR #2700, opened 2021-03-27
Replace android.widget.Button with just Button (COMMUNITY)

fynngodau:
This small PR is intended to reduce the diff to CCTG by implementing a change that increases the codebase's compatibility with Android 5.

I noticed that Android 5 does not render background tint correctly if buttons are added to a layout in the app by using `android.widget.Button` instead of just `Button`. I have no clue why this happens but since adjusting the layout declarations accordingly doesn't have any drawbacks, I felt you might be able to merge this.

#### Screenshots for demonstration (Error report screen, CCTG, Android 5)

| Before | After |
|---|---|
| Screenshot_1616849752 | Screenshot_1616857963 |

Thanks in advance! ❤️

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/bugreporting_debuglog_fragment.xml
+++ b/Corona-Warn-App/src/main/res/layout/bugreporting_debuglog_fragment.xml
@@ -164,7 +164,7 @@
 
         </androidx.constraintlayout.widget.ConstraintLayout>
 
-        <android.widget.Button
+        <Button
             android:id="@+id/toggle_send_error_log"
             style="@style/buttonPrimary"
             android:layout_width="match_parent"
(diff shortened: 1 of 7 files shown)


mtwalli (approved):
Thank you for the PR!. the answer is [here](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:appcompat/appcompat/src/main/java/androidx/appcompat/app/AppCompatViewInflater.java;l=73?q=AppCompatView&ss=androidx%2Fplatform%2Fframeworks%2Fsupport). When `Button` is used in the layout the inflater will replace it with `AppCompatButton` which is supporting tinting on old Android versions while `android.widget.Button` won't be replaced which causes the issue you mentioned

d4rken (approved):

d4rken:
Thanks Fynn

fynngodau:
@mtwalli Thanks for the link, we didn't see the connection :slightly_smiling_face:

Issue #2701, opened 2021-03-28
Malformed ripple effect in contact journal


fynngodau:
## Describe the bug

The (i) button next to the Note field in both People as well as Places view in contact journal has a malformed ripple effect. The effect should be circular like the (i) icon, but it actually is elliptic.

## Steps to reproduce the issue

Open contact journal, open any day, add a place, hold (i) icon. The following ripple effect is shown:

![Screenshot_1616949710](https://user-images.githubusercontent.com/16943720/112759881-46a65f00-8ff5-11eb-987c-44bcc5d87821.png)

## Expected behaviour

The ripple effect should be the same shape as the (i) icon.

## Technical details

- Device name: emulator and others
- Android version: 6, 10
- App version: 1.15.1

---
Internal Tracking ID: [EXPOSUREAPP-6002](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6002)

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-6002](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6002)

d4rken:
Fynn's PR was merged into 2.1, so fixed "soon" :grin:, thanks!

MikeMcC399:
@dsarkar 
Do you want to add the [Fix 2.1 label](https://github.com/corona-warn-app/cwa-app-android/labels/Fix%202.1) to this issue?

heinezen:
@MikeMcC399 Done

PR #2702, opened 2021-03-28
Fix ripple effect for i button in circumstances text field (COMMUNITY) (#2701, EXPOSUREAPP-6002)

fynngodau:
This PR addresses the problem described in #2701.

Additionally, I removed some padding to the end (right) of the button because that felt better to me. If you don't like that I can modify the PR to add the removed padding again.

| state | before | after |
|:--|---|---|
| idle | Screenshot_1616950651 | Screenshot_1616950790 |
| ripple shown (holding) | ripple shown before | Screenshot_1616950827

---
Internal Tracking ID: [EXPOSUREAPP-6002](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6002)


fynngodau:
@harambasicluka No worries, I saw that you all are being very focused at the moment – and the 2.0 event registration features are coming together really well, you can do it! Best of luck 😊

harambasicluka (approved):

BMItr (approved):
tested on oneplus 8.
Ripple effect is better now 👌 
I think there are other elements than 'space' especially for constraint layouts. but it won't hurt

PR #2705, opened 2021-03-29
Protocol Buffer update for release 2.0.x at Mon Mar 29 07:25:06 UTC 2021

Corona-Warn-App-Technical-User:
Lukas: 
The following needed to be done to get the build running again: 
- Remove Signature Validation
- Replace SignedTraceLocation with QrCodePayload
- Deactivate some not-anymore-needed tests
- Lots of other adaptions


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/QRCodeUriParser.kt
```diff
@@ -15,7 +15,7 @@ class QRCodeUriParser @Inject constructor() {
      * https://e.coronawarn.app/c1/SIGNED_TRACE_LOCATION_BASE32
      * HTTPS://E.CORONAWARN.APP/C1/SIGNED_TRACE_LOCATION_BASE32
      */
-    fun getSignedTraceLocation(maybeUri: String): ByteString? = URI.create(maybeUri).run {
+    fun getQrCodePayload(maybeUri: String): ByteString? = URI.create(maybeUri).run {
```
BMItter, line 18: What should happen if create throws an exception?

BMItter: sorry, I saw too late that you just renamed it..

BMItr (approved):
nothing to complain.
lgtm 👌

d4rken (approved):

PR #2713, opened 2021-03-29
Camera permission card (EXPOSUREAPP-5786)

mtwalli:
- Camera permission card in Check-Ins Screen
- Created settings for this specifically, using system permission API did not provide away to know if the user has permanently
   denied the permission . Additionally it does not differentiate between initial permission state and after user denial.
- Refactored CheckInsViewModel and CheckInFragment logic into smaller code chunks 

### Testing 
- Run the App and play with the permissions , Card should be only visible incase user has denied the permission permanently.
- Test from Submission Flow and from Check-In  flow where Camera permission is requested

### TODO
- [x] Add some tests


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
```diff
@@ -42,11 +48,15 @@ class CheckInsFragment : Fragment(R.layout.trace_location_attendee_checkins_frag
         factoryProducer = { viewModelFactory },
         constructorCall = { factory, savedState ->
             factory as CheckInsViewModel.Factory
-            factory.create(savedState = savedState, deepLink = navArgs.uri)
+            factory.create(
```
d4rken, line 51: Why did you refactor this?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
```diff
@@ -57,33 +67,77 @@ class CheckInsFragment : Fragment(R.layout.trace_location_attendee_checkins_frag
...
+
+            is CheckInEvent.OpenDeviceSettings -> openDeviceSettings()
+        }
+    }
+
+    private fun updateViews(items: List<CheckInsItem>) {
+        checkInsAdapter.update(items)
+        binding.apply {
+            val index = items.indexOfFirst { it is CameraPermissionVH.Item }
+            swipeExcludedPositions.replaceAll(listOf(index))
```
d4rken, line 123: This doesn't feel very modular :frowning:.
Can't we do something like `SwipeExtension.getMovementFlags(...) { viewHolder isSwipable }`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
```diff
@@ -57,33 +67,77 @@ class CheckInsFragment : Fragment(R.layout.trace_location_attendee_checkins_frag
...
         }
+    }
 
+    private fun openDeviceSettings() {
+        startActivity(
+            Intent(
+                Settings.ACTION_APPLICATION_DETAILS_SETTINGS,
+                Uri.parse("package:" + BuildConfig.APPLICATION_ID)
+            )
+        )
```
d4rken, line 137: Please catch the ActivityNotFoundException here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
```diff
@@ -0,0 +1,25 @@
...
+    @AppContext private val context: Context
+) {
+    private val prefs by lazy {
+        context.getSharedPreferences("camera.permission.settings", Context.MODE_PRIVATE)
+    }
+
+    val isCameraDeniedPermanently = prefs.createFlowPreference(
+        key = PREFS_KEY_CAMERA_PERMISSION_DENIED,
+        defaultValue = false
+    )
```
d4rken, line 20: When is this set from `true` to `false`?

Can't we solve this without an extra preference? Just by checking `ActivityCompat.shouldShowRequestPermissionRationale`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
```diff
@@ -0,0 +1,25 @@
...
+import de.rki.coronawarnapp.util.preferences.createFlowPreference
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class CameraPermissionSettings @Inject constructor(
+    @AppContext private val context: Context
+) {
+    private val prefs by lazy {
+        context.getSharedPreferences("camera.permission.settings", Context.MODE_PRIVATE)
```
d4rken, line 14: Should we use a "permission" specific setting, there may be more permission related settings in the future :thinking: 
Maybe we should just call it "CameraSettings"?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
mtwalli: Do you expect that any App won't have a settings screen in Android 🤔? If yes what do you want to show to the user when he clicks on the card and nothing happens?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
mtwalli: Why do you think it is not modular. Is not the adapter itself doing the same? I don't like that we have to implement many interfaces for the data item and the holders just for that . Most of the screens does have to exclude swiping items , your suggestion will require more work for evey consumer to implement those interfaces one for the item which is OK and currently happening and one more for the holder.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
mtwalli: What refactoring ? You mean the formatting?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
mtwalli: Can you please elaborate more ? The class is called CameraSettings and the string is called camera.permission.settings? What should be exactly the change?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
mtwalli: Not with a combination with calling requestPermission in this screen which will trigger the system dialog earlier. Additionally this function returns false when user never asked for a permission and after denial . It is not helping in this case. If you have an idea let me know?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
mtwalli: I don't thin that Camera settings is suitable here. We don't have a camera feature to have a settings for it like instagram for example. We just doing this for permission and I think reflect what it does

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
mtwalli: It is not set from false again. In this case the user should have been done this from the device settings and I don't know any way to connect both. Still the user has actually denied it permanently before and as far as i know we can not show system the dialog anymore can we? . Everything will be reset only when user clears the data including the permissions

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
d4rken: Yes, seemed very random :grin:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
d4rken: > Do you expect that any App won't have a settings screen in Android thinking? 

Yes, I sadly have come to the conclusion that any activity you launch, that is outside of your app, will eventually cause an `ActivityNotFoundException` on some device.
It will be rare, but we shouldn't crash the app. Showing the exception in a toast would be fine for me.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
d4rken: This isn't a blocker for me. Up to you.
I prefer list items to be self-sufficient and not requiring top-levels item specific logic.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
d4rken: > Still the user has actually denied it permanently before and as far as i know we can not show system the dialog anymore can we?

We can, after the user went to the app settings and changed it, that's the idea behind showing the camera permission card, isn't it?

>  Everything will be reset only when user clears the data including the permissions.

We need to check if we have camera permission or if `shouldShowRequestPermissionRationale` returns true,  and then reset the `isCameraDeniedPermanently` to false, would this work? :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/permission/CameraPermissionSettings.kt
d4rken: Just thinking out loud whether this needs a single preference. The file "camera.permission.settings" would be a "not nice" name for anything, whether we can give this any re-use potential as "CameraSettings" or "PermissionSettings" :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/scan/ScanCheckInQrCodeFragment.kt
```diff
@@ -95,9 +95,11 @@ class ScanCheckInQrCodeFragment :
         ) {
             if (shouldShowRequestPermissionRationale(Manifest.permission.CAMERA)) {
                 showCameraPermissionRationaleDialog()
+                viewModel.setCameraDeniedPermanently(false)
```
d4rken, line 98: When going to system app settings and changing it to "ask every time" or "allow while app is open", after returning to the CWA, the camera permission card is still shown until I leave/reload the screen.

This code triggers on the scan QR Code fragment, but the user returns to the `CheckInsFragment`, we need to re-check on resume there.

d4rken (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/scan/ScanCheckInQrCodeFragment.kt
mtwalli: I don't think the issue is here nor with using `shouldShowRequestPermissionRationale ` in `CheckInsFragment `
Please check 4d1581e9da62dfba5354f49f81e009e4c13ffd18

d4rken (dismissed):

chiljamgossow (dismissed):

d4rken (approved):

chiljamgossow (approved):

janetback (approved):
UA reviewed

janetback (commented):
UA reviewed

PR #2726, opened 2021-03-31
Fix event registration organizer flow (DEV)

jurajkusnier:
**Fix:** After QR code is created and users end up in the QR codes list screen back navigation should redirect them to the home screen.


LukasLechnerDev (approved):
Thanks for the quick fix!

ralfgehrer (approved):

PR #2727, opened 2021-03-31
QR Code Poster (EXPOSUREAPP-5962)

mtwalli:
- Create location poster PDF
- Enable printing , Sharing
- Delete testing screens, Please use the actual screens 

### Testing
- Test against mock server
- Create location -> Choose Print option from overflow menu
- Create location -> QR Code details screen -> Print button
- Check poster and try printing and sharing 

### TODO
- [x] Finalise positioning values from server
- Text Box font - not provided yet
- Create poster on the fly (New ticket 🎫 probably or even after MVP)

After merging https://github.com/corona-warn-app/cwa-app-android/pull/2737
- [x] Use Location Url to generate QR Code 
- [x] Extend testability for TraceLocation


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/eventregistration/ui/EventRegistrationTestFragment.kt
```diff
@@ -36,29 +35,6 @@ class EventRegistrationTestFragment : Fragment(R.layout.fragment_test_eventregis
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
 
-        with(binding) {
```
LukasLechnerDev, line 39: Why did you remove all the TestMenu Screens? Do you see no benefits in still having them for e.g. trying something out in a quick way?

LukasLechnerDev: If we remove them, we also need to remove the "Events" cards.

mtwalli: Yes all removed. I kept what is needed for testing such as 
1- CheckIn matching 
2- Last locations cards

This was temporarily for testing now we have all in place ,please use the actual ones  additionally some screens were causing a conflict in the generated NavDirections

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/PosterTemplateProvider.kt
```diff
@@ -0,0 +1,64 @@
...
+import java.io.FileOutputStream
+import javax.inject.Inject
+import kotlin.math.roundToInt
+
+class PosterTemplateProvider @Inject constructor(
+    private val posterTemplateServer: QrCodePosterTemplateServer,
+    @AppContext private val context: Context
+) {
+    @Suppress("BlockingMethodInNonBlockingContext")
+    suspend fun template(): Template {
```
LukasLechnerDev, line 21: I think we should run this suspend function `withContext(Dispatchers.IO)`. I know, it is currently already called from IO, but it's best practice for suspend functions to be callable from the main thread without blocking.

mtwalli: this is a utility class ,so it is the responsibility of the called to do that , which is already happening

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterViewModel.kt
```diff
@@ -0,0 +1,119 @@
...
+                append(traceLocation.description)
+                appendLine()
+                append(traceLocation.address)
+            }
+            posterLiveData.postValue(
+                Poster(qrCode, template, textInfo)
+            )
+        } catch (e: Exception) {
+            Timber.d(e, "Generating poster failed")
+            posterLiveData.postValue(Poster())
```
LukasLechnerDev, line 98: It's a bit confusing to post an empty poster in the error case. Maybe use an `error` Livedata property here as we have it in other `ViewModel`s ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterFragment.kt
```diff
@@ -0,0 +1,156 @@
...
+        viewModel.sharingIntent.observe(viewLifecycleOwner) {
+            onShareIntent(it)
+        }
+    }
+
+    private fun QrCodePosterFragmentBinding.bindPoster(poster: Poster) {
+        Timber.d("poster=$poster")
+        progressBar.hide()
+
+        val template = poster.template ?: return // Exit early
```
LukasLechnerDev, line 68: Should we show the user that something went wrong?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterViewModel.kt
mtwalli: What is the issue with that? the Poster is empty and you can see one line below that the error is reported already. this is just to give a hint to the UI to stop the progress

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterFragment.kt
mtwalli: We do

mtwalli: Please test it ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterViewModel.kt
```diff
@@ -0,0 +1,119 @@
...
+     * as a sharing [FileSharing.ShareIntentProvider]
+     */
+    @Suppress("BlockingMethodInNonBlockingContext")
+    fun createPDF(view: View) = launch(context = dispatcher.IO) {
+        try {
+            val weakViewRef = WeakReference(view) // Accessing view in background thread
+            val directory = File(view.context.cacheDir, "poster").apply { if (!exists()) mkdirs() }
+            val file = File(directory, "cwa-qr-code.pdf")
+
+            val weakView = weakViewRef.get() ?: return@launch // View is not existing anymore
```
LukasLechnerDev, line 55: Do you think the view could not exist anymore because the user has left the screen? 

What about calling `ensureActive()` instead to check if the viewModelScope wasn't cancelled already?

axelherbstreith:
Tested it on different screen size (4.5, 5.7 and 6.3"). Looks really impressive! 💯 

image

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterViewModel.kt
mtwalli: It is not about the ViewModel ,the view model survives in configuration changes. So it could be that the fragment view is destroyed ,but not the ViewModel. this one point, the other point is the some work is being done in another thread and user could decide to leave the screen and the view will destroyed too. to avoid such cases I hold a weak reference to the view just in case

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/PosterTemplateProvider.kt
LukasLechnerDev: I am still the opinion, that its best practise to use `withContext(Dispatchers.IO)` here, as you e.g. can read [here](https://elizarov.medium.com/blocking-threads-suspending-coroutines-d33e11bf4761) - Section "Suspending Convention".

This convention shifts the responsibility about the Thread which is used from the caller to the suspend function itself.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/poster/QrCodePosterViewModel.kt
LukasLechnerDev: The issue is that it is not consistent with most of the other `ViewModels` in our App, in which we use `errorEvent` properties instead of returning empty data for error cases. Its not a blocker for me, so you can leave it as it is if you want.

mtwalli: I think we discussed this before, error reposting in the app is `e.report`

LukasLechnerDev:
Looks very good on my Pixel 3A and Samsung S20+, both Android 11. 

Great work 🚀

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/PosterTemplateProvider.kt
mtwalli: Addressed

LukasLechnerDev (dismissed):

axelherbstreith (dismissed):

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1968,4 +1968,10 @@
     <string name="trace_location_attendee_camera_card_button">Einstellungen öffnen</string>
     <!--  YMSG: Trace location onboarding image description-->
     <string name="trace_location_onboarding_content_description">Drei Personen an einem Stehtisch, zwei von ihnen schauen auf ihr Smartphone.</string>
+    <!-- XMIT: Trace location poster print -->
+    <string name="trace_location_organiser_poster_print">"Drucken"</string>
+    <!-- XMIT: Trace location poster share -->
+    <string name="trace_location_organiser_poster_share">"Share"</string>
```
janetback, line 1974: Please change to "Teilen". Thanks.

janetback (approved):
UA approved.

axelherbstreith (approved):

PR #2707, opened 2021-03-29
Warning package download & scheduling (EXPOSUREAPP-5695,EXPOSUREAPP-5696)

chiljamgossow:
This PR adds the download of warnings for tracelocations, wires up the download/calculation logic to be regularly executed. Some additional refactoring and and polishing too.

* Adds a worker to regularly launch a `Task` that downloads `TraceWarningPackage`'s and processes them if they have not yet been processed.
* The worker calls the `TraceWarningPackageSyncTool`
	* It will clean up inconsistent data states and revoke data if necessary.
    * It compares: `Our Check Ins` <> `What The Server Has` <> `What We Have Already Processed` and download any missing data.
    * New data is `unprocessed` and thus be available to the `CheckInWarningMatcher`.
* Unprocessed packages are given to the `CheckInWarningMatcher`
	* It will compare them against local `CheckIn`s  and possible generate a high risk state by adding data to `PresenceTracingRiskRepository`.
	* Processed packages are marked, deleted and will not be processed again.

This PR contains #2739.

### Testing hints
* The test menu has entries to execute the task which would trigger both download and matching.
* Filter log for `Warning` and `TraceWarning`.

PR #2728, opened 2021-04-01
Risk state (EXPOSUREAPP-5697)

chiljamgossow:
Most changes are renaming only in order to differentiate between the risk results.

There are now two sets of risk results, one for exposure windows (ew) and one for presence tracing (pt). They are all available in the RiskLevelStorage interface.

The risk calculation for ew remains untouched. It is still the basis for all analytics and survey logic.

The two sets are combined to CombinedEwPtRisk, to be used for the home tracing card and details. 
Changes in this combined risk should trigger the risk level change notification (see RiskLevelChangeDetector).

The home card will always show the highest risk that was calculated, that means if there is one high risk check-in or encounter, it will turn red. Calculation failed will only show if both calculations failed.

How to test: check whether the old risk calculation with exposure windows still works as before

Once the warning downloads are available, the risk from check-ins should also show up.

Missing: all unit tests -> follow up PR


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
```diff
@@ -51,37 +61,70 @@ class PresenceTracingRiskRepository @Inject constructor(
             presenceTracingRiskCalculator.calculateAggregatedRiskPerDay(it)
         }
 
-    suspend fun replaceAllMatches(list: List<CheckInWarningOverlap>) {
-        deleteAllMatches()
-        addAll(list)
+    internal suspend fun reportSuccessfulCalculation(list: List<CheckInWarningOverlap>) {
+        traceTimeIntervalMatchDao.insert(list.map { it.toEntity() })
+        val fifteenDaysAgo = timeStamper.nowUTC.minus(Days.days(15).toStandardDuration()).toLocalDateUtc()
```
BMItter, line 66: (fifteenDaysAgo) create a fun for this, that you dont have to repeat it elsewhere/below

BMItr (dismissed):
lgtm 👌 
one optional comment

BMItr (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/PresenceTracingRiskRepository.kt
```diff
@@ -51,37 +64,76 @@ class PresenceTracingRiskRepository @Inject constructor(
...
+    internal suspend fun deleteStaleData() {
+        traceTimeIntervalMatchDao.deleteOlderThan(fifteenDaysAgo.millis)
+        riskLevelResultDao.deleteOlderThan(fifteenDaysAgo.millis)
     }
 
-    suspend fun deleteStaleMatches() {
-        val endTime = timeStamper.nowUTC.minus(Days.days(15).toStandardDuration())
-        traceTimeIntervalMatchDao.deleteOlderThan(endTime.millis)
+    internal suspend fun markPackageProcessed(warningPackageId: String) {
+        // TODO
```
jurajkusnier, line 85: when/where this will be implemented?

chiljamgossow: When @d4rken has finished https://github.com/corona-warn-app/cwa-app-android/pull/2707 due to a missing db field

jurajkusnier (approved):

vaubaehn:
I was thinking about error notifications, if either EW calculations or PT calculations for any reason (and there are some...) permanently fail in the future...
Band-aid to get aware of problems and to start debugging: https://github.com/corona-warn-app/cwa-wishlist/issues/497
Possible Long-term solution: https://github.com/corona-warn-app/cwa-wishlist/issues/497#issuecomment-831509949 // https://github.com/corona-warn-app/cwa-app-android/pull/2995#discussion_r622277201

PR #2737, opened 2021-04-04
Trace Location URL and ID (EXPOSUREAPP-6137, EXPOSUREAPP-6182)

mtwalli:
### New
- Generate Trace Location Url [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/6e45edc4df5a8ced091a179bcd40da9bcc162c85/docs/spec/event-registration-client.md#organizer-qr-code-generation)
- Calculate  Trace Location Id  [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/6e45edc4df5a8ced091a179bcd40da9bcc162c85/docs/spec/event-registration-client.md#calculate-tracelocation-id)

### Change
- Updated deep link intent filter to work with the new url requirements 
- Updated QrCode Validation to use `qrCodeDescriptors` from configs and support Base64 and Base32 encoding validation
- Updated Unit Tests and clean them up 

### Testing 
- Test against mock server 
- Create  Trace location 
- Go to test menu -> Event Registration screen
- Scroll to the card `Last Organiser Location`
- You should see location, locationUrl and locationId data displayed for the last created Location 
- If you click on the link the App should respond to and open to ConfirmCheckIn 
- After confirming a CheckIn is added to the list 
- Now go back again to the test menu of the event registration , a card should be displayed for the `Last Attendee Location` (ChecKIn) with the same data above , since Location is the same of organiser and attendee , the data should be exactly the same except the `id` which is Table specific. Data should varies in case you scanned another location which will come next by https://github.com/corona-warn-app/cwa-app-android/pull/2727 .

Please note that url  validation supports both Base32 and Base64 ,but the not the creation which supports only Base64


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -181,7 +181,8 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
     private fun navigateByIntentUri(intent: Intent?) {
         val uri = intent?.data ?: return
         Timber.i("Uri:$uri")
-        navController.navigate(CheckInsFragment.createCheckInUri(uri.toString()))
+        val rootUri = uri.toString().replace(".app/?", ".app?")
```
mtwalli, line 184: The actual uri does not include `/` in `.app/?`, but the system adds it when it is scanned from Google Lens.
The link still valid ,but it won't match any `regexDescriptors`  from `AppConfig` and aborts the CheckIn process 🙅🏼

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/TraceLocation.kt
```diff
@@ -21,12 +23,45 @@ data class TraceLocation(
...
+        AUTHORITY.plus(base64Url)
+    }
+
+    /**
+     *  Returns a byte sequence that serves as an identifier for the trace location.
+     *  The ID is the byte representation of SHA-256 hash.
+     */
+    @IgnoredOnParcel
+    val locationId: ByteString by lazy {
+        val cwaDomain = CWA_GUID.toByteArray()
```
LukasLechnerDev, line 46: Rename `cwaDomain` => `cwaGuid`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/confirm/ConfirmCheckInViewModel.kt
```diff
@@ -88,23 +88,27 @@ class ConfirmCheckInViewModel @AssistedInject constructor(
...
-        checkInEnd = checkInEnd,
-        completed = completed,
-        createJournalEntry = createJournalEntry,
-    )
+    ): CheckIn {
+        val traceLocationId = verifiedTraceLocation.traceLocation.locationId
+        return CheckIn(
+            traceLocationId = traceLocationId,
+            // sha256() Hash of locationId ,which is itself a sha256() Hash
+            traceLocationIdHash = traceLocationId.sha256(),
```
LukasLechnerDev, line 96: Maybe also add the hashing to `TraceLocation` where we also have the calculation and hashing of the `traceLocationId`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
LukasLechnerDev: I am no regex expert, but I think it should be possible to modify our descriptors so that both versions match (with and without the `/`). Then we wouldn't need this...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/TraceLocation.kt
mtwalli: cwaDomain is the result of CWA_GUID.toByteArray() Please check specs as it says so: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/6e45edc4df5a8ced091a179bcd40da9bcc162c85/docs/spec/event-registration-client.md#calculate-tracelocation-id

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ProtoBuff.kt
```diff
@@ -0,0 +1,7 @@
+package de.rki.coronawarnapp.util
```
LukasLechnerDev, line 1: `ProtoBuff.kt` => `ProtoBuf.kt` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/confirm/ConfirmCheckInViewModel.kt
mtwalli: I added what is common to the TraceLocation , locationIdHash is more for checkIn than Location , and it is accessed only once on  confirmation afaik, is it required somewhere else?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/TraceLocation.kt
LukasLechnerDev: okay :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
mtwalli: The descriptor comes from the config. @mlenkeit  can you update the regex to match both?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/confirm/ConfirmCheckInViewModel.kt
LukasLechnerDev: `locationId` is also only needed for check-ins, or? I just think that the logic for `locationId` and `locationIdHash` should be in the same place.

mtwalli: ✅

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ProtoBuff.kt
mtwalli: ✅

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
mtwalli: Aligned with Max , Config should match both

LukasLechnerDev (dismissed):
Good work, nice tests, thumbs up :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ProtoBuf.kt
```diff
@@ -0,0 +1,7 @@
+package de.rki.coronawarnapp.util
+
+import com.google.protobuf.ByteString
+import okio.ByteString.Companion.toByteString
+
+fun okio.ByteString.toProtoByteString(): ByteString = ByteString.copyFrom(toByteArray())
+fun ByteString.toOkioByteString() = toByteArray().toByteString()
```
d4rken, line 7: Can you add a small unit test for this?

Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -85,7 +85,6 @@
 
                 <data
                     android:host="e.coronawarn.app"
-                    android:pathPrefix="/"
```
d4rken, line 88: Why did this change?

d4rken (commented):
Looks good, nice tests, few comments/questions, no blockers.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/PresenceTracingConfigMapper.kt
```diff
@@ -95,7 +91,8 @@ class PresenceTracingConfigMapper @Inject constructor() : PresenceTracingConfig.
                 revokedTraceLocationVersions = revokedTraceLocationVersionsList.orEmpty(),
                 riskCalculationParameters = riskCalculationParameters,
                 submissionParameters = submissionParameters,
-                plausibleDeniabilityParameters = plausibleDeniabilityParameters
+                plausibleDeniabilityParameters = plausibleDeniabilityParameters,
+                qrCodeDescriptors = qrCodeDescriptorsOrBuilderList
```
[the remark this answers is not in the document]
mtwalli: I did not see any specs for that. Can you align with Max first? The regex in this case is dynamic I don't know how do you want to check if valid or not ?

Corona-Warn-App/src/main/AndroidManifest.xml
mtwalli: the url changed and does not have `/` a prefix fix leaving it prevents the app from responding to deeplinks

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/PresenceTracingConfigMapper.kt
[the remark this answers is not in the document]
d4rken: Just compile it, if it is not valid regex, it would throw an exception? /cc @mlenkeit

d4rken (approved):
lgtm, thanks!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/PresenceTracingConfigMapper.kt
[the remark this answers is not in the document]
d4rken: Let's merge it as is and do a follow up if necessary.

LukasLechnerDev (approved):

PR #2741, opened 2021-04-05
Return default or most recent cached poster (EXPOSUREAPP-6079)

jurajkusnier:
**Update QR Template**
- If any other errors occur during the API request the client falls back to the most recent cached template. 
- If no cached template is available, the client falls back to the default template

**Testing**
- Test Menu > Event Registration > QR Code and poster creation > Download poster template
- Feel free to stop mock server / disable internet / clean cached files by device explorer and watch logs


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import android.content.Context
+import de.rki.coronawarnapp.util.di.AppContext
+import timber.log.Timber
+import java.io.File
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class QrCodePosterTemplateCache @Inject constructor(@AppContext private val context: Context) {
```
mtwalli, line 11: Why new class? Lukas already used Retrofit `Cache` to cache the response in app cache directory `context.filesDir `, I think we should also rely on the response headers. Please double check the specs?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateServer.kt
```diff
@@ -1,60 +1,67 @@
...
+                exportBinary,
+                SignatureValidation.parseTEKStyleSignature(exportSignature)
             )
+
+            if (!hasValidSignature) {
+                throw QrCodePosterTemplateInvalidResponseException(message = "Invalid Signature!")
+            }
+
+            templateCache.saveTemplate(exportBinary)
+            exportBinary
```
mtwalli, line 61: Could you please do me a favour and return the file directly along with other response data?
Currently I need to convert those bytes again into a temporary file and delete afterwards. It would be great if you can return a response object such as:
```kotlin
TemplateResponse(
   templateFile : File, // File path in cache directory
   offsetX : Float,
   offsetY : Float,
  ...
)
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val posterTemplateDir = File(context.filesDir, "events")
+
+    fun getTemplate(): ByteArray {
+        return File(posterTemplateDir, FILE_NAME).let { file ->
+            if (file.exists()) {
+                Timber.d("Loading template from $file")
+                file.readBytes()
+            } else {
+                Timber.d("Loading default template")
+                context.assets.open("default_qr_code_poster_template.pdf").readBytes()
```
LukasLechnerDev, line 22: As the default poster, we have to use the binary of the Poster template **protobuf message**, not only the PDF itself (Similar to AppConfig). You can find it here: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/event-registration-mvp/docs/spec/event-registration-client.md#protocol-buffer-message-qrcodepostertemplate

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
```diff
@@ -0,0 +1,41 @@
...
+@Singleton
+class QrCodePosterTemplateCache @Inject constructor(@AppContext private val context: Context) {
+
+    private val posterTemplateDir = File(context.filesDir, "events")
+
+    fun getTemplate(): ByteArray {
+        return File(posterTemplateDir, FILE_NAME).let { file ->
+            if (file.exists()) {
+                Timber.d("Loading template from $file")
+                file.readBytes()
```
LukasLechnerDev, line 19: I think all file operations should be done on the IO Dispatcher.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
LukasLechnerDev: Yes, we could implement these caching cases by implementing an okhttp interceptor instead of managing caching ourself. 

I think it can be realised by adding an Interceptor that returns the cached response if we have no internet connection (like [here](https://stackoverflow.com/questions/23429046/can-retrofit-with-okhttp-use-cache-data-when-offline)) and extend it so that it also returns the cached response when the backend returns an unsuccessful response.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateServer.kt
LukasLechnerDev: You shouldn't call this method directly, but instead call `downloadQrCodePosterTemplate()` which returns the protobuf that contains these values.

mtwalli: Still I have to convert the PDF bytes into file right? if you have the file already cached, It would be nice to return the path directly

LukasLechnerDev: Ah, now I get it. Yes makes sense.

jurajkusnier: but in case I remove my `QrCodePosterTemplateCache` implementation and use only okhttp cache, as you suggested in the previous comment, we wouldn't have direct access to the file anyway.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
jurajkusnier: The default template is from protobuf message, I just changed `.bin` to `.pdf` since it actually is pdf file,
but I can change it back to `.bin` and add `.sha256` to have the same format as in the app config

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateServer.kt
mtwalli: 👍🏼

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
mtwalli: Please consider having a default data as in here https://github.com/corona-warn-app/cwa-app-tech-spec/blob/a71ca36e09998a2ab92e764ade732e4546e30db7/docs/spec/event-registration-client.md#protocol-buffer-message-qrcodepostertemplate

jurajkusnier: yes, the default template is from `default_qr_code_poster_template_android` in spec

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateCache.kt
jurajkusnier: yes, custom implementation was overkill, okhttp cache seems to work as expected out of the box. It can handle error states and offline mode as well.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/DefaultQrCodePosterTemplateSanityCheck.kt
```diff
@@ -0,0 +1,39 @@
...
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.robolectric.RobolectricTestRunner
+import org.robolectric.annotation.Config
+import testhelpers.BaseTest
+import testhelpers.EmptyApplication
+
+@Config(sdk = [Build.VERSION_CODES.P], application = EmptyApplication::class)
+@RunWith(RobolectricTestRunner::class)
+class DefaultQrCodePosterTemplateSanityCheck : BaseTest() {
```
mtwalli, line 19: Please add `*Test` in the name ?

mtwalli:
Could you please update `QrCodePosterTemplateServerTest` still expecting xml file. Maybe @LukasLechnerDev  can help on how to update it?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/DefaultQrCodePosterTemplateSanityCheck.kt
jurajkusnier: that's the pattern we use
- `DefaultAppConfigSanityCheck`
- `DefaultQrCodePosterTemplateSanityCheck`

I guess in that case it makes sense to rename both tests. What do you suggest?
`DefaultQrCodePosterTemplateSanityCheckTest` or `DefaultQrCodePosterTemplateTest` ?

mtwalli: Whatever you like, I just need to distinguish them as `Test` classes

jurajkusnier: renamed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/events/server/qrcodepostertemplate/QrCodePosterTemplateServer.kt
```diff
@@ -1,60 +1,66 @@
...
+                SignatureValidation.parseTEKStyleSignature(exportSignature)
             )
+
+            if (!hasValidSignature) {
+                throw QrCodePosterTemplateInvalidResponseException(message = "Invalid Signature!")
+            }
+
+            exportBinary
+        } catch (exception: Exception) {
+            Timber.e(exception, "Response is not successful, trying to load template from cache")
```
mtwalli, line 62: Since you already handled the exception, maybe `Timber.d(e,"Message")` so in the logs it is distinguished for uncaught errors

jurajkusnier: ok

mtwalli (approved):
LGTM , tested 
1- on other env where poster is not provided and it return the default one 
2- on Local where Poster is served
3- when mock server is down 

All cases passed 👍🏼

LukasLechnerDev (approved):
LGTM!

Tested all the scenarios and everything worked as expected. 

Good work 👍

PR #2752, opened 2021-04-06
Only use check-ins of last 15 days (EXPOSUREAPP-5910)

LukasLechnerDev:
Even though we have a worker that deletes all stale check-ins, it's still possible to have stale check-ins in the database because the worker runs only once a day.

This PR adds another function to `CheckInRepository`, called `checkInWithinRetention`. This function only returns the check-ins which end date isn't older than 15 days. 

This new function is now used:
- `CheckInViewModel`: To not show outdated check-ins
- `MainActivityViewModel`: To not include outdated check-ins in the badge-count in Bottom Nav
- `SubmissionTask`: to not submit outdated check-ins
- `CheckInWarningMatcher`: To not match outdated check-ins

I also refactored the retention logic of `TraceLocations` by also adding a `traceLocationsWithinRetention` property to `DefaultTraceLocationRepository`.


vaubaehn:
@LukasLechnerDev 
Hi Lukas, some short questions to understand the idea behind this PR:
Why did you decide on 'hiding' the stale check-ins until the worker is run, instead of cleaning the database when the app is openend/going to foreground? Does database cleaning take much time or could cause any conflict in the opened app? Could there be any scenario, where a stale check-in could cause a problem (bug) but cannot easily be spotted by the users, because it's hidden?

LukasLechnerDev:
Hi @vaubaehn ! Great questions! 

> Why did you decide on 'hiding' the stale check-ins until the worker is run, instead of cleaning the database when the app is openend/going to foreground?

We decided that doing it on every app start is not enough (apps can be started and running for a long time) and doing it every time the app is brought to the foreground is too much :). Either way, we still could have outdated check-ins as some time will pass from the point in time when the user opens the app or brings it to foreground and then sees the check-ins or submits the check-ins. 

Therefore, I think it is best to just load the up-to date check-ins whenever we need them and also let a worker run daily to clean up all outdated checkins. 

> Does database cleaning take much time or could cause any conflict in the opened app?

I don't think it takes that much time. One issue we could have if we clean the db when the user is currently in the check-ins screen could be that suddenly some outdated check-ins disappear. 

 > Could there be any scenario, where a stale check-in could cause a problem (bug) but cannot easily be spotted by the users, because it's hidden?

I can't think of such a scenario ...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
```diff
@@ -24,10 +26,23 @@ class CheckInRepository @Inject constructor(
...
         .allEntries()
         .map { list -> list.map { it.toCheckIn() } }
 
+    /**
+     * Returns check-ins that are within the retention period. Even though we have a worker that deletes all stale
+     * check-ins it's still possible to have stale check-ins in the database because the worker only runs once a day.
+     */
+    val checkInsWithinRetention: Flow<List<CheckIn>> = allCheckIns.map { checkInList ->
+        checkInList.filter { checkIn ->
+            isWithinRetention(checkIn, timeStamper)
```
mtwalli, line 42: please pass a common instant , every time timeStamper.newUTC is called it is getting a new instance.
Maybe we should think of away to avoid this common mistake?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRetention.kt
```diff
@@ -0,0 +1,15 @@
+package de.rki.coronawarnapp.eventregistration.checkins
+
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.seconds
+import de.rki.coronawarnapp.util.TimeStamper
+import java.util.concurrent.TimeUnit
+
+private const val CHECK_IN_RETENTION_DAYS = 15
+private val CHECK_IN_RETENTION_SECONDS = TimeUnit.DAYS.toSeconds(CHECK_IN_RETENTION_DAYS.toLong())
+
+fun isWithinRetention(checkIn: CheckIn, timeStamper: TimeStamper): Boolean {
```
mtwalli, line 10: Please add unit test for it? and maybe make it and extension for` checkIn`

`CheckIn. isWithinRetention(now)`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRetention.kt
```diff
@@ -0,0 +1,15 @@
...
+    val retentionThreshold = (timeStamper.nowUTC.seconds - CHECK_IN_RETENTION_SECONDS)
+    return checkIn.checkInEnd.seconds >= retentionThreshold
+}
+
+fun isOutOfRetention(checkIn: CheckIn, timeStamper: TimeStamper) = !isWithinRetention(checkIn, timeStamper)
```
mtwalli, line 15: some tests please

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/CheckInCleaner.kt
```diff
@@ -17,18 +16,12 @@ class CheckInCleaner @Inject constructor(
 
     suspend fun cleanUp() {
         Timber.d("Starting to clean up stale check-ins.")
-        val retentionThreshold = (timeStamper.nowUTC.seconds - RETENTION_SECONDS)
         val checkInsToDelete = checkInRepository.allCheckIns.first()
-            .filter {
-                it.checkInEnd.seconds < retentionThreshold
+            .filter { checkIn ->
+                isOutOfRetention(checkIn, timeStamper)
             }
```
mtwalli, line 22: Before filtering please get the time now and pass the instance , with every checkin in a new instant is retrieved in `filter`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationCleaner.kt
```diff
@@ -17,19 +15,13 @@ class TraceLocationCleaner @Inject constructor(
 
     suspend fun cleanUp() {
         Timber.d("Starting to clean up stale trace locations.")
-        val retentionThreshold = (timeStamper.nowUTC.seconds - RETENTION_SECONDS)
         traceLocationRepository.allTraceLocations.first()
-            .filter {
-                it.endDate != null && it.endDate.seconds < retentionThreshold
+            .filter { traceLocation ->
+                isOutOfRetention(traceLocation, timeStamper)
```
mtwalli, line 20: same here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationRetention.kt
```diff
@@ -0,0 +1,21 @@
...
+
+import de.rki.coronawarnapp.eventregistration.checkins.qrcode.TraceLocation
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.seconds
+import de.rki.coronawarnapp.util.TimeStamper
+import java.util.concurrent.TimeUnit
+
+private const val TRACE_LOCATION_RETENTION_DAYS = 15
+private val TRACE_LOCATION_RETENTION_SECONDS = TimeUnit.DAYS.toSeconds(TRACE_LOCATION_RETENTION_DAYS.toLong())
+
+fun isWithinRetention(traceLocation: TraceLocation, timeStamper: TimeStamper): Boolean {
```
mtwalli, line 11: 1- `TraceLocation. isWithinRetention` ?
2- Pass the instant so it can be common between all TraceLocations?
3- some unit tests would be great ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationRetention.kt
```diff
@@ -0,0 +1,21 @@
...
+    val retentionThreshold = (timeStamper.nowUTC.seconds - TRACE_LOCATION_RETENTION_SECONDS)
+    return if (traceLocation.endDate == null) {
+        true
+    } else {
+        traceLocation.endDate.seconds >= retentionThreshold
+    }
+}
+
+fun isOutOfRetention(traceLocation: TraceLocation, timeStamper: TimeStamper) =
```
mtwalli, line 20: here too?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRepository.kt
LukasLechnerDev: Fixed. The only idea that I have to avoid this mistake would be to add some comments in the `TimeStamper`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRetention.kt
LukasLechnerDev: Added.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRetention.kt
LukasLechnerDev: Added.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/CheckInCleaner.kt
LukasLechnerDev: good point. Fixed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationCleaner.kt
LukasLechnerDev: fixed too

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationRetention.kt
LukasLechnerDev: Fixed all three points.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationRetention.kt
LukasLechnerDev: ✅

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/CheckInRetention.kt
LukasLechnerDev: Making it an extension is a very good improvement, thanks!

vaubaehn:
@LukasLechnerDev 
Hi Lukas, very convincing answers!
Thanks a lot for your thorough reply, hope it didn't steal much of your time.
I can follow you in all points, and I now understand why increasing complexity a bit here is a good idea.
The only critical point for possible issues then is the consistent usage of `checkInWithinRetention` function and `traceLocationsWithinRetention` property always and everywhere (but db cleaning), and not to mix up with `allCheckIns`, right? Could it then be helpful for future 'CWA newbies' to add a short comment on when to use `allCheckIns` and when to use `checkInWithinRetention` and `traceLocationsWithinRetention` to the basic repos (i.e., `CheckInRepository.kt`) to avoid confusion and mix up?

I'm very excitedly looking forward for release 2.0, pressing all thumbs and wishing it to become a great success!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/storage/retention/TraceLocationRetention.kt
```diff
@@ -0,0 +1,20 @@
...
+import org.joda.time.Instant
+import java.util.concurrent.TimeUnit
+
+private const val TRACE_LOCATION_RETENTION_DAYS = 15
+private val TRACE_LOCATION_RETENTION_SECONDS = TimeUnit.DAYS.toSeconds(TRACE_LOCATION_RETENTION_DAYS.toLong())
+
+fun TraceLocation.isWithinRetention(now: Instant): Boolean {
+    val retentionThreshold = (now.seconds - TRACE_LOCATION_RETENTION_SECONDS)
+    return if (endDate == null) {
+        true
```
mtwalli, line 14: do delete locations that does not have an end data? should it be `false` it we keep them

LukasLechnerDev:
@vaubaehn  Thanks for your feedback. You are right, it's definitely useful to document these functions better. I have added a commit (see above 👆)

Thank you for your wishes! We are also very excited for `2.0` 🚀

mtwalli (dismissed):
Thank you! 😊

jurajkusnier (approved):

mtwalli (approved):

PR #2753, opened 2021-04-06
Paragraph format & dark mode icons (EXPOSUREAPP-6188, EXPOSUREAPP-6211)

harambasicluka:
image

1. Go to create Qr Code and check legal text in the information fragment.
2. Check if the bold text in the grey box has the same alignment as the one below.

image

1. Go to My Check-Ins.
2. Check if the icons look like every where else and also have a working dark mode


BMItr (approved):
lgtm.
tested on Samsung S10 (dark mode)

mtwalli (approved):

PR #2760, opened 2021-04-07
Fix navigation from submission flow back to home screen (EXPOSUREAPP-6227)

SamuraiKek:
Fixes navigation issue that would pop up the trace location info screen instead of the home fragment when coming back from the submission flow.


harambasicluka (dismissed):
LGTM, tested on a Xiamoi Mi9 (Android 9) with positive tests on WRU-XD.

mtwalli (dismissed):

axelherbstreith (approved):

harambasicluka (approved):

PR #2763, opened 2021-04-07
Organizer Flow - Small Polishing (DEV)

axelherbstreith:
- Increased padding for the collapsed fab (doesn't collide with the list item anymore) -> Set default value of 72dp to all fabs in the application (including contact diary) 
- Fixed an issue where popBackStack was triggered when pressing anywhere on the toolbar in the "My QR code List"


harambasicluka (dismissed):

mtwalli (approved):
👍🏼

harambasicluka (approved):

PR #2765, opened 2021-04-07
Set correct traceLocationId on submission (EXPOSUREAPP-6070)

d4rken:
Now that `traceLocationId` format is alligned, pick the right and set it on the protobuf during submission.

Please review the test cases.


mtwalli (approved):
LGTM

chris-cwa (approved):

PR #2766, opened 2021-04-07
Align auto-checkout behavior and prevent editing of active checkins (DEV)

d4rken:
* Remove check-in monitor limit, execute for all changes (not only for changes to `isCompleted`)
* Process overdue automatic-checkouts when the `allCheckins` database table changes.
* Prevent the UI from showing a negative check-in duration
* Don't allow editing active check-ins
* Handle check-ins with default duration of 0. Scheduling an alarm 0 seconds in the future does not work. Treat as "overdue" and  handle in `processOverDueCheckouts()`


kolyaopahle (approved):
Noice 👍

harambasicluka (approved):

PR #2768, opened 2021-04-07
Improve CheckIns transitions  (DEV)

mtwalli:
- Improve transition from Check-ins Fragment to its destinations 
- this will avoid glitches that was happening before 

### Testing
Check before and after the changes


SamuraiKek (approved):
Looks good. Tested on Nokia 3 (Android 9).

axelherbstreith (approved):

PR #2770, opened 2021-04-08
Automatic contact journal entry creation (EXPOSUREAPP-5943)

BMItr:
Automatic contact journal entry creation (EXPOSUREAPP-5943).

Tests and little changes are missing. (DraftMode will be removed when finished)
update: 
- Tests adjusted.
- New Tests will be added in this PR or a follow Up PR.


chris-cwa (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
```diff
@@ -0,0 +1,76 @@
...
+import javax.inject.Inject
+import kotlin.math.roundToLong
+
+@Reusable
+class ContactJournalEntryCreator @Inject constructor(
+    private val diaryRepository: ContactDiaryRepository
+) {
+
+    suspend fun createEntry(checkIn: CheckIn) {
+        if (checkIn.createJournalEntry) {
```
d4rken, line 25: I'd prefer that the `if (checkIn.createJournalEntry)' would happen in `CheckOutHandler`
or
Renaming this `tryCreateEntry` or `maybeCreateEntry` and then using an early `return` if `!createJournalEntry`, but the first approach seems cleaner.

chris-cwa (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
```diff
@@ -0,0 +1,76 @@
...
+            Timber.d("Creating journal entry for %s", this)
+
+            // 1. Create location if missing
+            val traceLocationIdString = checkIn.traceLocationId.utf8()
```
mtwalli, line 29: could please use `checkIn.traceLocationId.base64()` as everywhere else?

d4rken, line 29: To prevent future bugs due to mistakes when comparing the traceLocationId please align this with the `traceLocationId` handling of the rest of the app.

i.e. use a ByteString everywhere outside the database, and for the database base64 encode.

```
interface ContactDiaryLocation : HasStableId {
    val locationId: Long
    var locationName: String
    val phoneNumber: String?
    val emailAddress: String?
    val traceLocationID: okio.ByteString? // (or use our TraceLocationId typealias)
}
```

then just compare the bytestrings directly, and when storing the tracelocationID in the contact diary table encode as base64 (`traceLocationId.base64()`).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
```diff
@@ -0,0 +1,76 @@
...
+                .find { it.traceLocationID == traceLocationIdString } ?: checkIn.toLocation(traceLocationIdString)
+
+            // 2. Split CheckIn by Midnight UTC
+            val splitCheckIns = checkIn.splitByMidnightUTC()
+            Timber.d("Split %s into %s ", this, splitCheckIns)
+
+            // 3. Create LocationVisit
+            splitCheckIns
+                .map { it.toLocationVisit(location) }
+                .forEach { diaryRepository.addLocationVisit(it) }
```
mtwalli, line 40: would that work?
```
splitCheckIns.forEach{
   diaryRepository.addLocationVisit(
     it.toLocationVisit(location)
   )
}
```

d4rken (changes requested):
Works well :+1:.

Please align the `traceLocationId` handling.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
BMItter: will be handled in a follow up PR

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
```diff
@@ -0,0 +1,76 @@
...
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toLocalDateUtc
+import kotlinx.coroutines.flow.first
+import org.joda.time.Duration
+import org.joda.time.Seconds
+import timber.log.Timber
+import javax.inject.Inject
+import kotlin.math.roundToLong
+
+@Reusable
+class ContactJournalEntryCreator @Inject constructor(
```
mtwalli, line 20: Maybe better naming `ContactJournalCheckInEntryCreator ` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+
+        return nameParts.joinToString(separator = ", ")
+    }
+
+    private fun CheckIn.toLocationVisit(location: ContactDiaryLocation): ContactDiaryLocationVisit {
+        // Use Seconds for more precision
+        val durationInMinutes = Seconds.secondsBetween(checkInStart, checkInEnd).seconds / 60.0
+        val duration = (durationInMinutes / 15).roundToLong() * 15
```
mtwalli, line 68: what does 15 mean here? could you please name at least what `duration` here is ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
BMItter: it was already mentioned above! please read

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
d4rken, line 40: Something with the ID check and visit generation is still off.

When checking into the same location twice.
* Scan QR, checkout after a few seconds
* Scan QR, checkout after a few seconds.
I endet up with two location entries, on overview, a single location on the day and no visit.
When I try to create a visit for that location, it doesn't expand the entry, but generates multiple visits on the overview.

Screenshot from 2021-04-08 17-47-11
Screenshot from 2021-04-08 17-46-53
Screenshot from 2021-04-08 17-46-44
Screenshot from 2021-04-08 17-46-36
Screenshot from 2021-04-08 17-52-35

d4rken (changes requested):
Multiple check-in seems to have a few undesired edge-cases.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
mtwalli: apparently we wrote at the same time :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
BMItter: name is given in the tech-spec.
would like to let it as it is for now (checkin context is clear)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
BMItter: check this for details: https://github.com/corona-warn-app/cwa-app-tech-spec/pull/30/commits/693e85f79a918bef112d63693fd36a16b7ec8d99

adjusted.

BMItter: part has been improved.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreatorTest.kt
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.jupiter.api.Test
+import testhelpers.BaseTest
+
+class ContactJournalCheckInEntryCreatorTest : BaseTest() {
+
+    @MockK lateinit var contactDiaryRepo: ContactDiaryRepository
+
+    private val testCheckIn = CheckIn(
+        id = 42L,
+        traceLocationId = "traceLocationId1".encode(),
```
mtwalli, line 27: should it be `encodeBase64` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
mtwalli: It is not clear for me what 15 here means , 15 mins 15 sec ?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreatorTest.kt
BMItter: yes. but as I wrote advanced testing is coming and variable is not relevant yet

mtwalli (approved):
Just small comments, not blocker for me

chris-cwa (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalEntryCreator.kt
BMItter: "Duration column is set by calculating the time difference in minutes between Check-in StartDate and Check-in EndDateand rounding it to the closest 15-minute duration."

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreator.kt
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    private suspend fun CheckIn.toLocation(): ContactDiaryLocation {
+        val location = DefaultContactDiaryLocation(
+            locationName = locationName(),
+            traceLocationID = traceLocationId
+        )
+        Timber.d("Created new location %s and adding it to contact journal db", location)
+        return diaryRepository.addLocation(location) // Get location from db cause we need the id autogenerated by db
+    }
```
d4rken, line 52: Please use a different name or pattern here. `toLocation` suggests conversions (compare all other cases of `toABC`), but this actually performs a database write operation, :ghost:.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreator.kt
```diff
@@ -0,0 +1,104 @@
...
+        if (traceLocationStart != null && traceLocationEnd != null) {
+            if (traceLocationStart.millis > 0 && traceLocationEnd.millis > 0) {
+                val formattedStartDate = traceLocationStart.toUserTimeZone().toString(DateTimeFormat.shortDateTime())
+                val formattedEndDate = traceLocationEnd.toUserTimeZone().toString(DateTimeFormat.shortDateTime())
+                nameParts.add("$formattedStartDate - $formattedEndDate")
+            }
+        }
+
+        return nameParts.joinToString(separator = ", ")
+    }
```
d4rken, line 67: Now this could be a `toLocationName` or `val CheckIn.locationName` :wink:

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/CheckOutHandlerTest.kt
```diff
@@ -68,7 +83,37 @@ class CheckOutHandlerTest : BaseTest() {
...
         // TODO cancel auto checkouts
     }
+
+    @Test
+    fun `Creates entry if create journal entry is true`() = runBlockingTest {
+        createInstance().apply {
+            checkOut(42)
+        }
+
+        updatedCheckIn?.createJournalEntry shouldBe true
```
d4rken, line 100: No `rofl?.copter shouldBe XYZ` in tests please. If it should not be null, force a check `!!`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/CheckOutHandlerTest.kt
```diff
@@ -68,7 +83,37 @@ class CheckOutHandlerTest : BaseTest() {
...
+        }
+    }
+
+    @Test
+    fun `Does not create entry if create journal entry is false`() = runBlockingTest {
+        createInstance().apply {
+            checkOut(43)
+        }
+
+        updatedCheckIn?.createJournalEntry shouldBe false
```
d4rken, line 113: No `rofl?.copter shouldBe XYZ` in tests please. If it should not be null, force a check `!!`.

d4rken (approved):
Thanks for the changes.
Few comments, would be happy if you would address them in a follow up PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreator.kt
BMItter: :ghostbusters

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreator.kt
```diff
@@ -0,0 +1,104 @@
...
+
+    private suspend fun List<CheckIn>.createMissingLocationVisits(location: ContactDiaryLocation):
+        List<ContactDiaryLocationVisit> {
+            Timber.d(
+                "createMissingLocationVisits(location=%s) for %s",
+                location,
+                this.joinToString(prefix = System.lineSeparator(), separator = System.lineSeparator())
+            )
+            val existingLocationVisits = diaryRepository.locationVisits.firstOrNull() ?: emptyList()
```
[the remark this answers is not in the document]
BMItter: In this case it won't make a difference. If it's null the test will fail anyways

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreator.kt
BMItter: In this case it won't make a difference. If it's null the test will fail anyways

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/CheckOutHandlerTest.kt
BMItter: In this case it won't make a difference. If it's null the test will fail anyways

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/CheckOutHandlerTest.kt
BMItter: In this case it won't make a difference. If it's null the test will fail anyways

PR #2771, opened 2021-04-08
Make the event registration checkin duration picker open on the whole row (DEV)

kolyaopahle:
### Description
This PR changes the click handle of the checkin duration picker to the whole row of the card to allow for easier selection
(currently not on the edit screen as we have different ui's there)

### Steps to reproduce
1. Open the checkin screen
2. Check if the duration picker is easier to access


Corona-Warn-App/src/main/res/layout/fragment_confirm_check_in.xml
```diff
@@ -206,6 +206,7 @@
                         android:background="?android:attr/listDivider" />
 
                     <androidx.constraintlayout.widget.ConstraintLayout
+                        android:id="@+id/confirm_checkin_settings_card_checkout_time_row"
                         android:layout_width="match_parent"
                         android:layout_height="wrap_content"
                         android:orientation="horizontal">
```
mtwalli, line 212: it is not related to the fix. there is no `orientation`  for `ConstraintLayout`

mtwalli (dismissed):

axelherbstreith (dismissed):

Corona-Warn-App/src/main/res/layout/fragment_confirm_check_in.xml
kolyaopahle: got rid of it

mtwalli (approved):

PR #2774, opened 2021-04-08
Event Registration: Restrict check-in time (EXPOSUREAPP-6177)

kolyaopahle:
### Description
This Pr adds a warning text to the edit checkin screen that informs the user why the save button is disabled,
also fixed an issue where having a 0 duration checkin could be saved when opening and closing the timepicker instantly

### Steps to reproduce
1. check into event
2. try to edit the event
3. check if event can be longer than 24h
4. check if checkout can be before checkin


mtwalli:
when start and end time are the same I see the error message, is this right?
device-2021-04-08-135434

harambasicluka:
> when start and end time are the same I see the error message, is this right?

Not sure how to handle this, the min default duration as discussed in teams is 15min. So it should be aligned from my point of view. @mlenkeit what do you think?

harambasicluka (approved):
For me it's good as it is, works like a charm. 

But if we get a fast decision on the concern of @mtwalli it can be also included. Or maybe delivered in another PR.

kolyaopahle:
> when start and end time are the same I see the error message, is this right?
> device-2021-04-08-135434
> 

I assume this is intended as checkout has to be after the checkin according to the message

mtwalli (approved):

PR #2783, opened 2021-04-08
International hotline number (EXPOSUREAPP-6011)

SamuraiKek:
Adds an extra  international phone number for the "Technical Hotline" and "Request Tan" screens. 
I've had to remove the "Call" button from the "Request TAN" screen because it wouldn't make sense to have it with two phone numbers in the UI. The behaviour is also more in line with the "Technical Hotline" screen.

- Technical Hotline number: +49 30 498 75401
- Request TAN number: +49 30 498 75402


SamuraiKek:
Closing this PR and opening #2789 to avoid rebasing on 2.1

Issue #2761, opened 2021-04-07
QR code time for event moved 2 hours forward


MikeMcC399:
## Technical details

- Device name: Google Pixel 3a emulator
- Android version: 11
- App version: [v2.0.0-RC2](https://github.com/corona-warn-app/cwa-app-android/releases/tag/v2.0.0-RC2)

## Describe the bug

The time defined when creating a QR code for an event is displayed as 2 hours later when the event is saved.

## Steps to reproduce the issue

1. Open CWA, scroll down to "CREATE QR CODE" and select it
2. Tap on "+ QR CODE"
3. Tap on "Cultural event"
4. Enter a description text and a place text
5. Define start and end times and make a note of them
6. Tap "SAVE"
7. Examine the screen with the header "My QR Codes" and note that the times are two hours later than the ones entered in step 5.

![Create QR code](https://user-images.githubusercontent.com/66998419/113860377-400ea900-97a6-11eb-85e2-540ad81e8011.png)

![Saved event](https://user-images.githubusercontent.com/66998419/113860390-43099980-97a6-11eb-815a-22672ba24909.png)

## Expected behavior

The times entered should be the same as the times later displayed after a QR code for an event is created.

## Possible Fix

It looks like there may be a confusion between UTC and local times. The local time zone is currently CEST which is UTC+2.

---
Internal Tracking ID: [EXPOSUREAPP-6249](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6249)

d4rken:
Nice catch. I agree, it's an issue with UTC conversion.

We should either adjust for the users local time within the ViewModel, or change the picker to use users current timezone already on the generated choices, and then convert them to UTC before storing them.
Later one seems cleaner, expect the UI to be in the users timezone, and then convert to UTC before saving.

/cc @jurajkusnier 

![Screenshot from 2021-04-07 18-03-46](https://user-images.githubusercontent.com/1439229/113898397-22543a80-97cc-11eb-807a-3b31402a98f0.png)

https://github.com/corona-warn-app/cwa-app-android/blob/ea87861d91b73a5408fc05f7c56713fb8697cee8/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/organizer/create/TraceLocationCreateFragment.kt#L162-L174

dsarkar:
[...]

PR #2786, opened 2021-04-09
Fix QR code timezone (EXPOSUREAPP-6249)

jurajkusnier:
PR fixes issue where QR code time for event moved 2 hours forward


MikeMcC399:
I successfully verified the fix in this PR linked to issue https://github.com/corona-warn-app/cwa-app-android/issues/2761.

## Verification

1. Open CWA, scroll down to "CREATE QR CODE" and select it
2. Tap on "+ QR CODE"
3. Tap on "Cultural event"
4. Enter a description text and a place text
5. Define start and end times and make a note of them
6. Tap "SAVE"
7. Examine the screen with the header "My QR Codes" and compare to the times entered in step 5.

## Expected behavior

The times entered should be the same as the times later displayed after a QR code for an event is created. ✔️ 

## Environment

- Device name: Google Pixel 3a emulator
- Android version: 11

d4rken (approved):
:+1: 

Can we add a unit test for this? 

image

image

jurajkusnier:
> 👍
> 
> Can we add a unit test for this?

This should be covered by an instrumentation test in the following PR

chris-cwa (approved):

PR #2788, opened 2021-04-09
Adjust EW/PT riskresult combination (EXPOSUREAPP-6169)

d4rken:
* The home screen uses a "CombinedRiskResult" consisting of the PT and EW risk result, the combination behavior was adjusted in this PR. A failed calculation from either PT or EW, will cause the combined result to also have the state "failed", so either risk calculation failing should show the error card with the retry button on the home screen.
* The EW risk result has a value `matchedKeyCount` to detect "low risk, despite encounters", which is used in the risk details screen to display an additional info box, and show a link to the FAQ. There is now a new `matchedRiskCount`  which is the combination of `matchedKeyCount` (EW) and `checkInOverlapCount` (PT).
* Some refactoring as I incorrectly renamed two methods previously.
* Tracing progress status now also reflects download/riskcalculation of the `PresenceTracingWarningTask`.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/TracingProgress.kt
```diff
@@ -2,9 +2,15 @@ package de.rki.coronawarnapp.tracing
 
 sealed class TracingProgress {
 
-    object Idle : TracingProgress()
+    object Idle : TracingProgress() {
+        override fun toString(): String = "TracingProgress.Idle"
```
chris-cwa, line 6: is this necessary?

LukasLechnerDev:
I think it not good to show the user a failed risk calculation if one of the two fails. He could have a high risk because of EW, but since the PT fails, he sees the failed Risk card instead of the high-risk card and so isn't aware that he has a high risk.

harambasicluka:
> I think it not good to show the user a failed risk calculation if one of the two fails. He could have a high risk because of EW, but since the PT fails, he sees the failed Risk card instead of the high-risk card and so isn't aware that he has a high risk.

@mlenkeit what do you think?

d4rken:
> I think it not good to show the user a failed risk calculation if one of the two fails. He could have a high risk because of EW, but since the PT fails, he sees the failed Risk card instead of the high-risk card and so isn't aware that he has a high risk.

I discussed this with Max ( /cc @mlenkeit ), I think both variants are valid, if desired we can change this back in a follow up PR on monday.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/TracingProgress.kt
d4rken: Not necessary, but nice, otherwise it prints to debug log like `abcdefasdas.asd.asda.sd@asdasdkasdk.TracingProgress.Idle`.

LukasLechnerDev (approved):

chris-cwa (approved):

mlenkeit:
> > I think it not good to show the user a failed risk calculation if one of the two fails. He could have a high risk because of EW, but since the PT fails, he sees the failed Risk card instead of the high-risk card and so isn't aware that he has a high risk.
> 
> I discussed this with Max ( /cc @mlenkeit ), I think both variants are valid, if desired we can change this back in a follow up PR on monday.

Was originally discussed with POs. Matthias is right.

PR #2791, opened 2021-04-09
Adapt auto-checkout default values (EXPOSUREAPP-6243)

LukasLechnerDev:
This PR fine-tunes the the default automatic-checkout value in the ConfirmCheckout Screen:

image

More information in the Jira Ticket!


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/AutoCheckoutHelper.kt
```diff
@@ -0,0 +1,68 @@
+package de.rki.coronawarnapp.eventregistration.checkins.qrcode
```
chris-cwa, line 1: could we not call it helper? maybe DefaultAutoCheckoutLength?

chris-cwa (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/AutoCheckoutHelper.kt
```diff
@@ -0,0 +1,68 @@
...
+
+        if (minutesUntilEndDate > maxDefaultAutoCheckOutLengthInMinutes) {
+            return maxDefaultAutoCheckOutLengthInMinutes
+        }
+
+        return roundToNearest15Minutes(minutesUntilEndDate)
+    }
+}
+
+private fun roundToNearest15Minutes(minutes: Int): Int {
```
BMItter, line 60: What about a case like 7.5 minutes? It would be 0. Better use Seconds for the calculation...
maybe also as extension function 🐰

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/AutoCheckoutHelper.kt
LukasLechnerDev: I can change it, but what's wrong with "helper"?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/AutoCheckoutHelper.kt
LukasLechnerDev: Input is `Int`, so 7.5 is not possible 😉. I discussed it with Max, rounding on minute-level is sufficient. 
I added some comments and test to make this function easier to understand.

chris-cwa (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/eventregistration/checkins/qrcode/AutoCheckoutHelper.kt
BMItter: all right.

BMItr (approved):
lgtm.

PR #2797, opened 2021-04-09
Prevent matching our own CheckIns (EXPOSUREAPP-6305)

d4rken:
Prevent matching our own CheckIns, by marking submitted CheckIns in the database and skipping them during matching.


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/risk/calculation/OverlapTest.kt
```diff
@@ -262,13 +262,32 @@ class OverlapTest : BaseTest() {
             traceWarningPackageId = id
         )!!.roundedMinutes shouldBe 5
     }
+
+    @Test
+    fun `returns null if it matches our own ssubmitted CheckIn`() {
```
chiljamgossow, line 267: typo submitted

chiljamgossow (approved):
This way of handling it will also obscure other people's warnings. I suppose this is accepted.

jurajkusnier (approved):
LGTM

Issue #482, opened 2020-06-16
Reduce size of compiled / uploaded app


IT-Cru:
## Current Implementation
In Google PlayStore the Corona-Warn-App is displayed with a size of 31 MB.
In Apple App Store the Corona-Warn-App is displayed with a size of 18,9 MB.

## Suggested Enhancement
I think many people will not understand why a simple app which will not track any personal data or something else from them is so big. Also not so powerful devices should be supported, because not all people have latest power devices. I think many users will not install an app with such a huge size.

- [ ] Reduce size of production required assets like images, icons, logos
- [ ] Remove non-production required assets from app source
- [ ] Remove all develop / test related stuff from production app
- [ ] Reduce usage of not really required libs or use smaller ones
- [ ] Check used install package type Android App Bundles or APK
- [ ] Maybe optimize CI jobs / pipelines

## Expected Benefits
More people will install a clean and smaller app.

Internal Tracking ID: EXPOSUREAPP-1872

SebastianWolf-SAP:
The app simply needs the space right now. If you have concrete steps for improvements, contributions are welcome. But for now, we close this issue as too abstract.

Mit freundlichen Grüßen/Best regards, 
SW
Corona Warn-App Open Source Team

corneliusroemer:
Some of the recommendations were quite explicit: Reduce size of assets?

And yes, 30MB is big. Once it's running smoothly, try to get it down in a week or two.

Phoenix616:
@SebastianWolf-SAP please stop closing issues that aren't resolved, thanks.

corneliusroemer:
@SebastianWolf-SAP Please don't add wontfix prematurely. Surely you will work on this app for the next few months while the pandemic is ongoing and continue to improve it.

You did a good job getting it out quickly. Now the optimisation begins.

ironjan:
@SebastianWolf-SAP please remember the first few weeks after the cwa-repositories were published step by step: Lots of issues  were closed with "we'll do that later, please wait". Soon after, lots of duplicates were created for the closed issues - most likely because the original problem still persisted and there were no open issues that tracked it. For one thing, this causes unnecessary work for the repository maintainers. But worse: it may lead to frustration on the side of contributors. 

In this sense, I would suggest to close #527 as duplicate of #482 and reopen #482 since 
 * the issue actually lists "concrete steps for improvements" and
 * "contributions are welcome".


----
 
[...]

Issue #265, opened 2020-06-07
[UI] Elements missing from images based on theme


IndianaDschones:
## Describe the bug
The images on the screen where a user can choose how to enter the TAN (QR-Code, manually, request the TAN) are missing some details based on the theme color, see pictures below.

The back-arrow is also not visible on the black theme, but this might be a seperate issue...

There might be other views with similar error, but i noticed it on this screen.

## Expected behaviour
The pictures should show identical content, regardless of the theme.

## Steps to reproduce the issue
If your phone has the option to change the system color from light to dark, do it. Open the app in both themes and see the different behaviour.

![dark (2)_LI](https://user-images.githubusercontent.com/6652142/83980903-90fd0900-a919-11ea-973d-e9a584af5709.jpg)
![light (2)_LI](https://user-images.githubusercontent.com/6652142/83980905-92c6cc80-a919-11ea-902a-424b7010c643.jpg)

## Technical details

- Smartphone with Android 10
- Build from dev-branch https://github.com/corona-warn-app/cwa-app-android/commit/8dd22995d2d609c9e1ec96e356c3faf551c6c3b1

## Additional context

---
Internal Tracking ID: EXPOSUREAPP-1968

marcmuschko:
Closing this issue as it was resolved in PR #288 
Please reopen if it still seems to be malfunctioning! 
Thanks and best,
Marc

IndianaDschones:
Hi @marcmuschko 

i have a device running Android 6 (Marshmallow) and there are the fingers from the hand holding the smartphone on the qr-code and the "still no tan" card rectangular, see screenshot.

On my Android 10 (Q) device this bug isn´t present.

However on both devices the header (Selection) including the back-arrow is scrolling up. The header is sticky on all other screens. Is this intended?

![selection-header](https://user-images.githubusercontent.com/6652142/84206724-0f8baf00-aab0-11ea-9fe2-5aa0b870d0a1.png)

Unfortunately i don´t have a option to reopen this issue.

Fabian-K:
Hi @IndianaDschones,

I can´t reproduce the "still no tan" issue on Android 6. Do you still face the issue on your device using the current dev branch?

Thanks, Fabian

IndianaDschones:
The fingers are still missing from the first and third card. If you cannot reproduce it on a device running Marshmallow, maybe it is something device specific?
I´ve got a Motorola Moto G (2nd Generation)

[...]

PR #2800, opened 2021-04-10
Use vector drawables support library (COMMUNITY)

fynngodau:
### Description

Set `vectorDrawables.useSupportLibrary = true` in `build.gradle`.

This causes the Android build system not to pregenerate all image resources for different sizes, thereby drastically reducing APK size.

To avoid crashes, additionally:

* pregenerate the splash screen logo
* always use `app:srcCompat` instead of `android:src`

### Steps to reproduce

Compile APK with and without these changes and compare `res` folder.

To the left: CWA build without support library for vector drawables; to the right: CWA build with support library for vector drawables.

image

How to test: ensure nothing breaks.

If you know any reason that speaks against setting this, please let me know.

### Related issues

https://github.com/corona-warn-app/cwa-app-android/issues/265#issuecomment-757444223
#482


fynngodau:
The pre-generated PNGs are not generated correctly, even: for instance, this is a pre-generated PNG from the latest official release APK (1.15.2 `/res/drawable-mdpi-v4/ic_information_illustration_technical.png`):

ic_information_illustration_technical

MikeMcC399:
@fynngodau 
This issue is also mentioned in https://github.com/corona-warn-app/cwa-app-android/issues/265#issuecomment-757444223 with Internal tracking ID EXPOSUREAPP-1968. Maybe you could cross-check if you would also be fixing this one?

fynngodau:
I noticed this required two more changes:

* pregenerating the splash screen logo, because Android 6 does not want to render it itself (probably)
* always use `app:srcCompat` instead of `android:src`

I added the according changes and changed the base to `release/2.1.x`. I could not find any further crashes on Android 6 or otherwise.

@MikeMcC399 Indeed, that issue is related. It should be fixed with this PR because the assets would no longer be incorrectly pre-generated, but instead be rendered on the device (and, I believe, correctly).

mtwalli (dismissed):
@fynngodau thank you very much!, this PR hits  two birds with one stone 😃.
it did reduce the APK size significantly 

From:
Screenshot 2021-04-10 at 23 38 16

To:
Screenshot 2021-04-10 at 23 00 58

Additionally fixes the linked issues above for older versions.


FYI splash icon uses Vector Drawable gradient as `fillColor` which is only supported from API 24+, that's why you had to bundle the icons manually.  While this is a downside of this approach , I don't think it is a big deal for one icon considering the benefits, and I don't expect that the logo icon will change any time soon.

fynngodau:
@mtwalli Thanks for this insightful information :heart:

d4rken (approved):
lgtm

mtwalli (approved):

d4rken:
Thank you @fynngodau

PR #2801, opened 2021-04-11
Fix camera settings item blinking (EXPOSUREAPP-6337)

mtwalli:
- With the introduction of `intervalFlow` in CheckIns screen causes a refresh every second, when Camera item is displayed it had blinking effect to avoid that this PR create only one item lazily 
- Fix typo


d4rken (dismissed):
Works for me, as the CameraItem was only instantiated once in `init`.  The other solution would be to use `HasPayloadDiffer`.

mtwalli:
> Works for me, as the CameraItem was only instantiated once in `init`. The other solution would be to use `HasPayloadDiffer`.

I did try it as :
```       
 override fun diffPayload(old: Any, new: Any): Any? = null
``` 
did not work for me , not sure if this how it should be ?

d4rken:
> did not work for me , not sure if this how it should be ?

You need to return a non-null value.

https://github.com/corona-warn-app/cwa-app-android/blob/c9cabdcdcf8db7e964df6c9643979b2367577008/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/diffutil/SmartDiffUtil.kt#L24-L39

Other cards we have use this pattern:

https://github.com/corona-warn-app/cwa-app-android/blob/c9cabdcdcf8db7e964df6c9643979b2367577008/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestErrorCard.kt#L34-L36

then

https://github.com/corona-warn-app/cwa-app-android/blob/c9cabdcdcf8db7e964df6c9643979b2367577008/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestErrorCard.kt#L22-L29

The idea being that we want to replace any stale on-click listeners that may reference something that no longer exists due to recreation on lifecycle events. Not sure if it's necessary in this case, but using the payload mechanism tells the RecyclerView that we will take care of any layout that is necessary, so you may be able to just return `Any()` as payload diff.

mtwalli:
> > did not work for me , not sure if this how it should be ?
> 
> You need to return a non-null value.
> 
> https://github.com/corona-warn-app/cwa-app-android/blob/c9cabdcdcf8db7e964df6c9643979b2367577008/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/diffutil/SmartDiffUtil.kt#L24-L39
> 
> Other cards we have use this pattern:
> 
> https://github.com/corona-warn-app/cwa-app-android/blob/c9cabdcdcf8db7e964df6c9643979b2367577008/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestErrorCard.kt#L34-L36
> 
> then
> 
> https://github.com/corona-warn-app/cwa-app-android/blob/c9cabdcdcf8db7e964df6c9643979b2367577008/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/ui/homecards/TestErrorCard.kt#L22-L29
> 
> The idea being that we want to replace any stale on-click listeners that may reference something that no longer exists due to recreation on lifecycle events. Not sure if it's necessary in this case, but using the payload mechanism tells the RecyclerView that we will take care of any layout that is necessary, so you may be able to just return `Any()` as payload diff.

In this case , current solution is sufficient, item does not change and click listener does not any data

d4rken (approved):

harambasicluka:
@mtwalli how do I get this card? Tried it on an Xiamoi Mi (Android 9) and Oneplus 6t (Android 10) by clicking deny in the scanning creen.

harambasicluka:
> @mtwalli how do I get this card? Tried it on an Xiamoi Mi (Android 9) and Oneplus 6t (Android 10) by clicking deny in the scanning screen.

Ok found it, you have deny once and then the second time you get asked if you don't want to be asked again. 

Works on both devices :)

harambasicluka (approved):

PR #2802, opened 2021-04-11
Clean up (DEV)

mtwalli:
⬆️


Corona-Warn-App/build.gradle
```diff
@@ -173,7 +173,8 @@ android {
                     "-Xno-kotlin-nothing-value-exception",
                     "-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi",
                     "-Xuse-experimental=kotlinx.coroutines.FlowPreview",
-                    "-Xuse-experimental=kotlin.time.ExperimentalTime"
+                    "-Xuse-experimental=kotlin.time.ExperimentalTime",
+                    "-Xopt-in=kotlin.RequiresOptIn"
```
d4rken, line 177: Did this PR add something that uses it? Can't spot it :thinking:

d4rken (approved):
lgtm

Corona-Warn-App/build.gradle
mtwalli: Not in here , it was part of Swipe feature, I used `OptIn` annotation which is still in preview

axelherbstreith (approved):

PR #2810, opened 2021-04-12
Improve risk display in reset and error cases (EXPOSUREAPP-6356)

d4rken:
* https://github.com/corona-warn-app/cwa-app-android/commit/b290ddac793c1f78067848929beb4492ca25d0b6 fixes the initial calculation failed state on app reset by changing the "fallback value if there is no data" to LOW_RISK instead of CALCULATION_FAILED, which became an issue after changing CALCULATION_FAILED to be the dominant state.
* https://github.com/corona-warn-app/cwa-app-android/pull/2810/commits/02cbccae93ce6aacb072586960a0c33273778276 which changes the TracingRepo to call both tracing tasks when there is an error. Otherwise pressing the retry button wouldn't help if it was the presence tracing that failed last.

For testing you'd want to reset the app via in app reset, and play a bit with failed risk states (i.e. no internets).


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -91,29 +93,17 @@ class TracingRepository @Inject constructor(
...
-                )
+    fun refreshRiskResult() = scope.launch {
+        Timber.tag(TAG).d("refreshRiskResults()")
+
+        riskWorkScheduler.runRiskTasksNow()
+
+        taskController.submit(
+            DefaultTaskRequest(
+                RiskLevelTask::class,
+                originTag = "TracingRepository.refreshDiagnosisKeys()"
```
chiljamgossow, line 104: Should the tag be updated as well?

chiljamgossow (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
d4rken: Yes, currently only affects logging though.

d4rken: Done.

chiljamgossow (approved):

chris-cwa (approved):
LGTM!

BMItr (approved):
code lgtm 👌 
nothing to complain

PR #2811, opened 2021-04-12
Increase auto-checkout text granularity (display minutes) (EXPOSUREAPP-6331)

d4rken:
Create a checkout with auto checkout after 1.5h.

Before: No minutes :fearful: 
Now: Minutes :cocktail:


mtwalli (approved):
LGTM
Screenshot 2021-04-12 at 16 51 45

harambasicluka (approved):

PR #2812, opened 2021-04-12
Event Creation - Self check-in flow (EXPOSUREAPP-6299)

jurajkusnier:
Fix self-check-in flow


chiljamgossow (approved):
Tested, LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
```diff
@@ -53,7 +53,8 @@ class CheckInsFragment : Fragment(R.layout.trace_location_attendee_checkins_frag
             factory as CheckInsViewModel.Factory
             factory.create(
                 savedState = savedState,
-                deepLink = navArgs.uri
+                deepLink = navArgs.uri,
+                cleanHistory = navArgs.cleanHistory
```
mtwalli, line 57: I think it is better if you call it `clearHistory`

mtwalli (approved):
Not a blocker , I think clearHistory is more suitable than cleanHistory

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/eventregistration/attendee/checkins/CheckInsFragment.kt
jurajkusnier: I have nothing against using `clear` vs `clean` it could be refactored later

PR #2814, opened 2021-04-13
Add unit test for PresenceTracingRiskRepository (EXPOSUREAPP-6332)

chiljamgossow:
Unit tests
Refactoring/ clean up 
Ensures only data from last 14 days are used
Correct assignment of latest matches to risk result


d4rken (approved):
lgtm

BMItr (approved):

PR #2818, opened 2021-04-13
Skeletons for new test result architecture (EXPOSUREAPP-6007)

d4rken:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepo.kt
```diff
@@ -0,0 +1,50 @@
...
+import de.rki.coronawarnapp.coronatest.storage.CoronaTestStorage
+import de.rki.coronawarnapp.coronatest.type.CoronaTest
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.emptyFlow
+import timber.log.Timber
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class CoronaTestRepo @Inject constructor(
```
mtwalli, line 15: Please call it repository to be aligned with all ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepo.kt
```diff
@@ -0,0 +1,50 @@
...
+    private val server: CoronaTestServer,
+) {
+
+    val coronaTests: Flow<Set<CoronaTest>> = emptyFlow()
+
+    /**
+     * When this returns and there was no exception, the test was registered and a valid registrationToken obtained.
+     * Your new test should be available via **coronaTests**.
+     */
+    suspend fun registerTest(request: CoronaTestQRCode) {
```
mtwalli, line 26: should it return `registrationToken` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/CoronaTestStorage.kt
```diff
@@ -0,0 +1,23 @@
...
+@Singleton
+class CoronaTestStorage @Inject constructor() {
+
+    suspend fun load(): Set<CoronaTest> {
+        Timber.tag(TAG).d("load()")
+        throw NotImplementedError()
+    }
+
+    suspend fun save(tests: Set<CoronaTest>) {
+        Timber.tag(TAG).d("saveload(tests=%s)", tests)
```
mtwalli, line 17: `"save"`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/CoronaTestStorage.kt
```diff
@@ -0,0 +1,23 @@
...
+
+@Singleton
+class CoronaTestStorage @Inject constructor() {
+
+    suspend fun load(): Set<CoronaTest> {
+        Timber.tag(TAG).d("load()")
+        throw NotImplementedError()
+    }
+
+    suspend fun save(tests: Set<CoronaTest>) {
```
mtwalli, line 16: Do you always save a set of test?, I think it should be for one test

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/antigen/RapidAntigenCoronaTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+    override val type: CoronaTest.Type = CoronaTest.Type.RAPID_ANTIGEN
+
+    enum class State {
+        PENDING,
+        INVALID,
+        POSITIVE,
+        NEGATIVE,
+        EXPIRED,
+
```
mtwalli, line 24: extra line

axelherbstreith (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/pcr/PCRCoronaTest.kt
```diff
@@ -0,0 +1,25 @@
...
+) : CoronaTest {
+
+    override val type: CoronaTest.Type = CoronaTest.Type.PCR
+
+    enum class State {
+        PENDING,
+        INVALID,
+        POSITIVE,
+        NEGATIVE,
+        EXPIRED,
```
mtwalli, line 23: this was called `REDEEMED` not expired

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/CoronaTestQRCode.kt
```diff
@@ -0,0 +1,30 @@
+package de.rki.coronawarnapp.coronatest.qrcode
+
+import android.os.Parcelable
+import de.rki.coronawarnapp.coronatest.type.CoronaTest
+import kotlinx.parcelize.Parcelize
+import org.joda.time.Instant
+
+sealed class CoronaTestQRCode : Parcelable {
```
mtwalli, line 8: does it have to be Parcelable ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepo.kt
d4rken: I don't see an immediate need, we are more flexible if all data retrieval the UI needs to do, is done via the `coronaTests` flow.
This will make it easier for us to make test registration async in the future (without internet).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/CoronaTestStorage.kt
d4rken: It's always a set, it's a snapshot of the current tests. The tests will live inside a flow  while the app is running,  via `HotDataFlow`, always in memory. On data change the snapshot will be updated via `save` and when the app was killed, they will be restored via `load`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/CoronaTestQRCode.kt
d4rken: Not guaranteed in the final version, but for the purpose of us parallelising work, it should be for now. Navigation is not fully clear yet, and when switching to different screens @axelherbstreith has the option of passing the scanned QR Code along and completing the flow without requiring to scan again, or to persist the data to disk, giving him a lot of flexibility when working on the UI flow.

mtwalli (approved):
👍🏼

axelherbstreith (approved):

PR #2819, opened 2021-04-13
Poster Text -  Roboto font (DEV)

mtwalli:
Set text font family as robots , so it is not changed to the default system font
Screenshot 2021-04-13 at 13 46 30


harambasicluka (dismissed):

axelherbstreith (approved):
Lgtm

harambasicluka (approved):

PR #2824, opened 2021-04-13
Corona test result fetching, storage & wiring (EXPOSUREAPP-6007)

d4rken:
* This PR replaces the mix of SharedPreference properties used to store the data related to the current PCR test with a `CoronaTestRepository` that allows to work with implementations of `CoronaTest`.
* Each `CoronaTest` has a `CoronaTestProcessor` that hides the test specific actions
* Any test related actions are performed via the repository, access to the current test data is only possible via `Flow<Set<CoronaTest>>` to allow threadsafe observation from various areas of the app.
* `CoronaTestMigration` will take the old data and construct a `data class PCRTest : CoronaTest` out of it, after that has been stored in the new `CoronaTestStorage` the old attributes are deleted.
* The tests themselves are stored in memory via a `HotDataFlow<Set<CoronaTest>>`. This flow is observed and on every data change a new snapshot of the data classes is serialized and persisted. The last snapshots is loaded from storage and used to initialize `HotDataFlow` if our app is killed.
* On migration of old data, the user may see the app refreshing the tests for a short moment or `PENDING` if  there is no internet until we were able to determine the current test result state for the migrated test data.

- [ ] Unit tests for migration
- [ ] Unit tests for processors
- [ ] Unit tests for repo
- [ ] Unit tests for storage
- [ ] Implement `State` determination within new test data classes
- [ ] Refactor remaining access to old test attributes
- [ ] Implement specific repository methods for updating test properties (viewed/consent etc)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTestProcessor.kt
```diff
@@ -0,0 +1,14 @@
...
+interface CoronaTestProcessor {
+
+    val type: CoronaTest.Type
+
+    suspend fun create(data: CoronaTestQRCode): CoronaTest
+
+    suspend fun markSubmitted(test: CoronaTest): CoronaTest
+
+    suspend fun pollServer(test: CoronaTest): CoronaTest
+}
```
mtwalli, line 21: Can we use generics ? 
```kotlin 
interface CoronaTestProcessor<C : CoronaTestQRCode, T : CoronaTest> {

    val type: CoronaTest.Type

    suspend fun create(data: C): T

    suspend fun markSubmitted(test: T): T

    suspend fun pollServer(test: T): T
}
```

and then in derived Processors we won't need to cast:

```kotlin 
class PCRProcessor @Inject constructor(
    private val verificationServer: VerificationServer,
) : CoronaTestProcessor<CoronaTestQRCode.PCR, PCRCoronaTest> {

    override val type: CoronaTest.Type = CoronaTest.Type.PCR

    override suspend fun create(data: CoronaTestQRCode.PCR): PCRCoronaTest {
    }

    override suspend fun markSubmitted(test: PCRCoronaTest): PCRCoronaTest {
       
    }

    override suspend fun pollServer(test: PCRCoronaTest): PCRCoronaTest {
    }
}
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/migration/CoronaTestMigration.kt
```diff
@@ -0,0 +1,85 @@
...
+        val testResultNotificationSent = tracingSettings.isTestResultAvailableNotificationSent
+        Timber.tag(TAG).v("testResultNotificationSent=%s", testResultNotificationSent)
+
+        val legacyPCRTest = PCRCoronaTest(
+            testGUID = LEGACY_GUID,
+            registrationToken = token,
+            registeredAt = devicePairingSuccessfulAt,
+            testResult = when (isAllowedToSubmitKeys) {
+                true -> CoronaTestResult.PCR_POSITIVE
+                else -> CoronaTestResult.PCR_OR_RAT_PENDING
```
mtwalli, line 59: What about other testResult cases such as Negative, ...?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/migration/CoronaTestMigration.kt
```diff
@@ -0,0 +1,85 @@
...
+import de.rki.coronawarnapp.coronatest.type.RegistrationToken
+import de.rki.coronawarnapp.coronatest.type.pcr.PCRCoronaTest
+import de.rki.coronawarnapp.storage.TracingSettings
+import de.rki.coronawarnapp.submission.SubmissionSettings
+import de.rki.coronawarnapp.util.CWADebug
+import timber.log.Timber
+import javax.inject.Inject
+
+@Reusable
+class CoronaTestMigration @Inject constructor(
```
mtwalli, line 15: `PCRTestMigration` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/migration/CoronaTestMigration.kt
```diff
@@ -0,0 +1,85 @@
...
+        tracingSettings.deleteLegacyTestData()
+    }
+
+    companion object {
+        private const val TAG = "CoronaTestMigration"
+
+        /**
+         * We only use this for identification
+         */
+        private const val LEGACY_GUID = "legacy-guid"
```
mtwalli, line 83: "legacy-pcr-guid" ?  As I understand this GUID is used  the HotFlow Map , it is better to distinguish it as PCR

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepositoryExtensions.kt
```diff
@@ -0,0 +1,26 @@
+package de.rki.coronawarnapp.coronatest
+
+import de.rki.coronawarnapp.coronatest.type.CoronaTest
+import de.rki.coronawarnapp.coronatest.type.pcr.PCRCoronaTest
+import de.rki.coronawarnapp.coronatest.type.rapidantigen.RapidAntigenCoronaTest
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+
+val CoronaTestRepository.latestPCRT: Flow<PCRCoronaTest?>
```
mtwalli, line 10: Since this our implementation, and it is part of the repository core functionality I prefer  to move this to the repo as regular properties not an extensions what do you think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationServer.kt
```diff
@@ -96,5 +102,7 @@ class VerificationServer @Inject constructor(
         const val PADDING_LENGTH_BODY_TAN = 31 + VERIFICATION_BODY_FILL
         const val PADDING_LENGTH_BODY_TAN_FAKE = 31 + VERIFICATION_BODY_FILL
         const val DUMMY_REGISTRATION_TOKEN = "11111111-2222-4444-8888-161616161616"
+
+        const val TAG = "VerificationServer"
```
mtwalli, line 106: private

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/CoronaTestStorage.kt
```diff
@@ -1,23 +1,64 @@
...
 
+    var coronaTests: Collection<CoronaTest>
+        get() {
+            Timber.tag(TAG).d("load()")
+            val raw = prefs.getString(PKEY_TESTDATA, null) ?: return emptySet()
+            val tests: List<CoronaTest> = gson.fromJson(raw, typeToken)
+            tests.forEach {
+                Timber.tag(TAG).v("Loaded: %s", it)
+                requireNotNull(it.testGUID)
+            }
```
mtwalli, line 49: I think you can print it directly , I assume it will crash anyway if the parser tried to assign nullable value to non nullable GUID property

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTest.kt
```diff
@@ -2,14 +2,29 @@ package de.rki.coronawarnapp.coronatest.type
...
     val testGUID: CoronaTestGUID
-    val registrationToken: RegistrationToken
     val type: Type
-    val isRefreshing: Boolean
+
+    val registeredAt: Instant
+    val registrationToken: RegistrationToken
+
+    val isProcessing: Boolean
+    val lastError: Throwable?
```
mtwalli, line 15: Do you need to persist those values?

mtwalli:
In general looks good , I like the separation for  handling test lifecycle

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTestProcessor.kt
d4rken: That was my first implementation :smile:. I tried that for an hour and gave up. I didn't manage to get the magic combination of `@JvmSupressWildcards` working. You are very welcome to try it, I'm very curious to know what I overlooked when trying it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/migration/CoronaTestMigration.kt
d4rken: The existing data does not does not tell us about any other cases. We only know that when the user is allowed to submit keys, they had a positive result, in all other cases we have to set it to `PENDING` so the app polls the server for the actual result again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/migration/CoronaTestMigration.kt
d4rken: The only mandatory requirement for it, is to be different from an actual corona test guid (which are UUIDs). I'll use your suggestion though.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepositoryExtensions.kt
d4rken: I'd prefer to keep the allowed I/O operations on the repo minimal. This is just a helper method to filter the tests for a specific one. Adding specific flows for each test type to the repo feels like API bloat.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTest.kt
d4rken: Not in storage, but in memory.
These reflect state and errors during polling (will rename name later when less conflict potential).
If the app dies and we are no longer in memory, then on next launch it's initialised again (when our app dies we are also no longer polling the server :grin:).

The error is used by the UI currently to show a dialog somewhere so "per app session" works ok here too. The subclasses should have this annotated with `@Transient`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/CoronaTestStorage.kt
d4rken: In certain edge cases it is possible to get invalid data, i.e. writing "empty data", and GSON will happily created a class instance and set non-nullable properties to null. This will only throw an exception when accessed.

Printing the object via Timber does not guarantee that the properties are evaluated, especially as the `it` will only be `it.toString()` if logging is actually enabled.

So the safer bet is to just access a specific attribute that should never be null.

In the previous case (I think exposure detection tracker), the issue of "empty data being deserialized" was likely caused writing to files directly and a buffer not being flushed. That's why I also used SharedPreference here, so we don't do direct file access. The previous issue was very very difficult to debug (https://github.com/corona-warn-app/cwa-app-android/pull/1715) due to our limited crashing tracking abilities, and was hotfixed in production.

To not be bitten by that again I'm doing the check here, to crash early, instead when the values are accessed later, which generates weird stacktraces.

chris-cwa (dismissed):

mtwalli:
I got a crash while registering a test:
```
2021-04-16 11:58:09.770 27053-27144/de.rki.coronawarnapp.test E/AndroidRuntime: FATAL EXCEPTION: DefaultDispatcher-worker-12 @coroutine#874
    Process: de.rki.coronawarnapp.test, PID: 27053
    java.lang.IllegalStateException: No test of type PCR available
        at de.rki.coronawarnapp.submission.SubmissionRepository$giveConsentToSubmission$1.invokeSuspend(SubmissionRepository.kt:50)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
        at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
```

d4rken:
> I got a crash while registering a test:

Currently expected, test registration will only work after additional PRs from other team members.
Because some data (like consent) is stored "per test", we need to forward the data until we can actually create the test (with QR Code). 

I can make it not crash, but I think it does not matter, because it will not be able to register a test until PRs from @chiljamgossow @axelherbstreith  and @SamuraiKek are merged.

SamuraiKek:
> I can make it not crash, but I think it does not matter, because it will not be able to register a test until PRs from @chiljamgossow @axelherbstreith and @SamuraiKek are merged.

I'm not sure if my PR would need to be merged for this to work. I think I would actually need to be able to register a test in order to test my PR.

d4rken:
I fixed the crashes, and registering a PCR test will now work, consent is hardcoded for now, this will be fixed by  @SamuraiKek .
Also fixed an issue with storing the tests, and a collision between `stableId`s on the homescreen.

chiljamgossow (approved):

chris-cwa (approved):

PR #2826, opened 2021-04-14
Fix button alignment (DEV)

d4rken:
Missing top padding if checkin texts are single row.

Before:
Screenshot from 2021-04-14 10-36-33

Now:
Screenshot from 2021-04-14 10-48-56


axelherbstreith (approved):
looks way better 👍

AlexanderAlferov (approved):

PR #2833, opened 2021-04-14
Contact Journal special selection not working (EXPOSUREAPP-6093)

BMItr:
This PR addresses (EEXPOSUREAPP-6093).
There was an annoying bug where the user was able to select multiple items for the journal overview screen by selecting too quick.

**How to test:**
try with many quick clicks to select a person or location inside the diary.
Before this change you were able to produces various items inside the overview screen sometimes.
With this fix the behavior is harmonized and not affected e.g. by the duration of the animation.

**Screenshot-before-fix:**
image


LukasLechnerDev (approved):
LGTM! Tested with Google Pixel 3A. Now I can meet Apollo only once per day.

chiljamgossow (commented):
I tested the new version. Creating multiple entries is not possible any more. However, old multiplied entries are still there and cannot be deleted unless I delete the person/location completely. Clicking on those creates more of them.

BMItter:
> I tested the new version. Creating multiple entries is not possible any more. However, old multiplied entries are still there and cannot be deleted unless I delete the person/location completely. Clicking on those creates more of them.

deleting old entries is not part of this ticket.
Create one and solve it if you want.

axelherbstreith (approved):
Nice fix! Quick fingers won't generate duplicated entries anymore 👍 Tested on Samsung Galaxy S8 with Android 10

Issue #2740, opened 2021-04-05
No diary entry shortly after midnight


DerVogel2020:
## Avoid duplicates
* [x ] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x ] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x ] Bug is not already reported in another issue

## Technical details

- Device name: Samsung Galaxy A5 2017
- Android version: 8
- App version: 1.15.2

## Describe the bug

When opening the diary shortly after midnight, there is no entry for today. See screenshot.
![20210405_073017.jpg](https://user-images.githubusercontent.com/65366950/113542219-19a11000-95e4-11eb-8d41-ebe8d9491c4f.jpg)
At about 1:10, there still was no entry.
At about 7:00, the entry appeared.

## Steps to reproduce the issue

Open the diary shortly after midnight. 

## Expected behaviour

Show the diary entry foe today at 0:01.
## Additional context

Maybe it's because of the time change at end of March.

JIRA: [6092](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6092)

Ein-Tim:
@DerVogel2020 

What happens if you 

a) close the app 
b) switch to the main screen and go back to the contact diary? 

Is there still no entry?

MikeMcC399:
@DerVogel2020 

It looks like the calendar is working on UTC time not local time (CEST = UTC+2 currently).

I enabled flight mode, disabled automatic date and time, then set the clock to tomorrow (Tuesday April 6, 2021)
00:10 then
01:10 then
02:10

It was only when I got to 2 AM that the calendar entry for Tuesday appeared. I tried Force Stopping the app each time to make sure it was not a refresh issue.

Afterwards I set the clock back again to 01:10 and the Tuesday entry disappeared.

Great catch!

DerVogel2020:
> @DerVogel2020 
> 
> What happens if you 
> 
> a) close the app 
> b) switch to the main screen and go back to the contact diary? 
> 
> Is there still no entry? 

Tried both after manually changing date and time. 
The entry didn't appear.

DerVogel2020:
> @DerVogel2020 
> 
> It looks like the calendar is working on UTC time not local time (CEST = UTC+2 currently).
> 
[...]

PR #2835, opened 2021-04-15
Contact Journal time switching at midnight (EXPOSUREAPP-6092, EXPOSUREAPP-6451)

BMItr:
This PR addresses Exposureapp-6092 (Contact journal is switching too late).

**Problem-Description:**
Contact journal is switching /generating a new day one hour late (sometimes only many hours later as reports mention).
This creates a situation where user are not able to edit their diary for the new day e.g. right after midnight.

**How to test:**
- timetravel e.g. to 23:58
- switch to the CWA APP
- check if the Contact Journal creates a new Day at 0:00 (Journal Overview Screen)

->should happen automatically without reopening the diary now (with this fix)

**Screenshot:**
diary_overview_better


MikeMcC399:
Link back to #2740.

LukasLechnerDev (approved):
LGTM. Works like a charm. Tested on Samsung S20+

chris-cwa (approved):

PR #2836, opened 2021-04-15
Fix risk calc (EXPOSUREAPP-6363)

chiljamgossow:
The bug was caused by the ConfigChangeDetector which has cleared the table with matches, resulting in the disappearance of risk encounters.

On config changes, existing matches need to be kept and a new calculation must be triggered with new config

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
@@ -1,6 +1,7 @@
 package de.rki.coronawarnapp.appconfig
 
 import androidx.annotation.VisibleForTesting
+import de.rki.coronawarnapp.presencetracing.risk.execution.PresenceTracingWarningTask
 import de.rki.coronawarnapp.risk.RiskLevelSettings
 import de.rki.coronawarnapp.risk.RiskLevelTask
 import de.rki.coronawarnapp.risk.storage.RiskLevelStorage
@@ -46,8 +47,17 @@ class ConfigChangeDetector @Inject constructor(
         val oldConfigId = riskLevelSettings.lastUsedConfigIdentifier
         if (newIdentifier != oldConfigId) {
             Timber.tag(TAG).i("New config id ($newIdentifier) differs from last one ($oldConfigId), resetting.")
-            riskLevelStorage.clear()
-            taskController.submit(DefaultTaskRequest(RiskLevelTask::class, originTag = "ConfigChangeDetector"))
+            riskLevelStorage.clearResults()
+            taskController.submit(
+                DefaultTaskRequest(RiskLevelTask::class, originTag = "ConfigChangeDetector")
+            )
+            taskController.submit(
+                DefaultTaskRequest(
+                    PresenceTracingWarningTask::class,
+                    arguments = PresenceTracingWarningTask.Arguments(true),
+                    originTag = "ConfigChangeDetector"
+                )
+            )
         } else {
             Timber.tag(TAG).v("Config identifier ($oldConfigId) didn't change, NOOP.")
         }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
@@ -54,7 +54,6 @@ class ConfigChangeDetector @Inject constructor(
             taskController.submit(
                 DefaultTaskRequest(
                     PresenceTracingWarningTask::class,
-                    arguments = PresenceTracingWarningTask.Arguments(true),
                     originTag = "ConfigChangeDetector"
                 )
             )
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapper.kt
@@ -6,6 +6,8 @@ import de.rki.coronawarnapp.risk.DefaultRiskLevels.Companion.inRange
 import de.rki.coronawarnapp.risk.RiskState
 import de.rki.coronawarnapp.risk.mapToRiskState
 import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
 import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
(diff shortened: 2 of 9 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
```diff
@@ -46,8 +47,17 @@ class ConfigChangeDetector @Inject constructor(
...
-            riskLevelStorage.clear()
-            taskController.submit(DefaultTaskRequest(RiskLevelTask::class, originTag = "ConfigChangeDetector"))
+            riskLevelStorage.clearResults()
+            taskController.submit(
+                DefaultTaskRequest(RiskLevelTask::class, originTag = "ConfigChangeDetector")
+            )
+            taskController.submit(
+                DefaultTaskRequest(
+                    PresenceTracingWarningTask::class,
+                    arguments = PresenceTracingWarningTask.Arguments(true),
```
d4rken, line 57: We need to add the arguments in other places with 

```
            taskController.submit(
                DefaultTaskRequest(
                    PresenceTracingWarningTask::class,
```

too, otherwise the `arguments as Arguments` will through a cast exception.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapper.kt
```diff
@@ -8,12 +8,19 @@ import de.rki.coronawarnapp.risk.mapToRiskState
...
 
+@Singleton
 class PresenceTracingRiskMapper @Inject constructor(
     private val configProvider: AppConfigProvider
 ) {
     private var presenceTracingRiskCalculationParamContainer: PresenceTracingRiskCalculationParamContainer? = null
 
+    fun clearConfig() {
+        Timber.tag(TAG).i("Clearing config params.")
+        presenceTracingRiskCalculationParamContainer = null
```
d4rken, line 21: I don't think this is threadsafe:

```kotlin
 private suspend fun getRiskCalculationParameters(): PresenceTracingRiskCalculationParamContainer? {
        if (presenceTracingRiskCalculationParamContainer == null) {
            presenceTracingRiskCalculationParamContainer =
                configProvider.currentConfig.first().presenceTracing.riskCalculationParameters
            Timber.d(presenceTracingRiskCalculationParamContainer.toString())
// <------------------------ what if now clearConfig() is called?
        }
        return presenceTracingRiskCalculationParamContainer
    }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/storage/PresenceTracingRiskRepository.kt
```diff
@@ -144,9 +144,14 @@ class PresenceTracingRiskRepository @Inject constructor(
         get() = timeStamper.nowUTC.minus(Days.days(15).toStandardDuration())
 
     suspend fun clearAllTables() {
+        Timber.i("Deleting all matches and results.")
         traceTimeIntervalMatchDao.deleteAll()
         riskLevelResultDao.deleteAll()
     }
+
+    suspend fun clearResults() {
+        riskLevelResultDao.deleteAll()
```
d4rken, line 153: Extra log line here please.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/ConfigChangeDetector.kt
d4rken: You were right and the arguments may not be necessary, so my comment is obsolete once refactored.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapper.kt
chiljamgossow: I hope the mutex does the trick

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapper.kt
```diff
@@ -46,11 +59,15 @@ class PresenceTracingRiskMapper @Inject constructor(
         getRiskCalculationParameters()?.normalizedTimePerCheckInToRiskLevelMapping
 
     private suspend fun getRiskCalculationParameters(): PresenceTracingRiskCalculationParamContainer? {
-        if (presenceTracingRiskCalculationParamContainer == null) {
-            presenceTracingRiskCalculationParamContainer =
-                configProvider.currentConfig.first().presenceTracing.riskCalculationParameters
-            Timber.d(presenceTracingRiskCalculationParamContainer.toString())
+        mutex.withLock {
```
d4rken, line 62: I think we could get compiler nullchecks from kotlins and return non-null here, can you try:

```kotlin
private suspend fun getRiskCalculationParameters(): PresenceTracingRiskCalculationParamContainer = mutex.withLock {
        presenceTracingRiskCalculationParamContainer.let {
            if (it == null) {
                val newParams = configProvider.currentConfig.first().presenceTracing.riskCalculationParameters
                Timber.d("New params %s", newParams)
                presenceTracingRiskCalculationParamContainer = newParams
                newParams
            } else {
                it
            }
        }
    }
```

d4rken (commented):
Looks good, small code improvement maybe.

d4rken (approved):
LGTM!

harambasicluka (commented):
Did the device test, works as expected! 🚀

kolyaopahle (approved):
LGTM 👍 i assume that `configProvider.currentConfig.first().presenceTracing.riskCalculationParameters` is guranteed to not be null by the app config parser which would error before we would ever get here right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapper.kt
```diff
@@ -6,51 +6,70 @@ import de.rki.coronawarnapp.risk.DefaultRiskLevels.Companion.inRange
...
-        getRiskCalculationParameters()?.normalizedTimePerCheckInToRiskLevelMapping
+        getRiskCalculationParameters().normalizedTimePerCheckInToRiskLevelMapping
 
-    private suspend fun getRiskCalculationParameters(): PresenceTracingRiskCalculationParamContainer? {
-        if (presenceTracingRiskCalculationParamContainer == null) {
-            presenceTracingRiskCalculationParamContainer =
-                configProvider.currentConfig.first().presenceTracing.riskCalculationParameters
-            Timber.d(presenceTracingRiskCalculationParamContainer.toString())
+    private suspend fun getRiskCalculationParameters(): PresenceTracingRiskCalculationParamContainer = mutex.withLock {
+        presenceTracingRiskCalculationParamContainer.let {
```
BMItter, line 62: Simply use
presenceTracingRiskCalculationParamContainer ?: run { ... }

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/risk/calculation/PresenceTracingRiskMapperTest.kt
```diff
@@ -100,5 +101,26 @@ class PresenceTracingRiskMapperTest : BaseTest() {
         }
     }
 
+    @Test
+    fun `config is requested only once`() {
+        runBlockingTest {
```
BMItter, line 106: mapper.apply { ... } for style points

BMItter:
lgtm! two little comments..

PR #2860, opened 2021-04-19
Wire up new consent (EXPOSUREAPP-6532)

mtwalli:
## Important Notes
- CheckIns screen should not be displayed if there are no completed check-ins , meaning the flow should be like as it is at the moment
- Android back action should behave as `X` button and the behaviour is explained below

## Case 1: User has given CWA submission consent (this is not the access TEKs ENS Consent)
* Start at `SubmissionTestResultAvailableFragment`
* Press NEXT, If TEK consent is available
	* From `SubmissionTestResultAvailableFragment` go to `CheckInsConsentFragment`
* Press NEXT, If TEK consent is not available
	* Get TEK consent on this screen (TEKHistoryUpdater)
	* If user shared TEKS then go to `CheckInsConsentFragment`
	* If user didn't share TEKs, stay
* On `CheckInsConsentFragment` and (`SubmissionRepository.hasViewedTestResult==false`)
	* If the user presses X
		* Show the same abort dialog as on `SubmissionTestResultAvailableFragment`
			* If the user choses "don't show test result" -> go to home screen
			* If the user choses "show test result" -> user stays on the same screen (so they can give consent) (to advance to the test result screen, the user has to either press NEXT or SKIP)
	* If the user presses NEXT
		* (because at least 1 entry is selected) we update the checkins in the database
		* We set `AutoSubmission.updateMode(MONITOR)`
		* We navigate to `SubmissionTestResultConsentGivenFragment`
	* If the user presses SKIP
		* We show a confirmation dialog 
			* If user chooses "Share" we stay on the same screen
			* If the user choses don't share go to the `SubmissionTestResultConsentGivenFragment`
				* We set `AutoSubmission.updateMode(MONITOR)`

## Case 2: User has not given CWA submission consent
 * Start at `SubmissionResultPositiveOtherWarningNoConsentFragment`
 * Press AGREE, If TEK consent is available
 	* We go to `CheckInsConsentFragment`
 * Press AGREE, if TEK is not available
 	* Get TEK consent on this screen (TEKHistoryUpdater)
 	* Then go to `CheckInsConsentFragment` if user shared TEKs
 	* If user didn't share TEKs, stay on screen
* On `CheckInsConsentFragment` and (`SubmissionRepository.hasViewedTestResult==true`)
	* If the user presses X
		* Show same dialog as SKIP option
	* If the user presses NEXT
		* (because at least 1 entry is selected) we update the checkins database
		* We set `AutoSubmission.updateMode(MONITOR)`
		* We go to `SubmissionResultReadyFragment`
	* If we press SKIP
		* We show a confirmation dialog 
			* If the user presses "DONT SHARE"
				* We set `AutoSubmission.updateMode(MONITOR)`
				* Go to `SubmissionResultReadyFragment`
			* If the user presses "SHARE" we stay on the screen

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInRepositoryExtension.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInRepositoryExtension.kt
@@ -1,4 +1,4 @@
-package de.rki.coronawarnapp.ui.presencetracing.attendee.checkins.common
+package de.rki.coronawarnapp.presencetracing.checkins.common
 
 import de.rki.coronawarnapp.eventregistration.checkins.CheckIn
 import de.rki.coronawarnapp.eventregistration.checkins.CheckInRepository
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModel.kt
@@ -8,23 +8,30 @@ import dagger.assisted.AssistedFactory
 import dagger.assisted.AssistedInject
 import de.rki.coronawarnapp.eventregistration.checkins.CheckIn
 import de.rki.coronawarnapp.eventregistration.checkins.CheckInRepository
+import de.rki.coronawarnapp.submission.SubmissionRepository
+import de.rki.coronawarnapp.submission.auto.AutoSubmission
+import de.rki.coronawarnapp.ui.presencetracing.attendee.checkins.common.completedCheckIns
 import de.rki.coronawarnapp.util.coroutine.DispatcherProvider
+import de.rki.coronawarnapp.util.ui.SingleLiveEvent
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactory
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.combine
+import kotlinx.coroutines.flow.first
 import timber.log.Timber
 
 class CheckInsConsentViewModel @AssistedInject constructor(
     @Assisted private val savedState: SavedStateHandle,
     dispatcherProvider: DispatcherProvider,
-    checkInRepository: CheckInRepository,
+    private val checkInRepository: CheckInRepository,
+    private val submissionRepository: SubmissionRepository,
+    private val autoSubmission: AutoSubmission
 ) : CWAViewModel(dispatcherProvider) {
 
     private val selectedSetFlow = MutableStateFlow(initialSet())
 
     val checkIns: LiveData<List<CheckInsConsentItem>> = combine(
-        checkInRepository.checkInsWithinRetention,
+        checkInRepository.completedCheckIns,
         selectedSetFlow
     ) { checkIns, ids ->
         mutableListOf<CheckInsConsentItem>().apply {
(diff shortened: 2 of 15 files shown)


chiljamgossow (commented):
I tested and it seems correct, though I'm definitely confused by the flow and different consents. Sometimes it crashes after scanning a QR code (in SubmissionStateProvider else -> if (CWADebug.isDeviceForTestersBuild) throw IllegalStateException(eval.toString()) else TestPending) May not be related to these changes.

harambasicluka:
> after scanning a QR code (in SubmissionStateProvider else -> if (CWADebug.isDeviceForTestersBuild) throw IllegalStateException(eval.toString()) else TestPending) May not be related to these changes.

Yes that's not related, @chris-cwa is working on a fix, could maybe share with him how you produced the crash? He'll add some more log statements do understand better whats happening here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInRepositoryExtension.kt
```diff
@@ -0,0 +1,13 @@
+package de.rki.coronawarnapp.presencetracing.checkins.common
+
+import de.rki.coronawarnapp.eventregistration.checkins.CheckIn
+import de.rki.coronawarnapp.eventregistration.checkins.CheckInRepository
+import kotlinx.coroutines.flow.map
+
+/**
+ * Returns completed [CheckIn]s only
+ */
+val CheckInRepository.completedCheckIns
```
kolyaopahle, line 10: couldn't we just add this to the `CheckInRepository` does it need to be an extension function?

kolyaopahle (commented):
Code looks good to me, will test on device and approve if nothing explodes :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInRepositoryExtension.kt
d4rken: I like this pattern.
It's easy to test, and the Repository itself does not need to know about every extra usecase, keeping the API it offers slim.
Having many flows in a repository, makes it difficult for other tests too, as we would have to mock multiple flows, could be subscribed too, instead of just mocking a single flow (and the extension function does not need to be mocked).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModel.kt
```diff
@@ -33,17 +40,58 @@ class CheckInsConsentViewModel @AssistedInject constructor(
...
         val idsWithConsent = selectedSetFlow.value
         checkInRepository.updateSubmissionConsents(
             checkInIds = idsWithConsent,
             consent = true,
         )
-        // TODO persist selection and proceed to submitting check-ins
+
+        if (submissionRepository.hasViewedTestResult.first()) {
+            Timber.d("Navigate to SubmissionResultReadyFragment")
+            events.postValue(CheckInsConsentNavigation.ToSubmissionResultReadyFragment)
```
BMItter, line 58: use something like val event = when (....)
events.postValue(event)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModel.kt
```diff
@@ -33,17 +40,58 @@ class CheckInsConsentViewModel @AssistedInject constructor(
...
+        } else {
+            Timber.d("Navigate to SubmissionTestResultConsentGivenFragment")
+            events.postValue(CheckInsConsentNavigation.ToSubmissionTestResultConsentGivenFragment)
```
BMItter, line 61: also in the following..

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModelTest.kt
```diff
@@ -226,18 +236,94 @@ class CheckInsConsentViewModelTest : BaseTest() {
     }
 
     @Test
-    fun shareSelectedCheckIns() {
-        // TODO test navigation
+    fun `Confirming cancel goes to home screen`() {
+        val viewModel = createViewModel()
```
BMItter, line 240: createViewModel().apply{...} ?
(or run...)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModel.kt
mtwalli: is not that already happening here ? or what do you mean?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModel.kt
mtwalli: Here too?

kolyaopahle (approved):
ui flow worked on a S9 👍

BMItter:
UI flow also tested successfully with OnePlus 8

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/consent/CheckInsConsentViewModelTest.kt
mtwalli: > createViewModel().apply{...} ?
> (or run...)

done

chiljamgossow (approved):

BMItr (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInRepositoryExtension.kt
kolyaopahle: hmm true, i was mainly concerned about fragmenting responsibilities into too many places and making it hard to figure out where an implementation is coming from, but i guess ides should help with that 😄

PR #2865, opened 2021-04-19
Antigene Test Result Notification (EXPOSUREAPP-6039)

chris-cwa:
new notification service for rat tests - code review only, this can be tested after other PRs are merged


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/TestResultAvailableNotificationService.kt
```diff
@@ -1,25 +1,26 @@
...
 import de.rki.coronawarnapp.util.notifications.setContentTextExpandable
 import kotlinx.coroutines.flow.first
 import timber.log.Timber
-import javax.inject.Inject
 import javax.inject.Provider
 
-class TestResultAvailableNotificationService @Inject constructor(
-    @AppContext private val context: Context,
+open class TestResultAvailableNotificationService(
+    private val context: Context,
```
mtwalli, line 17: @AppContext  ?

mtwalli (commented):
LGTM , please fix the test? then I will approve

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/notification/TestResultAvailableNotificationService.kt
chris-cwa: this class won't be injected, only subclasses of it, so no injection annotations needed, i guess?

mtwalli (approved):
Thanks

d4rken (approved):
lgtm

PR #2867, opened 2021-04-19
Fix wrongly colored QR code icon corners (COMMUNITY)

fynngodau:
Please see these screenshots:

Before | After
---|---
Screenshot_1618835386 | Screenshot_1618835408

As you can see, there currently is some incorrect white space in the top corners of the QR code icon gradient background. This can be seen better while holding down the card, but the color is in fact a different one than the card's background while it is resting.

This is happening because `R.drawable.trace_location_view_cardhighlight_background` is set as a background for `TraceLocationCardHighlightView`, while `containerView` draws the real background (above its parent view) (set through `setCaption`). Removing the incorrect background fixes this issue.

Please excuse the CCTG branding, the issue of course also happens with CWA just the same.


AlexanderAlferov:
Thank you for bringing this. This is amazing how you actually spotted this! 🔍 
I just have tested your solution on Samsung S8 and I could confirm that it works like a charm.

Screenshot with 740% zoom in image editor:
Screenshot 2021-04-19 at 16 47 17

AlexanderAlferov (approved):

mtwalli (approved):

Issue #2683, opened 2021-03-23
Integrate battery optimization dialog into onboarding


fynngodau:
### Current situation

After completing the onboarding, users are facing a series of dialogs that ask them to disable battery optimizations (along with some additional information on exposure notifications). Those dialogs are merely a popup, potentially annoying and not looking important to users.

| Dialog with information on exposure notifications | Dialog asking users to disable battery optimizations | Additional dialog if battery optimizations are disabled |
|---|---|---|
| ![Screenshot_1616846248](https://user-images.githubusercontent.com/16943720/112720022-5aba6580-8efc-11eb-92bd-54ae479a5d25.png) | ![Screenshot_1616846253](https://user-images.githubusercontent.com/16943720/112720025-5c842900-8efc-11eb-9700-67c0e3ba88e1.png) | ![Screenshot_1616846261](https://user-images.githubusercontent.com/16943720/112720026-5db55600-8efc-11eb-8ab0-e1066b300d01.png) |

### Suggested improvement

The onboarding flow should contain a page that asks users to disable battery optimizations.

### Expected benefits

The onboarding flow feels better when it is actually complete when it seems to be done, instead of being followed by more popup dialogs. The rate of users who disable battery optimizations might improve, which would be good.

### See also
#2682 Warn about deactivated battery optimization on main screen
#1606 Improve UX for disabling battery optimization (an issue about using `ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS`)

For any readers having trouble disabling the battery optimization, there are some instructions in the FAQ
https://www.coronawarn.app/en/faq/#no_risk_update about how to do this.

---
Internal Tracking ID: [EXPOSUREAPP-5965](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5965)

MikeMcC399:
@fynngodau 

This is very similar to the enhancement request #1606 which you already quoted.

Maybe the devs could fill us in on the decision which led to this function being removed? I guess it was because it wasn't reliable across all vendors.

I can quite imagine that most users will ignore the dialog, as you say. It's difficult for somebody unfamiliar with the concepts to understand what is being requested and what needs to be done.

It was interesting to read the feedback from @thomasaugsten in https://github.com/corona-warn-app/cwa-app-android/issues/2682#issuecomment-805142879 about how other methods are being used to mitigate the issue of battery optimization.

fynngodau:
[...]

PR #2868, opened 2021-04-19
Make exposure logging functionality popup not cancelable (COMMUNTIY)

fynngodau:
For reference, we're looking at this popup:

exposure_logging_functionality_popup

Contrary to other popups that are shown upon initially opening the home screen (#2683 contains screenshots of all of them), this popup is cancelable. However, if it is dismissed, it will be shown again frequently. Therefore, it would be better to make it non-dismissable.

Long-term, this should be integrated into onboarding in my opinion, like https://github.com/corona-warn-app/cwa-app-android/issues/2683.


d4rken (approved):
lgtm

harambasicluka (approved):

PR #2873, opened 2021-04-19
Incompatibility warning card on main screen (EXPOSUREAPP-5464)(COMMUNITY)

d4rken:
Shows warning if device hardware is not supporting BLE scanning. Card opens FAQ link on click.

How to test:
Use some old phone running Android 6 or the emulator.
Eg. Nexus 5 displays the warning. Fairphone 2 is mentioned in the ticket.

Continuation of #2481 by @fynngodau 
Rebased and refactored.

Also see comments in #2481

---
Internal Tracking ID: [EXPOSUREAPP-5464](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5464)

device-2021-04-26-093630


fynngodau:
Thank you so much for picking this one up!

vaubaehn:
@fynngodau 
I'm very happy that this will go into 2.2. At that time, Event Registration (=presence tracing) will already be the second main feature. I wonder, if the currently proposed texts for the cards of this PR might get confusing for the less advanced users, and if they might be irritated that the app as a whole is not working, eventhough the incompatibility warning refers to RPI receiving/submitting bluetooth functionality only.
Would it make sense to adapt the texts?
For (a very quick and dirty) example: Unfortunately, your device is only partially compatible with Exposure Notifications `that work with bluetooth`. You can be notified for risk contacts `that were in range of your bluetooth receiver `but won't be able to warn others `in the same way`. `Exposure notifications from registered events (Event Registration) and warning others in that way will work as expected.`
And: Unfortunately, your device is not compatible with Exposure Notifications `that work with bluetooth`. `Exposure notifications from registered events (Event Registration) and warning others in that way will work as expected.`
(the challenge is, to be as precise and as short as possible...)
This might be even more important, when support for Android 5 is about to come, as more people with compatibility limitations will adopt CWA, imho.
What do you think?

fynngodau:
@vaubaehn Thanks for the great hint. In fact I had already [changed](https://codeberg.org/corona-contact-tracing-germany/cwa-android/commit/8b860285572bd1cda98cbf1e55430b4dea57aacf) the strings for CCTG to the following:

* Unfortunately, your device is only partially compatible with Bluetooth-based Exposure Notifications. You can be notified for risk contacts but won't be able to warn users outside of locations and events you checked in to.
* Unfortunately, your device is not compatible with Bluetooth-based Exposure Notifications. Only Check-In functionality is supported on your device.

(Though looking at it now, it should be "of" risk contacts, not "for".)

vaubaehn:
@fynngodau 
That's perfect imho!

fynngodau:
A good change would be to add an "open externally" icon to the card to indicate that it can be clicked. You can use this commit for that: https://codeberg.org/corona-contact-tracing-germany/cwa-android/commit/a22275e6e75af1b75b8141bed07626c14153d940

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1962,7 +1962,23 @@
     <!-- XACT: Button/Dialog label to cancel something-->
     <string name="generic_action_abort">"Cancel"</string>
     <!-- XACT: Button/Dialog label to remove something-->
-    <string name="generic_action_remove">"Remove"</string>
+    <string name="generic_action_remove">Remove</string>
```
harambasicluka, line 1965: @chiljamgossow we need german strings, english strings in this file won't be recognized by the translation system. Please sync with @janetback for the correct strings.

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1962,7 +1962,23 @@
...
+        partially compatible with Bluetooth-based Exposure Notifications. You can be notified of
+        risk contacts but won\'t be able to warn users outside of locations and events you have
+        checked in to.
+    </string>
+    <string name="incompatible_scanning_not_supported">Unfortunately, your device is not compatible
+        with Bluetooth-based Exposure Notifications. Only check-In functionality is supported on
+        your device.
+    </string>
+
+    <string name="incompatible_link">"https://www.coronawarn.app/en/faq/#incompatibility_warning"</string>
```
harambasicluka, line 1981: @chiljamgossow we need german strings, english strings in this file won't be recognized by the translation system. Please sync with @janetback for the correct strings.

chiljamgossow (dismissed):
Well done 😉

Corona-Warn-App/src/main/res/values/strings.xml
chiljamgossow: We want German in the English string file?

harambasicluka: No, we don't provide english texts this come from our translators. We provide german texts and then get the english ones. To satisfy our Linter we still need the empty strings here or you can put the germans one in (to not cause any crashes) after you also added them to the german string file. The same process as for all other strings applies here.

chiljamgossow: removed

chiljamgossow (approved):

PR #2875, opened 2021-04-19
Event duplication wrong time (EXPOSUREAPP-6439)

jurajkusnier:
- Fixing wrong date/time displayed in `QrDetailFragment`, `TraceLocationsFragment`, `TraceLocationCreateFragment` 
- Add instrumentation tests


.circleci/config.yml
```diff
@@ -162,6 +162,7 @@ commands:
             -no-boot-anim
             -memory 2048
             -nojni
+            -timezone Europe/Berlin
```
mtwalli, line 165: why , I recall some situations where the different timezone in CI helped us to spot some issues and you already doing that in the tests that require this time zone

mtwalli (commented):
LGTM, tested and I confirm the fix

.circleci/config.yml
jurajkusnier: I was testing multiple approaches, I guess this is not necessary anymore, I'll try to remove it and check if the test pass

mtwalli (approved):

BMItr (approved):
lgtm, tested on pixel 2 device

PR #2879, opened 2021-04-20
Update Deletion Warning Screen Wiring (DEV)

axelherbstreith:
Fixes an issue that prevented to pass the submission flow when there is already a existing test of the same type

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningViewModel.kt
@@ -20,6 +20,7 @@ import de.rki.coronawarnapp.ui.submission.viewmodel.SubmissionNavigationEvents
 import de.rki.coronawarnapp.util.ui.SingleLiveEvent
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactory
+import kotlinx.coroutines.flow.first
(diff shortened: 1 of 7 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningViewModel.kt
```diff
@@ -36,7 +37,8 @@ class SubmissionDeletionWarningViewModel @AssistedInject constructor(
     val registrationError = SingleLiveEvent<CwaWebException>()
 
     fun deleteExistingAndRegisterNewTest() = launch {
-        coronaTestRepository.removeTest(coronaTest.identifier)
+        val currentTest = submissionRepository.testForType(coronaTest.type).first()
+        coronaTestRepository.removeTest(currentTest!!.identifier)
```
mtwalli, line 41: should you catch the error here ? so, this would not crash for the user

mtwalli (approved):
LGTM , just one question

SamuraiKek (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningViewModel.kt
axelherbstreith: I think it is not needed in this case since I only navigate to the warning deletion screen if there is another test of the same type stored.

mtwalli (dismissed):

mtwalli (approved):

SamuraiKek (approved):

PR #2882, opened 2021-04-20
Test registration polish (EXPOSUREAPP-6007)

d4rken:
* Fixes a crash on exception within `CoronaWarnRepository` as exceptions could not be effectively caught. Small refactoring to the error handling behavior of `HotDataFlow<T>` to remove this :foot::gun:.
    * Crashed when scanning a second test of the same type (e.g. via test menu)
* Add registration date display to PCR and RAT test cards.
* Fix `PCRCoronaTest?.toSubmissionState()` for submitted test state.
* Change polling conditions, polling the server is only blocked if we are in the state `testResult==EXPIRED` state.
* Update testResultReceivedAt timestamp only in final states (POS/NEG/EXPIRED).
* `ShareTestResultNotificationService` now listens to `CoronaTestRepository` and can make more decisions on it's own. This needs a follow work to evaluate behavior in multi test scenarios.
* Adapt `isSubmissionAllowed`  from `result==positive` to `result==positive && !submitted`


chiljamgossow (dismissed):
Tested. Seems to fix the crash after scanning an invalid RAT code.

chiljamgossow (approved):

AlexanderAlferov (approved):
LGTM 🚀

PR #2886, opened 2021-04-21
RAT registration via link (EXPOSUREAPP-6313)

chiljamgossow:
RAT QR code registration via link

Registration can now be triggered from the consent screen as well in case the qr code is already present.

How to test:
- Click on a RAT qr code (link) -> CWA should be offered, open up on the consent screen and move forward without having to pass the scan screen
- Register a test (pcr, ra) via scan -> should work as before
- Scan a check-in qr code with camera app-> should still work as before


Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -79,14 +79,20 @@
...
                 <category android:name="android.intent.category.DEFAULT" />
                 <category android:name="android.intent.category.BROWSABLE" />
 
                 <data
                     android:host="e.coronawarn.app"
                     android:scheme="https" />
+
+                <data
+                    android:host="s.coronawarn.app"
+                    android:pathPattern=".*"
```
mtwalli, line 91: I think the filter here is not needed?  it works without it

chiljamgossow: it does not work without

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/QrCodeRegistrationStateProcessor.kt
```diff
@@ -0,0 +1,74 @@
...
+            submissionRepository.removeTestFromDevice(type = coronaTestQRCode.type)
+            showRedeemedTokenWarning.postValue(Unit)
+        } catch (err: Exception) {
+            registrationState.postValue(RegistrationState(ApiRequestState.FAILED))
+            err.report(ExceptionCategory.INTERNAL)
+        }
+
+    private fun checkTestResult(testResult: CoronaTestResult) {
+        if (testResult == CoronaTestResult.PCR_REDEEMED) {
+            throw InvalidQRCodeException()
```
mtwalli, line 67: Please add some message to the exception? Let us borrow  it from https://github.com/corona-warn-app/cwa-app-android/pull/2894/commits/8083fef55febe73b5fc488c4f114ab65c4137078#diff-58ba20a44ecdb979466f8e5c5038e91330761f46f7834a97e5f8555650975418R118

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
```diff
@@ -23,10 +33,14 @@ class SubmissionConsentFragment : Fragment(R.layout.fragment_submission_consent)
...
     private val viewModel: SubmissionConsentViewModel by cwaViewModels { viewModelFactory }
     private val binding: FragmentSubmissionConsentBinding by viewBindingLazy()
+    private val navArgs by navArgs<SubmissionConsentFragmentArgs>()
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         binding.viewModel = viewModel
+        navArgs.qrCode?.let {
+            viewModel.qrCode = it
+        }
```
mtwalli, line 43: how do you feel about passing this the viewModel as `Assisted` parameter ?

Corona-Warn-App/src/main/AndroidManifest.xml
mtwalli: Ok, for me I did work , maybe you can add it also the other link if that is made work on your device ? who know what other devices have the same behaviour .

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
```diff
@@ -46,11 +60,75 @@ class SubmissionConsentFragment : Fragment(R.layout.fragment_submission_consent)
...
+                            )
+                    )
+                }
             }
         }
         viewModel.countries.observe2(this) {
             binding.countries = it
         }
+
+        viewModel.showRedeemedTokenWarning.observe2(this) {
```
mtwalli, line 78: btw you can use `observe` method now, with the latest version of the lifecycle library we will get a lint error if we tired to pass `this`. In Dialog fragments it is Ok to use `this` though . just mentioning it ,you don't have to change it 
cc/ @d4rken

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
```diff
@@ -46,11 +60,75 @@ class SubmissionConsentFragment : Fragment(R.layout.fragment_submission_consent)
...
+            if (ValidationState.INVALID == it) {
+                showInvalidQrCodeDialog()
+            }
+        }
+
+        viewModel.registrationState.observe2(this) { state ->
+            binding.progressSpinner.visibility = when (state.apiRequestState) {
+                ApiRequestState.STARTED -> View.VISIBLE
+                else -> View.GONE
+            }
```
mtwalli, line 100: `binding.progressSpinner.isVisible = state.apiRequestState == ApiRequestState.STARTED` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentViewModel.kt
```diff
@@ -42,12 +55,50 @@ class SubmissionConsentViewModel @AssistedInject constructor(
...
+                        consentGiven = true
+                    )
+                )
+            } else {
+                qrCodeRegistrationStateProcessor.startQrCodeRegistration(coronaTestQRCode, true)
+            }
+        } catch (err: InvalidQRCodeException) {
+            qrCodeValidationState.postValue(QrCodeRegistrationStateProcessor.ValidationState.INVALID)
+        }
+    }
```
mtwalli, line 100: Please add some logs to help us identify the behaviour if we got some issues?

mtwalli (commented):
Works really nice! 👍, Tested on Samsung S8, Android 9 
1- Before Onboarding 
2- After Onboarding

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
chiljamgossow: Personal opinion: Too complicated for such a simple task

LukasLechnerDev:
Works really nice. But don't we need a different text in the consent screen in this case? It says things like "Scannen Sie nun per QRCode Ihren eigenen Test ..." and "Jeder Test kann nur einmal gescannt werden. " Although no scanning happens in this flow.

chiljamgossow:
> Works really nice. But don't we need a different text in the consent screen in this case? It says things like "Scannen Sie nun per QRCode Ihren eigenen Test ..." and "Jeder Test kann nur einmal gescannt werden. " Although no scanning happens in this flow.

Yes, the text does not make sense. Should probably be adapted to work for both scenarios. Will ask for a solution tomorrow.

LukasLechnerDev (approved):
LGTM! I focused mainly on device testing. (Samsung S20, Android 11)

New functionality works and the existing one still behaves as expected. 

Good work 👍

Corona-Warn-App/src/main/AndroidManifest.xml
mtwalli: Should I do it in another PR ? or will you do it here? 
```
 <data  android:host="e.coronawarn.app"
android:pathPattern=".*"
```
             android:scheme="https" />

chiljamgossow: Did you try both patterns? With and without an extra /

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentViewModel.kt
chiljamgossow: There is a log statement and the validator also logs errors

Corona-Warn-App/src/main/AndroidManifest.xml
mtwalli: Sorry, maybe I was not clear, I mean you to add `android:pathPattern=".*"` to other intent filter of the event registration link, if that  path pattern works on your device. I am suggesting to use the same both intent filter , what do you think ?

chiljamgossow: Let's do it in separate PR. I use a Xiaomi and without .* it just opens the browser.

mtwalli (approved):
Thanks, nice work , please let me know if I need make the changes for Event registration filter myself

PR #2889, opened 2021-04-21
Rapid Antigen Test: Change state to Outdated (EXPOSUREAPP-6010)

kolyaopahle:
### Description
This PR enables the outdated state for RA Tests, if a test is older than the hoursToDeemTestOutdated from appconfig

The current version stores the interval until the test becomes outdated inside of the test data structure (because of this a test will respect the outdated time from the appconfig at time of scanning and not at time of viewing the test)

### Steps to reproduce
1. Scan a RA Qr code
2. wait hoursToDeemTestOutdated (48 on mock server)
3. check if test is marked outdated


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
```diff
@@ -51,20 +52,30 @@ data class RACoronaTest(
...
-        CoronaTestResult.RAT_NEGATIVE -> State.NEGATIVE
-        CoronaTestResult.RAT_POSITIVE -> State.POSITIVE
-        CoronaTestResult.RAT_INVALID -> State.INVALID
-        CoronaTestResult.RAT_REDEEMED -> State.REDEEMED
-        else -> throw IllegalArgumentException("Invalid RAT test state $testResult")
-    }
+    private fun isOutdated(nowUTC: Instant) = testedAt.plus(outdatedAfter).isBefore(nowUTC)
+
+    fun getState(nowUTC: Instant) =
+        if (isOutdated(nowUTC)) {
```
d4rken, line 67: The techspec graph says that a test can go from outdated to expired.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RapidAntigenCoronaTestExtensions.kt
```diff
@@ -30,7 +31,6 @@ fun RACoronaTest?.toSubmissionState(nowUTC: Instant = Instant.now()) = when {
         NEGATIVE -> TestNegative(testRegisteredAt = registeredAt)
         REDEEMED -> TestInvalid
         PENDING -> TestPending
-        // TODO: Should be updated once the logic for OUTDATED tests is in
-        OUTDATED -> TestNegative(testRegisteredAt = registeredAt)
+        OUTDATED -> TestOutdated
```
d4rken, line 34: We don't show a date in this state?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/CoronaTestConfig.kt
```diff
@@ -9,10 +9,10 @@ interface CoronaTestConfig {
 }
 
 data class CoronaRapidAntigenTestParametersContainer(
-    val hoursToDeemTestOutdated: Int = DEFAULT_HOURS
+    val hoursToDeemTestOutdated: Long = DEFAULT_HOURS
 ) {
     companion object {
-        const val DEFAULT_HOURS = 48
+        const val DEFAULT_HOURS = 48L
```
BMItter, line 15: Just add the type 🏎️

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
```diff
@@ -51,20 +52,30 @@ data class RACoronaTest(
     @SerializedName("dateOfBirth")
     val dateOfBirth: LocalDate?,
 
+    @SerializedName("outdatedAfter")
+    val outdatedAfter: Duration,
```
d4rken, line 56: A test is outdated after `testedAt` + `TimeFromConfigConfigTime`.
`testedAt` is fixed, when the test was scanned.
`TimeFromConfigConfigTime` is dynamic and can change whenever the config changes.

I think it would be better to not  store a fixed value `outdatedAfter`, but have a function to call with the current arguments.

e.g. `fun getState(nowUTC, currentConfig): State`

edit:
Thinking about it, the `testedAt` is from an external time source. When determining the `OUTDATED`, we should probably also check `currentConfigisDeviceTimeCorrect`? /cc @mlenkeit

d4rken (changes requested):
Some discussions, needs some clarification too :ghost:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RapidAntigenCoronaTestExtensions.kt
kolyaopahle: I did not create the outdated card, but from looking at the initialisation code i cant see any place where a time is expected

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
kolyaopahle: sure, i was not sure which way makes more sense, should a test that was made when 48 hours were the config expire on the next day if the config is changed to be 12 hours?

checking device time probably makes sense, how would we handle a mismatch?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
kolyaopahle: hmm does it make a difference for the app? Id prefer to not make the state logic even more edgecasy

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
mlenkeit: I agree with Matthias, dynamic evaluation would make sense here.

Device time check is currently not expected for this I think but let me double check...

kolyaopahle: sounds good, removed the outdatedAfter property and switched check to use appconfig
ill keep this unresolved untill there is info on the time check @mlenkeit

mlenkeit: No need to check the device time.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
kolyaopahle: the logic is now implemented as @mlenkeit suggested
`testResult == CoronaTestResult.PCR_NEGATIVE && isOutdated(nowUTC, testConfig) -> Outdated`
otherwise its based on the old state calculation

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RapidAntigenProcessor.kt
```diff
@@ -28,7 +28,7 @@ import javax.inject.Inject
 @Reusable
 class RapidAntigenProcessor @Inject constructor(
     private val timeStamper: TimeStamper,
-    private val submissionService: CoronaTestService,
+    private val submissionService: CoronaTestService
```
d4rken, line 31: Please re-add the commas. Since we upgraded Kotlin we can use the trailing commas, and they prevent merge conflicts if multiple people add new lines to the constructor.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RapidAntigenProcessor.kt
```diff
@@ -50,7 +50,7 @@ class RapidAntigenProcessor @Inject constructor(
             testedAt = request.createdAt,
             firstName = request.firstName,
             lastName = request.lastName,
-            dateOfBirth = request.dateOfBirth,
+            dateOfBirth = request.dateOfBirth
```
d4rken, line 53: :point_up:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -208,48 +211,58 @@ class HomeFragmentViewModel @AssistedInject constructor(
...
-                HomeFragmentDirections
-                    .actionMainFragmentToSubmissionTestResultPendingFragment(testType = CoronaTest.Type.RAPID_ANTIGEN)
-            )
-        }
-        is SubmissionStateRAT.TestPending -> RapidTestPendingCard.Item(state) {
-            routeToScreen.postValue(
-                HomeFragmentDirections
-                    .actionMainFragmentToSubmissionTestResultPendingFragment(testType = CoronaTest.Type.RAPID_ANTIGEN)
-            )
+    private suspend fun coronaTestParameters() = appConfigProvider.getAppConfig().coronaTestParameters
```
d4rken, line 214: Please turn this into a `Flow` subscribing to `appCOnfigProvider.latestConfig`, we do not need to recheck the config on UI level via `getAppConfig` the calls happening within other Tasks are sufficient, and if the config changes, we ideally want the card to be updated too. To prevent flickering we could offer the add `distinctUntilChanged`.

d4rken (commented):
Some minor optimizations please :relaxed:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
kolyaopahle: moved the appconfig resolution to the homeItems flow combine

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/mapping/CoronaTestParametersDistinctTest.kt
```diff
@@ -0,0 +1,25 @@
...
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.test.runBlockingTest
+import org.junit.Test
+import testhelpers.BaseTest
+
+class CoronaTestParametersDistinctTest : BaseTest() {
+
+    @Test
+    fun `can we use distinctUntilChanged on CoronaTestParameters`() = runBlockingTest {
```
d4rken, line 16: :grin:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
```diff
@@ -64,14 +65,22 @@ data class RACoronaTest(
...
-        RAT_POSITIVE -> State.POSITIVE
-        RAT_INVALID -> State.INVALID
-        RAT_REDEEMED -> State.REDEEMED
-        else -> throw IllegalArgumentException("Invalid RAT test state $testResult")
-    }
+    private fun isOutdated(nowUTC: Instant, testConfig: CoronaTestConfig) =
+        testedAt.plus(testConfig.coronaRapidAntigenTestParameters.hoursToDeemTestOutdated).isBefore(nowUTC)
+
+    fun getState(nowUTC: Instant, testConfig: CoronaTestConfig) =
+        if (testResult == CoronaTestResult.PCR_NEGATIVE && isOutdated(nowUTC, testConfig)) {
```
d4rken, line 72: **RA**_NEGATIVE

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
```diff
@@ -64,14 +65,22 @@ data class RACoronaTest(
...
+            State.OUTDATED
+        } else {
+            when (testResult) {
+                PCR_OR_RAT_PENDING -> State.PENDING
```
d4rken, line 76: `PCR_OR_RAT_PENDING || RA_PENDING`

PCR has one pending state, RA has two :shrug:

d4rken (changes requested):
Small bug, once fixed, works well on device.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
kolyaopahle: fixed 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RACoronaTest.kt
kolyaopahle: should be less broken now 😄

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/mapping/CoronaTestParametersDistinctTest.kt
kolyaopahle: 😘

d4rken (approved):

jurajkusnier (approved):

PR #2890, opened 2021-04-21
Checkin - overnight check ins without date on the second day (EXPOSUREAPP-6453)

BMItr:
This PR adresses Exposureapp-6453 (missing date on the second day..)

Steps / Hot to Test:
- create event and check in: for the auto checkout choose a duration which ends on the next day
- on the next day check the My Check-ins list (or time travel)

Expected result:
- the exact check in and check out Date and Time is visible

before result:
- the second date ist missing - only the time is displayed

screenshot (before):
image

Note:
Same Day Checkout -> Only one day should be displayed (no second date)
Different Day Checkout -> 2 day should be displayed


AlexanderAlferov (approved):
LGTM 🚀 

Tested on Samsung S8:
Screenshot 2021-04-21 at 17 38 47

Quick tip for testers: after time traveling you would probably need to wait for about 1 minute to see any changes.

axelherbstreith (approved):
Looks good 👍

PR #2898, opened 2021-04-21
Be more lenient when parsing json from RapidAntigen QRCodes (DEV)

d4rken:
+ We allow empty strings and treat them as null.
+ some extra unit tests


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/RapidAntigenQrCodeExtractor.kt
```diff
@@ -78,15 +50,57 @@ class RapidAntigenQrCodeExtractor @Inject constructor() : QrCodeExtractor<Corona
...
+                return rawHash
+            }
+
+        val createdAt: Instant
+            get() {
+                if (rawTimestamp == null || rawTimestamp <= 0) throw InvalidQRCodeException("Timestamp is invalid")
+                return Instant.ofEpochSecond(rawTimestamp)
+            }
+
+        val firstName: String?
```
BMItter, line 76: what about blank Strings?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/RapidAntigenQrCodeExtractor.kt
```diff
@@ -78,15 +50,57 @@ class RapidAntigenQrCodeExtractor @Inject constructor() : QrCodeExtractor<Corona
...
+            get() {
+                if (rawFirstName.isNullOrEmpty()) return null
+                return rawFirstName
+            }
+
+        val lastName: String?
```
BMItter, line 82: what about blank Strings?...

BMItr (approved):

chiljamgossow (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/RapidAntigenQrCodeExtractor.kt
harambasicluka: @d4rken I think blank strings can be used for the birthdays, not sure about the last name.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/RapidAntigenQrCodeExtractor.kt
d4rken: I don't know. I guess `‎‎‎‏‏‎ ‎‏‏‎ ‎‏‏‎ ‎‏‏‎ ‎    ,  ‏‏‎ ‎‏‏‎ ‎‏‏‎ ‎‏‏‎ ‎    ` should not be excluded from taking RA tests? :shrug: /cc @mlenkeit

PR #2913, opened 2021-04-22
Fix sporadic crash due to IllegalArgumentException on quick navigation (EXPOSUREAPP-6610)

d4rken:
There is a fragment racecondition if you navigate to another fragment and quickly popBackStack().
Our uiHandler.post { } will not have executed for some reason.
In that case we manually null the old viewBinding, to allow for clean recreation.

Can be reproduced by adding "popBackStack()" to the target fragment in onResume or onCreate.

```
  at de.rki.coronawarnapp.util.ui.ViewBindingProperty.getValue (ViewBindingExtensions.kt:2)
  at de.rki.coronawarnapp.ui.main.home.HomeFragment.getBinding (HomeFragment.kt)
  at de.rki.coronawarnapp.ui.main.home.HomeFragment.onViewCreated (HomeFragment.kt:2)
```


AlexanderAlferov (approved):
LGTM 🚀

LukasLechnerDev (approved):

Issue #2880, opened 2021-04-20
PRIO - ENS Timeout? Outdated exposure checks: CWA uses the same diagnosis key package several times in ProvideDiagnosisKeys to ENS


vaubaehn:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details
- Device name: LG-G4 (LGH-815)
- Android version: 6.0
- App version: 1.15.2 (but also other versions had been affected before)
- ENS version: 18210915000 (but also other versions had been affected before)

## Describe the bug
At some days, CWA is providing the same previously downloaded diagnosis key package to ENS, resulting in outdated exposure checks. This already happened to me last year, and is now occuring again.
My assumption is, that a timeout of ENS and an insufficient handling of the error in CWA is the reason, because I'm experiencing the problem only in times with high incidence rates, and at the moment the DK packages contain 500k+ diagnosis keys.
**So, probably many people might be affected currently without knowing it (unless they don't check the logs in ENS UI carefully).**
As you can see from my attached ENS logs (see below) this happened on 10 days since April 1st.
Also, on April 18th at around 9:00 am, CWA correctly indicated, that exposure matching failed (for more that 36 hours):
<img alt="Screenshot_2021-04-18-08-57-47" width="250" src="https://user-images.githubusercontent.com/68376758/115393398-ee244500-a1e1-11eb-90e3-99844ae39b80.png">   <img alt="Screenshot_2021-04-18-08-58-15" width="250" src="https://user-images.githubusercontent.com/68376758/115394088-b23daf80-a1e2-11eb-8ad6-e187cee00c86.png">
When tapping on "Erneut Starten", the spinning wheel of downloading was visible for <1s, then the green card appeared, but exposure matching didn't start actually (as you can see from the logs).

## Steps to reproduce the issue
1. Have an old, slow phone, preferrably Android 6, as there is only one single exposure check per day currently. Wait for rising incidence rates.
2. Let CWA do its job for a couple of days.
[...]

PR #2914, opened 2021-04-22
Fix potentially failing background RiskLevel calculation on slow devices (DEV)

d4rken:
Run the RiskLevelTask within the `ExposureStateUpdateWorker` via `submitBlocking` to retain elevated priority by keeping the worker running, otherwise the system may kill our process while the RiskLevelTask is still running, because the worker isn't.

This came up while debugging #2880.
It is not the cause for the weird ENS log entries reported by @vaubaehn , but may be the cause for seeing the error card on the home screen: The risk level task didn't run reliably as on his device the risk level task takes longer than the system giving us time, due to the worker waiting for the task.

### Testing
* Get an ExposureStateUpdateReceiver that causes the RiskLevelTask to execute while the app is in the background, i.e. get one risklevel calculation while the app was not open.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ExposureStateUpdateWorker.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/ExposureStateUpdateWorker.kt
@@ -12,6 +12,7 @@ import de.rki.coronawarnapp.exception.reporting.report
 import de.rki.coronawarnapp.risk.RiskLevelTask
 import de.rki.coronawarnapp.task.TaskController
 import de.rki.coronawarnapp.task.common.DefaultTaskRequest
+import de.rki.coronawarnapp.task.submitBlocking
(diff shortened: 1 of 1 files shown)


chiljamgossow (approved):

BMItr (approved):

vaubaehn:
@d4rken 
Hi Matthias, sad news - error card still behaves weird: Testing around last night, after pulling CWA to foreground, risk state/card changed (very quickly) from green to error. "Erneut versuchen" triggered CWA (according to the texts on the error card) to a) download DKs ("Bitte warten") and to b) provide them to ENS ("Prüfung läuft") extremly quickly one after the other (1s or less for both).
I could repeat this 6 or 7 times, and after the last "erneut versuchen" the risk card changed to green finally, without any log entry of exposure checking inside ENS UI for that time -> so ENS didn't take over anything.
Beside this, ENS exposure checking seems to work fine for me for now!

PR #2921, opened 2021-04-22
Fix problematic properties during serialization (EXPOSUREAPP-6650, part1)

d4rken:
GSON uses unsafeinit and will initialize a data class without any "val some = thing" being initialized that are in the class body.
So the `RACoronaTest` class, after deserialization would just have `val type: CoronaTest.Type = null` despite being not nullable.

This seems to be a common footgun with GSON, I think we had that issue before. I've removed all uses of `@Transient` in class bodies and replace it with Kotlin property getters.


chiljamgossow (approved):
Works, tests are persisted

SamuraiKek (approved):
Yep, it works. Nice!

PR #2923, opened 2021-04-22
Test result screens polishing (DEV)

SamuraiKek:
Propagates the test type throughout the submission flow and adjusts the test result texts accordingly. 

- Fixes a bug where the app would hang after getting a positive result and the test result screen would not show up.
- Maybe a few more bugs reported so far.


chiljamgossow (approved):
Nice! Fixes crash and many other bugs

axelherbstreith (approved):

PR #2927, opened 2021-04-23
Reduce Logs for production build (EXPOSUREAPP-6642)

kolyaopahle:
### Description
This PR removes some of the logging from `CheckinWarningMatcher` in builds for testers as this produces a huge amount of noise in their logs without much added value

### Steps to reproduce
1. Check into some event
2. Get an positive package submitted
3. Check if tester builds are logging lots of stuff about Matches
4. Check if there is still too much noise in the logs and if anything else should be disabled

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/CheckInWarningMatcher.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/calculation/CheckInWarningMatcher.kt
@@ -5,6 +5,7 @@ import de.rki.coronawarnapp.presencetracing.checkins.CheckIn
 import de.rki.coronawarnapp.presencetracing.checkins.split.splitByMidnightUTC
 import de.rki.coronawarnapp.presencetracing.warning.storage.TraceWarningPackage
 import de.rki.coronawarnapp.server.protocols.internal.pt.TraceWarning
+import de.rki.coronawarnapp.util.CWADebug
 import de.rki.coronawarnapp.util.coroutine.DispatcherProvider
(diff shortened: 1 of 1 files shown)


LukasLechnerDev:
Will this be merged to `2.1`? The story has a fix version `2.2`

LukasLechnerDev:
Doesn't the story say to ONLY log in debug builds?

kolyaopahle:
> Will this be merged to `2.1`? The story has a fix version `2.2`

true, will retarget

> Doesn't the story say to ONLY log in debug builds?

i am honestly not sure, the feedback came from the testers and the logging is disabled in prod anyways isn't it?

harambasicluka:
> > Will this be merged to `2.1`? The story has a fix version `2.2`
> 
> true, will retarget
> 
> > Doesn't the story say to ONLY log in debug builds?
> 
> i am honestly not sure, the feedback came from the testers and the logging is disabled in prod anyways isn't it?

Log will be activated in 2.2, so it should be only deactivated for production logging to not have huge log files. But we still want to see this in the logs we receive from our testers.

kolyaopahle:
> > > Will this be merged to `2.1`? The story has a fix version `2.2`
> > 
> > 
> > true, will retarget
> > > Doesn't the story say to ONLY log in debug builds?
> > 
> > 
> > i am honestly not sure, the feedback came from the testers and the logging is disabled in prod anyways isn't it?
> 
> Log will be activated in 2.2, so it should be only deactivated for production logging to not have huge log files. But we still want to see this in the logs we receive from our testers.

ahhh got it, fixed 👍

harambasicluka (approved):

chris-cwa (approved):

PR #2934, opened 2021-04-23
Fix hourly key package download on Android 6.0 devices (EXPOSUREAPP-6657)

### LINKED ISSUE #2881: No 'hourly' exposure checks since CWA 1.7 despite Wifi connection available 24 hrs.
## Avoid duplicates
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue - I think.

## Technical details
- Device name: LG-G4 (LGH-815)
- Android version: 6.0
- App version: 1.7.x - 1.15.2
- Affected ENS versions: <= 18210915000

## Describe the bug
[Since CWA version 1.7 hourly fetching of diagnosis keys and matching for exposures was introduced.](https://www.coronawarn.app/en/blog/2020-11-25-corona-warn-app-version-1-7/) For repeated exposure matching several times a day (up to 6 times), a wifi connection with internet is necessary.
Unfortunately, for me this doesn't work, despite wifi connection available 24 hours per day. Only one single exposure matching is carried out daily.
After 1.7 was introduced, I found many reports in Google Play Store reviews of people who were also affected by this problem. Obviously all of them were also running Android 6. Shame on me, I'm a bit late with my bug report. I think there were no more bug reports lately, as for everyone at least one exposure check per day is working, and the 'hourly feature' has just become forgotten for the people who are using Android 6.
The problem popped up for me again, as I noticed that with increasing diagnosis keys (500k+) I obviously run into a 'silent ENS timeout' again (which I will report in a separate issue), and CWA was using the same DK package to try to match exposures for > 24 hours, so I didn't have any timely exposure check for some time.
If I remember correctly, I read somewhere, that it's possible to trigger manual fectching of DKs and starting exposure matching by turning off battery optimization. This also does not work for me.

1. Have an old phone with Android 6 and any version of CWA between 1.7.x and 1.15.2 installed.
2. Let it do its job for a couple of days. Be aware to have internet access via wifi all the time.
3. Go to ENS UI via CWA > Settings > App-Information > tap on ENF version number > select "Überprüfungen auf mögliche Begegenungen" from the 3-dot-menu up right in the screen [ > enter your screen lock pin/password if you have one]
4. Find that for every day there is just a single entry in the log. If there are two entries, they are the same packages (compare checksum), but that's a different issue... => #2880 

## Expected behaviour
Hourly fetching of diagnosis key packages and matching for exposures (up to 6 times per day) should also work for Android 6 and below.

## Possible Fix
This is just a wild guess: Could it be, that the way how you check for 'Wifi Enabled' is API level dependent and that the method works for Android 7+ only?

## Additional context
I uploaded my last ENF logs. Feel free to have a look. Again: please be aware, that in my case more than one entry per day is due to failed exposure matching (time out?) and exposure matching is carried out with the same package like before.

[20210416_all-exposure-checks.txt](https://github.com/corona-warn-app/cwa-app-android/files/6342951/20210416_all-exposure-checks.txt)
[20210420_all-exposure-checks.txt](https://github.com/corona-warn-app/cwa-app-android/files/6342952/20210420_all-exposure-checks.txt)

@d4rken We could wait for release 2.2 with enabled Debug Log Sharing to fix this. I'd be happy to provide my logs. Would the logger be able to find the cause in this case? But if you or one of your colleagues already have any idea what might be the reason of the problem, then just go ahead ☺️ .

----
Internal Tracking-ID: [EXPOSUREAPP-6657](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6657)

d4rken:
A lot of Android 6.0 devices do not set `NET_CAPABILITY_NOT_METERED` correctly on their NetworkConnection.
This PR adds a workaround for Android 6 (< API 24) to check for wifi status, and if it's a wifi type connection, assume unmetered.

This likely fixes #2881


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
```diff
@@ -101,7 +104,14 @@ class NetworkStateProvider @Inject constructor(
...
         val isMeteredConnection: Boolean
-            get() = isFakeMeteredConnection || !(capabilities?.hasCapability(NET_CAPABILITY_NOT_METERED) ?: false)
+            get() {
+                val unMetered = if (BuildVersionWrap.hasAPILevel(Build.VERSION_CODES.N)) {
+                    capabilities?.hasCapability(NET_CAPABILITY_NOT_METERED) ?: false
+                } else {
+                    capabilities?.hasTransport(NetworkCapabilities.TRANSPORT_WIFI) ?: false
+                }
+                return isFakeMeteredConnection || !unMetered
+            }
```
d4rken, line 114: This is the key change.
The rest is testing and some refactoring because the `ApiLevel` pattern I added some time ago just is not useful enough. I replaced it with a mockable Kotlin object.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
```diff
@@ -117,7 +120,14 @@ class NetworkStateProvider @Inject constructor(
         private val isFakeMeteredConnection: Boolean = false
     ) {
         val isMeteredConnection: Boolean
-            get() = isFakeMeteredConnection || !(capabilities?.hasCapability(NET_CAPABILITY_NOT_METERED) ?: false)
+            get() {
+                val unMetered = if (BuildVersionWrap.hasAPILevel(Build.VERSION_CODES.N)) {
```
chris-cwa, line 124: can we use positive var names and move the negation operator from the return statement to the assignment?

chris-cwa (approved):
LGTM!

BMItr (approved):

Issue #2856, opened 2021-04-17
Some buttons have a cut-off shadow


fynngodau:
### Describe the bug

Throughout the app, some buttons have a shadow that is cut off. Here's a list of all those that I could find.

In the screenshots, I am holding down the button so you can notice the cut-off shadow better.

Screenshot | Description | Screenshot | Description
---|:---|---|:--
![Screenshot_1618679790](https://user-images.githubusercontent.com/16943720/115121149-6e179880-9fb1-11eb-8af5-84f2e4ac06df.png) | Create QR code card has a cut-off shadow |  ![Screenshot_1618679750](https://user-images.githubusercontent.com/16943720/115121116-54765100-9fb1-11eb-8940-6a1f1864f203.png) | Activate exposure logging button has a cut-off shadow
![Screenshot_1619261256](https://user-images.githubusercontent.com/16943720/115956115-47021f00-a4fb-11eb-96b1-039c48cedc60.png) | The Your Risk Status fragment's Activate Exposure Logging button and Update button are also affected. | ![Screenshot_1618680352](https://user-images.githubusercontent.com/16943720/115121394-d0bd6400-9fb2-11eb-8537-844ce7058976.png) | "Change prioritized background activity" card's button has a cut-off shadow
![Screenshot_1618679650](https://user-images.githubusercontent.com/16943720/115121087-1a0cb400-9fb1-11eb-82ae-70786d1cdf60.png) | All buttons in the error report screen are cut off at the sides. Best seen at the bottom-most one, which is also cut off downwards.
| | I don't have a screenshot for this, but I think the Turn On Exposure Check button on the reenable risk card is also affected. | | The same applies to the positive test result card.

### Steps to reproduce the issue

View one of the buttons. Touching and holding them increases their elevation, so you can see that the shadow is cut off.

### Expected behavior

The shadow should not be cut off. For example, these shadows are *not* cut off:

![Screenshot_1618680600](https://user-images.githubusercontent.com/16943720/115121497-5d682200-9fb3-11eb-95dc-49e1d2e5554b.png) | ![Screenshot_1618680623](https://user-images.githubusercontent.com/16943720/115121500-60631280-9fb3-11eb-8f80-d9a9c30ca149.png) | ![Screenshot_1618680606](https://user-images.githubusercontent.com/16943720/115121501-622cd600-9fb3-11eb-968b-e372a9a077fd.png)
---|---|---

### Possible Fix

[...]

PR #2942, opened 2021-04-24
Fix button shadows (COMMUNITY)

fynngodau:
This addresses the problem described in #2856.

Demonstration:

Screenshot_1619264002 | Screenshot_1619264011
---|---

It works by telling parents not to clip their children to its padding through `clipToPadding="false"`. At first, I thought this would require restructuring each affected layout, and was happy to learn that this is not the case and that the fix is so simple.

---
 Internal Tracking ID: [EXPOSUREAPP-6543](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6543)

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/bugreporting_debuglog_fragment.xml
+++ b/Corona-Warn-App/src/main/res/layout/bugreporting_debuglog_fragment.xml
@@ -97,6 +97,7 @@
         android:paddingTop="8dp"
         android:paddingEnd="24dp"
         android:paddingBottom="16dp"
+        android:clipToPadding="false"
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintStart_toStartOf="parent">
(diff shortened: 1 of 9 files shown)


dsarkar:
@fynngodau Thanks.  Internal Tracking ID: [EXPOSUREAPP-6543](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6543)

---

harambasicluka:
@fynngodau I just checked our mockups, our buttons shouldn't have a shadow O.o
Here is an example from the mockups:
image

So I would suggest to close this PR and if you want to fix it remove the shadow from our button styling.

fynngodau:
@harambasicluka Does that apply to all of the buttons? For the button on the card (which is already elevated itself), that makes sense, but for other buttons, in my opinion they really should have a shadow to look Material.

harambasicluka:
@fynngodau I think that are personal preferences. Also Google mentions that a high contrast is enough: https://material.io/design/environment/elevation.html#depicting-elevation
image

This would be a question to our designers, so if you want I can forward your suggestion to our designers. But for now I would say this PR solves a problem which shouldn't be there as our buttons shouldn't have an elevation. So I would still stick to my initial suggestion :)

PR #2945, opened 2021-04-24
Improve PPA more info fragment layout (COMMUNITY)

fynngodau:
The PPA more information layout (Detailed Information about Data Sharing) has a few issues which this PR addresses.

Before | After
---|---
Layout does not always have space to the right. Screenshot_1619265762 | Layout always has space to the right. Screenshot_1619268633
The layout uses seemingly random padding values. | The layout uses values from `dimens.xml` like `spacing_small`, `spacing_normal`, …. One exception is the space between bullet dot and text, which is still `13dp`.
The layout uses `ConstraintLayout`. Each view has to reference the view above it. | The layout uses `LinearLayout`, `TableLayout` and `TableRow`. Views no longer need to reference the view above them. The layout becomes simpler to read and add to.

Of course, the text in this view remains unchanged.

---
Internal Tracking ID: [EXPOSUREAPP-6712](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6712)


dsarkar:
@fynngodau Thanks. Internal Tracking ID: EXPOSUREAPP-6712

harambasicluka (approved):
Works nicely and fixes the described bug.
Thanks :)

BMItter:
@fynngodau 
thx again, works nicely 👌 
But as written in your other PR, using tableviews/row & linear is not the best choice nowadays.
It would be nice if you could change your approach using constraint flow layout.. its very easy to use, fast & avoids nesting...
You will be surprised ;-)

d4rken (approved):

harambasicluka:
@fynngodau we had an internal discussion. We would still like to see the implementation @BMItter recommended, but as your solution fixes a bug we'll merge it. So feel free to open a follow up PR.

fynngodau:
I'm not certain whether I'll be able to rework the whole file again.

PR #2947, opened 2021-04-24
Fix overlapping views in tracing settings fragment (COMMUNITY)

### LINKED ISSUE #2940: UI/Display error with text at location enabling screen
## Avoid duplicates
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)

## Technical details

- Device name: Fairphone 3 (FP3)
- Android version: 10 (Stock ROM)
- App version: 2.0.3

## Describe the bug
When the location services are disabled in Android system settings, the screen that asks the user to enable that has a display error with overlapping text.

0. Disable location services completely.
1. On the home page click on “location services disabled” (”Standortdienste deaktiviert”)
2. You get to a screen titled with “risk assessment” (”Risikoermittlung”).
3. Up until the menu for other countries it looks all right. And further down the static explanation text (”Um zu erkennen…”) also looks ok. Only the big interactive part/prompt with the button for the user is absolutely distorted.

The texts “Ermittlungszeitraum aktivieren” (”enable detection period”) and “Standortverwendung erlauben” (”enable location access”) seem to overlay each other:

promptGpsReenable

BTW, the button works though. :)

(English translations are my guess, I have not double-checked whether these are

## Expected behaviour

If I had to guess, I’d say this has been introduced [with the latest update](https://github.com/corona-warn-app/cwa-app-android/releases/tag/v2.0.3).
Using the app in German language.
Screen size [is according to Wikipedia](https://en.wikipedia.org/wiki/Fairphone_3):
5.65 in (144 mm)
1080×2160 px FullHD+
427 ppi

Reproducible in dark and light mode.

---
Also reproducible with CCTG v2.0.3: https://codeberg.org/corona-contact-tracing-germany/cwa-android/issues/120 (Same bug text copied to both issues, only images changed.)

---
Internal Tracking ID: [EXPOSUREAPP-6711](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6711)

fynngodau:
This PR fixes #2940.

The problem was that one of the views was not correctly aligned to be below the other.

Since fixing this within a `ConstraintLayout` is confusing (specifically because at least one of the affected views is hidden in reality), I found the best solution to be, again, replacing the `ConstraintLayout` with a `LinearLayout`, which automatically enforces all its children to be below each other.

---
Internal Tracking ID: [EXPOSUREAPP-6711](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6711)


BMItter:
@fynngodau 
Greetings,
and thank you for your contribution.
I checked your changes.
We are using Constraint Layouts all over the app because its fast and easy to change & maintain.
Replacing ConstraintLayout with LinearLayout is a little bit "old style" even if it works from ui perspective.
As you are aiming for automatic enforcement of all children-views to be below each other I would recommend using a more modern approach,  like constraint layout **flows**. Constraint Layout’s Flow is very useful for displaying views in one direction as much as possible and wrap to next line if space is not enough. Also hiding views is not a problem..

fynngodau:
@BMItter

One of the reasons I don't like `ConstraintLayout`s is specifically because I've experienced them to be slow (as in #1854), though I also believe you were using a quite outdated version back then.

I furthermore don't see a reason to make the layout more complicated than needed. `LinearLayout` may have been around for a while, but it is a perfect fit for rendering multiple views below or next to each other.

BMItter:
> @BMItter
> 
> One of the reasons I don't like `ConstraintLayout`s is specifically because I've experienced them to be slow (as in #1854), though I also believe you were using a quite outdated version back then.
> 
> I furthermore don't see a reason to make the layout more complicated than needed. `LinearLayout` may have been around for a while, but it is a perfect fit for rendering multiple views below or next to each other.

The low performance was mainly caused by the HTML parsing (Html.fromHtml) not by the constraint layout itsef. Using Constraint Flow is also more easy. You can also set a view gabs easily and don't have to put margin and padding allover..
Give it a try 👌

fynngodau:
> The low performance was mainly caused by the HTML parsing (Html.fromHtml) not by the constraint layout itsef.

That's not true, see the PR I linked. However, since this was [a major version of `ConstraintLayout` ago](https://github.com/corona-warn-app/cwa-app-android/pull/2063/files#diff-264ff2dbc2ac999887d86a9dc8b73fb2380a6e76aeeda85b480bc7a5dc1470b9R316), this is probably no longer current anyhow.

> You can also set a view gabs easily and don't have to put margin and padding allover

Right, here I added a `marginHorizontal` to each view to closely imitate the behavior of the `layout_constraintEnd_toStartOf` and `layout_constraintStart_toStartOf` lines. A better solution would have been to move this to the parent.

I still don't see why using a `ConstraintLayout` here, even with using its Flow feature, would make the layout better.

BMItter:
> > The low performance was mainly caused by the HTML parsing (Html.fromHtml) not by the constraint layout itsef.
> 
> That's not true, see the PR I linked. However, since this was [a major version of `ConstraintLayout` ago](https://github.com/corona-warn-app/cwa-app-android/pull/2063/files#diff-264ff2dbc2ac999887d86a9dc8b73fb2380a6e76aeeda85b480bc7a5dc1470b9R316), this is probably no longer current anyhow.
> 
> > You can also set a view gabs easily and don't have to put margin and padding allover
> 
> Right, here I added a `marginHorizontal` to each view to closely imitate the behavior of the `layout_constraintEnd_toStartOf` and `layout_constraintStart_toStartOf` lines. A better solution would have been to move this to the parent.
> 
> I still don't see why using a `ConstraintLayout` here, even with using its Flow feature, would make the layout better.

I guess you misunderstood my comment.. I don't want you to use a nested Layout there.. Just replace the nested layout with an flow, so you dont have performance hit of a nested Linear or Constraint Layout...

fynngodau:
@BMItter Indeed, I did not understand that this is what you would like me to do.

I have tried to use Flow for this purpose here: https://github.com/corona-warn-app/cwa-app-android/commit/9354770a3dd8f7ad9a2795e0e6c9fb0ccc986cd0

However, it is not working correctly, as there is a huge space at the bottom of all the views, but only if one of the `include` layouts is shown, unless I set the width to a value different from `0dp` (match constraints). Maybe you could give feedback why this is happening, and what I can do to fix it? Thanks.

BMItter:
> @BMItter Indeed, I did not understand that this is what you would like me to do.
> 
> I have tried to use Flow for this purpose here: [9354770](https://github.com/corona-warn-app/cwa-app-android/commit/9354770a3dd8f7ad9a2795e0e6c9fb0ccc986cd0)
> 
> However, it is not working correctly, as there is a huge space at the bottom of all the views, but only if one of the `include` layouts is shown, unless I set the width to a value different from `0dp` (match constraints). Maybe you could give feedback why this is happening, and what I can do to fix it? Thanks.

Hi Fynngodau,
regarding 9354770 you are on a good track!
Try the following:
- use for flow_mode 'chain'
- orientation to horizontal
- flow_maxElementsWrap = 1

something like this:

_<androidx.constraintlayout.helper.widget.Flow
    android:id="@+id/settings_tracing_status"
    android:layout_width="0dp"
    android:layout_marginTop="@dimen/spacing_small"
    android:orientation="horizontal"
    app:flow_maxElementsWrap="1"
    app:flow_verticalBias="0"
    app:constraint_referenced_ids="settings_tracing_status_location,settings_tracing_status_bluetooth,risk_details_period_logged_headline,risk_details_period_logged_subtitle,risk_details_period_logged_body_notice,risk_details_period_logged_days"
    app:flow_verticalStyle="packed"
    app:flow_wrapMode="chain"
    app:layout_constraintEnd_toEndOf="@id/guideline_end"
    app:layout_constraintStart_toStartOf="@id/guideline_start"
    app:layout_constraintTop_toBottomOf="@id/settingsInteroperabilityRow" />_

fynngodau:
@BMItter Thanks for the feedback. It does not make a lot of sense to me to set the orientation to `horizontal` when the views are aligned `vertical`ly in reality, and it is not clear to me why my commit did not work, but with your suggestion it is now working and I pushed the change.

fynngodau:
I noticed that the linter is failing, because it thinks that the `TextView`s that are referenced by the Flow layout are not constrainted. That is quite silly, as they are not "jump[ing] to (0,0) at runtime" unlike the linter is claiming, and the views are also not marked red in my IDE. There seems to be no proper solution to this (https://stackoverflow.com/questions/63695251/missingconstraints-lint-error-when-using-flow#comment116741917_63911511). I guess we could `ignore` this lint for the affected views?

BMItter:
@fynngodau 
Hi, yes this lint bug will once be gone.
Could you just try to suppress it for now? (e.g. tools:ignore="MissingConstraints")
We could merge your PR today.
But I can also do a commit...

BMItr (approved):
lgtm!
thx for your commitment! 👌 
ps: I've suppressed the ci warning because its simply not right.

axelherbstreith (approved):

axelherbstreith:
Looks nice 💯 Thank you for fixing it!

PR #2844, opened 2021-04-15
Compat changes after 2.0 (COMMUNITY)

fynngodau:
Two compat changes that were overlooked when creating the new code for 2.0:

* 1x `getColor` → `getColorCompat` (like #1924)
* 1x `android.widget.Button` → `Button` (like #2700)

PR #2026, opened 2021-01-03
Use ContextCompat when accessing color resources (EXPOSUREAPP-4419) (#1924)

fynngodau:
As discussed in #1924, this PR replaces calls to `Context.getColor` with calls to the compat class `ContextCompat.getColor`.

It does not introduce any functional difference to the application, but reduces maintenance workload for our fork of the app, which has Android 5 support.

The tests were changed accordingly: depending on how the tests worked, some now `verify` that `ContextCompat.getColor` is being called instead of `Context.getColor`, some mock both `Context.getColor` as well as `Context.getResources().getColor` (one of which the compat class always calls).

Internal Tracking ID: [EXPOSUREAPP-4419](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-4419)

Issue #1799, opened 2020-12-03
[Discussion] Android 5.0 (lollipop) Support


felischmi2003:
Danke, dass ihr den Source Code öffentlich zugänglich gemacht habt.

Aber trotzdem habe ich ein problem...
Ich besitze ein Gerät mit Android 5.0 und die App ist aber erst mit 6.0.x kompatieble.

Wie kann ich diesen Code umändern um die auch für ältere Geräte verfügbar machen kann?

Mit freundlichen Grüßen,
Felix Schmitt

---
Internal Tracking ID: [EXPOSUREAPP-6360](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6360)

MikeMcC399:
@felischmi2003 

es gibt leider technische Probleme mit Android 5.0 (API 21) - siehe https://github.com/corona-warn-app/cwa-app-android/issues/649 und https://github.com/corona-warn-app/cwa-app-android/issues/1446#issuecomment-716675650 
`Manifest merger failed : uses-sdk:minSdkVersion 21 cannot be smaller than version 23 declared in library [androidx.security:security-crypto:1.0.0-rc03] 
`

Offiziell wird die Corona-Warn-App erst ab Android 6.0 (API 23) unterstützt - siehe https://www.coronawarn.app/de/faq/#minimum_requirements.

felischmi2003:
@MikeMcC399 
Oh, Ok Danke für die Info

fynngodau:
@MikeMcC399 There is a more recent version of this `androidx.security:security-crypto` available that supports API version 21. Support was added [in 1.1.0-alpha1](https://developer.android.com/jetpack/androidx/releases/security#security-crypto-1.1.0-alpha01).

You could use it like I did in this commit to the repo of our fork: https://codeberg.org/corona-contact-tracing-germany/cwa-android/commit/f420799562f2e203cfaa941914d815700c2e2c13

With that obstacle out of the way, it would only be a few minor issues that would need to be fixed for Android 5 support as far as the CWA client is concerned. I'm currently also planning to tackle the ones I haven't had time for yet in my fork.

MikeMcC399:
@fynngodau 
> There is a more recent version of this androidx.security:security-crypto available that supports API version 21. Support was added in 1.1.0-alpha1.

I see there is some more information on https://developer.android.com/jetpack/androidx/releases/security. 

[...]

PR #2955, opened 2021-04-25
Support Android 5 (COMMUNITY)

fynngodau:
Implemets #1799.

Thanks to multiple compatibility PRs that were already merged (#2026, #2700, #2844), not a lot of changes were left.

Unfortunately, I had to disable the splash screen for Android 5 & 5.1, as the image is always stretched to fit the screen width and height without respecting its aspect ratio, looking very broken.

With this change, all visual bugs on Android 5 should be fixed, *except* that the ripple effect on the dispatcher cards is missing.

The PR also hides the Background Priority setting below Android 6, as such a system setting did not exist there.

---
Internal Tracking ID: [EXPOSUREAPP-6360](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6360)


dsarkar:
@fynngodau Thanks. Internal Tracking ID: EXPOSUREAPP-6360

harambasicluka:
Hi @fynngodau this needs to be discussed with a lot of stakeholders. I'll keep you updated here 🚀 
And as always thanks for all your PRs ❤️

harambasicluka:
Current idea is to merge this without the `minSdkVersion`. So we need some approval to change this but making our code work for older versions shouldn't be a problem so that it can be used in some forks... :)

fynngodau:
@harambasicluka I've pushed a change that undoes the change to `minSdkVersion` again. It is up to you which of the other changes you would like in your codebase – namely, https://github.com/corona-warn-app/cwa-app-android/pull/2955/commits/823356a6bf16e30a74e17734f220732090051b09 is quite the hack.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ConnectivityHelper.kt
```diff
@@ -74,11 +76,17 @@ object ConnectivityHelper {
...
-        val caps: NetworkCapabilities? = manager.getNetworkCapabilities(activeNetwork)
-        return caps?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) ?: false
+        return if (BuildVersionWrap.hasAPILevel(Build.VERSION_CODES.M)) {
+            val activeNetwork = manager.activeNetwork
+            val caps: NetworkCapabilities? = manager.getNetworkCapabilities(activeNetwork)
+            caps?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) ?: false
+        } else {
+            val activeNetworkInfo = manager.activeNetworkInfo
+            activeNetworkInfo != null && activeNetworkInfo.isConnected
+        }
```
mtwalli, line 89: maybe we could polish it a bit and make it more readable ? as follows:
```kotlin
fun isNetworkEnabled(context: Context): Boolean {
        val manager = context.getSystemService<ConnectivityManager>() ?: return false
        return when {
            Build.VERSION.SDK_INT >= Build.VERSION_CODES.M -> {
                val activeNetwork = manager.activeNetwork
                val caps: NetworkCapabilities? = manager.getNetworkCapabilities(activeNetwork)
                caps?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) ?: false
            }
            else -> {
                val activeNetworkInfo = manager.activeNetworkInfo
                activeNetworkInfo != null && activeNetworkInfo.isConnected
            }
        }
   }
```
and not using the version wrapper in this case , then we don't have to annotate it with `SuppessLint`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DialogHelper.kt
```diff
@@ -111,7 +111,7 @@ object DialogHelper {
         textView.linksClickable = true
         textView.movementMethod = LinkMovementMethod.getInstance()
         textView.setPadding(paddingStartEnd, paddingLeftRight, paddingStartEnd, paddingLeftRight)
-        textView.setTextAppearance(R.style.body1)
+        textView.setTextAppearance(context, R.style.body1)
```
mtwalli, line 114: `TextViewCompat.setTextAppearance` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/PowerManagement.kt
```diff
@@ -18,7 +21,12 @@ class PowerManagement @Inject constructor(
...
-        get() = powerManager.isIgnoringBatteryOptimizations(context.packageName)
+        @SuppressLint("NewApi")
+        get() = if (BuildVersionWrap.hasAPILevel(android.os.Build.VERSION_CODES.M)) {
+            powerManager.isIgnoringBatteryOptimizations(context.packageName)
+        } else {
+            true
+        }
 
     val toBatteryOptimizationSettingsIntent =
         Intent(Settings.ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS)
```
mtwalli, line 32: Same here I think we should use the SDK version checks directly, and provide check also for `toBatteryOptimizationSettingsIntent `

and remove remove `disableBatteryOptimizationsIntent` since it is not used otherwise it has to have a check also

```kotlin
    val isIgnoringBatteryOptimizations
        get() = if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
            powerManager.isIgnoringBatteryOptimizations(context.packageName)
        } else {
            true
        }

    val toBatteryOptimizationSettingsIntent =
        if (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.M) {
            Intent(Settings.ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS)
        } else {
            Intent() // Need to check if this empty intent is not really triggering anything ?
        }
```

Corona-Warn-App/src/main/res/drawable/splash_screen.xml
```diff
@@ -4,9 +4,4 @@
     <!-- background with solid color -->
     <item android:drawable="@color/colorBackground" />
 
-    <!-- logo -->
-    <item
-        android:drawable="@drawable/ic_splash_logo_round"
-        android:gravity="center" />
-
```
mtwalli, line 11: does this mean there is no logo on below Android M devices splash?

Corona-Warn-App/src/main/res/drawable/tan_input_digit.xml
```diff
@@ -7,12 +7,14 @@
         </shape>
     </item>
     <item
-        android:height="@dimen/submission_tan_input_digit_stroke"
-        android:gravity="bottom">
+        android:top="-4dp"
+        android:right="-4dp"
+        android:left="-4dp">
```
mtwalli, line 12: what is the reason behind this? I prefer to create a drawable for Lollipop devices specifically if this causing issue on it

Corona-Warn-App/src/main/res/drawable/tan_input_digit_error.xml
```diff
@@ -7,12 +7,14 @@
         </shape>
     </item>
     <item
-        android:height="@dimen/submission_tan_input_digit_stroke"
-        android:gravity="bottom">
+        android:top="-4dp"
+        android:right="-4dp"
+        android:left="-4dp">
```
mtwalli, line 12: same here ?

Corona-Warn-App/src/main/res/layout/home_fragment_layout.xml
```diff
@@ -18,7 +18,7 @@
         android:orientation="vertical"
         tools:context=".ui.main.MainActivity">
 
-        <Toolbar
+        <androidx.appcompat.widget.Toolbar
```
mtwalli, line 21: `MaterialToolbar` please ?

Corona-Warn-App/src/main/res/layout/home_fragment_layout.xml
```diff
@@ -34,7 +34,7 @@
                 app:srcCompat="@drawable/ic_main_header"
                 bind:cwaContentDescription="@{@string/accessibility_logo}"
                 tools:ignore="ContentDescription" />
-        </Toolbar>
+        </androidx.appcompat.widget.Toolbar>
```
mtwalli, line 37: `MaterialToolbar` please ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
```diff
@@ -90,34 +92,46 @@ class NetworkStateProvider @Inject constructor(
...
+                        } catch (e: Exception) {
+                            Timber.tag(TAG).e(e, "Failed to determine link properties.")
+                            null
+                        }
+                    },
+                    assumeMeteredConnection = testSettings.fakeMeteredConnection.value
+                )
+            }
+        } else {
+            // Most state information is not available
```
mtwalli, line 120: Maybe we can split it into small functions and adjusted it a bit ?

```kotlin
 private val currentState: State
        get() = when {
            Build.VERSION.SDK_INT >= Build.VERSION_CODES.M -> api23NetworkState()
            else -> {
                // Most state information is not available
                State(
                    activeNetwork = null,
                    capabilities = null,
                    linkProperties = null,
                    assumeMeteredConnection = testSettings.fakeMeteredConnection.value ||
                        ConnectivityManagerCompat.isActiveNetworkMetered(manager)
                )
            }
        }

    @RequiresApi(Build.VERSION_CODES.M)
    private fun api23NetworkState() = manager.activeNetwork.let { network ->
        State(
            activeNetwork = network,
            capabilities = network?.let {
                try {
                    manager.getNetworkCapabilities(it)
                } catch (e: SecurityException) {
                    Timber.tag(TAG).e(e, "Failed to determine network capabilities.")
                    null
                }
            },
            linkProperties = network?.let {
                try {
                    manager.getLinkProperties(it)
                } catch (e: Exception) {
                    Timber.tag(TAG).e(e, "Failed to determine link properties.")
                    null
                }
            },
            assumeMeteredConnection = testSettings.fakeMeteredConnection.value
        )
    }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
```diff
@@ -90,34 +92,46 @@ class NetworkStateProvider @Inject constructor(
...
-                        manager.getLinkProperties(it)
-                    } catch (e: Exception) {
-                        Timber.tag(TAG).e(e, "Failed to determine link properties.")
-                        null
-                    }
-                },
-                isFakeMeteredConnection = testSettings.fakeMeteredConnection.value
+                activeNetwork = null,
+                capabilities = null,
+                linkProperties = null,
```
mtwalli, line 124: are there any alternatives on pre API 23 for these nullables ?

mtwalli (commented):
Thank you for the effort! , I would like to suggest the following changes ⬇️

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/PowerManagement.kt
fynngodau: We could annotate `toBatteryOptimizationSettingsIntent` as `@RequiresApi(Build.VERSION_CODES.M)`?

Corona-Warn-App/src/main/res/drawable/tan_input_digit.xml
fynngodau: The width and height attributes are only supported from Android 6 on; see commit message here: https://github.com/corona-warn-app/cwa-app-android/pull/2955/commits/823356a6bf16e30a74e17734f220732090051b09

Agreed, will put into separate file.

Corona-Warn-App/src/main/res/drawable/splash_screen.xml
fynngodau: Yes, otherwise it will look similar to this (see https://github.com/corona-warn-app/cwa-app-android/pull/2955/commits/d7c66b66f35704c592b0cceff440933664ebacd1):
Screenshot_1619782755

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/details/items/behavior/BehaviorInfoRow.kt
```diff
@@ -61,7 +61,7 @@ class BehaviorInfoRow @JvmOverloads constructor(
     }
 
     fun setBackgroundTint(@ColorInt color: Int) {
-        ViewCompat.setBackgroundTintList(iconBackground, ColorStateList.valueOf(color))
+        iconBackground.background.setColorFilter(color, PorterDuff.Mode.SRC_OVER)
```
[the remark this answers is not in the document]
fynngodau: That's what the code contained originally but it is not working :(

Lollipop screenshot with `ViewCompat.setBackgroundTintList`:

image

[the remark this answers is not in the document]
fynngodau: Probably a platform bug on 5.0, as in 5.1 emulator this is working correctly. I added a change that uses the original call except on 5.0

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/device/PowerManagement.kt
fynngodau: Empty Intent causes `ActivityNotFoundException: No Activity found to handle Intent {  }`. Therefore adding a check at this place does not make sense, as both intents would lead to a crash on Android < 6 (besides that this method should not be called there)

Corona-Warn-App/src/main/res/drawable/tan_input_digit.xml
```diff
@@ -1,5 +1,12 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!--
+        This asset is only used below Android 6. These Android versions
+        do not support the `android:height` and `android:width` attributes,
+        making tricks necessary.
+    -->
```
vaubaehn, line 8: 👍

Corona-Warn-App/src/main/res/drawable/tan_input_digit_error.xml
```diff
@@ -1,5 +1,13 @@
 <?xml version="1.0" encoding="utf-8"?>
 <layer-list xmlns:android="http://schemas.android.com/apk/res/android">
+
+    <!--
+        This asset is only used below Android 6. These Android versions
+        do not support the `android:height` and `android:width` attributes,
+        making tricks necessary.
+    -->
```
vaubaehn, line 8: 👍

Corona-Warn-App/src/main/res/drawable/splash_screen.xml
mtwalli: I think this will work, can you please try on your side?
```xml
<layer-list xmlns:android="http://schemas.android.com/apk/res/android">

    <!-- background with solid color -->
    <item android:drawable="@color/colorBackground" />

    <item>
        <bitmap
            android:gravity="center"
            android:src="@drawable/ic_splash_logo_round" />
    </item>

</layer-list>
```

I tired it looks like this:
Screenshot 2021-04-30 at 15 59 55

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
fynngodau: A problem here is that `BuildVersionWrap.SDK_INT` can be mocked in the tests while this is not possible (or I don't know how) when comparing the value directly. Therefore I changed this again to use `BuildVersionWrap` and `@SuppressLint`.

mtwalli:
In `ErrorDialog` file line 38 we need to change the textApperance too 
```
       - setTextAppearance(R.style.body1)
       + TextViewCompat.setTextAppearance(this, R.style.body1)
```

Corona-Warn-App/src/main/res/drawable/splash_screen.xml
fynngodau: Thanks, it looks remarkably similar to what was in the file originally but your solution fixes the problem :tada: will integrate into PR & remember for the future.

fynngodau: I found no disadvantage in doing it this way, therefore I removed the separation of assets between Android versions again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
fynngodau: We could additionally store the active `NetworkInfo` object (deprecated), but this would be a bit confusing; and since the fields currently doesn't seem to be needed anywhere else, it might be better to find out how to extract one specific thing that is needed once it is needed.

mtwalli: Sounds good 👍

d4rken (changes requested):
Changes look good.

Would like a few more tests, and we need to guard against `androidx.security` issues.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ConnectivityHelper.kt
```diff
@@ -76,9 +77,17 @@ object ConnectivityHelper {
      */
     fun isNetworkEnabled(context: Context): Boolean {
         val manager = context.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
-        val activeNetwork: Network? = manager.activeNetwork
-        val caps: NetworkCapabilities? = manager.getNetworkCapabilities(activeNetwork)
-        return caps?.hasCapability(NetworkCapabilities.NET_CAPABILITY_VALIDATED) ?: false
+        return when {
+            Build.VERSION.SDK_INT >= Build.VERSION_CODES.M -> {
```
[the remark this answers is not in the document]
mtwalli: BuildVersionWrap does not remove the warning and we need to SuppessLint error, I prefer that way because it offers both

Corona-Warn-App/src/main/AndroidManifest.xml
```diff
@@ -4,7 +4,7 @@
     package="de.rki.coronawarnapp"
     tools:ignore="LockedOrientationActivity">
 
-    <uses-sdk tools:overrideLibrary="com.google.zxing.client.android" />
+    <uses-sdk tools:overrideLibrary="com.google.zxing.client.android,androidx.security" />
```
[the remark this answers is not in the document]
mtwalli: I agree specially that official API 21 support from Google still in alpha https://developer.android.com/jetpack/androidx/releases/security#security-crypto-1.1.0-alpha01.
I would love to test that on real Android 5 devices as well as ENF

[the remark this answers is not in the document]
fynngodau: Android 5 support is not available in the version that you are using, as you are using `1.0.0-rc03`. It would be an option to update to this version, but I wouldn't consider it to be the best one for your project.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ConnectivityHelper.kt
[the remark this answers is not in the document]
d4rken: But can we test it without `BuildVersionWrap`?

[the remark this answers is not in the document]
mtwalli: Yes, we can either:
1- Using the following `Robolectric Util`  - which we already have -
```kotlin
 ReflectionHelpers.setStaticField(Build.VERSION::class.java, "SDK_INT", 23)
```
or
2- Creating our own helper to have less boilerplate code:
```kotlin
object SDKVersion {
    fun set(newValue: Int) {
        val field = Build.VERSION::class.java.getField("SDK_INT")
        field.isAccessible = true
        val modifiersField: Field = Field::class.java.getDeclaredField("modifiers")
        modifiersField.isAccessible = true
        modifiersField.setInt(field, field.modifiers and Modifier.FINAL.inv())
        field.set(null, newValue)
    }
}
```
and then in test cases set the required version `SDKVersion.set(21)`

I tried both ways and seems fine. If we agree about this approach, I can do it in another PR ,and totally remove the wrapper if no issue arises

Corona-Warn-App/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
mtwalli: Unfortunately we won't be able to use alpha versions here,  I would suggest to revert this change and keep our minSDK = 23 and we can merge this PR. In our side this would not be a problem as we still guard our code to run on API 23+.

By the changes in here we have increased the compatibility with pre Android M versions. We can lower our minSDK when all blockers are addressed. @d4rken what do think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
```diff
@@ -90,34 +93,50 @@ class NetworkStateProvider @Inject constructor(
...
-                    } catch (e: Exception) {
-                        Timber.tag(TAG).e(e, "Failed to determine link properties.")
-                        null
-                    }
-                },
-                isFakeMeteredConnection = testSettings.fakeMeteredConnection.value
-            )
+        @SuppressLint("NewApi")
+        get() = when {
+            BuildVersionWrap.hasAPILevel(Build.VERSION_CODES.M) -> api23NetworkState()
```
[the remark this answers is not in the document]
fynngodau: I might need some help with this. I wrote the following test:

    @Test
    fun `current state is correctly determined by method for API level 23`() = runBlockingTest2(ignoreActive = true) {
        val instance = createInstance(this)
        instance.networkState.first()

        verify { instance.api23NetworkState() }
    }

and I am getting: `java.lang.AssertionError: Verification failed: call 2 of 7: Any(child of context#3#10).getActiveNetwork()) was not called`

Do you know why my test is not working?

Corona-Warn-App/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
fynngodau: I added a change like @d4rken suggested: f5a137cd9ea445e4bbcb6e761eb6f2a7860b17e3.

This appears to be the only place from which `EncryptedPreferencesMigration` can be called, so the code from the `androidx.security` library that doesn't work with < 23 won't be invoked on those devices.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ConnectivityHelper.kt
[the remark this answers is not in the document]
d4rken: Isn't just using the wrapper object much easier? No reflection necessary? Works well with `BuildConfigWrap` already.

Corona-Warn-App/src/main/AndroidManifest.xml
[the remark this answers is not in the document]
d4rken: Thanks, looks good.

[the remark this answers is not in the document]
mtwalli: Thanks you for addressing this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/network/NetworkStateProvider.kt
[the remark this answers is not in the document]
d4rken: Looks a bit like a nested verify call? i.e. verify on a mock of a mock or something like that.

The PR is in a good place currently, and we are going to merge it now.

@mtwalli  or me will create a small follow up PR with additional unit tests. I'll ping you on the unit test when we get it working :grin:

mtwalli (approved):
Thank you very much!, We will improve  the test coverage and SDK version wrapper points in a follow up PR

d4rken:
Thanks a lot for all your effort on this @fynngodau. I'll leave the comments open and address them in a follow up PR on our end.

We'll keep the minAPI at 23 for now, prepare internal testing and discuss when we can lower the API requirement.

d4rken (approved):

PR #2958, opened 2021-04-26
Fix number of days with encounters calculation (EXPOSUREAPP-6703)

chiljamgossow:
Count distinct dates with risk encounters from EW and PT only
-> risk encounters on the same day should only be counted once
Risk calculation is not affected, only the displayed number of days may be too high as they are counted twice


d4rken (approved):
lgtm

PR #2960, opened 2021-04-26
Refactor worker scheduling (EXPOSUREAPP-6007)

d4rken:
@chris-cwa & me refactored the worker scheduling, to fix issues and align the RA/PCR test result polling.

There is now a scheduler class for each worker, that listens to one or more flows, and queues or cancels the periodic workers.

* There were various inconsistencies with regards to the combination of presence tracing risk calculation and exposure window risk calculation.
    * Presence tracing should not have been disabled when exposure window based tracing was disabled.
    * When either workers were stopped, in `MainActivity.onResume` we rescheduled the workers which then aborted by themselves again, unnecessarily consuming resources.

`CoronaWarnApplication` will setup the schedulers that determine worker behavior:
```kotlin
        Timber.v("Setting up risk work schedulers.")
        exposureWindowRiskWorkScheduler.setup() // Checks tracing, onboarding and for positive results
        presenceTracingRiskWorkScheduler.setup() // Checks onboarding and for positive results

        Timber.v("Setting up test result work schedulers.")
        pcrTestResultScheduler.setup() // Checks for test and final test state
        raTestResultScheduler.setup() // Checks for test and final test state
 
        Timber.v("Setting up test result available notification services.")
        pcrTestResultAvailableNotificationService.setup() // Checks for test and if notification was already sent
        raTestResultAvailableNotificationService.setup() // Checks for test and if notification was already sent
```

There are no more function calls from various locations in code. If there needs to be a change to worker behavior, the necessary information to make that decision needs to be exposed so that the "schedulers" can observe a flow that contains this information.

### TODO

- [x] Fix unit tests


BMItr (approved):
lgtm 👌 
its more clear now, how the workers are scheduled
Basic testing successful, but I will run deeper testing later..

harambasicluka (approved):

PR #2965, opened 2021-04-26
Fix RA test state mapping when submission is done (EXPOSUREAPP-6004)

d4rken:
The screens that are shown when the submission done cards are clicked seems to be not implemented?


AlexanderAlferov (dismissed):
LGTM 🚀

chiljamgossow (approved):

PR #2968, opened 2021-04-26
No test result notification for tans (EXPOSUREAPP-4287)

d4rken:
For TANs we immediately consume the notification flag.


chiljamgossow (approved):
It's a lie but works

jurajkusnier (approved):

PR #2972, opened 2021-04-27
Fix test result notifications and handle unexpected testresults (EXPOSUREAPP-6715)

d4rken:
* Fixes test result available notifications not being visible (they were directly canceled, now we check if the result was viewed before canceling the notification)
* Adjusts test result state mapping. Instead of throwing an exeception, we map the unexpected state to the test specific invalid state `*_INVALID`. This  has relatively minor effects on polling, but if the server ever has an error and returns the wrong PCR/RA test result, we do not loose the registration token. Instead we  end up with the `INVALID` state and the server has the chance to deliver a correct result on polling. While hopefully unlikely, this should reduce user frustration (due to one time QR use).

To test notifications:
* Launch mock server in interactive mode
* Go to test menu, scan a test.
* Return PENDING from mock server
* Press refresh again
* Close the home through home button
* Now return POSITIVE or NEGATIVE answer from mock server
* You should see a notification and clicking it should take you to the right result page.


chiljamgossow (commented):
I tested, but was forwarded to test result invalid although mock server returned positive. Same on the home card, shows as invalid despite choosing positive on mock server side. Maybe I did not test correctly.

chris-cwa (approved):
other than naming i like it!

BMItr (approved):
nothing to complain 👌

Issue #2953, opened 2021-04-25
Error 9002 - "Connection reset." appears once per day.


sfadschm:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ]
* [x] Bug is specific for Android
* [x] Bug is not already reported in another issue

## Technical details

- Device name: Google Pixel 3a
- Android version: 11
- App version: 2.0.4

## Describe the bug
When opening the app for the first time of the day, the error 9002 "Connection reset" occurs.
This error only occurs since the 2.0.4 update but is reproducible every day. It is only triggered when opening the app for the first time.
I tried triggering it by switching from WLAN to LTE during daytime, resetting our Router connection, turning the phone off and on, but did not have any success.

As far as I can tell, functionality of the app is not touched by this, as after confirming the error, contacts get checked like normally.

## Steps to reproduce the issue
![Screenshot_20210425-094951](https://user-images.githubusercontent.com/40514119/115985435-26de6880-a5ac-11eb-9fb1-1c70b733b99f.png)
![Screenshot_20210425-081521](https://user-images.githubusercontent.com/40514119/115985447-2c3bb300-a5ac-11eb-8b72-3034452db74e.png)

1. Have CWA App closed overnight.
2. Wake up in the morning.
3. Open CWA App.
4. Get error.

## Expected behaviour
The App opens without any error.

dsarkar:
@sfadschm Tanks for reporting. We created an Internal Tracking ID: [EXPOSUREAPP-6714](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6714). The development team will have a look at this issue. Any further news regarding this issue will be published here. Best wishes, DS

---
Corona-Warn-App Open Source Team

schuhmi2:
Ditto, I'm experiencing this too
OnePlus 6T android 10
CoronaWarn app v 2.0.4

sfadschm:
Addition:
Just got the same error a second time today.
Still cannot trigger it intentionally.

MikeMcC399:
This issue is listed in the resolved section of the FAQs https://www.coronawarn.app/en/faq/#cause9002_timeout which refers also to issue https://github.com/corona-warn-app/cwa-app-android/issues/998.

I have seen it occurring sporadically on an emulated Google Pixel 3a device using Android 11. Emulators are generally slower than real physical devices, so this would be consistent with a timeout issue.

sfadschm:
[...]

PR #2973, opened 2021-04-27
Connection Reset error handling (EXPOSUREAPP-6784)

d4rken:
A test is invalid on 40X errors (client errors), instead of checking for  `!is CwaServerError`, we check for `is CwaClientError`.
This causes the error case for `SocketException` to be mapped to PENDING.

### Testing
Test the behavior by either playing with your firewall, or throwing an exception here:

Screenshot from 2021-04-27 13-40-14


d4rken:
Related to #2953

mlenkeit (approved):
✅  Fix confirmed. Prior to this, a connection error while communicating led to the Redeemed card. Now, it stays in Pending. Verified end-to-end by producing connection error in the communication with the server.

chiljamgossow (approved):

chiljamgossow:
Tested

PR #2976, opened 2021-04-27
Fix missing line in red warning card (EXPOSUREAPP-6775)

chiljamgossow:
The data field used to retrieve EW days with risk encounters was null. So it is added from a different source now.

Would be cleaner to refactor EW risk calculations and data structures, but no time 😞 

How to test:
Fake ENF risk encounters via test menu. Check that first line on the red card is there.


BMItr (dismissed):
code lgtm.
tested on oneplus 8 device

BMItr (approved):

LukasLechnerDev (approved):
LGTM

PR #2977, opened 2021-04-27
Add 'submit test' reminder for RAT tests (EXPOSUREAPP-6599)

LukasLechnerDev:
### Current Implementation
We already had a reminder for PCR tests:
- A user registers a positive PCR test
- but doesn't share his keys
- => A reminder-notification to share his keys is scheduled in 2 hours
- => User sees this reminder max 2 times

### This PR ...
- ... now implements the same behavior for RAT tests too

### How to test
- Scan PCR or RAT Tests but don't share keys
- Timetravel 2h into the future 
- Notification should be received
- When clicking on it, you should be able to share the keys for the respective test
- Maybe also test that this notification doesn't show up if you share the keys in the first place


chris-cwa (approved):
LGTM!

chiljamgossow:
I land on the TestResultAvailable screen after clicking on the notification. Is that correct?

LukasLechnerDev:
> I land on the TestResultAvailable screen after clicking on the notification. Is that correct?

Yes, I think so. I didn't change this behavior, it was like this before.

chiljamgossow (approved):

PR #2979, opened 2021-04-27
Fix app crash when removing RAT test (EXPOSUREAPP-6772)

jurajkusnier:
- Fixing the wrong test type during removing
- Popup with connection error was not reproducible


chiljamgossow (approved):
Tested, no crash, test removed.

BMItr (approved):
👌

PR #2980, opened 2021-04-27
Improve error handling during CoronaTest result retrieval (EXPOSUREAPP-6784)

d4rken:
Only HTTP400 errors are mapped to TestInvalid state.

This is the `2.1.x` fix for #2973

Draft because we currently don't show popups for these errors anymore, but we should.
* We either need to grab them in the test processors and report them
* Or listen for test results with `lastError != null` in the UI.
Not yet sure which approach is better.

I'll add error dialogs with another PR.


BMItr (approved):
lgtm,
look forward to follow-up-dialog pr

chiljamgossow (approved):

PR #2993, opened 2021-04-28
Adjust deadman notification scheduler (EXPOSUREAPP-6600)

LukasLechnerDev:
Refactor the Deadman Notification scheduler to our new way of how workers work. This way, we don't schedule work from several places within the app, but only from `DeadmanNotificationScheduler`

Basically, `deadmanNotificationScheduler.setup()` is called at app start and then the scheduler observes if
- onBoarding was done
- a positive test result is registered
- enf tracing is enabled

=> We schedule the Deadman Notification if onBoarding was done, enf tracing is activated and no positive test result is registered. 
=> I didn't change the worker itself, so it should behave as before.


jurajkusnier (dismissed):
LGTM
Tested on Pixel 4 (Android 11)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationScheduler.kt
```diff
@@ -4,16 +4,55 @@ import androidx.work.ExistingPeriodicWorkPolicy
...
 
+    fun setup() {
+        Timber.i("setup() DeadmanNotificationScheduler")
+
+        combine(
+            onboardingSettings.isOnboardedFlow,
+            coronaTestRepository.coronaTests,
+            enfClient.isTracingEnabled
+        ) { isOnboarded, coronaTests, isTracingEnabled ->
+            val noPositiveTestRegistered = coronaTests.none { it.isSubmissionAllowed }
```
d4rken, line 37: Should we use `isPositive`?  Because isSubmissionAllowed will be false again after submission.

```kotlin
    override val isSubmissionAllowed: Boolean
        get() = isPositive && !isSubmitted
```

d4rken (commented):
Love it, removing all that code from different places :ok_hand:.

Small question? :point_down:

BMItr (dismissed):
lgtm!
tested on oneplus 8

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationScheduler.kt
LukasLechnerDev: Good point. Thanks! Changed it to use `isPositive` now.

jurajkusnier (approved):

d4rken (approved):

PR #2994, opened 2021-04-28
Implement more secure RAT QR Code design (EXPOSUREAPP-6776)

jurajkusnier:
To test with **cwa-app-cli** don't forget to pull the latest changes.
You can generate invalid RAT Qr codes using:
```
cwa rat --invalid fn
cwa rat --invalid ln
cwa rat --invalid dob
```


chiljamgossow (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/RapidAntigenQrCodeExtractor.kt
```diff
@@ -113,6 +131,16 @@ class RapidAntigenQrCodeExtractor @Inject constructor() : QrCodeExtractor<Corona
...
             if (!complete) throw InvalidQRCodeException("QRCode contains incomplete personal data: $raw")
         }
+
+        private fun requireValidHash() {
+            val isQrCodeWithPersonalData = firstName != null && lastName != null && dateOfBirth != null
+            val generatedHash =
+                "${raw.dateOfBirth}#${raw.firstName}#${raw.lastName}#${raw.timestamp}#${raw.testid}#${raw.salt}"
+                    .toSHA256()
+            if (isQrCodeWithPersonalData && !generatedHash.equals(hash, true)) {
+                throw InvalidQRCodeException("Generated hash doesn't match QRCode hash")
```
LukasLechnerDev, line 141: Please add a log statement when the hash matching fails.

chiljamgossow (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/qrcode/RapidAntigenQrCodeExtractor.kt
jurajkusnier: I guess this should be done globally as this is not the only exception that could be thrown. Doing it looks out of the scope for the current task.

LukasLechnerDev (approved):
The functionality looks good 👍  Tested it with the cli tool. Some unit tests in an upcoming PR would be great.

PR #2995, opened 2021-04-28
Improve error handling and popups (DEV, EXPOSUREAPP-6518)

d4rken:
* This fixes random error popup during onboarding if ENS is not available. The error will only be shown when pressing "activate tracing" during onboarding or in our apps tracing settings.
    * The previous error came from the globally suscribed `DefaultTracingStatus.isTracingEnabled: Flow<Boolean>`, this no longer throws errors, but depending on the errors, slows down polling or if permanent, aborts. Problems with tracing will be displayed by the `TracingPermissionHelper` on the screens where you can enable tracing.
* The `HomeFragment` display errors related to `CoronaTestRepository` actions
    * Once per error type, once per test type for the duration of the app session
    * Each test error shows the test type in the title
 * I introduced initial plumbing for better error handling via [`KnownExceptions`](https://github.com/corona-warn-app/cwa-app-android/pull/2995/files#diff-ffe7392a9200dd478d103008fc3eec5b996012ba264a7c5793b399a288ec95daR8)
     * Based on the `Throwable` a specific exception that has extra error text can be detected
     * Currently only wired up for ENS API Error 17 (ENS not installed or outdated)
     * Wired into `Throwable.tryHumanReadableError(context: Context): HumanReadableError`
     * There is a helper method to show an error in a dialog `error.toErrorDialogBuilder(context).show()`
     * In most cases we should be able to just go `viewmodel.errors.observe { it.toErrorDialogBuilder(context).show() }`


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/exceptions/KnownExceptions.kt
```diff
@@ -0,0 +1,29 @@
...
+    ENS_NOT_INSTALLED {
+        override fun matches(throwable: Throwable): Boolean = throwable is ApiException && throwable.statusCode == 17
+
+        override fun getHRE(context: Context): HumanReadableError = HumanReadableError(
+            title = "${context.getString(R.string.errors_generic_details_headline)}: 3\n" +
+                context.getString(R.string.errors_generic_headline),
+            description = context.getString(R.string.errors_google_update_needed)
+        )
+    },
+}
```
d4rken, line 25: My idea is for future updates when we have more time:
* Expand this, consume all cases 
* Replace `ExceptionReporter.kt` with a more central flow that can be subscribed too, hooking somewhere into the error classes by @BMItter 
* Subscribe to this errorFlow in MainActivity
* Remove the whole "application accesses activity lifecycle to setup `ErrorReportReceiver` , then remove the whole `ErrorReportReceiver`
* Now we have a single flow that supplies error data, show popups for now, in the future future, replace popup showing with a homescreen card for errors.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/tracing/DefaultTracingStatus.kt
```diff
@@ -84,23 +84,34 @@ class DefaultTracingStatus @Inject constructor(
...
                 break
+            } catch (e: ApiException) {
+                emit(false)
+                if (e.statusCode == 17) {
+                    // No ENS installed, no need to keep polling.
+                    Timber.tag(TAG).v("No ENS available, aborting polling, assuming permanent.")
+                    break
+                } else {
+                    Timber.tag(TAG).v("Polling failed, will retry with backoff.")
+                    delay(POLLING_DELAY_MS * 5)
```
mtwalli, line 104: does `retry` operator suits this case ? instead of `delay` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/tracing/DefaultTracingStatus.kt
```diff
@@ -111,8 +122,14 @@ class DefaultTracingStatus @Inject constructor(
 
     private suspend fun isEnabled(): Boolean = suspendCoroutine { cont ->
         client.isEnabled
```
mtwalli, line 124: can we just remove the whole `suspendCoroutine` and use `await()`?

harambasicluka (dismissed):
Tested on a OnePlus 6t, fix works as promised 🌮

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/ui/ErrorDialog.kt
```diff
@@ -0,0 +1,60 @@
...
+            paddingStartEnd,
+            paddingLeftRight,
+            paddingStartEnd,
+            paddingLeftRight
+        )
+        setTextAppearance(R.style.body1)
+        setLinkTextColor(context.getColorStateListCompat(R.color.button_primary))
+        setTextIsSelectable(!textHasLinks)
+    }
+    setView(textView)
```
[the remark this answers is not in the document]
d4rken: Yes, but I didn't have time for that and took the previous code we had for this.
You are more than welcome to improve this in a follow up PR for 2.2.x.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepositoryExtensions.kt
```diff
@@ -32,3 +35,27 @@ val CoronaTestRepository.isRiskCalculationNecessary: Flow<Boolean>
     get() = coronaTests.map { tests ->
         tests.none { it.isPositive }
     }
+
+// This in memory to not show duplicate errors
+// CoronaTest.lastError is also only in memory
+private val consumedErrors = mutableMapOf<String, Throwable?>()
```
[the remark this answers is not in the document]
d4rken: `distinctUntil*` only remembers it for the subscription, if we leave the HomeFragment and return it is reset, but we do not want to display the same error again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/nearby/modules/tracing/DefaultTracingStatus.kt
d4rken: I don't see a way to use the flow retry operator here without completely changing the polling logic here, this would be too much for the scope of the PR, but I'd happy to discuss changing it in another PR for 2.2 if you see potential.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepositoryExtensions.kt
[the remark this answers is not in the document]
d4rken: We could solve this by using `.stateIn` or `.shareIn` but this has potentially more site effects I currently don't want to wrap my head around for 2.1. :cold_sweat:

mtwalli (approved):

harambasicluka (approved):

MikeMcC399:
@d4rken 

Thank you very much for this PR! As a side effect it makes it much more convenient for community members running builds from Android Studio on a Google account without the benefit of the account being allowlisted for Exposure Notifications by Google. 

I'm able to run the release/2.1.x branch with no error! 😄 

An error only occurs if I try to activate exposure logging, which I can easily avoid doing.

PR #3006, opened 2021-04-29
Fix auto-checkout being processed before wrong risk task (DEV)

d4rken:
Fix auto-checkouts being processed before EW risk calculation, they should be processed before PT risk calculation. Checkouts only impact PT, not EW, risk calculation.


LukasLechnerDev (dismissed):
LGTM

mtwalli (dismissed):

LukasLechnerDev (approved):

mtwalli (approved):

PR #3011, opened 2021-04-29
Vaccination Certificate (feature main branch)

d4rken:
WIP :tea:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinatedPerson.kt
```diff
@@ -0,0 +1,32 @@
+package de.rki.coronawarnapp.vaccination.core
+
+import org.joda.time.Instant
+import org.joda.time.LocalDate
+
+data class VaccinatedPerson(
+    val vaccinationCertificates: Set<VaccinationCertificate>,
+    val proofCertificates: Set<ProofCertificate>,
+    val isRefreshing: Boolean,
```
chiljamgossow, line 9: What does refreshing mean in this context?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinationCertificate.kt
```diff
@@ -0,0 +1,27 @@
...
+interface VaccinationCertificate {
+
+    val firstName: String
+    val lastName: String
+    val dateOfBirth: LocalDate
+
+    val vaccinatedAt: Instant
+
+    val vaccinationName: String
+    val vaccinationManufacturer: String
```
chiljamgossow, line 16: vaccineName and vaccineManufacturer

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinatedPerson.kt
d4rken: That this data is currently being updated, e.g. polling servers for a new ProofCertificate.
Use case:
User presses refresh button in UI, UI wants to display a progress spinner.
This can't be only handled on viewmodel level, as there will also be automatic refreshes/updates via workers, that could overlap with the user having the UI open.

chiljamgossow: Can we call it isUpdatingData or the like to make more clear? My first association was refreshing the vaccination.

chiljamgossow:
Looks well structured overall, thanks for the preparation

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinatedPerson.kt
d4rken: Yes, will do in another PR.

PR #3014, opened 2021-04-29
Fix positive test abort condition for ENS key provision (EXPOSUREAPP-6879)

d4rken:
We want to abort if there is a positive test, and not just when we are allowed to submit (which is transient).

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
@@ -114,9 +114,9 @@ class DownloadDiagnosisKeysTask @Inject constructor(
             // remember version code of this execution for next time
             settings.updateLastVersionCodeToCurrent()
 
-            val isAllowedToSubmitKeys = coronaTestRepository.coronaTests.first().any { it.isSubmissionAllowed }
-            if (isAllowedToSubmitKeys) {
-                Timber.tag(TAG).i("task aborted, positive test result")
+            val isPositive = coronaTestRepository.coronaTests.first().any { it.isPositive }
+            if (isPositive) {
+                Timber.tag(TAG).i("Task aborted, positive test result available.")
                 return object : Task.Result {}
             }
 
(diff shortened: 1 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -114,10 +114,10 @@ class DownloadDiagnosisKeysTask @Inject constructor(
             // remember version code of this execution for next time
             settings.updateLastVersionCodeToCurrent()
 
-            val isAllowedToSubmitKeys = coronaTestRepository.coronaTests.first().any { it.isSubmissionAllowed }
-            if (isAllowedToSubmitKeys) {
-                Timber.tag(TAG).i("task aborted, positive test result")
-                return object : Task.Result {}
+            val isPositive = coronaTestRepository.coronaTests.first().any { it.isPositive }
```
LukasLechnerDev, line 117: Naming suggestion `isPositive` => `hasPositiveTest` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/execution/PresenceTracingWarningTask.kt
```diff
@@ -89,7 +90,13 @@ class PresenceTracingWarningTask @Inject constructor(
 
             presenceTracingRiskRepository.reportCalculation(successful = true)
 
-            return Result(calculatedAt = nowUTC)
+            return Result()
+        }
+
+        val isPositive = coronaTestRepository.coronaTests.first().any { it.isPositive }
```
LukasLechnerDev, line 96: Naming suggestion `isPositive` => `hasPositiveTest` ?

LukasLechnerDev (approved):
LGTM. 

One renaming suggestion.

mtwalli (approved):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/execution/PresenceTracingWarningTask.kt
d4rken: Hm, feels wrong. We are asking the `CoronaTest` if it has a positive test? The `CoronaTest` IS the positive test :smile:

PR #3016, opened 2021-04-29
Adjust test polling behavior 21 days  after registration (EXPOSUREAPP-6882)

d4rken:
* Map HTTP 400 errors to *_REDEEMED after 21 days (server has deleted data set).
* Stop test polling after 21 days if we are in a final state (*_REDEEMED).

* See ticket and related tech spec changes. 2.1.x hotfix target.
* Check test cases, right PCR/RAT results?
* Testable via mockserver and timetravel

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTestExtensions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTestExtensions.kt
@@ -4,5 +4,5 @@ import org.joda.time.Duration
 import org.joda.time.Instant
 
 fun CoronaTest.isOlderThan21Days(nowUTC: Instant): Boolean {
-  return Duration(registeredAt, nowUTC).standardDays > 21
+    return Duration(registeredAt, nowUTC).standardDays > 21
 }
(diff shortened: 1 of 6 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTestExtensions.kt
```diff
@@ -0,0 +1,8 @@
+package de.rki.coronawarnapp.coronatest.type
+
+import org.joda.time.Duration
+import org.joda.time.Instant
+
+fun CoronaTest.isOlderThan21Days(nowUTC: Instant): Boolean {
+    return Duration(registeredAt, nowUTC).standardDays > 21
```
BMItter, line 7: Personal preference, but i would use a default param here

BMItr (changes requested):
NEVER! 
(sry, wrong PR 🥇 )

BMItr (approved):
lgtm!

chiljamgossow (approved):
Looks correct to me

mlenkeit (approved):
LGTM, verified for both PCR and RAT.

PR #3019, opened 2021-04-30
Settings toggle (eg. tracing) (EXPOSUREAPP-6871)

chiljamgossow:
Use a custom view that has only one click event receiver to avoid inconsistencies in states and simplify layout.

Screen adapted as they use the same view:
SubmissionYourConsent
Tracing Settings
Notification Settings
PPA settings

How to test: Toggle on the views to make sure they still work.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/SwitchRowView.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    private val switch by lazy { findViewById<SwitchMaterial>(R.id.switch_view) }
+    private val title by lazy { findViewById<TextView>(R.id.switch_title) }
+    private val subtitleView by lazy { findViewById<TextView>(R.id.switch_subtitle) }
+
+    init {
+        LayoutInflater.from(context).inflate(R.layout.switch_row, this, true)
+    }
+
+    fun setChecked(turnedOn: Boolean?) {
```
LukasLechnerDev, line 25: Why `Boolean?` and not `Boolean` ?

Corona-Warn-App/src/main/res/layout/fragment_settings_notifications.xml
```diff
@@ -17,16 +17,15 @@
         android:layout_width="match_parent"
         android:layout_height="match_parent">
 
-        <include
+        <androidx.appcompat.widget.Toolbar
```
LukasLechnerDev, line 20: Is there a reason why we use `Toolbar` instead of `MaterialToolbar` here?

LukasLechnerDev (dismissed):
LGTM. Just 2 questions. 

Nice Refactoring. 
Tested on Samsung S20+; All switches working as expected

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/SwitchRowView.kt
chiljamgossow: For data binding. Can be null initially.

Corona-Warn-App/src/main/res/layout/fragment_settings_notifications.xml
chiljamgossow: Nope, plain ignorance.

d4rken (approved):
lgtm

LukasLechnerDev (approved):

PR #3025, opened 2021-04-30
Fix being able to change optional data donation entries if sharing is disabled (EXPOSUREAPP-5489)

d4rken:
* Rows are now gone when sharing is disabled
* Reuse MoreInformationView in layout for consistency.
* Extend MoreInformationView to offer toggle.

Screenshot from 2021-04-30 12-08-22
Screenshot from 2021-04-30 12-07-57


axelherbstreith (approved):
LGTM 👍 Tested on Huawei P30 with Android 10

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/views/MoreInformationView.kt
```diff
@@ -14,9 +17,12 @@ class MoreInformationView @JvmOverloads constructor(
     context: Context,
     attrs: AttributeSet? = null,
     defStyleAttr: Int = 0
-) : ConstraintLayout(context, attrs, defStyleAttr) {
+) : ConstraintLayout(context, attrs, defStyleAttr), Checkable {
```
LukasLechnerDev, line 20: Naming again :) Should me think about a different name now that the functionality of this view has increased? Maybe something like `SettingsRowView`?

LukasLechnerDev (approved):
LGTM! Maybe rename `MoreInformationView` into something more appropriate.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/views/MoreInformationView.kt
d4rken: Chilja has created something that is more like the preference/settings views I think:
https://github.com/corona-warn-app/cwa-app-android/pull/3019

We can merge both first, as each improve existing code already, and then evaluate re-using/renaming. Let's see when we have time this sprint :cold_sweat:

PR #3026, opened 2021-04-30
Refactor Contact Diary Work Scheduler (DEV)

LukasLechnerDev:
Work is now scheduled in the same way as for other workers. 

Now only `.setup()` is called at app start and the scheduler now observes changes of the onboarding status on its own and so we don't have to schedule the work from other places within the app.


chris-cwa (approved):

d4rken (approved):
:+1:

PR #3027, opened 2021-04-30
Fix flaky unit test (DEV)

d4rken:
* Fix unreliable test for dev environment change fragment
* Remove class that was `smartLiveData` that was only used in that viewmodel and unnecessary
* Add missing license text


chiljamgossow (approved):
Thanks. This was annoying.

jurajkusnier (approved):
LGTM

PR #3049, opened 2021-05-03
Display risk calculated at on risk cards (DEV)

chiljamgossow:
To prevent confusion, the timestamp shown is changed to the one from the in-app calculation.


chiljamgossow:
obsolete

PR #3051, opened 2021-05-03
Fix risk calc timestamp (EXPOSUREAPP-6857)

chiljamgossow:
Currently we show the timestamp from ENF exposure detection. However, this leads to confusion. 

So when the last result was due to a new PT calculation, the timestamp is taken from this calculation.


vaubaehn:
@chiljamgossow @d4rken 
Is there a way to track both timestamps (ENS/PT)? This might help monitoring correct functioning/debugging a lot, e. g. #2880 #2881 , when ENS exposure detection is not working. One possibility would be to move some of those information to the App information screen (e.g. "Last ENS exposure detection: 03.05.2021 19:43 /n Last PT exposure detection: 03.05.2021 19:45" or alike...

harambasicluka:
> @chiljamgossow @d4rken
> Is there a way to track both timestamps (ENS/PT)? This might help monitoring correct functioning/debugging a lot, e. g. #2880 #2881 , when ENS exposure detection is not working. One possibility would be to move some of those information to the App information screen (e.g. "Last ENS exposure detection: 03.05.2021 19:43 /n Last PT exposure detection: 03.05.2021 19:45" or alike...)
> 
> From what I understand from this PR, the timestamp of the last successful risk calculation is shown.
> This would mean, if ENS if failing always (for any reason), the risk card reflects the PT risk calculation only (and the other way round). Therefore it would be good to know both timestamps, to be able to see whether both calculations are running correctly.

I like your suggestion, this PR firstly fixes an issue. I would recommend to open a wishlist issue where more details can be discussed.

vaubaehn:
@harambasicluka 
Hi Luka,
> I would recommend to open a wishlist issue where more details can be discussed.

I will do so then. Thanks, and enjoy the rest of the day!

PR #3053, opened 2021-05-03
Updated Pending Result Screen (EXPOSUREAPP-6897)

axelherbstreith:
Adapted the Pending Test result screen for rapid antigen test according to Figma while leaving the screen for PCR tests as it currently is.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingViewModel.kt
```diff
@@ -27,7 +27,7 @@ import timber.log.Timber
 class SubmissionTestResultPendingViewModel @AssistedInject constructor(
     dispatcherProvider: DispatcherProvider,
     private val submissionRepository: SubmissionRepository,
-    @Assisted private val testType: CoronaTest.Type
+    @Assisted val testType: CoronaTest.Type
```
LukasLechnerDev, line 30: You could access the type via `navArgs.testType` in the `ViewModel`. Then this property could stay private.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingFragment.kt
```diff
@@ -60,11 +62,22 @@ class SubmissionTestResultPendingFragment : Fragment(R.layout.fragment_submissio
             }
         }
 
+        binding.apply {
```
LukasLechnerDev, line 65: Its a bit hard to figure out which view is displayed in which case 😅.

what about keeping it simple like

```
when (viewModel.testType) {
    CoronaTest.Type.PCR -> {
        testResultPendingStepsContactDiaryResult.isVisible = false
        testResultPendingStepsWaitingPcrResult.isVisible = true
        // and so on
    }
    CoronaTest.Type.RAPID_ANTIGEN -> {
        testResultPendingStepsContactDiaryResult.isVisible = true
        testResultPendingStepsWaitingPcrResult.isVisible = false
        // and so on
   }
}
```

LukasLechnerDev (changes requested):
2 optional suggestions. 

Why I request changes is that afaik we are not showing the RAT results in the diary yet (I think we will implement this in the next sprint (Version `2.4`)) So I think we need to remove the "Ihr Testergebnis wird dem Kontakt-Tagebuch hinzugefügt" Entry for now?

axelherbstreith:
@LukasLechnerDev nice catch! The row is hidden now 👍

LukasLechnerDev (dismissed):
LGTM! Thanks for the fix :)

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1068,9 +1068,21 @@
     <!-- YTXT: Body text for for results next steps  -->
     <string name="submission_test_result_steps_added_body">"Your test has been stored in the Corona-Warn-App."</string>
     <!-- XHED: Page headline for pending test result next steps  -->
-    <string name="submission_test_result_pending_steps_waiting_heading">"Your test result is not available yet."</string>
+    <string name="submission_test_result_pending_steps_waiting_pcr_heading">"Your test result is not available yet."</string>
```
harambasicluka, line 1071: Just add the strings here, and do not change the old identifiers. With that you don't have to change other string files.

InkenTownsend (approved):
Correct according to https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6897. UA reviewed.

chiljamgossow (approved):
LGTM

PR #3056, opened 2021-05-03
support umlauts in RAT QR codes (EXPOSUREAPP-7034)

jurajkusnier:
- Fix umlauts support (First Name and Last Name, e.g. "Rüdiger", "Müller".)

### Testing
test scanning RAT QR code generated by mock server, new flag has been added `--flag umlaut`


harambasicluka:
no client issue, fix is not needed.

PR #3057, opened 2021-05-03
Test not found exception (EXPOSUREAPP-6870)

mtwalli:
- Avoid crashing for end user, just return current set of tests
- Many places are not handling potential crash in UI


mtwalli:
Closing this one , I will create a follow up PR for EXPOSUREAPP-6870

PR #3058, opened 2021-05-03
Vaccination Repository Part 1 (EXPOSUREAPP-6729)

d4rken:
* Storage logic
* Initial drafts stored data (some specs are still TBD)
* Additional wiring & structure

Doesn't really do anything yet as server side is not ready yet, but it supplies an initial data set.
Review architecture, comment on potential issues, once all concerns are addressed, we should merge this as other tickets need this to integrate.

Draft:
- [x] More unit tests


d4rken:
I'll perform a follow up PR with more logic and unit tests. So this is mostly about structure.

chris-cwa (approved):
i like it!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinationCertificate.kt
```diff
@@ -1,20 +1,24 @@
...
+
+    val vaccineName: String
+    val vaccineManufacturer: String
+    val medicalProductName: String
+
+    val doseNumber: Int
+    val totalSeriesOfDoses: Int
+
+    val certificateIssuer: String
+    val certificateCountry: Country
```
chiljamgossow, line 20: Ist that correct? Shouldn't it be a String or ISO code?

chiljamgossow (approved):
LGTM, just one comment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/qrcode/VaccinationCertificateV1.kt
```diff
@@ -0,0 +1,42 @@
...
+) {
+
+    data class NameData(
+        @SerializedName("fn") val familyName: String?,
+        @SerializedName("fnt") val familyNameStandardized: String,
+        @SerializedName("gn") val givenName: String?,
+        @SerializedName("gnt") val givenNameStandardized: String?,
+    )
+
+    data class VaccinationData(
```
mtwalli, line 20: How this will scale for `n` certificates ? is not it better to save it a DB instead ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinationCertificate.kt
```diff
@@ -1,20 +1,24 @@
...
-    val vaccineManufacturer: String,
-    val chargeId: String,
-    val certificateIssuer: String,
-    val certificateCountry: Country,
-    val certificateId: String,
-) {
-    val identifier: VaccinatedPersonIdentifier get() = ""
+interface VaccinationCertificate {
+    val firstName: String?
+    val lastName: String
```
mtwalli, line 8: How about  aligning naming here with Tech specs ? `givenName and familyName `

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/repository/VaccinationRepository.kt
```diff
@@ -1,85 +1,207 @@
...
+    init {
+        internalData.data
+            .onStart { Timber.tag(TAG).d("Observing test data.") }
+            .onEach { vaccinatedPersons ->
+                Timber.tag(TAG).v("Vaccination data changed: %s", vaccinatedPersons)
+                storage.personContainers = vaccinatedPersons.map { it.data }.toSet()
+            }
+            .catch {
+                it.reportProblem(TAG, "Failed to snapshot vaccination data to storage.")
+                throw it
```
mtwalli, line 72: so this will crash the App then , I would like to change this approach here since you already saw  from CoronaTestRepo we had to do many changes after the feature was already shipped

d4rken: If there is an exception when we serialize the data, it should never have been created.
This is a case of "should never happen", and I don't see a way to recover from this.
If we save it, we corrupt the previously successfully stored data, and "all" is lost.
If we crash the can restart again with a valid data set.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinationCertificate.kt
d4rken: I always find this very confusing and I'm more used to first/last name. If the other devs do not share this opinion, then I'm fine to refactor this in a follow up PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/qrcode/VaccinationCertificateV1.kt
d4rken: There is a lot of changes happening to the attributes, storing it as JSON makes it easier to add/remove attributes. Database changes will require migrations each time.

The worst case scenario seems to be:
*  2 parents? + a few children 4?
* 1-3 vaccinations per
* 1-2 proof certificates per

Still seems small enough to handle via JSON, and we don't need any complex queries on the data, and we have to load the data each time anyways to show on the home screen.

I'd prefer to go ahead with JSON for now, should the requirements change unexpectedly, we can just exchange the storage layer.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/VaccinationCertificate.kt
d4rken: `VaccinationCertificate` is user facing, we internally work with strings, I initially thought that we might need to display some kind of country flag.

I'd put this up to @mtwalli  or @LukasLechnerDev  to refactor in a future PR if a string is more useful here.

chiljamgossow: It was more about the value set we cover. We should accept all countries.

PR #3061, opened 2021-05-04
Home Screen - Registration card (EXPOSUREAPP-6724)

jurajkusnier:
- Add register certificate card design on the home screen
- Implementation will be included in future PR
image


fynngodau:
Quick question, since I don't have access to the mockups: will there also be a tab for displaying the certificate?

jurajkusnier:
> Quick question, since I don't have access to the mockups: will there also be a tab for displaying the certificate?

The certificate will be displayed after clicking on another new home screen card.

fynngodau:
@jurajkusnier Thanks for the info. I will open a wishlist item with my thoughts regarding tabs

chiljamgossow (dismissed):
LGTM

Corona-Warn-App/src/main/res/layout/home_vaccination_registration_card.xml
```diff
@@ -0,0 +1,72 @@
...
+            app:layout_constraintTop_toBottomOf="@+id/title" />
+
+        <androidx.appcompat.widget.AppCompatImageView
+            android:id="@+id/icon"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:importantForAccessibility="no"
+            android:layout_marginEnd="16dp"
+            app:layout_constraintEnd_toEndOf="parent"
+            app:layout_constraintTop_toTopOf="@+id/body"
```
axelherbstreith, line 47: please only use `@+id/` for `android:id`, else use `@id/ `

jurajkusnier: true, I reused an older card as template

axelherbstreith (approved):
LGTM

chiljamgossow (approved):

PR #3064, opened 2021-05-04
Remove internet requirement for EW risk calculation (EXPOSUREAPP-6948)

d4rken:
We don't need internet for the risk calculation and using parameters from a cached config will be sufficient.
This will likely dramatically reduce the amount white "calculation failed" cards being displayed.


kolyaopahle (approved):
nice 👍 i was able to trigger a calculation from dev menu without internet connection enabled 🚀

chiljamgossow (approved):
LGTM

PR #3065, opened 2021-05-04
Rapid Test - QR-Code from Schnelltestprofile cannot be scanned by PoC if fields contain Umlauts (EXPOSUREAPP-7016)

kolyaopahle:
### Description
RAT Profile qr codes were not encoded as utf-8 which is the only format the testcenter barcode library supports

### Steps to reproduce
1. Create a RAT Profile
2. Scan the QR Code in the testcenter frontent
3. Check if the code was able to be scanned


harambasicluka (dismissed):
Tested on the Schnelltest Platform on WRU 🚀

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/qrcode/QrCodeGenerator.kt
```diff
@@ -38,7 +38,8 @@ class QrCodeGenerator @Inject constructor(
 
         val hints = mapOf(
             EncodeHintType.ERROR_CORRECTION to correctionLevel,
-            EncodeHintType.MARGIN to margin
+            EncodeHintType.MARGIN to margin,
+            EncodeHintType.CHARACTER_SET to Charsets.UTF_8
```
mtwalli, line 42: this would cause a crash on older Android versions, did you test it on Android 6 ?
I had it before and remove it on purpose. I would suggest to add SDK version check and or make it configurable 
as this also affecting Presence Tracing Qr-Codes

kolyaopahle: No i did not, i do not know if i version check would help here, qr codes litterally cannot be parsed by the ui in a different encoding, is there a specific reason for the crash that we could work around?

kolyaopahle: should be fixed by the latest commit, turned out to just be a issue with .toString evaluation results

mtwalli (approved):

harambasicluka (approved):

PR #3072, opened 2021-05-04
No risk in journal - special switch (EXPOSUREAPP-7039)

BMItr:
This PR addresses Exposureapp-7039. ([Android] no High Risk in Kontakt-Tagebuch for CheckIns with exposure)

How to Test:

- Mobile - Check-in to an Event with two devices (QR is inside the ticket) with **switch off** (‘Nach dem Auschecken in mein Tagebuch eintragen‘)
- Mobile – Check-Out from the two Locations with both devices
- Report one device/event infected with testresult=2 (positive) and shares the diagnosis keys
- Check the second device (got warning with high risk) to have an entry inside the contact journal

Expected Result:
Attendee gets a red Risk Card and one entry in Kontakt-Tagebuch with ‘Erhöhtes Risiko aufgrund Ihrer Anwesenheit bei: ...(erhöhtes Risiko)‘

Screenshot:
before this adjustment there was no risk shown
Now it should look like this (if you use the QR Code inside the ticket)

image


LukasLechnerDev (approved):
Nice Fix! Works as described. Thanks for the good PR description. 👍

image

axelherbstreith (approved):

PR #3074, opened 2021-05-04
Corona test error handling (DEV)

mtwalli:
@d4rken  and I did some changes for better error handling of `CoronaTestNotFoundException`

- Handle all calls of `CoronaTestNotFoundException.modifyTest` in directly either in the view model or `SubmissionRepository`
- Move calls to mark test as viewed to onViewCreated instead of onResume
- One important change in `ShareTestResultNotificationService` is revert `catch` clause change 
why?
```kotlin
     try {
            coronaTestRepository.latestPCRT // This should not crash ,it is just a flow already constructed. if it throws an  exception ,it would have been caught in try,catch block
                .onEach {
                    if (it == null) {
                        resetSharePositiveTestResultNotification(PCR) // any exception is thrown in here won't be caught in try,catch block because it is launched in appScope
                    }
                   
                }
                .launchIn(appScope)
        }catch (e:Exception){
            Timber.e(e,"Failed to reset positive test result reminder for PCR test.")
        }
```

We kept the following form
```kotlin
         coronaTestRepository.latestPCRT // This should not crash ,it is just a flow already constructed and therefore no need to try and catch
            .onEach {
                if (it == null) {
                    resetSharePositiveTestResultNotification(PCR)
                }
            }
            // Catch errors in onEach block only - which is basically guarding showing the notification logic 
            .catch { it.reportProblem("Failed to reset positive test result reminder for PCR test.") }
            .launchIn(appScope)
```


BMItr (approved):
nothing to complain.
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
```diff
@@ -66,20 +67,20 @@ class SubmissionRepository @Inject constructor(
...
             coronaTestRepository.markAsViewed(identifier = test.identifier)
         }
     }
 
-    fun refreshTest(type: CoronaTest.Type? = null) {
+    suspend fun refreshTest(type: CoronaTest.Type? = null) {
         Timber.tag(TAG).v("refreshTest(type=%s)", type)
 
-        scope.launch {
+        withContext(scope.coroutineContext) {
```
LukasLechnerDev, line 83: This change is needed so that `refreshTest` rethrows the exception, right? Otherwise, we wouldn't be able to handle exceptions from the call site with try-catch - or?

LukasLechnerDev (approved):
LGTM. Interesting that we can't use try-catch with `launchIn`. Found out that it internally starts a new coroutine, and that's the issue.

One question though 👇

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
mtwalli: Yes, that is right , launching in appScope would not propagate crashes to try and catch. and our AppScope does not have an error handler and in this case it will crash the App

LukasLechnerDev: Okay. Thanks!

PR #3085, opened 2021-05-05
Adjust test availability constant (DEV)

d4rken:
The server returns a different result after 60, not 21 days.
This should currently only be able to happen when time traveling.


jurajkusnier (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationServer.kt
```diff
@@ -114,6 +115,12 @@ class VerificationServer @Inject constructor(
         const val PADDING_LENGTH_BODY_TAN_FAKE = 31 + VERIFICATION_BODY_FILL
         const val DUMMY_REGISTRATION_TOKEN = "11111111-2222-4444-8888-161616161616"
 
+        /**
+         * Test is available for this long on the server.
+         * After this period the server will delete it and return PENDING if the regtoken is polled again.
+         */
+        val TEST_AVAILABLBILITY = Duration.standardDays(60)
```
chiljamgossow, line 122: You could append _IN_DAYS to make the unit obvious, also there is a typo

chiljamgossow (approved):
LGTM, just a typo

PR #3091, opened 2021-05-05
Additional HTTP 304 check for remote config (EXPOSUREAPP-6298)

d4rken:
Don't retrieve the whole config if the server returns HTTP 304 Not Modified.


mtwalli (approved):
Looks good 👍

chris-cwa (approved):

Issue #2928, opened 2021-04-23
Android App: PDF generated for event has an awkward poster size, might lead to print errors


tik-boomer:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [?] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Google Pixel 3 XL
- Android version: 11,   Build RQ2A.210405.005
- App version: 2.03

## Describe the bug

The PDF poster generated for an event has a mismatch in PDF-file propertied between total size, and pages size.

## Steps to reproduce the issue

1. Generate a new event place, under Education / Bildungseinrichtung. 
2. Share the generated QR-code Poster by gmail.
3. download the QR-code poster using Adobe acrobat Reader
4. Start to print the document 
5. In the print dialoge chose actual size ("Tatsächliche Größe")
6. The print window will only include part of the qr-code poster (screenshot included)

## Expected behaviour

When opening the app generated QR-code poster PDF-file for printing, printing at actual size should print the complete image, in actual size, **on an A4 page**.

It seems to me that the document and image were actually generated much large than A4 size, at 508 x 718 mm. This way the local printout has to be scaled down to A4 first. Should not A4 be the standard, and easily printed with all printing presets --
[cwa-qr-code.pdf](https://github.com/corona-warn-app/cwa-app-android/files/6364731/cwa-qr-code.pdf)
![Event-QR print size](https://user-images.githubusercontent.com/66428107/115861664-54ec6d00-a433-11eb-92c6-126469174b88.png)

 and the people needing bigger printouts scale up? In the privacy concerns it is suggested to generate and print a new qr-code every day. Who has a printer of this format for daily routine printing? :-)

## Possible Fix

Simply generate the image and the PDF for full A4 format.

## Additional context

I see this with Android 11, butdo not have a IOS device, so I assume, but can not be sure, that this issue exists in the IOS app, too.

Attachments:
- Screen shot of print dialogue
- PDF file

---
Internal Tracking ID: [EXPOSUREAPP-6696](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6696)

Ein-Tim:
[...]

PR #3095, opened 2021-05-06
Device agnostic PDF size (EXPOSUREAPP-6696)

mtwalli:
- Export PDF files in A4 size regardless of device's display 
- Get the best of two worlds, display the PDF in high resolution and export it in a reasonable A4 quality
- Happy printing 🖨️  CWA community

### Testing 
Exporting PDF on any device should have `resolution = 595 x 842`


dsarkar:
fixes https://github.com/corona-warn-app/cwa-app-android/issues/2928

axelherbstreith (approved):
LGTM 👍 Tested on Huawei P30, printing option is now set to DIN A4 as default

d4rken (approved):

PR #3099, opened 2021-05-06
Vaccination Server - Retrofit interfaces & Module  (EXPOSUREAPP-6854, EXPOSUREAPP-6856, EXPOSUREAPP-7129)

BMItr:
This PR is related to EXPOSUREAPP-6854, EXPOSUREAPP-6856, EXPOSUREAPP-7129.
Vaccination Server.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/environment/EnvironmentSetup.kt
```diff
@@ -36,6 +38,8 @@ class EnvironmentSetup @Inject constructor(
         VERIFICATION_KEYS("PUB_KEYS_SIGNATURE_VERIFICATION"),
         DATA_DONATION("DATA_DONATION_CDN_URL"),
         LOG_UPLOAD("LOG_UPLOAD_SERVER_URL"),
+        VACCINATION_PROOF("VACCINATION_PROOF_SERVER_URL"),
+        VACCINATION_VALUE("VACCINATION_VALUE_SET_CDN_URL"),
```
d4rken, line 42: Let's pick a different name here, as "value sets" is just a specific data set that can be returned here, not sure what we could call it though, maybe just VACCINATION_CDN? Does the tech spec call it something useful?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/environment/EnvironmentSetup.kt
```diff
@@ -36,6 +38,8 @@ class EnvironmentSetup @Inject constructor(
         VERIFICATION_KEYS("PUB_KEYS_SIGNATURE_VERIFICATION"),
         DATA_DONATION("DATA_DONATION_CDN_URL"),
         LOG_UPLOAD("LOG_UPLOAD_SERVER_URL"),
+        VACCINATION_PROOF("VACCINATION_PROOF_SERVER_URL"),
```
d4rken, line 41: Please also extend `prod_environments.json` with a placerholder and create a PR on our internal repo to add placeholders/keys to our `test_environments.json`.

d4rken: When using values, use something that satisfies `requireValidUrl` e.g. `https://placeholder`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/server/proof/ProofCertificateServerData.kt
```diff
@@ -0,0 +1,5 @@
+package de.rki.coronawarnapp.vaccination.core.server.proof
+
+interface ProofCertificateServerData {
+    // Satisfy CI ¯\_(ツ)_/¯
```
d4rken, line 4: :smile:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/server/valueset/VaccinationValueSetApiV1.kt
```diff
@@ -0,0 +1,16 @@
...
+import retrofit2.Response
+import retrofit2.http.GET
+import retrofit2.http.Header
+import retrofit2.http.Path
+
+interface VaccinationValueSetApiV1 {
+
+    @GET("/version/v1/ehn-dgc/{lang}/value-sets")
+    suspend fun getValueSets(
+        @Header("If-None-Match") etag: String = "",
```
d4rken, line 13: Do we need to set that? Could  we just use `okhttp.Cache` which includes handing for this etag/304/if-none-match ?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/environment/EnvironmentSetupTest.kt
```diff
@@ -127,7 +127,9 @@ class EnvironmentSetupTest : BaseTest() {
         EnvironmentSetup.EnvKey.LOG_UPLOAD.rawKey shouldBe "LOG_UPLOAD_SERVER_URL"
         EnvironmentSetup.EnvKey.SAFETYNET_API_KEY.rawKey shouldBe "SAFETYNET_API_KEY"
         EnvironmentSetup.EnvKey.CROWD_NOTIFIER_PUBLIC_KEY.rawKey shouldBe "CROWD_NOTIFIER_PUBLIC_KEY"
-        EnvironmentSetup.EnvKey.values().size shouldBe 9
+        EnvironmentSetup.EnvKey.VACCINATION_PROOF.rawKey shouldBe "VACCINATION_PROOF_SERVER_URL"
```
d4rken, line 130: Please include the new keys in the json tests too.

d4rken (commented):
Looks good.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/server/valueset/VaccinationValueSetApiV1.kt
BMItter: had it first. liked it more this way.
but I reverted.
so:
adjusted :-)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/environment/EnvironmentSetup.kt
BMItter: adjusted.
will make a follow up pr for test_env -> internal repo

d4rken (approved):
lgtm

chris-cwa (approved):

PR #3101, opened 2021-05-06
Separate links for different types of incompatibility warnings (COMMUNITY)

fynngodau:
When writing an FAQ entry for the incompatibility warning card in https://github.com/corona-warn-app/cwa-website/pull/1211#discussion_r627494611, we noticed usability problems with my draft for the FAQ article we were discussing:

> […] it's not clear from the error message what type of incompatibility (full or partial) the user is facing. This can be a problem because it might leave the user more confused after reading the FAQ article than after reading the error messages.

The best option, to me, would be to add two separate FAQ article entries. However, due to the before-mentioned problem, it would be critical that users are linked to the correct article, so that they don't falsely assume that the app's (proximity-based) risk assessment is completely non-functional because they couldn't tell that they are only facing a "partial" and not a "full" incompatibility warning.

For this reason, I think it would be important to add this change to the 2.2.x release. I would appreciate your feedback if this is possible.

---
Internal Tracking ID: [EXPOSUREAPP-7118](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7118)


fynngodau:
I changed the link target to a shorter link anchor as @heinezen asked me.

harambasicluka:
This can't go into 2.2. Development is already closed. Only hotfixes get in.

Also please don't change other language files except german (values-de) and the defaults (values). Everything else will be handled by our translation system.

fynngodau:
@harambasicluka Alright, too bad.

> Also please don't change other language files except german (values-de) and the defaults (values). Everything else will be handled by our translation system.

Then lint fails, though, because of strings that exist in some languages but not in the defaults.

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -2018,6 +2018,8 @@
     <string name="incompatible_advertising_not_supported">"Your smartphone can only receive COVID-19 notifications via Bluetooth, but not send them. This means you can be warned of exposures through this interface, but you cannot warn others of exposures yourself. You can send and receive warnings that result from check-ins."</string>
```
MikeMcC399, line 2018: 1. Is it too late to suggest a change to the app message in English?
"Your smartphone can only receive COVID-19 notifications via Bluetooth, but not send them."
"Your smartphone cannot send or receive COVID-19 notifications via Bluetooth."

and use exactly the same term as the Google UI uses, i.e. "COVID-19 Exposure Notifications".

The German text is fine, since it uses the term "COVID-19-Benachrichtigungen" exactly like Google does.

MikeMcC399 (changes requested):
I made a suggestion about the English text so it is exactly the same as the term which Google uses.

Also, the labels like      all have a typo in them. "Incompitability" should be "Incompatability". I don't know if that can easily be fixed. Perhaps that will break the translation system?

ralfgehrer:
> @harambasicluka Alright, too bad.
> 
> > Also please don't change other language files except german (values-de) and the defaults (values). Everything else will be handled by our translation system.
> 
> Then lint fails, though, because of strings that exist in some languages but not in the defaults.

If you provide DE and DEFAULT strings, you should be able to remove all string file changes of other languages without lint complaining, or am I mistaken?

Background info: We have a translation process in place that does the translation of all keys. To avoid merge conflicts etc. we don't change other language files manually.

ralfgehrer:
@fynngodau many thanks. LGTM. I asked for UA text approval.

ChristinaTsigganos (dismissed):
UA approved

Corona-Warn-App/src/main/res/values/strings.xml
fynngodau: Sorry for late respone – since UA already approved this one I'd suggest to open a new PR / issue about this

fynngodau:
Ah, now that I resolved the merge conflict, the review disappeared again, sorry @ChristinaTsigganos.

Corona-Warn-App/src/main/res/values/strings.xml
MikeMcC399: @fynngodau 
Perhaps we could get a comment here from @ChristinaTsigganos before opening any new issue / PR?

ChristinaTsigganos (approved):
UA approved

ralfgehrer (approved):

MikeMcC399:
I had suggested changing:

"Your smartphone can only receive COVID-19 notifications via Bluetooth, but not send them."
to
"Your smartphone can only receive COVID-19 Exposure Notifications via Bluetooth, but not send them."

in the default / EN string
`incompatible_advertising_not_supported`

I will not pursue this further if others think that the existing proposed text in this PR is good.
My suggestion was to align the text with the Google UI text. I don't want to complicate the issue by adding a PR on top of an open PR.

ralfgehrer:
> I had suggested changing:
> 
> "Your smartphone can only receive COVID-19 notifications via Bluetooth, but not send them."
> to
> "Your smartphone can only receive COVID-19 Exposure Notifications via Bluetooth, but not send them."
> 
> in the default / EN string
> `incompatible_advertising_not_supported`
> 
> I will not pursue this further if others think that the existing proposed text in this PR is good.
> My suggestion was to align the text with the Google UI text. I don't want to complicate the issue by adding a PR on top of an open PR.

Thanks @MikeMcC399. I'd suggest we merge this PR as is but feel free to open a follow-up PR once merged.

mtwalli (approved):

PR #3107, opened 2021-05-07
Vaccination qr code extraction (EXPOSUREAPP-6726)

chiljamgossow:
Parses raw QR code and validates the input

Nothing to test yet, just check the code
Tech spec https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/vaccination-certificate/docs/spec/vaccination-certificate-client.md#scan-qr-code-to-register-a-vaccination-certificate


Corona-Warn-App/build.gradle
```diff
@@ -437,4 +437,7 @@ dependencies {
 
     // ANIMATIONS
     implementation "com.airbnb.android:lottie:3.5.0"
+
+    // HCert
+    implementation("com.augustcellars.cose:cose-java:1.1.0")
```
harambasicluka, line 442: Needs to be added to our internal tool, @AlexanderAlferov can help here.

chiljamgossow: Changed the lib, but still need to add it (where?)

d4rken (dismissed):
lgtm :+1:

Corona-Warn-App/build.gradle
chiljamgossow: I added a task to Jira to ask Katharina (7135)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/qrcode/VaccinationCertificateV1Parser.kt
```diff
@@ -0,0 +1,72 @@
...
+import com.upokecenter.cbor.CBORObject
+import de.rki.coronawarnapp.util.serialization.fromJson
+import de.rki.coronawarnapp.vaccination.core.qrcode.InvalidHealthCertificateException.ErrorCode.HC_CBOR_DECODING_FAILED
+import de.rki.coronawarnapp.vaccination.core.qrcode.InvalidHealthCertificateException.ErrorCode.VC_NO_VACCINATION_ENTRY
+import org.joda.time.Instant
+import javax.inject.Inject
+
+class VaccinationCertificateV1Parser @Inject constructor() {
+
+    private val keyEuDgcV1 = CBORObject.FromObject(1)
```
d4rken, line 13: Could we make this part of a companion object, as we may have multiple instances of `VaccinationCertificateV1Parser`, but these indexes seem static.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/qrcode/VaccinationCertificateV1Parser.kt
```diff
@@ -0,0 +1,72 @@
...
+        }
+        dateOfBirth
+        vaccinationDatas.forEach {
+            it.vaccinatedAt
+        }
+        return this
+    }
+}
+
+data class VaccinationCertificateHeader(
```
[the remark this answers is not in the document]
chiljamgossow: I think the header is not specific to the vaccination. I can move to its own file.

mtwalli (approved):
LGTM, Thanks!

d4rken (approved):

PR #3112, opened 2021-05-07
Remove build tools (DEV)

mtwalli:
- As Android studio suggests already this is not needed anymore. 
- For internal release builds it will pick  `30.0.3` or `30.0.2`

I have built release locally and seems fine


SamuraiKek (approved):
Works OK. Managed to also build a release locally.

jurajkusnier (approved):
LGTM

PR #3115, opened 2021-05-07
Fix duplicate test removal error case handling (EXPOSUREAPP-7124)

d4rken:
Refactor how the screen, that is shown when a duplicate test type is scanned, handles errors.
* In any error case, because the previous test was removed, go back to the submission dispatcher screen (where the user can select a new test type to add).
* Depending on the error type, show as good of an error dialog as we have available.
* Remove and merge the 1x,2x,3x `catch` clauses and remove the no longer being thrown `TransactionException`

This refactoring specifically aims to fix the edge cases, where the user could stay on the "duplicate test type" screen, after an error, and repeatedly press "continue" which would lead to errors because the previous test was already removed.

### Testing
* Reproduce the previous error case by scanning a RAT test twice, but on the second scan, on the "duplicate test warning" screen, have the mockserver return an error. When the error dialog shows, don't press a dialog button, just dismiss the error dialog by pressing somewhere.
* Both on the first scan, and if you replace an existing test, it should not be possible to save a new test, that is already in the state `PCR/RAT_REDEEMED` when you initially scan it.


chiljamgossow (approved):
Code LGTM (not tested)

axelherbstreith (approved):
Looks good 👍 Tested on Huawei P30 with both PCR and RAT and different test result states each

PR #3123, opened 2021-05-10
Material Alert (DEV)

mtwalli:
```kotlin
            - AlertDialog.Builder
            + MaterialAlertDialogBuilder
```


chris-cwa (approved):

d4rken (approved):
I had one error dialog where the body, the "setView" part had a different background color than the rest of the dialog. Could not reproduce it though. :thinking: 

@mtwalli could we use MD theming to now easily set the dialog button colors? We want our accent color blue, in most dialogs it's still the system "light blue" default color for the text.

mtwalli:
> I had one error dialog where the body, the "setView" part had a different background color than the rest of the dialog. Could not reproduce it though. 🤔
> 
> @mtwalli could we use MD theming to now easily set the dialog button colors? We want our accent color blue, in most dialogs it's still the system "light blue" default color for the text.

Yes, MD theme is coming next ,but it is not that straight forward. BUT it will fix many UI consistent issues
such as:
1- Alert dialog buttons colour
2- TimePicker buttons colour 
and coloring in general

PR #3133, opened 2021-05-10
Fix typos en: Check-In (closes #3132) (COMMUNITY)

### LINKED ISSUE #3132: typos en: Check-In in english translation
## Describe the bug

Sometimes Check-In is used instead of Check-in.
See also #3120 .

## Expected behaviour

For conistency and correctness reasons it should be "Check-in", like everywhere else.

## Possible Fix

Just change the texts.

DerVogel2020:
See #3132.


ChristinaTsigganos (approved):
UA has approved the German source text already, so translation receives the changes automatically, therefore we can close this PR from UA point of view.

mtwalli (approved):

PR #3137, opened 2021-05-10
Test result not updated when clicking on "Test anzeigen" (EXPOSUREAPP-7136)

jurajkusnier:
Update test results immediately after a user clicks on the "Test anzeigen" from the home screen card.
## Testing 
1.    Test registrieren > Pending PCR or RAT
1.    go to Homescreen = PCR or RAT show pending test result
1.    set the pending QR to e.g. negative
1.    on the Homescreen tap in PCR or RAT the button "Test anzeigen" 
### Expected result:

-    The negative test result is shown
-    It's not needed to click on button "Aktualisieren"


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/pending/SubmissionTestResultPendingViewModel.kt
```diff
@@ -32,6 +32,7 @@ class SubmissionTestResultPendingViewModel @AssistedInject constructor(
 
     init {
         Timber.v("init() coronaTestType=%s", testType)
+        updateTestResult()
```
mtwalli, line 35: How is it fixing update in `HomeScreen` ? this is the pending test result screen right?

jurajkusnier: From my understanding, this is not about the home screen but rather about test details.
I updated the description to be more clear.

mtwalli (dismissed):

chiljamgossow (commented):
This solution means that if you come from the scan screen there would be an unnecessary second backend call right after the first. Can it be triggered when one clicks on the card, maybe with a nav arg flag?

d4rken:
> This solution means that if you come from the scan screen there would be an unnecessary second backend call right after the first. Can it be triggered when one clicks on the card, maybe with a nav arg flag?

I also think we refresh too often, and we refresh every time we reach the home screen. Not refreshing again when entering the test details screen seems like a feature, not a bug. After all we have a "refresh" button specifically for that on the details screen?

The automatic refresh when entering the details was never an explicit requirement, but more of a "leftover" from previous test code where we didn't save the actual result, so we had to refresh when entering the screen.

We should discuss this again.

jurajkusnier:
> > This solution means that if you come from the scan screen there would be an unnecessary second backend call right after the first. Can it be triggered when one clicks on the card, maybe with a nav arg flag?
> 
> I also think we refresh too often, and we refresh every time we reach the home screen. Not refreshing again when entering the test details screen seems like a feature, not a bug. After all we have a "refresh" button specifically for that on the details screen?
> 
> The automatic refresh when entering the details was never an explicit requirement, but more of a "leftover" from previous test code where we didn't save the actual result, so we had to refresh when entering the screen.
> 
> We should discuss this again.

Yes, it seems unnecessary to me, however, we have this behavior on iOS already implemented so it makes sense to keep the behavior same on both platforms.

jurajkusnier:
> This solution means that if you come from the scan screen there would be an unnecessary second backend call right after the first. Can it be triggered when one clicks on the card, maybe with a nav arg flag?

Yes, it makes sense. Updated

chiljamgossow (approved):
LGTM

mtwalli (approved):

PR #3152, opened 2021-05-12
Merge if conditions in check extension (DEV)

d4rken:
Sonar debt.


d4rken:
Now detekt and sonar are disagreeing with each other :roll_eyes:, the previous solution ends up being easier to read. I'll tell sonar to put up with it...

PR #3153, opened 2021-05-12
Wire up full 'Register Vaccination' Flow (EXPOSUREAPP-7108, EXPOSUREAPP-6732)

LukasLechnerDev:
- Connect List Screen to Vaccination Repo instead of using mockdata
- Wire up navigation (HomeScreenCard -> Scan -> List -> Scan -> List -> Details)
- Remove list screen from the test menu

How to test: 
- Register some vaccinations from our "ImpfzertifikateBeispiele.zip"
- Check if everything works as expected, like in the video below

https://user-images.githubusercontent.com/10398034/117955322-397ccf80-b318-11eb-89ec-4a5d28013dd9.mp4


jurajkusnier (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/list/VaccinationListFragment.kt
```diff
@@ -116,11 +118,13 @@ class VaccinationListFragment : Fragment(R.layout.fragment_vaccination_list), Au
...
-        val behavior: AppBarLayout.ScrollingViewBehavior = params.behavior as (AppBarLayout.ScrollingViewBehavior)
+        val behavior: AppBarLayout.ScrollingViewBehavior =
+            layoutParamsRecyclerView.behavior as (AppBarLayout.ScrollingViewBehavior)
         behavior.overlayTop = (deviceWidth / divider) - 24
     }
 
     companion object {
-        fun navigationUri(personIdentifier: String) = "coronawarnapp://vaccination-list/$personIdentifier".toUri()
+        fun navigationUri(personIdentifierCode: String) =
+            "coronawarnapp://vaccination-list/$personIdentifierCode".toUri()
```
chiljamgossow, line 128: Do we need navigation by URI here? It does not work anymore eg. if the string contains a / so nav args would be preferable

mtwalli: The Vaccination list is part of a  nested graph. afaik we can not navigate by NavDirection from main graph (HomeScreen) into a nested destination (VaccinationList) ,unless we  use `navigate(Uri())` . Do you know if we can use another way?

chiljamgossow: We can merge this and I will adapt to the hashed identifier @d4rken is adding here https://github.com/corona-warn-app/cwa-app-android/pull/3160/files#diff-ef452868e9ae49a80d74fe4f985f31b350be978b3dd6363b9de6ab58ead989b2R29

mtwalli (approved):

PR #3159, opened 2021-05-12
Update default strings (DEV)

ralfgehrer:


DerVogel2020:
Are the changes of the capitalisation of "Check-In" intended?

Corona-Warn-App/src/main/res/values/event_registration_strings.xml
```diff
@@ -4,7 +4,7 @@
     ###################################### -->
 
     <!-- XHED: My check-ins: screen title -->
-    <string name="trace_location_checkins_title">"My Check-ins"</string>
+    <string name="trace_location_checkins_title">"My Check-Ins"</string>
```
mtwalli, line 7: There was a Community PR to do this change and it was UA approved. why do need to change it back?

ralfgehrer: I pinged UA to clarify. Thanks for pointing it out @DerVogel2020 and @mtwalli

NataliaLemmerth: Hi @mtwalli @DerVogel2020 @ralfgehrer, according to Duden, 'Check-in' is a correct spelling. And we have been trying to correct it everywhere in the App texts recently.

ralfgehrer: The English source still uses "Check-In" but this will be fixed with the next translation delivery. To not digress too much, I'd suggest we merge this default strings update and the spelling fix will be applied with the next translation delivery.

DerVogel2020: @ralfgehrer I think the default strings shouldn't be changed to a wrong text, since they are (regarding the "in") correct. See also https://www.oxfordlearnersdictionaries.com/definition/english/check-in_2
@NataliaLemmerth  @mtwalli We should discuss for the english translation the capitalisation of "check".

>  the spelling fix will be applied with the next translation delivery
@ralfgehrer Will the translation delivery update the default strings as well?

NataliaLemmerth: @DerVogel2020 As just discussed with our translator, if a compound or multipart noun is part of a title, all components of the noun are capitalized, and this is the case here. In German we decided to use 'Check-in ' for all occurrences.

DerVogel2020 (approved):
Ok. I just googled and capitalisation of all parts of a noun in a title seems correct in English.
Sorry for confusion 🙄

NataliaLemmerth (approved):
UA approved

mtwalli (approved):

PR #3164, opened 2021-05-12
NOOP vaccination data background update logic for now (EXPOSUREAPP-7251)

d4rken:
We don't need it now, but very likely need it again.
Keep it inactive in prod, keep it active but "noop" in tester builds to notice issues.


ralfgehrer:
Where did this PR's changes go? Did you merge it with the conflict fix @chiljamgossow?

LukasLechnerDev:
@ralfgehrer The commits here are also in this PR https://github.com/corona-warn-app/cwa-app-android/pull/3165 which is already merged. So I think we can close this one.

Issue #3086, opened 2021-05-05
Duration display in Event overflows box with Largest font


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue
Similar to https://github.com/corona-warn-app/cwa-app-android/issues/2610 for Contact Journal

## Technical details

- Device name: Pixel 3a emulator
- Android version: 11
- App version: 2.1.2
(also still present in 2.2.0-RC5)

## Describe the bug

If the Largest font size is chosen, then the duration value in "My Check-Ins" overflows out of the box which is supposed to contain it.

## Steps to reproduce the issue

1. Settings > Display > Advanced
2. Set Font size to Largest
3. Open CWA
4. Scroll down to CREATE QR CODE and tap it
5. Tap NEXT if necessary
6. Tap + QR CODE
7. Tap Retail
8. Enter A in Description and B in Place, Tap SAVE
9. Tap CHECK IN
10. Tap ACCEPT if necessary
11. Tap CHECK IN

Note that the duration value wraps to the next line and then flows out of the box.

![image](https://user-images.githubusercontent.com/66998419/117144676-7d208800-adb2-11eb-900a-592027358c2b.png)

## Expected behavior

The duration value should be contained in the Duration box for any font size selection, including Largest font size.

## Possible Fix

Layout is defined in 
`https://github.com/corona-warn-app/cwa-app-android/blob/release/2.1.x/Corona-Warn-App/src/main/res/layout/trace_location_attendee_checkins_item_active.xml`.

---
Internal Tracking ID: [EXPOSUREAPP-7199](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7199)

dsarkar:
@MikeMcC399 Thanks. Internal Tracking ID: [EXPOSUREAPP-7199](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7199)

dsarkar:
FYI PR https://github.com/corona-warn-app/cwa-app-android/pull/3168

MikeMcC399:
In the released version 2.3.2 the duration is completely contained within the box.

[...]

PR #3168, opened 2021-05-13
Fix duration text overflow in check ins screen. (EXPOSUREAPP-7199)

SamuraiKek:
Text used to stretch out on a second row when font size was set to max in Android settings.


dsarkar:
Fixes https://github.com/corona-warn-app/cwa-app-android/issues/3086

MikeMcC399:
I successfully verified the fix using the steps from #3086.

d4rken (approved):
lgtm

mtwalli (approved):
Tested on Samsung S9

PR #3171, opened 2021-05-14
Vaccination details (EXPOSUREAPP-7242)

mtwalli:
- Implement new vaccination details 
- Update screenshots test


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/details/VaccinationDetailsViewModel.kt
```diff
@@ -1,62 +1,74 @@
...
-        events.postValue(VaccinationDetailsNavigation.Back)
-    }
+    fun openFullScreen() = qrCodeText?.let { events.postValue(VaccinationDetailsNavigation.FullQrCode(it)) }
 
-    private fun findVaccinationDetails(vaccinatedPersons: Set<VaccinatedPerson>): VaccinationDetails {
-        val vaccinatedPerson = vaccinatedPersons.find { vaccinatedPerson ->
-            vaccinatedPerson
-                .vaccinationCertificates
-                .any { it.certificateId == vaccinationCertificateId }
+    private fun findVaccinationDetails(
```
LukasLechnerDev, line 46: I know that this was already there, but doesn't belong logic like this to the Repository? Shouldn't we have a function in the Repository like `fun getCertificate(certificateId) `?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/details/VaccinationDetailsFragment.kt
```diff
@@ -100,6 +99,17 @@ class VaccinationDetailsFragment : Fragment(R.layout.fragment_vaccination_detail
...
         )
+        // QrCode details
+        qrCodeCard.title.text = getString(
+            R.string.vaccination_qr_code_card_title,
+            certificate.doseNumber,
+            certificate.totalSeriesOfDoses
+        )
+        qrCodeCard.subtitle.text = getString(
+            R.string.vaccination_qr_code_card_subtitle,
+            certificate.vaccinatedAt.toDayFormat(),
```
LukasLechnerDev, line 110: Formatting should be `dd.MM.yy` instead of `dd.MM.yyyy`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/details/VaccinationDetailsViewModel.kt
```diff
@@ -1,62 +1,74 @@
...
         }
 
+        val certificate = person?.vaccinationCertificates?.find { it.certificateId == vaccinationCertificateId }
         return VaccinationDetails(
-            certificate = vaccinatedPerson?.vaccinationCertificates?.find {
-                it.certificateId == vaccinationCertificateId
-            },
-            isComplete = vaccinatedPerson?.getVaccinationStatus() == VaccinatedPerson.Status.COMPLETE
+            certificate = certificate,
+            isComplete = person?.getVaccinationStatus() == VaccinatedPerson.Status.COMPLETE,
```
LukasLechnerDev, line 56: I think this needs to be changed to `person?.getVaccinationStatus() == VaccinatedPerson.Status.IMMUNITY` so that the background is only blue after 14 days of the final vaccination have passed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/details/VaccinationDetailsViewModel.kt
mtwalli: it would make sense to move it to the repository, if it is commonly used. I think it is only logic that is related to this screen, or do you need that logic too?

NataliaLemmerth (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/details/VaccinationDetailsFragment.kt
```diff
@@ -85,7 +85,7 @@ class VaccinationDetailsFragment : Fragment(R.layout.fragment_vaccination_detail
         name.text = certificate.run { "$firstName $lastName" }
         birthDate.text = getString(
             R.string.vaccination_details_birth_date,
-            certificate.dateOfBirth.toString(format)
+            certificate.dateOfBirth.toDayFormat()
         )
         vaccinatedAt.text = certificate.vaccinatedAt.toString(format)
```
LukasLechnerDev, line 90: I think here we also need `.toDayFormat`. Figma shows 4 digits for the year.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/details/VaccinationDetailsViewModel.kt
LukasLechnerDev: I don't need this logic. Ok then let's leave it as it is.

LukasLechnerDev (approved):
LGTM 🚀

chiljamgossow (approved):
LGTM

NataliaLemmerth (approved):
UA approved

PR #3194, opened 2021-05-17
Vaccination registration screenshots (EXPOSUREAPP-6741)

jurajkusnier:
### Screenshots update

screenshots


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/vaccination/VaccinationConsentFragmentTest.kt
```diff
@@ -0,0 +1,73 @@
...
+    @MockK lateinit var vaccinationSettings: VaccinationSettings
+
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this, relaxed = true)
+
+        setupMockViewModel(
+            object : VaccinationConsentViewModel.Factory {
+                override fun create(): VaccinationConsentViewModel =
+                    VaccinationConsentViewModel(
```
mtwalli, line 43: You can mock the ViewModel itself, this way was required for fastlane

Corona-Warn-App/src/main/res/layout/vaccination_consent_fragment.xml
```diff
@@ -41,7 +41,7 @@
                 app:layout_constraintEnd_toEndOf="parent"
                 app:layout_constraintStart_toStartOf="parent"
                 app:layout_constraintTop_toTopOf="parent"
-                app:srcCompat="@drawable/ic_vaccination_consent" />
+                android:src="@drawable/ic_vaccination_consent" />
```
mtwalli, line 44: We should not use src ,Please have a look at https://github.com/corona-warn-app/cwa-app-android/pull/2800 ?

jurajkusnier: hm.. this was a fix because images are not displayed in the screenshot when using `app:srcCompat`
The second solution would be to change `ImageView` to `androidx.appcompat.widget.AppCompatImageView` to make it works but that looks ugly. Or do you have any other recommendations?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/ui/vaccination/VaccinationConsentFragmentTest.kt
jurajkusnier: updated

Corona-Warn-App/src/main/res/layout/vaccination_consent_fragment.xml
mtwalli: which Android version did you test? We take screenshots on android 29 if that works fine , I would say that is acceptable. Screenshots do not have work on every Android version ,but the App should

jurajkusnier: yes, I tested it with Android API 30 before. I'm reverting the changes.

mtwalli (approved):

d4rken (approved):

PR #3197, opened 2021-05-17
Improve VaccinationRepo tests and edge-case handling (EXPOSUREAPP-6729)

d4rken:
Episode 4...

* Fix being able to add duplicate certificates
* Fix crash on removal of last certificate
* Adjust error handling to use `InvalidHealthCertificateException` and it's mapped error codes for duplicate codes, missmatch on name or DOB.
* Add unit tests for everything

Known issues:
* Homescreen cards in state `IMMUNITY` are not working right yet, /cc @mtwalli I think you are on this?
* After deleting the last certificate, the details screen kicks you back to the list screen, which shows the deleted certificate. /cc @LukasLechnerDev ?

Testing:
* Scan different vaccinations into the app, remove them etc.


mtwalli (dismissed):
Lgtm, Tested the duplication case

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -332,9 +332,15 @@ class HomeFragmentViewModel @AssistedInject constructor(
...
-                    VaccinatedPerson.Status.IMMUNITY -> {
-                        throw NotImplementedError()
-                    }
+                    // TODO wrong card, just placeholder
+                    VaccinatedPerson.Status.IMMUNITY -> CompleteVaccinationHomeCard.Item(
+                        vaccinatedPerson = vaccinatedPerson,
+                        onClickAction = {
+                            popupEvents.postValue(
+                                HomeFragmentEvents.GoToVaccinationList(vaccinatedPerson.identifier.code)
+                            )
```
chiljamgossow, line 341: This would have to be the hashed code to work

d4rken: True, was just some glue code on my end to stop the app from crashing.
The card code will be overwritten by #3199 which should be merged today too.

chiljamgossow (approved):

mtwalli (approved):

PR #3199, opened 2021-05-18
Vaccination Home Cards (EXPOSUREAPP-7231)

mtwalli:
- Adjust vaccination home cards to cover new state


d4rken (commented):
Code looks good, probably should be merged after #3197 which will cause a few conflicts.

mtwalli:
> Code looks good, probably should be merged after #3197 which will cause a few conflicts.

Conflict resolved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/homecard/VaccinationHomeCard.kt
```diff
@@ -0,0 +1,52 @@
...
+    override val onBindData: VaccinationHomeCardBinding.(
+        item: Item,
+        payloads: List<Any>
+    ) -> Unit = { item, payloads ->
+        val curItem = payloads.filterIsInstance<Item>().singleOrNull() ?: item
+
+        personName.text = curItem.vaccinatedPerson.fullName
+        when (curItem.vaccinatedPerson.getVaccinationStatus()) {
+            VaccinatedPerson.Status.COMPLETE -> {
+                val days = curItem.vaccinatedPerson.getTimeUntilImmunity()!!.standardDays.toInt()
```
chiljamgossow, line 29: Can we use a fallback here instead of crashing the app?

Corona-Warn-App/src/main/res/values/vaccination_strings.xml
```diff
@@ -74,8 +74,17 @@
     <string name="vaccination_card_status_title">Digitaler Impfnachweis</string>
     <!-- XHED: Homescreen vaccination status card vaccination name -->
     <string name="vaccination_card_status_vaccination_name">SARS-CoV-2 Impfschutz</string>
-    <!-- XTXT: Homescreen card vaccination status label -->
+    <!-- XTXT: Homescreen card incomplete vaccination status label -->
     <string name="vaccination_card_status_vaccination_incomplete">Unvollständiger Impfschutz</string>
+    <!-- XTXT: Homescreen card complete vaccination status label -->
+    <plurals name="vaccination_card_status_vaccination_complete">
+        <item quantity="one">Vollständiger Impfschutz in %1$d Tag</item>
```
chiljamgossow, line 81: Can we just say "in einem Tag" and not "in 1 Tag"? -> UA

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/homecard/VaccinationHomeCard.kt
mtwalli: `COMPLETE` state should always have a `duration`. this should not crash in case it complete ,and if it does we would know that there is something wrong

Corona-Warn-App/src/main/res/values-de/vaccination_strings.xml
```diff
@@ -69,13 +69,22 @@
...
     <!-- XHED: Homescreen vaccination status card title -->
     <string name="vaccination_card_status_title">Digitaler Impfnachweis</string>
     <!-- XHED: Homescreen vaccination status card vaccination name -->
     <string name="vaccination_card_status_vaccination_name">SARS-CoV-2 Impfschutz</string>
-    <!-- XTXT: Homescreen vaccination status card vaccination status label -->
+    <!-- XTXT: Homescreen card incomplete vaccination status label -->
     <string name="vaccination_card_status_vaccination_incomplete">Unvollständiger Impfschutz</string>
+    <!-- XTXT: Homescreen card complete vaccination status label -->
+    <plurals name="vaccination_card_status_vaccination_complete" tools:ignore="UnusedQuantity">
+        <item quantity="one">Vollständiger Impfschutz in %1$d Tag</item>
```
chiljamgossow, line 80: Can we change to "in einem Tag" instead of "in 1 Tag" -> UA

mtwalli: Should be aligned with https://github.com/corona-warn-app/cwa-app-android/pull/3185/files#diff-fcc1b466984b0efa9d6f2acb9a2d2f9a16b4251afd27e101a23b7972fd6c3728R48 also

chiljamgossow (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/ui/homecard/VaccinationHomeCard.kt
LukasLechnerDev: Best solution here would be IMO to use a sealed class hierarchy for the Status instead of an enum and then have a `getTimeUntilImmunity()` for the `Immunity` data class. Would require some refactoring though...

Corona-Warn-App/src/main/res/values-de/vaccination_strings.xml
LukasLechnerDev: FYI: In the risk card, we also show the digit for eins => "Begegnungen mit niedrigem Risiko an %1$d Tag""

LukasLechnerDev (approved):
LGTM

Corona-Warn-App/src/main/res/values-de/vaccination_strings.xml
harambasicluka: @chiljamgossow you are right that would be correct german, but I assume it was decided like this on purpose like it i on the risk card as @LukasLechnerDev mentioned. The number should stand out so thats easier to "scan". But let's wait for our UA colleagues :)

ChristinaTsigganos: Thanks for the feedback, @chiljamgossow , but we'll go with the digit for eins :)

ChristinaTsigganos (approved):
UA approved

PR #3200, opened 2021-05-18
Incognito mode for all edit text fields (EXPOSUREAPP-6596)

chiljamgossow:
Please test add person, add location and the profile and check the keyboard behavior

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/contact_diary_add_location_fragment.xml
+++ b/Corona-Warn-App/src/main/res/layout/contact_diary_add_location_fragment.xml
@@ -4,7 +4,6 @@
     android:layout_height="match_parent"
     android:background="@color/colorBackground"
     android:fillViewport="true"
-    android:fitsSystemWindows="true"
     android:transitionName="contact_diary_shared_content">
 
     <androidx.constraintlayout.widget.ConstraintLayout
@@ -66,7 +65,7 @@
                 android:id="@+id/location_name_input_edit"
                 android:layout_width="match_parent"
                 android:layout_height="wrap_content"
-                android:imeOptions="actionNext"
+                android:imeOptions="actionNext|flagNoPersonalizedLearning"
                 android:inputType="textCapWords"
                 android:focusable="true"
                 android:maxLength="250" />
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/res/layout/contact_diary_add_location_fragment.xml
```diff
@@ -4,7 +4,6 @@
     android:layout_height="match_parent"
     android:background="@color/colorBackground"
     android:fillViewport="true"
-    android:fitsSystemWindows="true"
```
chiljamgossow, line 7: This was added by Ralph with ticket 5869. However, it broke the screen on Nokia device., I I removed it. Needs to be tested on Samsung.

AlexanderAlferov (approved):
Tested on Samsung S8. Was not able to reproduce previous issue (EXPOSUREAPP-5869). LGTM 🚀 

screenshot-1621330592112

Corona-Warn-App/src/main/res/layout/contact_diary_add_location_fragment.xml
mtwalli: Checked that on Samsung S9 and looks fine, I find two other screens where is flag is used also 
1- fragment_edit_check_in
2- fragment_confirm_check_in
and in MainActivity layout and this is intended and should actually apply for all screens hosted by it 

Are these two screens working fine on your device?

chiljamgossow: It was related to the keyboard, which does not come up on the two screen you mentioned. Thanks for checking though.

Corona-Warn-App/src/main/res/layout/rat_profile_create_fragment.xml
```diff
@@ -155,8 +155,8 @@
                     android:layout_width="match_parent"
                     android:layout_height="wrap_content"
                     android:focusable="true"
-                    android:imeOptions="actionNext"
-                    android:inputType="text" />
+                    android:imeOptions="actionNext|flagNoPersonalizedLearning"
+                    android:inputType="number" />
```
chiljamgossow, line 159: I changed this to numbers @mtwalli Would revert if that was intended to allow letters as well

mtwalli (approved):

Issue #3183, opened 2021-05-16
Timestamp of shared error report uses UTC


MikeMcC399:
## Avoid duplicates

* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] This report is for Android. iOS is not affected.
* [X] Bug is not already reported in another issue
Related to issue "CWA log filename contains characters reserved under Windows" https://github.com/corona-warn-app/cwa-app-android/issues/3158.

## Technical details

- Device name: Google Pixel 3a emulator
- Android version: 11
- App version: 2.2.1

## Describe the bug

The default filename used in "SHARE AND CONTINUE" of error reporting includes a UTC timestamp.

This is confusing because the name of a submitted error report uses the local time, which in Germany is never the same as UTC.

## Steps to reproduce the issue

1. In CWA Start Screen, tap three-dot icon
2. Tap "App Information", tap "Error Reports"
3. Tap "START"
4. Wait until "Recording Active" is shown
5. Tap "SHARE AND CONTINUE", note the name of the zip file
6. Select "Drive", tap "Save"
7. Tap "SEND ERROR REPORT"
8. Tap "AGREE AND SEND"
9. Tap ID History, note the time shown

---
Share filename shows UTC (currently CEST - 2 hours)
![image](https://user-images.githubusercontent.com/66998419/118394006-68f64980-b642-11eb-84b5-1709d5596970.png)

---
Submitted timestamp shows local time
![image](https://user-images.githubusercontent.com/66998419/118394112-fc2f7f00-b642-11eb-8179-b8ae83d7cf3a.png)

## Expected behavior

The filename for "SHARE AND CONTINUE" should be formed using the local time, not UTC time.

## Possible Fix

Convert UTC time to local time before forming the filename to be used in "SHARE AND CONTINUE".

----
Internal Tracking-ID: [EXPOSUREAPP-7298](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7298)

Ein-Tim:
There is no such timestamp in the filename under iOS, it's just called `cwa-log`.

heinezen:
@MikeMcC399 Thank you. We have reported it as [EXPOSUREAPP-7298](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7298)

----
Corona-Warn-App Open Source Team

dsarkar:
[...]

Issue #3158, opened 2021-05-12
CWA log filename contains characters reserved under Windows


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Pixel 3a emulator
- Android version: 11
- App version: 2.2.1 (and 2.3.0-RC0)

## Describe the bug

Attempting to unzip an error report from CWA using the Microsoft Windows built-in Extract command results in
`Error 0x80070057: The parameter is incorrect.`

## Steps to reproduce the issue

1. On the CWA Start Screen tap the three-dot symbol
2. Tap "App Information"
3. Tap "Error Reports"
4. Tap "SHARE AND CONTINUE"
5. Tap (Google) Drive - Save to Drive
6. Document title will be similar to "CWA Log 2021-05-12 14:07:40.331.zip"
7. Tap "Save"
8. On Microsoft Windows 10 system open https://drive.google.com/drive/my-drive
9. Right click and select Download
10. Downloaded filename is changed, for instance to "CWA Log 2021-05-12 14_07_40.331.zip"
11. Right-click downloaded file and select "Extract All..."
12. Click Extract
13. Note error message "Error 0x80070057: The parameter is incorrect."

![image](https://user-images.githubusercontent.com/66998419/117990014-0dc01080-b33d-11eb-81d6-5a9ade88c13d.png)

The error is caused by the zipped filename containing colons ":" which is a reserved character for Microsoft Windows filenames. See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions.

## Expected behavior

It should be possible to view the contents of a CWA Error Report using the built-in Microsoft Windows zip Extract command. Filenaming should be consistent with Microsoft Windows rules and not cause errors.

## Possible Fix

Replace ":" in the filename by "_".

The line of code seems to be https://github.com/corona-warn-app/cwa-app-android/blob/595ab630dfad5ac1e29021c084c400054193a498/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/internal/LogSnapshotter.kt#L36

## Workaround

Use [7-zip](https://www.7-zip.org/) to unzip the file.

[...]

PR #3202, opened 2021-05-18
FIx: Use local timestamp for error report filename

PhilippNowak96:
### Description

As described in #3183 UTC timestamp is used for the filename of the error report. This PR changes it to the users local timestamp.

---
Internal Tracking-ID: [EXPOSUREAPP-7298](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7298)


mtwalli (approved):
Lgtm

d4rken (approved):
I think this is more a feature, as the logging uses UTC timestamps and the filename matches the the timestamps of the last loglines available in that file, but I'm okay with changing it to align with the upload timestamp.

vaubaehn:
Ay, it's already merged... Otherwise I proposed to also take https://github.com/corona-warn-app/cwa-app-android/issues/3158 into account, while you're already having a hand on it...

PR #3204, opened 2021-05-18
Fix Datenabruf dialog not dismissed when completing submission (EXPOSUREAPP-7105)

SamuraiKek:
Making sure to dismiss the dialog before traveling to the next screen.
I could not reproduce the reported bug, but this code addition shouldn't change the behaviour.

Try and go through the submission process and make sure everything is working as expected.


AlexanderAlferov (approved):
Was not able to reproduce the error.
Submission flow works like intended.

chiljamgossow (approved):

PR #3210, opened 2021-05-18
Exclude resources from shrinking (DEV)

mtwalli:


chris-cwa (approved):

d4rken (approved):
Caused by newer R8 version due to upgraded buildtools:

https://issuetracker.google.com/issues/170709331

This workaround is the best solution for now.

PR #3219, opened 2021-05-19
Align Margins on Register Test Screen (EXPOSUREAPP-7348)

axelherbstreith:
Aligned the cards spacing between the cards to match the figma design


Corona-Warn-App/src/main/res/layout/fragment_submission_dispatcher.xml
```diff
@@ -107,7 +107,7 @@
                     layout="@layout/include_dispatcher_card"
                     android:layout_width="0dp"
                     android:layout_height="wrap_content"
-                    android:layout_marginTop="@dimen/spacing_normal"
+                    android:layout_marginTop="16dp"
```
mtwalli, line 110: In the design there should not be any margins for in the images it is aligned to bottom right corner
Screenshot 2021-05-19 at 13 01 06

axelherbstreith: is now fixed

mtwalli (dismissed):
Tested , Looks good

harambasicluka (changes requested):
image
the illustrations on the card are overlapping, therefore the border radius is gone O.o

mtwalli:
> image
> the illustrations on the card are overlapping, therefore the border radius is gone O.o

Good catch!, @axelherbstreith  I think we can make the parent layout of the items a `MaterialCard` and it will crop the content  with the rounded corners

axelherbstreith:
@mtwalli I fixed it now with changing the `ImageView` to `ShapeableImageView` and add a corner radius to the bottom right that fits the corder or the parent

mtwalli (approved):

harambasicluka (approved):

PR #3227, opened 2021-05-19
Coroutines 1.5.0 (👨‍💻)

mtwalli:
🆕  Upgrade Coroutines to `v1.5`

🆕  Changes:
```kotlin
- ConflatedBroadcastChannel // Deprecated 
+ MutableStateFlow // Not closable -> introduced new progress values for initial state and when it is finished 
```
 🧹  Small clean up

### DIFF
--- a/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/coronatest/ui/CoronaTestTestFragmentViewModel.kt
+++ b/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/coronatest/ui/CoronaTestTestFragmentViewModel.kt
@@ -1,6 +1,5 @@
 package de.rki.coronawarnapp.test.coronatest.ui
 
-import android.content.Context
 import androidx.lifecycle.asLiveData
 import com.journeyapps.barcodescanner.BarcodeResult
 import dagger.assisted.AssistedFactory
@@ -100,7 +99,7 @@ class CoronaTestTestFragmentViewModel @AssistedInject constructor(
     data class PCRTState(
         val coronaTest: PCRCoronaTest?
     ) {
-        fun getNiceTextForHumans(context: Context): String {
+        fun getNiceTextForHumans(): String {
             return coronaTest
                 ?.toString()
                 ?.replace("PCRCoronaTest(", "")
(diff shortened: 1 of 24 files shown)


d4rken (dismissed):
:+1:

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/coronatest/ui/CoronaTestTestFragmentViewModel.kt
```diff
@@ -100,7 +99,7 @@ class CoronaTestTestFragmentViewModel @AssistedInject constructor(
     data class PCRTState(
         val coronaTest: PCRCoronaTest?
     ) {
-        fun getNiceTextForHumans(context: Context): String {
```
BMItter, line 103: 👏

BMItr (dismissed):
lgtm! thx

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/task/TaskController.kt
```diff
@@ -63,15 +63,15 @@ class TaskController @Inject constructor(
...
+     * Don't re-use task requests, create new requests for each submission.
      * They contain unique IDs.
      */
     fun submit(request: TaskRequest) {
         if (!taskFactories.containsKey(request.type.java)) {
             throw MissingTaskFactoryException(request::class)
         }
         Timber.tag(TAG).i("Task submitted: %s", request)
-        taskQueue.offer(request)
+        taskQueue.trySend(request).isSuccess
```
LukasLechnerDev, line 74: 🤔  calling `isSuccess` is not necessary here?

LukasLechnerDev: I guess the auto-migration suggested it, since `offer()` returns a boolean. However, we don't use that boolean so we don't have to call `.isSuccess`

LukasLechnerDev: `trySend().isSuccess` returns `Boolean`
`trySend()` returns `ChannelResult`

But since the return type is not relevant in this situation, I don't see a reason to call `.isSuccess`

mtwalli: Ok, only for you, I will change it 🎁

LukasLechnerDev: ❤️

d4rken (approved):
lgtm if CI passes

LukasLechnerDev (approved):
image

=> Lukas is happy 🙂

PR #3230, opened 2021-05-19
Fix debug log censoring collisions (EXPOSUREAPP-7196)

d4rken:
@kolyaopahle found an issue for some uncensored log entries. The underlying cause was that if there is a collision between different censors, i.e. that two censor modules replace elements on the same line, the original string does not match what the other censor module expects. Example:

```kotlin
message = "Hello world

after module 1:
message = "Hello"

module 2 wants to replace "Hello world", but only finds "Hello" which is no match.
```

In this PR  @kolyaopahle  and me reworked the mechanism:
* The censoring mechanism to longer run in serial on the data that should be censored. All censor modules run at the same time.
* Each censor module now produces a `CensoredString(String,IntRage)`. With `String` being the changed message, and `IntRange` the start and end index of what was modified in the original log message.
* After all censors are done, the `DebugLogger` checks if more than one censor module modified the original message. If that is the case, the minimum and maximum index of modifications in the original log message is removed and replaced with a placeholder `<censoring-collision>`. This leaves us with some information that is useful for debugging and also removes any information that should have been censored.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
```diff
@@ -1,15 +1,44 @@
...
+            if (newer == null) return this
+
+            val range = when {
+                newer.range == null -> this.range
+                this.range == null -> newer.range
+                else -> min(this.range.first, newer.range.first)..max(this.range.last, newer.range.last)
+            }
+
+            return CensoredString(string = newer.string, range = range)
+        }
```
d4rken, line 29: Operator to make chaining censoring easier by being able to use "var old += new`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
```diff
@@ -1,15 +1,44 @@
...
+        fun CensoredString.censor(orig: String, replacement: String): CensoredString? {
+            val start = this.string.indexOf(orig)
+            if (start == -1) return null
+
+            val end = start + replacement.length
+            return CensoredString(
+                string = this.string.replace(orig, replacement),
+                range = start..end
+            )
+        }
```
d4rken, line 40: Acts like `String.replace` but tells us the start+end index of the match.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -160,10 +163,28 @@ class DebugLogger(
                 launch {
                     // Censor data sources need a moment to know what to censor
                     delay(1000)
-                    val censoredLine = bugCensors.get().fold(rawLine) { prev, censor ->
-                        censor.checkLog(prev) ?: prev
+
+                    val formattedMessage = rawLine.format()
```
d4rken, line 167: As we don't need to censor logtags, we pre compute the log message and only pass that to censors.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -160,10 +163,28 @@ class DebugLogger(
...
+                    val censored: Collection<BugCensor.CensoredString> = bugCensors.get()
+                        .map {
+                            async {
+                                it.checkLog(formattedMessage)
+                            }
+                        }
+                        .awaitAll()
+                        .filterNotNull()
+                        .filter { it.range != null }
```
d4rken, line 176: @kolyaopahle and me performance tested the censoring and we were at ~100ms if we run all censors in sequence. This now runs in parallel the `DebugScope` threads, reaching 10-20ms per logline.
For all bug censors that are stateful (i.e. do local caching), thread safe has to be taken into consideration though.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -160,10 +163,28 @@ class DebugLogger(
...
+
+                    val toWrite: String = when (censored.size) {
+                        0 -> formattedMessage
+                        1 -> censored.single().string
+                        else -> {
+                            val minMin = censored.minOf { it.range!!.first }
+                            val maxMax = censored.maxOf { it.range!!.last }
+                            formattedMessage.replaceRange(minMin, maxMax, CENSOR_COLLISION_PLACERHOLDER)
```
d4rken, line 184: Part that deletes whole sections if there was a censoring collision.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/LogLine.kt
```diff
@@ -13,15 +15,33 @@ data class LogLine(
...
         val time = Instant.ofEpochMilli(timestamp)
-        return "$time  ${priorityLabel(priority)}/$tag: $message\n"
+
+        val baseLine = "$time ${priorityLabel(priority)}/$tag: $message\n"
+
+        return if (throwable != null) {
+            baseLine + "\n" + getStackTraceString(throwable)
+        } else {
+            baseLine
+        }
```
d4rken, line 25: We previously overlooked that the actual stacktrace was not written to log :frowning:

d4rken:
Merge this after #3230

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CWADebug.kt
```diff
@@ -22,7 +22,7 @@ object CWADebug {
         if (isDebugBuildOrMode) System.setProperty("kotlinx.coroutines.debug", "on")
 
         if (isDeviceForTestersBuild) {
-            Timber.plant(Timber.DebugTree())
+//            Timber.plant(Timber.DebugTree())
```
LukasLechnerDev, line 25: Not intended I guess :)

d4rken: :fearful:  thanks

harambasicluka:
> Merge this after #3230

Recursion!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/internal/DebugLoggerScope.kt
```diff
@@ -9,7 +9,8 @@ import javax.inject.Qualifier
 import kotlin.coroutines.CoroutineContext
 
 object DebugLoggerScope : CoroutineScope {
-    val dispatcher = Executors.newSingleThreadExecutor(
+    val dispatcher = Executors.newFixedThreadPool(
+        4,
```
LukasLechnerDev, line 13: Why 4 here? Would it make sense to use Dispatchers.Default?

LukasLechnerDev (dismissed):
LGTM. Tested it on device and got the `<censoring-collision>` in the log.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/internal/DebugLoggerScope.kt
d4rken: We tested 1 to 16 threads, 4 yielded the most performance gain with the least amount of threads.
* 1 thread ~ 110ms per logline on avg
* 2 ~ 50ms
* 3  ~ 35ms
* 4 ~ 20ms
* 5 ~ 15-20ms
* 6 ~ 11-20ms
etc.

Using `Dispatchers.Default` would not be bad, but we decided against it to reduce impact on other app operations.

`Dispatchers.Default` provides a thread count equal to the number of cores, e.g. 4, if the logging mechanisms block for some reasons, the coroutines running on `Default` through out the app could starve as fewer or even no threads are available. I don't think this would happen, but just to be sure it can't and to reduce the impact of logging on the app and to make the behavior recorded by the logger more accurate to what is happening when it is not active, it is nicer to work on different threads here.

LukasLechnerDev: Makes Sense! Thanks for the in-depth explanation 👍

d4rken:
> > Merge this after #3230
> 
> Deadlock!

:smile:  I meant #3235

LukasLechnerDev (approved):
LGTM#2. Thank you for the new unit test 👍

mtwalli (approved):

PR #3231, opened 2021-05-19
Vaccination Server Test (EXPOSUREAPP-6890)

chris-cwa:
Unit tests for vaccination server.


mtwalli (dismissed):
Lgtm

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/vaccination/core/server/valueset/VaccinationServerTest.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * contains both binary and signature
+     */
+    private val exportZip = File("src/test/resources/export.zip")
+
+    /**
+     * binary is missing
+     */
+    private val invalidExportZip = File("src/test/resources/export_invalid.zip")
```
d4rken, line 32: The names are a bit generic.
Maybe `resources/vaccination/valueset_(default|invalid).zip`?

d4rken (commented):
Just the one comment on the filenames, otherwise nice test.

BMItr (dismissed):
if you want change the filename of the export.bin (as d4ken mentioned)
otherwise lgtm!

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/vaccination/core/server/valueset/VaccinationServerTest.kt
chris-cwa: renamed it

d4rken (approved):

mtwalli (approved):

PR #3232, opened 2021-05-19
Additional hint to find details of exposures in the contact journal (EXPOSUREAPP-5829)

chris-cwa:
This adds a small hint that user finds more details in contact journal.

How to test:
- go to details page
- hint should appear in case there is a low risk with encounters or high risk

- [ ] test it myself


Corona-Warn-App/src/main/res/layout/tracing_details_find_details_in_journal.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <data>
+        <variable
+            name="state"
+            type="de.rki.coronawarnapp.tracing.states.TracingFailed" />
+    </data>
```
mtwalli, line 9: can we please avoid using DataBinding here ?

chris-cwa: certainly. i don't think we need it

LukasLechnerDev:
I think the target branch should be set to `2.4`?

Corona-Warn-App/src/main/res/layout/tracing_details_find_details_in_journal.xml
```diff
@@ -0,0 +1,23 @@
+<?xml version="1.0" encoding="utf-8"?>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <androidx.constraintlayout.widget.ConstraintLayout
```
LukasLechnerDev, line 5: Since there is only one view here, we don't need a ViewGroup in this layout, see e.g. `trace_location_organizer_category_header`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/ui/details/items/additionalinfos/FindDetailsInJournalBox.kt
```diff
@@ -0,0 +1,31 @@
...
+            true
+        )
+    }
+
+    override val onBindData: TracingDetailsFindDetailsInJournalBinding.(
+        item: Item,
+        payloads: List<Any>
+    ) -> Unit = { _, _ -> }
+
+    object Item : AdditionalInformationItem
```
LukasLechnerDev, line 30: The item needs to be added to `TracingDetailsAdapter`

NataliaLemmerth (dismissed):
UA approved

harambasicluka (changes requested):
This is a 2.4 feature. please change the target.

chris-cwa:
wrong base - i have to create a new branch

Issue #3216, opened 2021-05-18
Contact Diary input of Person or Place cut-off / hidden by keyboard


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Pixel 3a emulator
- Android version: 11
- App version:  2.3.0-RC0 & 2.3.0-RC1 (pre-release versions) 

## Describe the bug

Add PERSON and add PLACE from Contact Diary input screen cut off at bottom when keyboard displayed.

This is a regression from CWA 2.2.1 whether I compare to the released version or to a version built myself in Android Studio (which shows 2.2.1-RC1).

## Steps to reproduce the issue

1. Tap Journal in CWA Start Screen
2. Tap "OPEN JOURNAL" if using for first time
3. Tap current day's entry in list
4. Tap "+ PERSON"
5. Note that e-mail field is only partially shown, and without label as soon as keyboard is displayed
6. Tap "X" to return
7. Tap "+ PLACE"
8. Note again that e-mail field is partially obscured

![image](https://user-images.githubusercontent.com/66998419/118705658-c93ee400-b818-11eb-981d-69a4f7c0efef.png)

## Expected behavior

When the Person input view is displayed, the e-mail view should be completely visible, and the SAVE button should be visible too, even when it is greyed out.

Same for the Place input view.

## Workaround

1. Tap three-dot symbol in keyboard
2. Tap "Floating"
3. All fields and the SAVE button are visible when the Floating Keyboard is displayed

---
Internal Tracking ID: [EXPOSUREAPP-7358](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7358)

MikeMcC399:
The outcome on the following device is even worse. Tapping in the name field of a new person opens the onscreen keyboard which then takes up the whole of the device's screen without showing the name field at all. After completing the entry of name, by tapping "Next" the keyboard moves on to the Phone field without showing it at all. So the user is left blind.

The issue is same whether creating a new person / place, or editing an existing person/place.

Device: Samsung Galaxy A5 (2017)
Android version: 8.0.0
App version: 2.3.0-RC0 (pre-release version)

[...]

PR #3233, opened 2021-05-20
Remove fitsSystemWindows from fragments (EXPOSUREAPP-7366,EXPOSUREAPP-7368)

mtwalli:
- Remove fitsSystemWindows from fragments, it is causing some overlay when keyboard is open
- this is flag is set now in the main activity layout and it should apply to all hosted fragments 
- One screen should have been fixed by @chiljamgossow  before , but it seems it is lost somehow


SamuraiKek (approved):
lgtm

dsarkar:
Fixes https://github.com/corona-warn-app/cwa-app-android/issues/3216

axelherbstreith (approved):

PR #3234, opened 2021-05-20
View binding polishing (DEV)

mtwalli:
- Remove lazy postfix, no need to call it `Lazy` because it is

Now you can use either way
```kotlin
    private val binding: FragmentVaccinationListBinding by viewBinding()
    private val binding by viewBinding<FragmentVaccinationListBinding>()
```


d4rken (approved):
lgtm

SamuraiKek (approved):

PR #3235, opened 2021-05-20
Make Error report filename Windows compatible (EXPOSUREAPP-7266) (COMMUNITY)

MikeMcC399:
This PR mitigates the issue "CWA log filename contains characters reserved under Windows" #3158 by replacing the reserved colon character ":" in the CWA Error Report zip file (log snapshot) filename by underscore "_".

If an attempt was made on a Microsoft Windows 10 machine to extract the contents of an uploaded CWA Error Report zip file using the built-in Compressed Folder Tools, Extract all command, it failed with the error message "Error 0x80070057: The parameter is incorrect.". The [7-Zip utility](https://www.7-zip.org/) on the other hand was able to deal with extracting the content to the Windows file system by changing the reserved colon characters in the filename to underscore characters. The fix in this PR follows the example of 7-Zip by using the same character replacement of colon by underscore.

An example snapshot filename prior to this PR is:
`CWA Log 2021-05-20 09:55:38.095.zip`
Using this PR, the snapshot filename changes to the following, for example:
`CWA Log 2021-05-20 09_55_38.095.zip`.

### Environment

Device name: Pixel 3a emulator
Android version: 11
Build Variant: deviceForTestersDebug

### Steps to Verify

1. On the CWA Start Screen tap the three-dot symbol
2. Tap "App Information"
3. Tap "Error Reports"
4. If Error Analysis shows Inactive, then tap START
5. Tap "SHARE AND CONTINUE"
5. Tap (Google) Drive - Save to Drive
6. Document title will be similar to "CWA Log 2021-05-20 09_55_38.095.zip"
7. Tap "Save"
8. On Microsoft Windows 10 system open https://drive.google.com/drive/my-drive
9. Right click and select Download
11. Right-click downloaded file in Windows Explorer and select "Extract All..."
12. Accept default folder and click Extract
13. Note that a folder such as "CWA Log 2021-05-20 09_55_38.095" is successfully created which contains a file such as "CWA Log 2021-05-20 09_55_38.095.txt".

## Test changes

This PR modifies LogSnapshotterTest.

- Previously LogSnapshotterTest failed when run in an Android Studio Windows environment with the following error:
`java.io.IOException: The filename, directory name, or volume label syntax is incorrect`
This PR modifies the filename so that LogSnapshotterTest now runs on Windows.

- PR https://github.com/corona-warn-app/cwa-app-android/pull/3202 changed the timestamp in the snapshot to local time instead of UTC, in order to match the way the timestamp is formed on a submitted error report.  This broke the test performed by LogSnapshotterTest if it was run in an Android Studio Ubuntu environment. It did however work in the GitHub CI environment. This PR adapts the test so that it takes account of the timezone change to the filename and it now runs correctly under Ubtunu.

## Acknowledgements

Thanks to @PhilippNowak96 for tips and code snippets to complete this PR.


PhilippNowak96:
@MikeMcC399 how does your `setup()` in `LogSnapshotterTest` look like? I tried it the following way and had no mismatch:

```
@BeforeEach
fun setup() {
    MockKAnnotations.init(this)

    every { context.cacheDir } returns cacheDir

    testDir.mkdirs()
    testDir.exists() shouldBe true

    every { debugLogger.runningLog } returns runningLogFake
    every { timeStamper.nowUTC.toUserTimeZone() } returns DateTime.parse("1970-01-01T00:00:00.000Z")

    runningLogFake.parentFile!!.mkdirs()
    runningLogFake.writeText("1 Doge = 1 Doge")
}
 ```

MikeMcC399:
@PhilippNowak96 
My apologies. I did not make your changes properly!
You suggestion works. I will just cross-check again on Ubuntu and then will push the correction here.

Thanks for your help!

d4rken (approved):
lgtm, thanks for the PR!

Windows :roll_eyes:

mtwalli (approved):

PR #3245, opened 2021-05-21
Reduce config check timeout for the update checker (EXPOSUREAPP-6101)

d4rken:
By 50% to reduce app blocking on launch due to bad internet or firewalls.


mtwalli (approved):
Lgtm

BMItr (approved):

PR #3253, opened 2021-05-21
Invalid QR dialog wrong button functions (EXPOSUREAPP-7341)

SamuraiKek:
Fixes an issue where the back button dismissed the dialog but the cancel button on the dialog would navigate you back to the previous screen. 
Now pressing the back button or the cancel button on the dialog will have the same effect.

Test by scanning an invalid corona test QR code.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DialogHelper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DialogHelper.kt
@@ -23,7 +23,8 @@ object DialogHelper {
         val cancelable: Boolean? = true,
         val isTextSelectable: Boolean = false,
         val positiveButtonFunction: () -> Unit? = {},
-        val negativeButtonFunction: () -> Unit? = {}
+        val negativeButtonFunction: () -> Unit? = {},
+        val cancelFunction: () -> Unit? = {}
     ) {
         constructor(
             context: Context,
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DialogHelper.kt
```diff
@@ -92,6 +92,7 @@ object DialogHelper {
                     ) { _, _ ->
                         dialogInstance.negativeButtonFunction()
                     }
+                    setOnCancelListener { dialogInstance.negativeButtonFunction() }
```
mtwalli, line 95: This affecting every dialog that uses `DialogHelper ` , right?
Can you make that change exclusively for the dialog in `SubmissionQRCodeScanFragment `?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DialogHelper.kt
```diff
@@ -93,6 +96,9 @@ object DialogHelper {
                         dialogInstance.negativeButtonFunction()
                     }
                 }
+                if (dialogInstance.cancelFunction != null) {
+                    setOnCancelListener { dialogInstance.cancelFunction() }
+                }
```
mtwalli, line 101: IDE is suggesting to simplify that to 
```kotlin
   setOnCancelListener { dialogInstance.cancelFunction() }
```

mtwalli:
Fix is not working as expected, please double check?

SamuraiKek:
> Fix is not working as expected, please double check?

Latest changes should address this.

AlexanderAlferov (approved):
LGTM 🚀

jurajkusnier (approved):

PR #3256, opened 2021-05-21
Risk Calculation cannot be restarted after time-travel (EXPOSUREAPP-6646)

LukasLechnerDev:
Risk Calculation cannot be restarted after time-travelling & going back to the correct time if there was a successful calculation in the last 6 hours. 

why? Because `DownloadDiagnosisKeyTask` checks if there was a successful calculation recently, and aborts the risk calculation if that's the case: `"task aborted, because detection was performed recently"`

Therefore, the user won't be able to see a successful risk card in the next 6h. 

My solution now is to check in the `TracingStateProvider` if a risk calculation was recently canceled because of this, and if so, it emits the last successfully calculated screen so that the UI can show this one. 

I am not 100% happy with the solution as it makes the code even more complex. Is there a better / simpler solution?


harambasicluka:
As this isn't a new bug in `2.3.x` we'll take this to `2.4.x`

PR #3259, opened 2021-05-21
Catch debug log FileNotFoundExceptions (DEV)

d4rken:
* On low storage, the system may clean caches without asking.
* If the parent dir of the debuglog is deleted, writing will fail with an exception.
* Catch that exception, and try to recreate the path+file.
* Write an initial log line that the log was deleted, then continue.
* If all fails, log the error to logcat and just skip.


mtwalli (approved):
Lgtm

AlexanderAlferov (approved):
LGTM 🚀

PR #3261, opened 2021-05-24
Contact Diary data is written in clear text into error log on data deletion (EXPOSUREAPP-7408)

LukasLechnerDev:
### Description
We have this problem that some of our censors read the data that needs to be censored directly from the database. However, when something is logged after the respective data is already removed, the censor can't censor anymore. 

This PR now updates these censors so that they keep the data in memory. Then, even after the data is removed from the database, we can still censor correctly. 

Actually, we had some tests that covered these scenarios, however, they were incorrect (false positive). 

### How to test
Delete
- Single Persons in the Contact Diary
- Single Locations in the Contact Diary
- CheckIns
- TraceLocations
=> Then export the error logs and check if all data got censored appropriately.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/submission/RACoronaTestCensor.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/submission/RACoronaTestCensor.kt
@@ -38,7 +38,7 @@ class RACoronaTestCensor @Inject constructor(
             .launchIn(debugScope)
     }
 
-    override suspend fun checkLog(message: String): CensoredString? {
+    override suspend fun checkLog(message: String): CensoredString? = mutex.withLock {
 
         var newMessage = CensoredString(message)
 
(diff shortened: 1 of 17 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/submission/RACoronaTestCensor.kt
```diff
@@ -10,48 +10,50 @@ import de.rki.coronawarnapp.bugreporting.debuglog.internal.DebuggerScope
...
     override suspend fun checkLog(message: String): CensoredString? {
 
-        val raCoronaTestFlow = coronaTestFlow.map { tests -> tests.filterIsInstance<RACoronaTest>() }.first()
-        val raCoronaTest = raCoronaTestFlow.firstOrNull() ?: return null
-
         var newMessage = CensoredString(message)
 
-        with(raCoronaTest) {
-            withValidName(firstName) { firstName ->
+        ratCoronaTestHistory.forEach { ratCoronaTest ->
```
d4rken, line 45: Do we need a lock here?

AlexanderAlferov (dismissed):
LGTM 🚀 
Tested on Samsung S8

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/submission/RACoronaTestCensor.kt
LukasLechnerDev: Yes. Thanks for catching that!

d4rken (approved):

AlexanderAlferov (approved):

PR #3266, opened 2021-05-25
Home screen navigation (DEV)

mtwalli:
- Enhance and unify HomeScreen navigation logic.
   - Before we many `LiveData`s handling navigation logic in Home screen and with every new feature, we are adding new ones
   - Now, I combined all of this logic in one place 
- Some rearranging for logic in the view model and the fragment to have more readability


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -205,4 +129,75 @@ class HomeFragment : Fragment(R.layout.home_fragment_layout), AutoInject {
             getButton(DialogInterface.BUTTON_POSITIVE).setTextColor(context.getColorCompat(R.color.colorTextTint))
         }
     }
+
+    private fun navigate(event: HomeFragmentEvents?) {
```
d4rken, line 133: Can this `when` be exhaustive? I see `null` is not handled?

d4rken (dismissed):
lgtm, did not test all paths on device.

chris-cwa (approved):
On Pixel 3 & 4 it seems you didn't break anything.

d4rken (approved):

PR #3267, opened 2021-05-25
Vaccination QR-Code isn't equal to scanned vaccination code (EXPOSUREAPP-7416)

mtwalli:
- Parameterise Qr-Code error correction level


BMItr (dismissed):
nothing to complain 👌

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/details/QrCodeDetailViewModel.kt
```diff
@@ -86,9 +97,15 @@ class QrCodeDetailViewModel @AssistedInject constructor(
         }
     }
 
-    fun openFullScreen() {
+    fun openFullScreen() = launch {
         traceLocation?.let {
-            routeToScreen.postValue(QrCodeDetailNavigationEvents.NavigateToFullScreenQrCode(it.locationUrl))
+            val correctionLevel = appConfigProvider.getAppConfig().presenceTracing.qrCodeErrorCorrectionLevel
```
d4rken, line 102: Do we need `getAppConfig`? Would the more friendly `latestAppConfig` also suffice?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/poster/QrCodePosterViewModel.kt
```diff
@@ -80,11 +82,13 @@ class QrCodePosterViewModel @AssistedInject constructor(
         try {
             val traceLocation = traceLocation()
             val template = posterTemplateProvider.template()
+            val correctionLevel = appConfigProvider.getAppConfig().presenceTracing.qrCodeErrorCorrectionLevel
```
d4rken, line 85: Can we use `latestAppconfig` instead of `getAppConfig`?

d4rken (commented):
lgtm, see comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/details/QrCodeDetailViewModel.kt
mtwalli: i used what was there before, Maybe @axelherbstreith  can answer that?

mtwalli: There is no `latestAppConfig`

axelherbstreith: Same as @mtwalli, I also reused the implementation of the QR code which was already there. I can't find `latestAppConfig`, do you maybe mean `currentAppConfig`? @d4rken

harambasicluka:
As this bug was newly introduced in 2.3 it would be nice to get it in to 2.3, not sure why the ticket has 2.4 as fix version :/ Do you think it's possible without a lot of work?

PR #3269, opened 2021-05-25
Adjust reference time for deadman notification (EXPOSUREAPP-7386)

d4rken:
This changes the reference time for the deadman scheduler to be the key package download timestamp.
Additionally it uses the keypackage day+hour identifier which are server based to prevent timetraveling from causing issues.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationTimeCalculation.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationTimeCalculation.kt
@@ -2,6 +2,7 @@ package de.rki.coronawarnapp.deadman
 
 import dagger.Reusable
 import de.rki.coronawarnapp.diagnosiskeys.storage.KeyCacheRepository
+import de.rki.coronawarnapp.diagnosiskeys.storage.pkgDateTime
 import de.rki.coronawarnapp.util.TimeStamper
 import kotlinx.coroutines.flow.first
 import org.joda.time.DateTimeConstants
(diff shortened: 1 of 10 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/deadman/DeadmanNotificationTimeCalculation.kt
```diff
@@ -29,10 +30,15 @@ class DeadmanNotificationTimeCalculation @Inject constructor(
...
+        val lastSuccess = keyCacheRepository.allCachedKeys()
+            .first()
+            .filter { it.info.isDownloadComplete }
+            .maxByOrNull { it.info.pkgDateTime }
+            ?.info
+
+        Timber.d("Last successful diagnosis key package download: $lastSuccess")
         return if (lastSuccess != null) {
-            getHoursDiff(lastSuccess).toLong()
+            getHoursDiff(lastSuccess.pkgDateTime.toInstant()).toLong()
```
d4rken, line 41: This is the key change.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/CachedKeyInfo.kt
```diff
@@ -89,3 +84,9 @@ data class CachedKeyInfo(
         @ColumnInfo(name = "completed") val isDownloadComplete: Boolean
     )
 }
+
+val CachedKeyInfo.pkgDateTime: DateTime
+    get() = when (type) {
+        CachedKeyInfo.Type.LOCATION_DAY -> day.toDateTimeAtStartOfDay(DateTimeZone.UTC)
+        CachedKeyInfo.Type.LOCATION_HOUR -> day.toDateTime(hour, DateTimeZone.UTC)
+    }
```
d4rken, line 92: To prevent confusing mocking of different values with the same effect, this is no longer a function that is part of the data class, so if you want to have a specific DateTime in a test you have to pass a specific day and hour to the data class.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/diagnosiskeys/download/CommonSyncToolTest.kt
```diff
@@ -160,3 +141,39 @@ abstract class CommonSyncToolTest : BaseIOTest() {
...
+internal val String.loc get() = LocationCode(this)
+internal val String.day get() = LocalDate.parse(this)
+internal val String.hour get() = LocalTime.parse(this)
+
+fun createMockCachedKeyInfo(
+    dayIdentifier: LocalDate,
+    hourIdentifier: LocalTime?,
+    isComplete: Boolean,
+    location: LocationCode = "EUR".loc,
+): CachedKeyInfo {
```
d4rken, line 154: We already had a mocking function that could be reused in other tests, I've done some refactoring to do that.

AlexanderAlferov (approved):
LGTM 🚀

BMItr (approved):
lgtm

PR #3280, opened 2021-05-26
Improve log censoring mechanics (EXPOSUREAPP-7455, EXPOSUREAPP-7196, EXPOSUREAPP-7514)

LukasLechnerDev:
This PR fixes multiple bugs caused by censoring multiple elements within the same log message. This PR fixes it by changing the censoring to work with lazy evaluated replacements:

### Idea

* Each `BugCensor` now produces a `CensorContainer` which is a combination of all "censorings" it wanted to do called `CensorAction`.
* Each `CensorAction` has an `CensorAction.range: IntRange` of what it wants to censor, and a `CensorAction.execute: (String) -> String` to actually do it.
* `CensorContainer.compile` creates what we previously knew as `CensoredString`, if there is a collision, it will replace the overlapping ranges with `<censor-collision/>` if there is none, if will apply all `CensorAction.execute` calls to the string and produce the final result.
* `DebugLogger`
    * Calls `compile` directly if there is only a single `BugCensor` that produced a `CensorContainer` and the others produced `null` (nothing to censor)
    * If there are multiple container produced by `BugCensor` it will combine them all into a single container and call `compile` on that.

### Fixes

* Fixes: Crash due to IndexOutOfBounds exception. Start+end index of the censored range was from the message after it has already been censored multiple times within one censor (and the replacements caused the string to be longer than the original). ( EXPOSUREAPP-7455)
* Fixes: Log message could contain uncensored elements if the logline contains the same `toCensor` element multiple times and there is a collision. Because the replacement range for the collision text was `firstIndex+length`, but it needs to be `firstIndex+lastIndex+length` (EXPOSUREAPP-7514, EXPOSUREAPP-7196)
* Fixes: Only partially censored entries if there is an overlap within the same censoring module, i.e. "John Berlin" who lives in "Berlin Kreuzberg".
* Improves: Partial censoring, if there no overlap texts be censored without collision.
* Improves: Censoring performance, if there is a collision we don't need to do individual string replacements, and just had to do `indexOf`.
* Improves: Censoring performance, we only work on the core logline, not the logline metadata data (e.g. timestamp)

Pandoras box of censoring :roll_eyes: :dizzy_face:

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
@@ -35,7 +35,7 @@ interface BugCensor {
             val start = this.string.indexOf(orig)
             if (start == -1) return null
 
-            val end = start + orig.length
+            val end = this.string.lastIndexOf(orig) + orig.length
             return CensoredString(
                 string = this.string.replace(orig, replacement),
                 range = start..end
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
@@ -7,7 +7,7 @@ interface BugCensor {
     data class CensorContainer(
         // Original String, necessary for correct censoring ranges
         val original: String,
-        val actions: List<Action> = emptyList()
+        val actions: Set<Action> = emptySet()
     ) {
 
         fun censor(toReplace: String, replacement: String): CensorContainer {
@@ -20,7 +20,7 @@ interface BugCensor {
 
             val newAction = Action(
                 range = start..end,
-                execute = { it.replace(toReplace, replacement) }
+                modifier = Action.SimpleReplace(toReplace, replacement)
             )
             return this.copy(actions = actions.plus(newAction))
         }
@@ -31,7 +31,7 @@ interface BugCensor {
 
             val isIntersecting = ranges.any { outer ->
                 ranges.any { inner ->
-                    outer != inner &&
+                    outer !== inner &&
                         (inner.contains(outer.first) || inner.contains(outer.last)) &&
                         (inner.last != outer.first && inner.first != outer.last)
                 }
@@ -46,7 +46,9 @@ interface BugCensor {
                 )
             } else {
                 CensoredString(
-                    censored = actions.fold(original) { notOriginal, action -> action.execute(notOriginal) },
+                    censored = actions.fold(original) { notOriginal, action ->
+                        action.execute(notOriginal)
+                    },
                     ranges = ranges
                 )
             }
@@ -56,8 +58,22 @@ interface BugCensor {
 
         data class Action(
             val range: IntRange,
-            val execute: (String) -> String
-        )
+            val modifier: StringModifier,
+        ) {
+
+            fun execute(original: String) = modifier.execute(original)
+
+            data class SimpleReplace(
+                val oldValue: String,
+                val newValue: String,
+            ) : StringModifier {
+                override val execute: (String) -> String = { it.replace(oldValue, newValue) }
+            }
+
+            interface StringModifier {
+                val execute: (String) -> String
+            }
+        }
 
         companion object {
             const val COLLISION_STRING = "<censor-collision/>"
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
@@ -2,9 +2,6 @@ package de.rki.coronawarnapp.bugreporting.censors
 
 interface BugCensor {
 
-    /**
-     * If there is something to censor a new log line is returned, otherwise returns null
-     */
     suspend fun checkLog(message: String): CensorContainer?
 
     data class CensorContainer(
@@ -32,11 +29,11 @@ interface BugCensor {
             val ranges = actions.map { it.range }
             if (ranges.isEmpty()) return null
 
-            val isIntersecting = ranges.any { outter ->
+            val isIntersecting = ranges.any { outer ->
                 ranges.any { inner ->
-                    outter != inner &&
-                        (inner.contains(outter.first) || inner.contains(outter.last)) &&
-                        (inner.last != outter.first && inner.first != outter.last)
+                    outer != inner &&
+                        (inner.contains(outer.first) || inner.contains(outer.last)) &&
+                        (inner.last != outer.first && inner.first != outer.last)
                 }
             }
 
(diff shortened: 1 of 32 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
```diff
@@ -1,47 +1,85 @@
 package de.rki.coronawarnapp.bugreporting.censors
 
-import kotlin.math.max
-import kotlin.math.min
-
 interface BugCensor {
 
     /**
      * If there is something to censor a new log line is returned, otherwise returns null
```
LukasLechnerDev, line 6: Please remove or update this outdated doc here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
```diff
@@ -1,47 +1,85 @@
...
+                execute = { it.replace(toReplace, replacement) }
+            )
+            return this.copy(actions = actions.plus(newAction))
+        }
+
+        fun compile(): CensoredString? {
+            val ranges = actions.map { it.range }
+            if (ranges.isEmpty()) return null
+
+            val isIntersecting = ranges.any { outter ->
```
LukasLechnerDev, line 35: Typo `outter` => `outer`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
```diff
@@ -1,47 +1,85 @@
...
+                ranges.any { inner ->
+                    outter != inner &&
+                        (inner.contains(outter.first) || inner.contains(outter.last)) &&
+                        (inner.last != outter.first && inner.first != outter.last)
+                }
             }
 
-            return CensoredString(string = newer.string, range = range)
+            return if (isIntersecting) {
+                val minMin = ranges.minOf { it.first }.coerceAtLeast(0).coerceAtMost(original.length)
```
LukasLechnerDev, line 44: I think we don't need `coerce` here, since we calculate the range bounds of the `original` LogLine here and so we should never get an `IndexOutOfBounds` exception.  🤔

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
```diff
@@ -1,47 +1,85 @@
...
+            )
+            return this.copy(actions = actions.plus(newAction))
+        }
+
+        fun compile(): CensoredString? {
+            val ranges = actions.map { it.range }
+            if (ranges.isEmpty()) return null
+
+            val isIntersecting = ranges.any { outter ->
+                ranges.any { inner ->
```
LukasLechnerDev, line 36: This would return false if we have an exact overlap -  e.g. 2 ranges with the exact same `start` and `end` values, since we check for `outer != inner` and this is true in this case. Can that happen? Is that a problem?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -165,26 +165,37 @@ class DebugLogger(
...
 
                     val toWrite: String = when (censored.size) {
                         0 -> formattedMessage
-                        1 -> censored.single().string
-                        else -> {
-                            val minMin = censored.minOf { it.range!!.first }
-                            val maxMax = censored.maxOf { it.range!!.last }
-                            formattedMessage.replaceRange(minMin, maxMax, CENSOR_COLLISION_PLACERHOLDER)
-                        }
+                        1 -> censored.single().compile()?.censored ?: formattedMessage
```
LukasLechnerDev, line 184: `compile` can never return null here, so maybe use `!!`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
```diff
@@ -165,26 +165,37 @@ class DebugLogger(
...
-                    val censored: Collection<BugCensor.CensoredString> = bugCensors.get()
+                    val censored: Collection<BugCensor.CensorContainer> = bugCensors.get()
                         .map {
                             async {
-                                it.checkLog(formattedMessage)
+                                try {
+                                    it.checkLog(formattedMessage)
+                                } catch (e: Exception) {
+                                    Log.e(TAG, "Error in censor module $it", e)
+                                    null
```
LukasLechnerDev, line 175: Wouldn't it be safer to return some String like `<censoring-error>` here? Returning `null` here means that no censoring happened and private data could end up in the log?

Corona-Warn-App/src/test/java/android/util/Log.java
```diff
@@ -0,0 +1,39 @@
+package android.util;
+
+public class Log {
```
LukasLechnerDev, line 3: Do we want to add this class in the repo or did you forget to not commit it?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
LukasLechnerDev: Maybe use `!==` instead 🤔

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
d4rken: Only if all censors return non-empty containers though, the "non-null" is not enforced by the `DebugLogger` but by other classes, to prevent crashes due to developer mistakes, I'd prefer to keep it null-safe here.

Corona-Warn-App/src/test/java/android/util/Log.java
d4rken: I wanted to keep for testing as it allows unit tests to pass if their use Log.d, seems nice :grin:.

BMItr (dismissed):
code looks good to me.
I will reapprove if you decide to make adjustments regarding Lukas' comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/BugCensor.kt
d4rken: I've changed it to `!==` and implement equals/hashcode for replacement actions, duplicate replacements should now also be filtered out by storing replacement actions in  a `Set` automatically.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/DebugLogger.kt
d4rken: I agree, will change.

BMItr (approved):

d4rken (approved):

PR #3287, opened 2021-05-26
Rapid Antigen Tests: Retrieve Test Result Timestamp from Verification Server (EXPOSUREAPP-6865)

kolyaopahle:
### Description
This pr adds fetching of the sampleCollectedAt parameter to the ra test backend and (if present) bases outdated check on this value

### Steps to reproduce
1. Scan a Ra test
2. Check if it provided a sampleCollectedAt value
3. check if outdated state is based on the right value

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationApiV1.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationApiV1.kt
@@ -31,7 +31,7 @@ interface VerificationApiV1 {
 
     data class TestResultResponse(
         @SerializedName("testResult") val testResult: Int,
-        @SerializedName("sc") val sc: Int?
+        @SerializedName("sc") val sampleCollectedAt: Int?
     )
 
     @POST("version/v1/testresult")
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationApiV1.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationApiV1.kt
@@ -30,7 +30,8 @@ interface VerificationApiV1 {
     )
 
     data class TestResultResponse(
-        @SerializedName("testResult") val testResult: Int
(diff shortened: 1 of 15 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationApiV1.kt
```diff
@@ -30,7 +30,8 @@ interface VerificationApiV1 {
     )
 
     data class TestResultResponse(
-        @SerializedName("testResult") val testResult: Int
+        @SerializedName("testResult") val testResult: Int,
+        @SerializedName("sc") val sc: Int?
```
SamuraiKek, line 34: Could you name it explicitly?

kolyaopahle: done :)

harambasicluka:
Should be testable with INT.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/server/VerificationServer.kt
```diff
@@ -51,7 +52,7 @@ class VerificationServer @Inject constructor(
 
     suspend fun pollTestResult(
         token: RegistrationToken
-    ): CoronaTestResult = withContext(Dispatchers.IO) {
+    ): Pair<CoronaTestResult, Instant?> = withContext(Dispatchers.IO) {
```
mtwalli, line 55: Cn you create a data class for that ?

mtwalli: Maybe 
```kotlin
data CoronaTestResultResponse(val testResult:Result, val timesamp:Int?)
```

So when it is used down below, it is clear which is which  instead of `first` and `second `

NataliaLemmerth (commented):
UA approved, no UI texts

mtwalli (approved):
Lgtm 👍🏼

BMItr (approved):
💚

PR #3301, opened 2021-05-27
UA typo fix (DEV)

ChristinaTsigganos:
changed "abrufen" to "abzurufen"

### DIFF
--- a/Corona-Warn-App/src/main/res/values-de/antigen_strings.xml
+++ b/Corona-Warn-App/src/main/res/values-de/antigen_strings.xml
@@ -7,7 +7,7 @@
     <!-- XHED: Register test homescreen card: title -->
     <string name="ag_homescreen_card_test_register_title">"Test registrieren"</string>
     <!-- XTXT: Register test homescreen card: body -->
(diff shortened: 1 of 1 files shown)


DerVogel2020:
In my opinion,  it should also be "und andere zu warnen".
Or both like before: You **can** fetch your test result and **can** warn others.

mtwalli (approved):

d4rken (approved):

d4rken:
> In my opinion, it should also be "und andere zu warnen".
> Or both like before: You can fetch your test result and can warn others.

What do you think @ChristinaTsigganos ?

NataliaLemmerth (approved):
UA approved

ChristinaTsigganos:
> > In my opinion, it should also be "und andere zu warnen".
> > Or both like before: You can fetch your test result and can warn others.
> 
> What do you think @ChristinaTsigganos ?

Thank you very much for your contribution, @DerVogel2020! We've adapted the text to make it easier to read.

SamuraiKek (approved):

PR #3308, opened 2021-05-27
Fix RAT profile continue button padding (COMMUNITY)

fynngodau:
When pressing Continue on the RAT profile screen, one immediately notices that the Agree button on the next screen is not at the same height as the Continue button that one has pressed.

https://user-images.githubusercontent.com/16943720/119865534-0cd1d600-bf1c-11eb-8661-9aff7a059362.mp4

Therefore, this PR adjusts the position of the Continue button to be equivalent to the position of the Agree button in `fragment_submission_consent`.

Before | After
---|---
Screenshot_1622134102 | Screenshot_1622134204

(I'm aware that the second screenshot contains CCTG branding, though this is of course not changed by this PR ;)


d4rken (approved):
lgtm, ty Fynn!

mtwalli (approved):

PR #3309, opened 2021-05-27
Extend corona test data structures with digital covid certificate related properties (EXPOSUREAPP-7487)

d4rken:
* Extend corona test data structures with digital covid certificate related properties 
* Some additional wiring, plumbing and tests for future work.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/pcr/PCRProcessor.kt
```diff
@@ -202,13 +202,22 @@ class PCRProcessor @Inject constructor(
...
-    override suspend fun updateConsent(test: CoronaTest, consented: Boolean): CoronaTest {
-        Timber.tag(TAG).v("updateConsent(test=%s, consented=%b)", test, consented)
+    override suspend fun updateSubmissionConsent(test: CoronaTest, consented: Boolean): CoronaTest {
+        Timber.tag(TAG).v("updateSubmissionConsent(test=%s, consented=%b)", test, consented)
         test as PCRCoronaTest
 
         return test.copy(isAdvancedConsentGiven = consented)
     }
 
+    override suspend fun updateDccConsent(test: CoronaTest, consented: Boolean): CoronaTest {
```
mtwalli, line 212: should the birthdate also passed for PCR test? or how it will be used ?

d4rken: Yes, I've adjusted the API.

mtwalli (approved):
Lgtm, thanks!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTest.kt
```diff
@@ -38,6 +38,15 @@ interface CoronaTest {
 
     val isResultAvailableNotificationSent: Boolean
 
+    // Is the digital green certificate supported by the point of care that issued the test
```
chiljamgossow, line 41: Are we now using DCC or DGC?

chiljamgossow (approved):
All good

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTest.kt
d4rken: We should use DCC as it is the official EU name, I will refactor in another PR. If you find it anywhere else, feel free to change it too :smile:

PR #3318, opened 2021-05-28
Fix images not shown in screenshots (DEV)

LukasLechnerDev:
Some images in `ImageView`s don't show up in our screenshot tests. 
We could fix this by replacing these `ImageView`s with `AppCompatImageView`s. 
However, there could be other issues as well, so I decided against this solution and instead now launch the fragments in our screenshot tests in an `AppCompatActivity`. 

The issue described was related to not using `AppCompatActivity`, because we used `FragmentScenario` which uses the non-`AppCompatActivity` called `EmptyFragmentActivity`. Unfortunately, its not possible to use custom `Activities` with `FragmentScenario` => https://issuetracker.google.com/issues/121347222

That's why we now use `ActivityScenario` and provide our own Empty `AppCompatActivity`.
Now all images show up. 

### TODO before merging

- [x] Revert https://github.com/corona-warn-app/cwa-app-android/pull/3318/commits/789d4dcd64e53da7742cd44f73adf6caf1a92420 (Run `firebase_screenshots` in CI pipeline)

### DIFF
--- a/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
+++ b/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
@@ -23,7 +23,7 @@ import testhelpers.Screenshot
 import testhelpers.SystemUIDemoModeRule
 import testhelpers.TestDispatcherProvider
 import testhelpers.launchFragment2
-import testhelpers.launchInEmptyActivity
+import testhelpers.launchFragmentInContainer2
 import testhelpers.takeScreenshot
 import tools.fastlane.screengrab.locale.LocaleTestRule
 
@@ -71,7 +71,7 @@ class DebugLogFragmentTest : BaseUITest() {
     @Screenshot
     @Test
     fun capture_inactive_screenshot() {
-        launchInEmptyActivity<DebugLogFragment>()
+        launchFragmentInContainer2<DebugLogFragment>()
         takeScreenshot<DebugLogFragment>()
     }
 
@@ -83,7 +83,7 @@ class DebugLogFragmentTest : BaseUITest() {
                 override fun create(): DebugLogViewModel = activeViewModel
             }
         )
-        launchInEmptyActivity<DebugLogFragment>()
+        launchFragmentInContainer2<DebugLogFragment>()
         takeScreenshot<DebugLogFragment>()
     }
 
--- a/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
+++ b/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
@@ -21,7 +21,7 @@ import testhelpers.BaseUITest
 import testhelpers.Screenshot
 import testhelpers.TestDispatcherProvider
 import testhelpers.launchFragment2
-import testhelpers.launchFragmentInContainer2
+import testhelpers.launchInEmptyActivity
 import testhelpers.takeScreenshot
 
 @RunWith(AndroidJUnit4::class)
@@ -61,7 +61,7 @@ class DebugLogFragmentTest : BaseUITest() {
     @Screenshot
     @Test
     fun capture_inactive_screenshot() {
-        launchFragmentInContainer2<DebugLogFragment>()
+        launchInEmptyActivity<DebugLogFragment>()
         takeScreenshot<DebugLogFragment>()
     }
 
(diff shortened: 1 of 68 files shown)


ChristinaTsigganos (dismissed):
UA approved, no text changes to review

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
```diff
@@ -71,7 +71,7 @@ class DebugLogFragmentTest : BaseUITest() {
     @Screenshot
     @Test
     fun capture_inactive_screenshot() {
-        launchFragmentInContainer2<DebugLogFragment>()
+        launchInEmptyActivity<DebugLogFragment>()
```
mtwalli, line 74: How about making `launchFragmentInContainer2` launch in our EmptyActivity by default ?

mtwalli:
@LukasLechnerDev Please run the firebase_screenshots job in CI to make sure these changes did not break anything?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
LukasLechnerDev: The documentation of `launchInMainActivity()` says

```
Note: In cases where screenshots are not required, it is better to use testing framework launcher.
This helper is convenient for screenshots when [BottomNavigationView] should be visible in the screenshot.
```

`launchFragmentInContainer2()` is the "testing framework launcher", right? 
So I thought we need: 
- `launchInEmptyActivity()` for Screenshot Tests
- `launchFragmentInContainer2()` for pure Instrumentation tests

Is that correct?

LukasLechnerDev:
@mtwalli Screenshots are now here: https://126387-268027139-gh.circle-artifacts.com/0/zips/CWA-Android-Screenshots-789d4dcd64e53da7742cd44f73adf6caf1a92420.zip. Images are now all visible as far as I can tell, but the weird "grey pressed state" issue is still there.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
mtwalli: `launchFragmentInContainer2`  is our version of `launchFragmentInContainer` -from testing API- 
what is `launchFragmentInContainer2` is  currently doing is passing our `AppTheme` . 
and it used actually for screenshots tests. 
I suggest that we change the implementation of `launchFragmentInContainer2` to call  internally `launchInMainActivity` and then tests using don’t have to know which activity is used , by doing that when the root issue is fixed in testing library we can remove the workaround and don't have to change it again everywhere , what do you think ?

mtwalli:
> @mtwalli Screenshots are now here: https://126387-268027139-gh.circle-artifacts.com/0/zips/CWA-Android-Screenshots-789d4dcd64e53da7742cd44f73adf6caf1a92420.zip. Images are now all visible as far as I can tell, but the weird "grey pressed state" issue is still there.

Thanks!, Did you figure out why the pressed state are appearing in screens ?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
mtwalli: Could you please also revert the changes that was made before in some layouts to use `AppCompatImageView` instead of `ImageView`. it should unnecessary now ?

LukasLechnerDev:
> > @mtwalli Screenshots are now here: https://126387-268027139-gh.circle-artifacts.com/0/zips/CWA-Android-Screenshots-789d4dcd64e53da7742cd44f73adf6caf1a92420.zip. Images are now all visible as far as I can tell, but the weird "grey pressed state" issue is still there.
> 
> Thanks!, Did you figure out why the pressed state are appearing in screens ?

No, unfortunately not. It doesn't happen locally on my devices, so its very hard to fix for me. I guess a potential solution would be to tweak the device on which the screenshots are taken in the CI pipeline (maybe try other Android Versions or a different device altogether, etc. )

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/bugreporting/DebugLogFragmentTest.kt
LukasLechnerDev: Changed all `AppCompatImageView`s to `ImageView`.

mtwalli (dismissed):

d4rken (approved):
lgtm

BMItr (approved):

Issue #1483, opened 2020-10-26
F-Droid release and reproducible builds


rugk:
## Avoid duplicates
* [x] This enhancement request has not already been raised before (**it has, but the situation has changed, see below**)
* [x] Enhancement request is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)
* [x] If you are proposing a new feature, please do so in [CWA-Wishlist](https://github.com/corona-warn-app/cwa-wishlist)

## Current Implementation

App can only be downloaded from the Google Play Store.

## Suggested Enhancement

Either:
* publish the app on the official F-Droid store (and/or)
* self-host a F-Droid repo and publish it there (like e.g. [Tagessschau did](https://service.tagesschau.de/app/repo/))

Then:
* Do built that (your APK) reproducibly.
* Optionally, [attach the APK](https://help.github.com/articles/creating-releases/) to your [GitHub releases](https://help.github.com/articles/about-releases/). (And/Or upload them elsewhere.)

## What has changed?

I saw https://github.com/corona-warn-app/cwa-app-android/issues/477 and https://github.com/corona-warn-app/cwa-documentation/issues/5, however, these were asked months ago.
The main reason for declining it [was that the api is from Google and requires Google Play Services anyway](https://github.com/corona-warn-app/cwa-documentation/issues/5#issuecomment-627848335). However, this is not the case anymore.
**Since September 2020, the Exposure Notification API can be used without Google Play Services on Android**, if you use microG, which is a 100% FLOSS replacement for these Google Play Services. The reason is **[microG added support for the API in v0.2.12.203315](https://github.com/microg/android_packages_apps_GmsCore/releases/tag/v0.2.12.203315)**.

## Expected Benefits

[F-Droid](https://f-droid.org/) is an Android app store specifically for free/libre open-source apps. It would be great if your app could be released there, as it is the number one for getting FLOSS Android apps for many people.
F-Droid also builds all apps from source ([optionally even reproducible](https://f-droid.org/en/docs/Reproducible_Builds/)), so downloads from there can be trusted.
This possibility then allows people to update the app both from Google Play and F-Droid as the signature is the same.

The [app developer FAQ](https://f-droid.org/en/docs/FAQ_-_App_Developers/) or [the quick start guide](https://f-droid.org/en/docs/Submitting_to_F-Droid_Quick_Start_Guide/) may help you to get started.
[...]

PR #3325, opened 2021-05-30
Remove top-level `action` elements from nav graph (COMMUNITY)

fynngodau:
We have discovered an issue that prevents the app from building reproducibly if there are top-level `action` elements present in the nav graph.

We noticed that the `navigation.safeargs.kotlin` library *sometimes* generates `action*` functions in classes where they are not needed, but *sometimes* not, for `action`s that are independent of a `fragment`. This made builds of our app non-reproducible.

We have reported the reproducibility bug to Google (https://issuetracker.google.com/issues/189498001). In the meantime, as a workaround, it appears to be necessary not to use top-level `action` elements.

You can find our issue with more details here: https://codeberg.org/corona-contact-tracing-germany/cwa-android/issues/135

Reproducible builds are also desired by CWA users (#1483, https://github.com/corona-warn-app/cwa-wishlist/issues/275). With this change, it should theoretically be possible again to build CWA in a reproducible way.

(However, for users to be able to actually reproduce CWA builds, further documentation would be required – e.g. we are building the app in a standardized docker container which uses `fdroidserver` tools in an attempt to avoid system-specific oddities in the build, and are explaining how to do that in https://codeberg.org/corona-contact-tracing-germany/cwa-android/src/branch/main/docs/rebuilding.md.)

### DIFF
--- a/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/greencertificate/GreenCertificateTestFragment.kt
+++ b/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/greencertificate/GreenCertificateTestFragment.kt
@@ -4,7 +4,6 @@ import android.annotation.SuppressLint
 import android.os.Bundle
 import android.view.View
 import androidx.fragment.app.Fragment
-import de.rki.coronawarnapp.NavGraphDirections
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.coronatest.type.CoronaTest.Type.PCR
 import de.rki.coronawarnapp.coronatest.type.CoronaTest.Type.RAPID_ANTIGEN
@@ -29,10 +28,14 @@ class GreenCertificateTestFragment : Fragment(R.layout.fragment_test_green_certi
 
         binding.apply {
             pcrScreen.setOnClickListener {
-                doNavigate(NavGraphDirections.actionSubmissionTestResultGreenCertificateFragment(PCR))
+                doNavigate(GreenCertificateTestFragmentDirections
+                    .actionGreenCertificateTestFragmentToRequestGreenCertificateFragment(PCR)
+                )
             }
             ratScreen.setOnClickListener {
-                doNavigate(NavGraphDirections.actionSubmissionTestResultGreenCertificateFragment(RAPID_ANTIGEN))
+                doNavigate(GreenCertificateTestFragmentDirections
+                    .actionGreenCertificateTestFragmentToRequestGreenCertificateFragment(RAPID_ANTIGEN)
+                )
             }
         }
     }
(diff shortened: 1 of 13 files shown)


mtwalli:
@fynngodau , in our app we v2.2.2 for safe args, in the issue filed to Google you mentioned v2.3.5, can you please explain this to me? 
Despite that, v2.2.2 and v2.3.5 seem to be consistent in adding global actions to all `*Directions` in the graph that has those global actions. I am not certain that this is the root cause , do you have the build failure reported somewhere in your side ?

Bubu:
@mtwalli We tried upgrading to see if this fixes the issue. In the end if made no difference either way (but I wanted to try the latest version before reporting it to google.)

fynngodau:
The build failure was that F-Droid could not reproduce our app anymore. The only info that the F-Droid build server gave us was that `classes2.dex` was not indetical (we had R8 disabled at first). To investiagte, I tried to rebuild the application on my local machine. This also failed for the same reason, but there we could find out the exact difference using `diffoscope`. You can see the output of this tool (the classes in `classes2.dex` disassembled and compared) in our issue 135 above.

It is still a mystery to us why this inconsistency occured, and why only with the two specific Direction classes you can see in the output of the tool.

Bubu:
There's been some activity on the google issue we filed:

```
It looks like those two destinations are present in two separate graphs - `nav_graph.xml` and `nav_graph_onboarding.xml`. This means that Safe Args will generate one `OnboardingDeltaAnalyticsFragmentDirections` from your `nav_graph.xml` and another from the `nav_graph_onboarding.xml` file - each attempting to add all valid actions from that destination (i.e., all of the global actions from the parent graphs).

What you're experiencing is the last generated version winning, overriding the other one. This is tracked in

[b/80332455](https://issuetracker.google.com/issues/80332455), which I'll mark this as a duplicate of.

As per [b/80332455#comment5](https://issuetracker.google.com/issues/80332455#comment5), you should only create a single destination for a given android:name, thus ensuring that Safe Args only generates that Directions class a single time. (You might consider using an <include> of a navigation graph that has just a single <fragment> destination). As an alternative workaround, you could also create an empty subclass for the Onboarding copies of those destinations just so that they have a different android:name.
```

(copying over the referenced comment:)
```
For a destination with an android:name, you'd want to include that android:name in only one place in your graph, referencing that shared destination from all of the places that need access to it (as a destination can access any sibling destinations of their parent graph).
```

So I suppose the recommended solution here looks somewhat different than what this MR is doing right now.

@mtwalli Do you/the cwa team want to follow up on this yourselves or should we attempt to implemented sugested solution by the google team?

mtwalli:
@Bubu , Thanks for the update, It would be great if you can implement the workaround and verify that it fixes the issue in your side?

mtwalli:
@fynngodau  should we close this PR for now?

fynngodau:
@mtwalli Sure, we’ll follow up with a fresh PR.

PR #3327, opened 2021-05-31
Risk calculation cannot be restarted after time travel (EXPOSUREAPP-6646)

LukasLechnerDev:
### Description
After setting the device time into the future, a `Risk State` with `CALCULATION_FAILED` is stored and the white error card is shown. After the user then sets the device time back to the correct time again, the white card is still there even after the user refreshes the risk calculation. 

The problem here is that when the device time is incorrect, the time of the device is taken as the `calculatedAt` property of the `RiskState`. When the user then sets the correct time again, the `Risk State` with the latest `calculatedAt` time is still the one with `CALCULATION_FAILED` and so the white card stays there. 

After some big workaround ideas, @d4rken and especially @chiljamgossow gave me some useful hints for a better solution. 

The solution proposed here is to simply use the server time instead of the incorrect device time for the `calculatedAt` property of the `Risk State`. This fixes the issue but I am not completely sure if we aren't introducing other issues with this change. 

### How to test
- Perform Risk calculation with the correct time and see green/red risk card appear
- Set the device time a couple of days into the future
- Go back to CWA and see the white card appear
- Set the device time back to the correct time
- Re-trigger the Risk calculation
- green/red risk card should appear again


chiljamgossow (dismissed):
LGTM

chiljamgossow (approved):

AlexanderAlferov (approved):
LGTM 🚀 
Tested on Samsung S8

PR #3328, opened 2021-05-31
 For RAT the test date is not set correct (EXPOSUREAPP-7555)

kolyaopahle:
### Description
This PR should contain the missing changes that allow rat test times to not only base outdated calculations on the new timestamp from the server,  but also display the timestamp to the user.

### Steps to reproduce
1. Scan a RAT qr code
2. Set its result and sampleCollectedAt time
3. Check if sampleCollectedAt time is displayed everywhere in the ui


BMItr (approved):
code lgtm.
tested on oneplus 8 with some RATs.
image
👌

LukasLechnerDev (approved):
👍 LGTM!

PR #3332, opened 2021-05-31
Test certificate repository, episode 1 (EXPOSUREAPP-7505)

d4rken:
`TestCertificateRepository` that handles creating/refreshing/deleting digital covid certificates for corona tests.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/CoronaWarnApplication.kt
```diff
@@ -138,6 +140,7 @@ class CoronaWarnApplication : Application(), HasAndroidInjector {
         Timber.v("Setting up test result available notification services.")
         pcrTestResultAvailableNotificationService.setup()
         raTestResultAvailableNotificationService.setup()
+        testCertificateRetrievalScheduler.setup()
```
d4rken, line 143: This makes the `TestCertificateRepository` active, with this nothing would happen.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/CovidCertificateConfig.kt
```diff
@@ -0,0 +1,16 @@
...
+
+    val testCertificate: TestCertificate
+
+    interface TestCertificate {
+        val waitAfterPublicKeyRegistration: Duration
+        val waitForRetry: Duration
+    }
+
+    interface Mapper : ConfigMapper<CovidCertificateConfig>
+}
```
d4rken, line 16: New config parameters.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+            }
+            .catch {
+                it.reportProblem(TAG, "Failed to snapshot TestCertificateContainer data to storage.")
+                throw it
+            }
+            .launchIn(appScope + dispatcherProvider.Default)
+    }
+
+    suspend fun requestCertificate(test: CoronaTest): TestCertificateContainer {
+        Timber.tag(TAG).d("createDccForTest(test.identifier=%s)", test.identifier)
```
d4rken, line 80: For an existing corona test, create a new test certificate entry. This just creates the entry to kick off automation.
No network requests happen yet.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                }
+            }
+        }
+
+        return refreshResults
+    }
+
+    private suspend fun registerPublicKey(
+        cert: TestCertificateContainer
+    ): TestCertificateContainer {
```
d4rken, line 203: Register our public key with the server.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+            }
+        } catch (e: Exception) {
+            Timber.tag(TAG).e("Failed to register public key for %s", cert)
+            throw e
+        }
+    }
+
+    private suspend fun obtainCertificate(
+        cert: TestCertificateContainer
+    ): TestCertificateContainer {
```
d4rken, line 244: Try to obtain the actual certificate. PublicKey registration and certificate retrieval are two steps, because if we manage to register our public key, but fail to get the certificate, we are still one step further.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                }
+
+                try {
+                    executeRequest()
+                } catch (e: Exception) {
+                    // TODO catch a specific error that reflects error code DGC_COMP_202
+                    delay(certConfig.waitForRetry.millis)
+                    executeRequest()
+                }
+            }
```
d4rken, line 278: The artificial delay due to the server not being able to make the certificate immediately available after registering out public key.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                    certificateReceivedAt = nowUtc,
+                )
+                CoronaTest.Type.RAPID_ANTIGEN -> (cert as RACertificateContainer).copy(
+                    testCertificateQrCode = extractedData.qrCode,
+                    certificateReceivedAt = nowUtc,
+                )
+            }.also {
+                it.qrCodeExtractor = qrCodeExtractor
+                it.preParsedData = extractedData.testCertificateData
+            }
```
d4rken, line 305: Decrypting and storing the retrieved components.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/TestCertificateContainer.kt
```diff
@@ -0,0 +1,114 @@
...
+    abstract val registeredAt: Instant
+    abstract val publicKeyRegisteredAt: Instant?
+    abstract val rsaPublicKey: RSAKey.Public?
+    abstract val rsaPrivateKey: RSAKey.Private?
+    abstract val certificateReceivedAt: Instant?
+    abstract val encryptedDataEncryptionkey: ByteString?
+    abstract val encryptedDccCose: ByteString?
+    abstract val testCertificateQrCode: String?
+
+    abstract val isUpdatingData: Boolean
```
d4rken, line 30: Currently both test certificate types (RA/PCR) have the same attributes, so we re-use an abstract class, but still treat them individually, as this may change in the future and then we don't have to migrate any data structures.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/common/TestCertificateRetrievalScheduler.kt
```diff
@@ -0,0 +1,116 @@
...
+    private val processedNewCerts = mutableSetOf<String>()
+
+    private val creationTrigger = testRepo.coronaTests
+        .map { tests ->
+            tests
+                .filter { it.isDccSupportedByPoc } // Only those that support it
+                .filter { it.isNegative } // Certs only to proof negative state
+                .filter { it.isDccConsentGiven && !it.isDccDataSetCreated } // Consent and doesn't exist already?
+        }
+        .distinctUntilChanged()
```
d4rken, line 44: We monitor the corona tests, and for each suitable test result, we will create a test certificate entry.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/common/TestCertificateRetrievalScheduler.kt
```diff
@@ -0,0 +1,116 @@
...
+            val isNew = processedNewCerts.contains(it.identifier)
+            if (isNew) processedNewCerts.add(it.identifier)
+            isNew
+        }
+
+        val hasWorkToDo = certificates.any { it.isPending }
+        Timber.tag(TAG).v("shouldPollDcc? hasNewCert=$hasNewCert, hasWorkTodo=$hasWorkToDo, foreground=$isForeground")
+        (isForeground || hasNewCert) && hasWorkToDo
+    }
+        .distinctUntilChanged()
```
d4rken, line 61: We also monitor our test certificates, and if the conditions are right, we will talk to the server to try to obtain the actual certificate components.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/common/TestCertificateRetrievalScheduler.kt
```diff
@@ -0,0 +1,116 @@
...
+            Timber.tag(TAG).d("Worker already queued, skipping requeue.")
+        }
+
+        Timber.tag(TAG).d("enqueueUniqueWork PCR_TESTRESULT_WORKER_UNIQUEUNAME")
+        workManager.enqueueUniqueWork(
+            WORKER_ID,
+            ExistingWorkPolicy.KEEP,
+            buildWorkerRequest()
+        )
+    }
```
d4rken, line 100: Test certificate retrieval happens via worker, if not riggered through the UI.
We may receive a negative test result while the app is in the background, and to not be killed off by the system, before we can get a certificate, we need to run the `TestCertificateRepo.refresh()` within a worker.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/CertificatePersonIdentifier.kt
```diff
@@ -8,7 +10,7 @@ import de.rki.coronawarnapp.vaccination.core.qrcode.VaccinationCertificateQRCode
 import org.joda.time.LocalDate
 import timber.log.Timber
 
-data class VaccinatedPersonIdentifier(
+data class CertificatePersonIdentifier(
```
d4rken, line 13: All DCC certs use the same name/DOB pattern as unique identifier, as we have to combine a few different types (per person) in a future sprint, small alignment here already.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/coronatest/CoronaTestTestData.kt
```diff
@@ -0,0 +1,167 @@
...
+import de.rki.coronawarnapp.vaccination.core.certificate.HealthCertificateHeader
+import okio.ByteString.Companion.decodeBase64
+import org.joda.time.Instant
+import javax.inject.Inject
+
+@Suppress("MaxLineLength")
+class CoronaTestTestData @Inject constructor(
+    private val qrCodeExtractor: TestCertificateQRCodeExtractor,
+    private val rsaKeyPairGenerator: RSAKeyPairGenerator,
+) {
```
d4rken, line 19: Test data isn't final yet, need to input some better values in a future PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+
+            val encryptionkey = rsaCryptography.decrypt(
+                toDecrypt = components.dataEncryptionKeyBase64.decodeBase64()!!,
+                privateKey = cert.rsaPrivateKey!!
+            )
+
+            val extractedData = qrCodeExtractor.extract(
+                decryptionKey = encryptionkey.toByteArray(),
+                encryptedCoseComponents = components.encryptedCoseTestCertificateBase64.decodeBase64()!!
+            )
```
chiljamgossow, line 289: I changed this so that both parameters are ByteArrays...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/TestCertificateContainer.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    @delegate:Transient
+    private val certificateData: TestCertificateData by lazy {
+        preParsedData ?: testCertificateQrCode!!.let { qrCodeExtractor.extract(it).testCertificateData }
+    }
+
+    val isPublicKeyRegistered: Boolean
+        get() = publicKeyRegisteredAt != null
+
+    val isPending: Boolean
```
chiljamgossow, line 45: maybe use a more specific naming so there is no confusion with test pending

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/common/TestCertificateRetrievalScheduler.kt
```diff
@@ -0,0 +1,116 @@
...
+
+    private fun buildWorkerRequest() =
+        OneTimeWorkRequestBuilder<TestCertificateRetrievalWorker>()
+            .setConstraints(
+                Constraints.Builder().setRequiredNetworkType(NetworkType.CONNECTED).build()
+            )
+            .setInitialDelay(TEST_RESULT_PERIODIC_INITIAL_DELAY, TimeUnit.SECONDS)
```
chiljamgossow, line 107: Can you append _SECONDS so it is clear what unit it is

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/TestCertificateQRCodeExtractor.kt
```diff
@@ -13,7 +13,7 @@ class TestCertificateQRCodeExtractor @Inject constructor() {
     fun extract(
         decryptionKey: ByteArray,
         encryptedCoseComponents: ByteString,
```
chiljamgossow, line 15: this a Byte Array now

chiljamgossow, line 15: This is a ByteArray now to be consistent

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                    identifier = identifier,
+                    registeredAt = test.registeredAt,
+                    registrationToken = test.registrationToken,
+                )
+            }
+            Timber.tag(TAG).d("Adding test certificate entry: %s", certificate)
+            mutate { this[certificate.identifier] = certificate }
+        }
+
+        return newData.values.single { it.registrationToken == test.registrationToken }
```
mtwalli, line 108: would that throw an exception if the registration token does not exist , or it is granted to have a value ?

if yes, please handle this the `TestCertificateRetrievalScheduler.setup()` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+
+            when (cert.type) {
+                CoronaTest.Type.PCR -> (cert as PCRCertificateContainer).copy(
+                    testCertificateQrCode = extractedData.qrCode,
+                    certificateReceivedAt = nowUtc,
+                )
+                CoronaTest.Type.RAPID_ANTIGEN -> (cert as RACertificateContainer).copy(
+                    testCertificateQrCode = extractedData.qrCode,
+                    certificateReceivedAt = nowUtc,
+                )
```
mtwalli, line 301: would that work?
```kotlin                  
                   when (cert) {
                        is PCRCertificateContainer -> it.copy()
                        is RACertificateContainer -> it.copy()
                    }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+            when (cert.type) {
+                CoronaTest.Type.PCR -> (cert as PCRCertificateContainer).copy(
+                    publicKeyRegisteredAt = nowUTC,
+                    rsaPublicKey = rsaKeyPair.publicKey,
+                    rsaPrivateKey = rsaKeyPair.privateKey,
+                )
+                CoronaTest.Type.RAPID_ANTIGEN -> (cert as RACertificateContainer).copy(
+                    publicKeyRegisteredAt = nowUTC,
+                    rsaPublicKey = rsaKeyPair.publicKey,
+                    rsaPrivateKey = rsaKeyPair.privateKey,
```
mtwalli, line 233:
```kotlin                  
                   when (cert) {
                        is PCRCertificateContainer -> it.copy()
                        is RACertificateContainer -> it.copy()
                    }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+        }
+
+        internalData.updateBlocking {
+            val certs = values.filter { workedOnIds.contains(it.identifier) }
+
+            mutate {
+                certs.forEach {
+                    this[it.identifier] = when (it.type) {
+                        CoronaTest.Type.PCR -> (it as PCRCertificateContainer).copy(isUpdatingData = false)
+                        CoronaTest.Type.RAPID_ANTIGEN -> (it as RACertificateContainer).copy(isUpdatingData = false)
```
mtwalli, line 192: it looks like the class hierarchy is sufficient to have a `copy` , would that work 
```kotlin                  
                   when (it) {
                        is PCRCertificateContainer -> it.copy()
                        is RACertificateContainer -> it.copy()
                    }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+        Timber.tag(TAG).d("refresh(identifier=%s)", identifier)
+
+        val refreshResults = mutableSetOf<RefreshResult>()
+
+        val workedOnIds = mutableSetOf<TestCertificateIdentifier>()
+
+        internalData.updateBlocking {
+            val toRefresh = values
+                .filter { it.identifier == identifier || identifier == null }
+                .filter { it.isPending }
```
mtwalli, line 126: can you merge the two `filters ` in one ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+
+            mutate {
+                toRefresh.forEach {
+                    workedOnIds.add(it.identifier)
+                    this[it.identifier] = when (it.type) {
+                        CoronaTest.Type.PCR -> (it as PCRCertificateContainer).copy(isUpdatingData = true)
+                        CoronaTest.Type.RAPID_ANTIGEN -> (it as RACertificateContainer).copy(isUpdatingData = true)
```
mtwalli, line 133: when (it) {
                        is PCRCertificateContainer -> it.copy()
                        is RACertificateContainer -> it.copy()
                    }

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                        RefreshResult(registerPublicKey(cert))
+                    } catch (e: Exception) {
+                        Timber.tag(TAG).e(e, "Failed to register public key for %s", cert)
+                        RefreshResult(cert, e)
+                    }
+                }
+
+            mutate {
+                refreshedCerts
+                    .filter { it.error == null }
```
mtwalli, line 156: does this mean that the returned certificates all has errors ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                        RefreshResult(obtainCertificate(cert))
+                    } catch (e: Exception) {
+                        Timber.tag(TAG).e(e, "Failed to retrieve test certificate for %s", cert)
+                        RefreshResult(cert, e)
+                    }
+                }
+
+            mutate {
+                refreshedCerts
+                    .filter { it.error == null }
```
mtwalli, line 179: erroneous certs only ? or ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
mtwalli: It would be nice if you can this comment as a Kotlin comment in code ? PR is gone, comment in there for later

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+        }
+    }
+
+    suspend fun clear() {
+        Timber.tag(TAG).i("clear()")
+        internalData.updateBlocking { emptyMap() }
+    }
+
+    companion object {
+        val TAG = TestCertificateRepository::class.simpleName!!
```
mtwalli, line 327: private

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/TestCertificateStorage.kt
```diff
@@ -0,0 +1,102 @@
...
+            }
+
+            return (pcrCerts + raCerts).also {
+                Timber.tag(TAG).v("Loaded %d certificates.", it.size)
+            }
+        }
+        set(value) {
+            Timber.tag(TAG).d("save(testCertificates=%s)", value)
+            prefs.edit {
+                value.filter { it.type == CoronaTest.Type.PCR }.run {
```
mtwalli, line 74: why do you need to check for the type here ? is not the custom Gson adapter should be doing that ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/TestCertificateContainer.kt
```diff
@@ -0,0 +1,114 @@
...
+import de.rki.coronawarnapp.vaccination.core.personIdentifier
+import de.rki.coronawarnapp.vaccination.core.qrcode.QrCodeString
+import de.rki.coronawarnapp.vaccination.core.server.valueset.VaccinationValueSet
+import de.rki.coronawarnapp.vaccination.core.server.valueset.getDisplayText
+import okio.ByteString
+import org.joda.time.Instant
+import org.joda.time.LocalDate
+import java.util.Locale
+
+abstract class TestCertificateContainer {
```
mtwalli, line 17: can it be a `sealed ` class ? in Kotlin 1.5 sealed classes can be in different files

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+        return newData.values.single { it.registrationToken == test.registrationToken }
+    }
+
+    data class RefreshResult(
+        val certificate: TestCertificateContainer,
+        val error: Exception? = null,
+    )
+
+    suspend fun refresh(identifier: TestCertificateIdentifier? = null): Set<RefreshResult> {
+        Timber.tag(TAG).d("refresh(identifier=%s)", identifier)
```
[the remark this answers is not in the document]
mtwalli: ☝🏼.`toKotlinDocPlease`  ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
```diff
@@ -0,0 +1,329 @@
...
+                }
+            }
+        }
+
+        internalData.updateBlocking {
+            Timber.tag(TAG).d("Checking for unregistered public keys.")
+
+            val refreshedCerts = values
+                .filter { workedOnIds.contains(it.identifier) }
+                .filter { !it.isPublicKeyRegistered }
```
mtwalli, line 144: two filters -> filter ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/TestCertificateStorage.kt
```diff
@@ -0,0 +1,102 @@
...
+        object : TypeToken<Set<RACertificateContainer>>() {}.type
+    }
+
+    var testCertificates: Collection<TestCertificateContainer>
+        get() {
+            Timber.tag(TAG).d("load()")
+
+            val pcrCerts: Set<PCRCertificateContainer> = run {
+                val raw = prefs.getString(PKEY_DATA_PCR, null) ?: return@run emptySet()
+                gson.fromJson<Set<PCRCertificateContainer>>(raw, typeTokenPCR).onEach {
```
mtwalli, line 51: GsonAdapter should be doing type checks ?

mtwalli (commented):
For public methods in the repo it would be helpful to annotate which one is expected to throw an exception ,so developers using it from UI layer can handle it accordingly

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/common/TestCertificateRetrievalScheduler.kt
d4rken: I've removed it. I don't see why we need a delay here.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/coronatest/storage/TestCertificateStorageTest.kt
```diff
@@ -0,0 +1,110 @@
...
+import testhelpers.extensions.toComparableJsonPretty
+import testhelpers.preferences.MockSharedPreferences
+import javax.inject.Inject
+
+@Suppress("MaxLineLength")
+class TestCertificateStorageTest : BaseTest() {
+    @MockK lateinit var context: Context
+    private lateinit var mockPreferences: MockSharedPreferences
+    @Inject lateinit var testData: CoronaTestTestData
+    @Inject lateinit var postProcessor: ContainerPostProcessor
```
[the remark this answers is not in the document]
d4rken: To test the actual `TestCertificateQRCodeExtractor`.
If we mock it all, then we don't know if it actually works in production.
If we manually construct it without DI, we can't be sure that it's the same `TestCertificateQRCodeExtractor` as it would be injected outside of unit tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    private fun DgcParameters.DGCParameters.mapCovidCertificateConfig(): CovidCertificateConfig.TestCertificate {
+        if (!this.hasTestCertificateParameters()) {
+            Timber.w("DCC config has no test certificate parameters.")
+            return TestCertificateConfigContainer()
+        }
+        return with(testCertificateParameters) {
+            TestCertificateConfigContainer(
+                waitAfterPublicKeyRegistration = waitAfterPublicKeyRegistrationInSeconds.let {
+                    if (it > 60 || it < 0) {
```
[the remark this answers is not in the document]
d4rken: That's nice, will change.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
d4rken: Every test should always have a registration token, but it may still throw an exception if there is already a test certificate entry for the same token. I'll add error handling to the scheduler :+1:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
d4rken: Yes, but then we would also need an `else ->`, while less, if someone adds a 3rd test type, we would not automatically get a compiler error here to notify us that we need to add extra handling here.

While less code, I think it's easier for future developers to add new test types if the compiler helps to fix things like this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/TestCertificateContainer.kt
d4rken: Nice! I didn't know that. :+1: 
I tried it just now and it seems it will only work if the files are in the same package, and currently they are split by package `pcr` and `rapidantigen`. :frowning:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
d4rken: Would prefer to keep it this way as performance is not an issue here, we will not have many entries.
The alternative seems less readable.

```kotlin
 .filter { (it.identifier == identifier || identifier == null) && it.isCertificateRetrievalPending  }
```

Semantically it's also nice: First we filter for all of those that are target (all or single identifier), then we filter for those that still need refreshing.

If this is a blocker for you, it can be changed though, I don't feel strongly about it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
d4rken: We filter out all `RefreshResult` that had an error so they stay the same as they previously were, while we update the entries that changed.

When refreshing multiple test certificates, each network request could throw an error, but if we successfully registered a public key or even got signed certificate data from the server, then we do not want to throw that away because an operation for a different certificate failed.

I just noticed though that the code is not adding to the top level `refreshResults`, I'll fix that.
The idea is that the top level `refreshResults` contain all operations (with and without error), while we only change the entries in `internalData` for values that were updated without error.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/TestCertificateRepository.kt
d4rken: Similar to the previous double filter, i think it serves readability two have it split.
We first filter for those that are targeted by this `refresh` call, then we filter for those that we need to work on in this step.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/storage/TestCertificateStorage.kt
d4rken: I think we would need to use `RuntimeTypeAdapterFactory` for this and for some reason I decided against it also in `CoronaTestStorage` :thinking:, likely because splitting the types seemed less fragile than using the `RuntimeTypeAdapterFactory`.

d4rken: If you see a way, we could discuss it on Monday.

PR #3345, opened 2021-06-01
DGC Detail Screen (EXPOSUREAPP-7478)

axelherbstreith:
- [x] Add Alert when deleting test certificate
- [x] Add Fullscreen Qr-Code
- [x] Wire Screen
- [x] Connect to Database


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/greencertificate/ui/certificates/details/CovidCertificateDetailsFragment.kt
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    private fun FragmentCovidCertificateDetailsBinding.onCertificateReady(
+        testCertificate: TestCertificate
+    ) {
+        qrCodeCard.subtitle.text = getString(
+            R.string.detail_green_certificate_card_subtitle,
+            testCertificate.testResultAt.toShortDayFormat(),
+            testCertificate.testResultAt.toShotTimeFormat(),
+        )
```
mtwalli, line 70: @d4rken  which date should be used here ?
Screenshot 2021-06-04 at 15 42 12

d4rken: `dr`, `testResultAt` would be correct AFAIK, I saw in another PR though that the `dr` attribute is optional now, so we need to clarify whether this is still desired and if there needs to be a fallback value :thinking: 

/cc @mlenkeit

mlenkeit: `sc` should be used here.

d4rken: > `sc` should be used here.

Thanks, so `sampleCollectedAt` @axelherbstreith

NataliaLemmerth (dismissed):
UA approved

BMItr (commented):
code looks good to me 👌 
just ping when you are done wiring ;-)

AlexanderAlferov (dismissed):
LGTM

AlexanderAlferov (approved):

mtwalli (approved):

NataliaLemmerth (approved):
UA approved

PR #3358, opened 2021-06-07
Error handling (EXPOSURERAPP-7508)

chiljamgossow:
ToDo 
- [X] server errors
- [X] test certificate decryption/decoding errors
- [ ] unit tests -> separate PR


Corona-Warn-App/build.gradle
```diff
@@ -441,6 +441,4 @@ dependencies {
 
     // HCert
     implementation("com.upokecenter:cbor:4.4.1")
-
-    implementation("bouncycastle:bcprov-jdk16:136")
```
d4rken, line 445: Why was this previously added and now removed?

chiljamgossow: Works just fine without

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/server/CovidCertificateApiV1.kt
```diff
@@ -15,16 +15,24 @@ interface CovidCertificateApiV1 {
     @POST("/version/v1/publicKey")
     suspend fun sendPublicKey(
         @Body requestBody: PublicKeyUploadRequest
-    )
+    ): Response<Any>
```
mtwalli, line 18: does it return anything ? would `Response<Unit>` work ?

mtwalli (dismissed):
Lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/server/CovidCertificateApiV1.kt
chiljamgossow: That's needs to be tested, what the server returns, Unit is likely fine

mtwalli (approved):

d4rken (approved):
lgtm

PR #3364, opened 2021-06-07
Default strings (DEV)

harambasicluka:


Corona-Warn-App/src/main/res/values/vaccination_strings.xml
```diff
@@ -91,8 +85,6 @@
         <item quantity="few">"Full vaccination protection in %1$d days"</item>
         <item quantity="many">"Full vaccination protection in %1$d days"</item>
     </plurals>
-    <!-- XTXT: Homescreen card complete vaccination status label -->
-    <string name="vaccination_card_status_vaccination_complete">"Gültig bis einschließlich %s"</string>
```
mtwalli, line 95: it seems that this string is still used in `ImmuneVaccinationCard`

mtwalli (dismissed):

mtwalli (approved):

PR #3366, opened 2021-06-07
App crash if details of increased/low with exposures risk card are opened (EXPOSUREAPP-7615)

kolyaopahle:
### Description

This PR fixes an issue that would occur when the FindDetailsInJournalBox was rendered on the exposure detail screen,
adding a layout tag around the xml layout seems to resolve the problem, although im not 100% sure why maybe @d4rken or @mtwalli can shed some light onto what is happening here.

### Steps to reproduce
1. Recalculate exposure risk with high level or with high due to multiple encounters
2. Check if info text is shown on the risk detail screen as in screenshot (currently only german text is available) and app does not crash


chiljamgossow (approved):
Works as promised

mtwalli (approved):
Lgtm

PR #3368, opened 2021-06-07
Error texts (DEV)

chiljamgossow:
https://github.com/corona-warn-app/cwa-app-tech-spec/pull/62/commits/ae987cda78e5a21b8caec76f1f878ebeecdfaac9


harambasicluka (commented):
Looks good, will ping our translation colleagues if there is any problem with the CDATA. What alternative could we offer? I think in other places we just escape the special characters.

chiljamgossow:
> Looks good, will ping our translation colleagues if there is any problem with the CDATA. What alternative could we offer? I think in other places we just escape the special characters.

I removed it. It was added automatically by AS

NataliaLemmerth (approved):
UA approved

mtwalli (approved):

PR #3374, opened 2021-06-08
Fixed DCC Server Module (EXPOSUREAPP-7457)

chris-cwa:
Added missing gson converter factory. Unit test shows that retrofit can now generate the API.


d4rken (approved):

chiljamgossow (approved):
LGTM

PR #3377, opened 2021-06-08
Fix name validation (EXPOSUREAPP-7569)

chiljamgossow:
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/test-certificate/docs/spec/dgc-overview-client.md#determining-the-display-name-of-the-person-from-a-dgc

checks explicitly for not null and not blank


BMItr (approved):
lgtm!

d4rken (approved):

PR #3378, opened 2021-06-08
TestCertificate bugfixes & more (EXPOSUREAPP-7505)

d4rken:
Fixes various issues around the test registration & certificate flow.

* Temporarily disable certificate pinning to allow testing, until server is sorted
* Fix date of birth hash being calculated and sent for RapidAntigen tests, it should only be sent for PCR tests.
* Fix handling of server responses for DCC components, `HTTP 202` has an empty body, Retrofit would throw an exception instead of returning the specific `HTTP 202` exception that we need to handle.
* Fix crash on opening the certificate list by refactoring TestCertificate storage. Using `copy` loses the `abstract var qrCodeExtractor`  as copy only considers data class constructor arguments, so we shouldn't do that :beer: 

For testing register a RA/PCR test with certificate via mock server or WRU


BMItr (approved):
👌

chiljamgossow (approved):

PR #3388, opened 2021-06-09
Bump nav versions and fix different versions (DEV)

mtwalli:
- We had different versions for Nav safe args and Navigation APIs 🤦🏼 -> now fixed 
- Bump to `2.3.5`


chris-cwa (approved):
Thank you!

BMItr (approved):

PR #3390, opened 2021-06-09
Certificates screenshots (EXPOSUREAPP-7444)

jurajkusnier:
Screenshots
CertificatesFragment(1)


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/CertificatesFragmentTest.kt
```diff
@@ -0,0 +1,190 @@
...
+@RunWith(AndroidJUnit4::class)
+class CertificatesFragmentTest : BaseUITest() {
+
+    @MockK lateinit var viewModel: CertificatesViewModel
+    @MockK lateinit var vaccinatedPerson: VaccinatedPerson
+
+    @Before
+    fun setup() {
+        MockKAnnotations.init(this, relaxed = true)
+        every { vaccinatedPerson.fullName } returns "Andrea Schneider"
```
mtwalli, line 46: Can we use `Max Mustermann` to align with all screenshots?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/CertificatesFragmentTest.kt
```diff
@@ -0,0 +1,190 @@
...
+        every { vaccinatedPerson.getMostRecentVaccinationCertificate.expiresAt } returns
+            DateTime.now().plusDays(365).toInstant()
```
mtwalli, line 48: Can we use specific dates? instead of `now` ?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/CertificatesFragmentTest.kt
```diff
@@ -0,0 +1,190 @@
...
+    private fun getVaccinationPendingCertScreenItems(): LiveData<List<CertificatesItem>> {
+        return MutableLiveData(
+            listOf(
+                HeaderInfoVaccinationCard.Item,
+                ImmuneVaccinationCard.Item(
+                    vaccinatedPerson = vaccinatedPerson,
+                    onClickAction = {}
+                ),
+                CovidTestCertificateErrorCard.Item(
+                    testDate = DateTime.now().toInstant(),
```
mtwalli, line 172: Please use specific dates?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/CertificatesFragmentTest.kt
```diff
@@ -0,0 +1,190 @@
...
+        return MutableLiveData(
+            listOf(
+                HeaderInfoVaccinationCard.Item,
+                ImmuneVaccinationCard.Item(
+                    vaccinatedPerson = vaccinatedPerson,
+                    onClickAction = {}
+                ),
+                CovidTestCertificateCard.Item(
+                    testDate = DateTime.now().toInstant(),
+                    testPerson = "Andrea Schneider"
```
mtwalli, line 148: `Max Mustermann` ?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/CertificatesFragmentTest.kt
```diff
@@ -0,0 +1,190 @@
...
+    private fun getVaccinationGreenCertScreenItems(): LiveData<List<CertificatesItem>> {
+        return MutableLiveData(
+            listOf(
+                HeaderInfoVaccinationCard.Item,
+                ImmuneVaccinationCard.Item(
+                    vaccinatedPerson = vaccinatedPerson,
+                    onClickAction = {}
+                ),
+                CovidTestCertificateCard.Item(
+                    testDate = DateTime.now().toInstant(),
```
mtwalli, line 147: specific dates please ?

DerVogel2020:
In the screenshots, there is English and German mixed.
Is this in the released version the same?

jurajkusnier:
> In the screenshots, there is English and German mixed.
> Is this in the released version the same?

Yes, but that's only because some strings are not translated yet.

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/CertificatesFragmentTest.kt
jurajkusnier: I'm using data according to the Figma design, but I can change the name if we are already using something else somewhere else.

harambasicluka:
> > In the screenshots, there is English and German mixed.
> > Is this in the released version the same?
> 
> Yes, but that's only because some strings are not translated yet.

That's a requirement the labels are `german / english`, also the text below.

mtwalli (approved):
👍🏼

jurajkusnier:
> > > In the screenshots, there is English and German mixed.
> > > Is this in the released version the same?
> > 
> > 
> > Yes, but that's only because some strings are not translated yet.
> 
> That's a requirement the labels are `german / english`, also the text below.

I think it was about `vaccination_card_status_vaccination_complete` in the previous screenshots

axelherbstreith (approved):
Lgtm 👍 Thank you for fixing the order of the name

PR #3401, opened 2021-06-09
Material Design 🎨 (DEV)

mtwalli:
- Move completely to MD Theme
-  `ProgressBar`  ->  `CircularProgressIndicator`
- Remove some redundant styles 

### Testing 
- Run the App and go through the screens 
- Check on older Android phones
- [Screenshots](https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/16202/workflows/d107eb6e-2f7c-4740-8298-aff20fb16aea/jobs/132021)

### TODO
- [x] Revert screenshots job


fynngodau:
Since buttons now have shadows again, could you please consider my changes from #2942?

mtwalli:
> Since buttons now have shadows again, could you please consider my changes from #2942?

Good point 👍, as Luka mentioned in the reference PR , there should not be shadows for buttons. I disabled the shadows here too. 

This PR is meant to just provide consistency, since we have mixed styles.

fynngodau:
Ah, I see @mtwalli, buttons still feel better than previously, thanks for PR!

You made the non-highlighted tabs in the bottom tab bar a bit darker. Since there are already so many tabs in the new version, I think it would be better if the non-selected tabs were not to catch the eye as much, so the screen looks less "full". For this reason I'd prefer if you could change this again.

Previously | With PR
---|---
Screenshot_1623305844 | Screenshot_1623305938

mtwalli:
> Ah, I see @mtwalli, buttons still feel better than previously, thanks for PR!
> 
> You made the non-highlighted tabs in the bottom tab bar a bit darker. Since there are already so many tabs in the new version, I think it would be better if the non-selected tabs were not to catch the eye as much, so the screen looks less "full". For this reason I'd prefer if you could change this again.

Done

axelherbstreith:
Nice refactoring 👍 Only one small remark from my side, the slider color looks a little bit to bright: 
image

axelherbstreith (approved):
Lgtm, thanks for fixing 👍

AlexanderAlferov (approved):
Tested on Samsung S8, Android 9. Overall looks good 🚀 
But some margins feels little bit off (e.g. onboarding ENF bottom buttons panel feels too big due to big margins between buttons), probably should be rechecked by designers later to tune.

fynngodau (approved):
Thanks. Haven't looked at the code changes in detail but the result lgtm!

PR #3403, opened 2021-06-10
Hot Fix date parsing (EXPOSUREAPP-7744)

chiljamgossow:
Should now be possible to scan this 
https://github.com/eu-digital-green-certificates/dgc-testdata/blob/main/BG/png/BG-1.png


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/certificate/VaccinationDGCV1.kt
```diff
@@ -39,9 +39,11 @@ data class VaccinationDGCV1(
...
-            get() = LocalDate.parse(dt)
+            get() = LocalDate.parse(dt.removeSuffix(DATE_SUFFIX))
     }
 
     val dateOfBirth: LocalDate
-        get() = LocalDate.parse(dob)
+        get() = LocalDate.parse(dob.removeSuffix(DATE_SUFFIX))
 }
+
+private const val DATE_SUFFIX = "T00:00:00"
```
d4rken, line 49: What if the time value is not `T00:00:00`?

Could we try date time parsing and then just take the date? e.g.

```kotlin
val vaccinatedAt: LocalDate
    get() = try {
        LocalDate.parse(dt)
    } catch (e: Exception) {
        try {
            LocalDateTime.parse(dt).toLocalDate()
        } catch (ignore: Exception) {
            throw e
        }
    }
```

chiljamgossow: The tech spec says it is T00:00:00, and nothing else, otherwise which time zone are we using here to convert?

d4rken: @mlenkeit What do you think?

chiljamgossow: I considered what you are proposing first, but decided otherwise because of these reasons

mlenkeit: Currently, we are only seeing `T00:00:00`. It seems to a ~~bug~~ feature of some C# JSON schema parsing lib (or something similar). Anyway, hard-coding for this is fine for now.

AlexanderAlferov (approved):
LGTM 🚀 
Tested on Samsung S8

d4rken (approved):

Issue #3413, opened 2021-06-10
No human-readable message when scanning SAME QR code with vaccine certificate scanner


vaubaehn:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue - it's quite closely related to #3394  , but @heinezen wanted me to open a new issue ;)

## Technical details

- Device name: LG-G4 (LGH-815)
- Android version: 6.0
- App version: 2.3.2

## Describe the bug
When scanning the SAME vaccination QR code, you get an error message saying "Das Impfzertifikat ist bereits in Ihrer App registriert - VC_ALREADY_REGISTERED". When clicking on "Details", a stacktrace is shown.

screenshots:
<img src="https://user-images.githubusercontent.com/68376758/121543989-a6fc4880-ca09-11eb-9c58-c41d752e085b.png" width="100" > <img src="https://user-images.githubusercontent.com/68376758/121544054-b67b9180-ca09-11eb-9c03-46f9bb1d7059.png" width="100" >

## Steps to reproduce the issue

1. Scan a vaccination certificate.
2. Scan the same vc again.
3. See error

## Expected behaviour
A stacktrace for a known exception is superfluous and may confuse users. The stacktrace should not be displayed in this case, especially as the stacktrace's dialog does not contain any "ok" or "dismiss" button.

## Possible Fix
Remove "details" button.

## Additional context
related: #3400 , #3394

----
Internal Tracking-ID: [EXPOSUREAPP-7768](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7768)

MikeMcC399:
@vaubaehn 
"overfluent" is not in my dictionary
but maybe "überflüssig" => "superfluous" or "unnecessary"?

vaubaehn:
@MikeMcC399 thanks, I changed it accordingly.

heinezen:
@vaubaehn Thank you.

Info is mirrored to Jira. It's tracked as an enhancement request, not as a bug though.

----
Corona-Warn-App Open Source Team

MikeMcC399:
@vaubaehn 
[...]

PR #3400, opened 2021-06-09
Hot fix faq invalid vc (EXPOSUREAPP-7709)

chiljamgossow:
device-2021-06-09-173604

Issue #3394, opened 2021-06-09
No human-readable message when scanning wrong QR code with vaccine certificate scanner


Michael-635:
## Technical details

- Device name: Samsung Galaxy A70
- Android version: 11
- App version: 2.3.2

## Describe the bug

When scanning some random QR code which is not a vaccination certificate, you get an error message saying "Dieser QR-Code ist kein gültiges Impfzertifikat - VC_PREFIX-INVALID". When clicking on "Details", a StackTrace is shown (probably intentional). To me, a StackTrace containing lines like

     kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)

is an information disclosure which in my opinion is a vulnerability and should be avoided.

## Steps to reproduce the issue

1. Click on "Hinzufügen" for "Impfzertifikat hinzufügen"
2. Scan a random QR code which is not a valid vaccination certificate
3. In the resulting error message "Fehler - Dieser QR-Code ist kein gültiges Impfzertifikat" click on "DETAILS"
4. StackTrace will be shown:

![image](https://user-images.githubusercontent.com/67007492/121358955-4e5a7c00-c933-11eb-97f1-0d8bdfd7e6cc.png)

## Expected behaviour

A human readable explanation should be shown which does not disclose any implementation details.

----
Internal Tracking-ID: [EXPOSUREAPP-7709](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7709)

Ein-Tim:
@Michael-635 Could you elaborate why this is a disclosure? The code is public on GitHub anyways. BTW AFAIK a StrackTrace like this is shown for all errors.

Michael-635:
Well, ok, then I guess it's just my training to never ever show a StackTrace to a user, which of course might not be valid for open source apps.

But still, even if it's not a disclosure, is it acceptable from a user experience perspective to present messages like this without an explanation to an end user, even after pressing details?

MikeMcC399:
@Michael-635 
I think you have a very valid point as the app is inconsistent in how it handles QR code scanning errors.

1. Scan an incorrect QR code with Register Your Test / Scan QR Code and the error message is "QR code is invalid" with **no** Details / Stack Trace.
2. Scan an incorrect QR code with Add Vaccination Certificate and the message includes the Details selection which shows the stack trace.

This could be considered an enhancement request to harmonize the error messages when an incorrect QR code is scanned.

[...]

PR #3414, opened 2021-06-10
Handle semi-invalid vaccination certificate QR-codes (EXPOSUREAPP-7755)

d4rken:
This introduces "modes" to qrcode data parsing.
* UI based parsing will be STRICT and show error codes, because we can show pop a dialog and abort.
* Storage based parsing will be LENIENT. If the user managed to scan something invalid, then we can't just delete it, and we also can't (shouldn't) crash the app. Future updates could introduce "error UI states" for this.

In this case, users scanned a vaccination qr code that contained more than 1 data set. Unclear whether this was just hacking around, or some (non DE?) sources generate such codes.

This is a test vaccination qr code that contain two data entries.
[**DO NOT SCAN THIS**](https://user-images.githubusercontent.com/1439229/121550411-01e46e80-ca0f-11eb-9fa1-5184e7979cab.png)
With your prod version :expressionless:, or your app will crash until the next update is out  :shrug:.

Before:
* Error dialog about more than one list item.
* Crash after dismissing dialog
* Crash on app open

```java
at kotlin.collections.ArraysKt___ArraysKt.single (_Arrays.kt:10)
at de.rki.coronawarnapp.vaccination.core.repository.storage.VaccinationContainer.getVaccination (VaccinationContainer.kt:1)
at de.rki.coronawarnapp.vaccination.core.repository.storage.VaccinationContainer$toVaccinationCertificate$1.getDoseNumber (VaccinationContainer.kt:1)
at de.rki.coronawarnapp.vaccination.core.VaccinatedPerson.getTimeUntilImmunity (VaccinatedPerson.kt:4)
at de.rki.coronawarnapp.vaccination.core.VaccinatedPerson.getVaccinationStatus$default (VaccinatedPerson.kt:4)
at de.rki.coronawarnapp.ui.main.home.HomeFragmentViewModel$special$$inlined$combine$1$3.invokeSuspend (Zip.kt:20)
at de.rki.coronawarnapp.ui.main.home.HomeFragmentViewModel$special$$inlined$combine$1$3.invoke (Zip.kt:2)
at kotlinx.coroutines.flow.internal.CombineKt$combineInternal$2.invokeSuspend (Combine.kt:26)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:3)
at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:15)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.kt:1)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt:12)
```

Now:
* Error dialog about qr code not being supported yet
* No crash on dismiss
* If you have already scanned the bad qrcode and update, your app should no longer crash. The app will show the newest out of multiple available data entries.

Screenshot from 2021-06-10 17-14-48


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/certificate/InvalidHealthCertificateException.kt
```diff
@@ -73,6 +75,9 @@ class InvalidHealthCertificateException(
             VC_NO_VACCINATION_ENTRY -> CachedString { context ->
                 context.getString(ERROR_MESSAGE_VC_NOT_YET_SUPPORTED)
             }
+            VC_MULTIPLE_VACCINATION_ENTRY -> CachedString { context ->
```
chiljamgossow, line 78: Here a plural would make sense -> ENTRIES

vaubaehn:
> DO NOT SCAN THIS

😱 

But seriously - while you're having your hands on it, will you consider to remove the  error dialog's "details" button here also?
See #3400 , #3413 , #3394

Ein-Tim:
I don't understand the problem here. Is the QR-Code [you shared](https://user-images.githubusercontent.com/1439229/121550411-01e46e80-ca0f-11eb-9fa1-5184e7979cab.png) invalid under Android, and if yes why? It works smoothly on my iOS device.

d4rken:
@vaubaehn 
> > DO NOT SCAN THIS
> 
> 
> 
> But seriously - while you're having your hands on it, will you consider to remove the error dialog's "details" button here also?
> See #3400 , #3413 , #3394

The details button is part of the general error dialog code that is used in many places.
Changing that is outside of the hotfix scope.
Whether all dialogs with a more specific error message should have the details button removed is something to discuss as part of a enhancement ticket.
Does every specific error type make the stacktrace obsolete, or are there cases where the stacktrace helps us fix it? The with error logs available, we could get the stacktrace that way still.

@Ein-Tim 

> I don't understand the problem with the QR-Code here. Why is the QR-Code you shared invalid under Android? It works smoothly on my iOS device.

The specifications say that a vaccination certificate looks like this

```json
{
  // Version (required)
  "ver": "1.0.0",
  // Name (required)
  "nam": {
    // Family name
    "fn": "d'Arsøns - van Halen",
    // Given name
    "gn": "François-Joan",
    // Standardized family name (required)
    "fnt": "DARSONS<VAN<HALEN",
    // Standardized given name
    "gnt": "FRANCOIS<JOAN"
  },
  // Date of birth (required)
  "dob": "2009-02-28",
  // Vaccination certificate (assumption: exactly one entry)
  "v": [
    {
      // Disease or agent targeted (required)
      "tg": "840539006",
      // Vaccine or prophylaxis (required)
      "vp": "1119349007",
      // Vaccine medicinal product (required)
      "mp": "EU/1/20/1528",
      // Marketing Authorization Holder (required)
      "ma": "ORG-100030215",
      // Dose Number (required)
      "dn": 2,
      // Total Series of Doses (required)
      "sd": 2,
      // Date of Vaccination (required)
      "dt": "2021-04-21",
      // Country of Vaccination (required)
      "co": "NL",
      // Certificate Issuer (required)
      "is": "Ministry of Public Health, Welfare and Sport",
      // Unique Certificate Identifier (required)
      "ci": "urn:uvci:01:NL:PlA8UWS60Z4RZXVALl6GAZ"
    }
  ]
}
```

The invalid certificate contains two elements in the `v` array.
Android internally used `List.single()` which caused the crash.
For everyone already affected, this fix changes it to show the newest of both elements.

According to @nickguendling
iOS currently just takes the first entry if there are multiple, and will get additional validiation in one of the next updates.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/vaccination/core/certificate/InvalidHealthCertificateException.kt
d4rken: Thanks, fixed.

vaubaehn:
@d4rken 

> The details button is part of the general error dialog code that is used in many places.
> Changing that is outside of the hotfix scope.

My full understanding. I naively thought, it could be easy to make use of @chiljamgossow 's code in #3400 , setting VC_MULTIPLE_VACCINATION_ENTRIES in the error list and using her error dialog builder, but it's probably much more complex :)

> Whether all dialogs with a more specific error message should have the details button removed is something to discuss as part of a enhancement ticket.
> Does every specific error type make the stacktrace obsolete, or are there cases where the stacktrace helps us fix it? With error logs available, we could get the stacktrace that way still. 🤔

I think good candidates for a more user-friendly/simpler error handling do not depend only on their specifity, but also how well-known they are, and that no "surprises" can be expected, like API 39508. I think everyone would agree that no stacktrace is needed for them, and I guess most vc errors could also be set on this list. But right, this is something that needs to be discussed in your whole team and could be part of a bigger refactoring of error handling. Afair, there are already open tickets for this, so there is no need to create a new issue for this, is there?

d4rken:
> My full understanding. I naively thought, it could be easy to make use of @chiljamgossow 's code in #3400 , setting VC_MULTIPLE_VACCINATION_ENTRIES in the error list and using her error dialog builder, but it's probably much more complex :)

That PR overwrites the default "neutral action" (the details -> stacktrace) of the error dialog builder on that specific screen. Yes we could do the same in this case, but we don't want to do this manually for every screen. We want an automatic solution :grin:.

The actual code may be as easy as "if(isKnownError) { setNeutralButton(null) }", but we need to think about the edge cases, and that we move all error handling into "nice and consistent" territory :coffee:.

> I think good candidates for a more user-friendly/simpler error handling do not depend only on their specifity, but also how well-known they are, and that no "surprises" can be expected, like API 39508. I think everyone would agree that no stacktrace is needed for them, and I guess most vc errors could also be set on this list. But right, this is something that needs to be discussed in your whole team and could be part of a bigger refactoring of error handling. 


> Afair, there are already open tickets for this, so there is no need to create a new issue for this, is there?

Discussion is always welcome. I'm not sure atm if there is a suitable ticket yet. I don't think we need to discuss each dialog individually, we'd need to discuss this on a general level.

We have a few links already, maybe "links to faq" should be part of the error dialog system and if such a link is present, automatically show that instead of the stacktrace button. 

There is also the issue that only some parts of the app use the `HasHumanReadableError` based exception system, for it all to work nicely together we need to use it everywhere, and get rid of the old CWA 1.0 error code stuff (e.g. CODE 900X).

chiljamgossow (approved):
LGTM

vaubaehn:
> That PR overwrites the default "neutral action" (the details -> stacktrace) of the error dialog builder on that specific screen. Yes we could do the same in this case, but we don't want to do this manually for every screen. We want an automatic solution 😁.

Automatic is good 🤖 Also, if I interpreted the code right after looking again, you would not get a specific "not yet supported" error here, but the general "invaild certificate" together with a link to the faq where nothing is explained, if you joined VC_MULTIPLE_VACCINATION_ENTRIES to `codesVcInvalid` , so...

> The actual code may be as easy as "if(isKnownError) { setNeutralButton(null) }", but we need to think about the edge cases, and that we move all error handling into "nice and consistent" territory ☕.

👍 

> We have a few links already, maybe "links to faq" should be part of the error dialog system and if such a link is present, automatically show that instead of the stacktrace button.

🚀 
 
> There is also the issue that only some parts of the app use the `HasHumanReadableError` based exception system, for it all to work nicely together we need to use it everywhere, and get rid of the old CWA 1.0 error code stuff (e.g. CODE 900X). I want to delete [this](https://github.com/corona-warn-app/cwa-app-android/blob/4855156f60b40cd9851e19e04a2044dbd24760c9/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/exception/reporting/ErrorCodes.kt#L3) 😄

😄 Yes... But as "I grew up here with them", I'd kind a miss them, although I never saw one of them for months... 😉 

> @chiljamgossow will improve some of that already for 2.5,

👍 

> but overall there are a few more 🚧 sites, and while refactoring that and nice architectures make me happy, other parties may see priorities else where 😉 .

There are other priorities? 😁 

Have a nice night, see you soon! 👋

mtwalli (approved):

Issue #3369, opened 2021-06-07
App Information does not scroll, version info truncated


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ ] Bug is not already reported in another issue
This issue is closely related to https://github.com/corona-warn-app/cwa-app-android/issues/3059 "App Information cut off at bottom on small screen (2)" with Internal Tracking ID: [EXPOSUREAPP-7040](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7040).

This issue is posted separately here due to different environments.

The original issue was https://github.com/corona-warn-app/cwa-app-android/issues/2302.

## Technical details

- Device name: Google Pixel emulator<br>5.0" 1080x1920 420dpi
- Android version: 10 (API 29) or Android 11 (API 30)
- App version: 2.3.2-RC2 & 2.4.0-RC7

## Describe the bug

The App Information menu fails to scroll down in some configurations, cutting off the bottom end of the display. 

Version | Result | Status
--- | --- | ---
2.2.1 | Version displayed | :heavy_check_mark:
2.3.2-RC2 deviceForTestersDebug | Version truncated | :x:
2.4.0-RC7 deviceForTestersDebug | Version truncated | :x:

The issue is sensitive to display size and font size.

The last line of the screenshot following should show `Version: 2.4.0-RC7`.

![2 4 0-scroll-issue](https://user-images.githubusercontent.com/66998419/121063015-171a8c80-c7c6-11eb-85cd-9266252aa82b.jpg)

### Setup
1. (Android) Settings > Display > Advanced
2. Set Font size to Large and Display size to Default

### In CWA
1. Start Screen, tap three-dot symbol
2. Tap "App Information"
3. Try to scroll down to bottom of view

## Expected behavior

Using an API 29 emulator the App Information view should display the complete view including the last line which is the Version of the app. If the view is too large then it should be possible to scroll down until the app version number is visible. 

It should be possible to scroll by tapping, holding and dragging anywhere in the scrolled view, however in this exact scenario it is only possible to scroll by using the area around the version number.

## Investigation

[...]

PR #3098, opened 2021-05-06
Fix app info cut on screen bottom (EXPOSUREAPP-7040)

SamuraiKek:
Fixes an issue where on some phones running older versions of Android, the bottom part of the "App Information" screen would be cut off and unscrollable. 
Issue is descirbed in #3059 and fix was also found by @fynngodau.
Steps to reproduce are quite finicky, best bet is to follow the last comments in #3059. [This](https://github.com/corona-warn-app/cwa-app-android/issues/3059#issuecomment-833459136) is how I've managed to reproduce it.

Thanks @vaubaehn, @fynngodau and @MikeMcC399 for the thorough investigation!

Issue #3059, opened 2021-05-03
App Information cut off at bottom on small screen (2)


vaubaehn:
## Avoid duplicates

* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [ ] Bug is not already reported in another issue - Bug was already reported in #2302, but can not be reproduced with that hardware configuration there, hence was not reopened.

## Technical details

- Device name: LG-G4 (LGH-815)
- Android version: 6.0
- App version: 2.1.2
- ENS version: 18211213000

## Describe the bug

The bottom of the App Information is cut off on a smaller device and cannot be viewed. The app version information/ENS version information is not displayed. The screen cannot be scrolled.

## Steps to reproduce the issue

1. Tap three-dot overflow menu > App Information
2. Try to scroll the view
![Screenshot_2021-05-03-21-15-07](https://user-images.githubusercontent.com/68376758/116921889-e8544800-ac54-11eb-93ad-4db8eccb50ac.png)

## Expected behaviour

It should be possible to view the version number by scrolling down. The version number is only partially displayed on this device.

## Possible Fix

Possibly similar to #2267? Although there were some other changes in #2501 and #2884 .

## Additional context

After installing the update to CWA 2.1.2, I opened CWA by chosing it from 'app selector' (square right down in device's menu). It started directly from the main screen (MainActivity) without displaying the 'New Features' screen... But I don't think that this messed up the scrolling issue in information fragment. The 'New Features' were displayed after closing CWA and reopening from app drawer.

---
Internal Tracking ID: [EXPOSUREAPP-7040](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7040)

MikeMcC399:
@vaubaehn 

I will check if I can reproduce with the Nexus 5 emulator on Android 6. Maybe we can find out why it works on Nexus 5, Android 11, but not on your environment.

vaubaehn:
@MikeMcC399 Thank you anyway. So it seems to be API level related (although I guess @fynngodau should have found out already, when preparing for Android 5 compatibility), or there is a different problem...

fynngodau:
I don't think this is, in general, API related.

[...]

PR #3416, opened 2021-06-10
Fix "App Information" scrolling for smaller devices (EXPOSUREAPP-7765) (COMMUNITY)

MikeMcC399:
### Description

This PR provides a workaround for a scrolling issue of App Information on smaller devices where the App Information almost, but not quite, fits the available display area. In this case part of the top of the version information is displayed at the bottom of the screen and the App Information cannot be scrolled to read the version information properly.

image

Issue https://github.com/corona-warn-app/cwa-app-android/issues/3369 and https://github.com/corona-warn-app/cwa-app-android/issues/3059 describe the issues. 

PR https://github.com/corona-warn-app/cwa-app-android/pull/3098 previously tried to address this issue, however there are still screen and font size combination which do not scroll properly.

The PR adds a `Space` view with height `dimen/spacing_huge` at the bottom of the `fragment_information.xml` layout. In situations where the platform appears to have difficulty deciding whether it has to offer scrolling or not, this extra space takes it out of the grey zone into the situation where it clearly needs to offer scrolling.

Perhaps at some stage we may understand the root cause better and have a better cure, however this workaround seems safe without undesired side-effects and it will help users, who have encountered this issue, in a timely manner.

### Steps to reproduce

- Use Google Pixel emulator (5.0" 1080x1920 420dpi)
- With Android version: 10 (API 29) 

#### Setup
1. (Android) Settings > Display > Advanced
2. Set Font size to Large and Display size to Default

#### In CWA
1. Start Screen, tap three-dot symbol
2. Tap "App Information"
3. Try to scroll to bottom of view

image

Compare outcome with test performed with Build Variant deviceForTestersDebug 2.4.0-RC7. In this build it is only possible to scroll App Information by grabbing the truncated version information and pulling it up. Scrolling by tapping and holding elsewhere in the view does not work.

The API 29 emulator has no Bluetooth, so no ENF version is displayed. Test outcomes will be different on API 30 with a release build, which will show an ENF version and thereby lengthen the App Information view.

Using the Android Studio debugger, I simulated the display with an ENF version, which then looks as follows with the view fully scrolled up:

image

---
Internal Tracking ID: [EXPOSUREAPP-7765](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7765)

### DIFF
--- a/Corona-Warn-App/src/main/res/layout/fragment_information.xml
+++ b/Corona-Warn-App/src/main/res/layout/fragment_information.xml
@@ -122,7 +122,14 @@
                     android:paddingBottom="@dimen/spacing_tiny"
                     tools:text="16000000"
                     tools:visibility="visible" />
+
+                <!-- Workaround for scrolling issue where view is
+                approximately as high as available space-->
+                <Space
+                    android:layout_width="match_parent"
+                    android:layout_height="@dimen/spacing_huge" />
             </LinearLayout>
+
         </ScrollView>
     </androidx.constraintlayout.widget.ConstraintLayout>
 </layout>


SamuraiKek (approved):
Looks like it's working. Hope this fixes it once and for all 👍

mtwalli (approved):
Thank you @MikeMcC399 !

PR #3418, opened 2021-06-11
Support Polish DCCs (EXPOSUREAPP-7776)

d4rken:
Attempt parsing to DateTime if simple Date parsing fails.
Allows more lenient handling of DCCs with different styles of date values (e.g. Sweden, Poland).


BMItr (approved):
nothing to complain!

chris-cwa (approved):
LGTM!

PR #3424, opened 2021-06-11
Missing text for test certificate on negative test result screen (EXPOSUREAPP-7753)

kolyaopahle:
### Description
This PR adds a new step to the bullet point list on both the negative pcr and ra test screens that informs the user about the test certificate associated with the test. It is only shown if the test is actually associated with a certificate and has two states, one for a pending and one for a fetched certificate.

### Steps to reproduce
1. scan a negative ra and pcr test without dgc support
2. check that both test screens look as usual
3. scan a negative ra and pcr test with dgc support
4. admire the newly added bullet point which should transition from pending to fetched state

### Screenshots
<p align="middle">
  
  
  
</p>


vaubaehn:
I'm always a little bit nervous, when the "remove button" at the bottom is that large and quite close to Android's (back) navigation buttons... 😉

kolyaopahle:
> I'm always a little bit nervous, when the "remove button" at the bottom is that large and quite close to Android's (back) navigation buttons... 😉

i quite agree, for now we at least have some kind of confirmation dialog to stop accidental deletion but i will forward the concern to the design team 😄

NataliaLemmerth (approved):
UA approved

vaubaehn:
While I'm really admiring the new bullet point, the new feature 'certificate as proof of negative test' is now not fitting too well to the before displayed text 'Befund negativ'/'Test entfernen' anymore:
If I want to use the test certificate to prove I'm not positive, I would not expect to be asked to remove this test (now) to be able to add a new test ('Bitte entfernen Sie...'). Do you understand what I mean?
This hint could be more open like 'Bevor Sie zu einem späteren Zeitpunkt einen neuen Test hinzufügen können, muss der letzte Test entfernt werden.' or similar. I guess, UI/UX can find a better wording.
What do you think?

harambasicluka:
> While I'm really admiring the new bullet point, the new feature 'certificate as proof of negative test' is now not fitting too well to the before displayed text 'Befund negativ'/'Test entfernen' anymore:
> If I want to use the test certificate to prove I'm not positive, I would not expect to be asked to remove this test (now) to be able to add a new test ('Bitte entfernen Sie...'). Do you understand what I mean? And the large 'remove button' is even strengthening the feeling I should immediately remove the test... By this text- and button design (remove test) is in contrast to intended function (wait for/use test certificate).
> The hint in the text could be more open like 'Bevor Sie zu einem späteren Zeitpunkt einen neuen Test hinzufügen können, muss der letzte Test entfernt werden.' or similar. I guess, UI/UX/UA can find a better wording in a follow-up PR.
> What do you think?

Thanks for your input, I forwarded it to the corresponding colleagues. And as suggested by you we won't handle it in this PR, but maybe in a follow-up PR.

mtwalli (approved):
lgtm

d4rken (approved):

PR #3425, opened 2021-06-11
Fix Screenshot for Play Store (EXPOSUREAPP-7807)

jurajkusnier:
- just renamed Max to Andrea to make everybody happy
### Known issues:
- some strings are still only in German
- translations should arrive on Monday
image


mtwalli (approved):

vaubaehn:
German language -> German time format? ("Test durchgeführt am...")

BMItr (approved):

PR #3426, opened 2021-06-11
Fix certificate LocalDate parsing being too lenient (DEV)

d4rken:
While #3418 made parsing a bit more lenient to accept Polish certificates, it's now a bit too lenient :facepalm:.
Missing month or day elements were assumed to be "start" (e.g. 2021 -> 2021-01-01).
While no such QR codes should be officially generated, better safe than sorry.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/Dcc.kt
```diff
@@ -1,9 +1,14 @@
...
 import com.google.gson.annotations.SerializedName
+import org.joda.time.DateTime
 import org.joda.time.LocalDate
+import org.joda.time.format.DateTimeFormat
+import org.joda.time.format.DateTimeFormatterBuilder
+import org.joda.time.format.ISODateTimeFormat
+import timber.log.Timber
 
-interface Dcc<PayloadType : Dcc.Payload> {
+abstract class Dcc<PayloadType : Dcc.Payload> {
```
chris-cwa, line 11: I'm just curious: What was wrong with being an interface?

d4rken: Interface doesn't allow backing fields, by making it an abstract class I can move the caching for `dateOfBirth` parsing into the `Dcc` class and reuse it for all types of certificates.

vaubaehn:
I found an updated schema here: https://github.com/ehn-dcc-development/ehn-dcc-schema
Looks like all related IT will need to adjust. Parsing with regex could be good to catch all future "adjustments".

mtwalli (approved):
Lgtm

BMItr (approved):
👌 safe

vaubaehn:
@d4rken 
> Missing month or day elements were assumed to be "start" (e.g. 2021 -> 2021-01-01).
> While no such QR codes should be officially generated, better safe than sorry.

As you can see here: https://github.com/ehn-dcc-development/ehn-dcc-schema/issues/78 , such QR codes are already issued, and NL won't probably the only country.

FYI: @mlenkeit

Issue #3354, opened 2021-06-02
duplicate rapid test registration - lost the test


Armin83:
I registered my test. (picture one)
I scanned the qr code with cwa app successfully. I had two test. the new rapid test and an old pcr test.
no result for 18 minutes ;)

![1](https://user-images.githubusercontent.com/19192087/120535447-43f42b80-c3e3-11eb-86ad-56eaa7f0485f.jpg)

Later i got a sms. you can see the result in cwa click the link (picture two)
I followed the link. I got the message test already registered.... i lost the test registration from the rapid test.
Only the old pcr test is still present.

![2](https://user-images.githubusercontent.com/19192087/120535441-435b9500-c3e3-11eb-806c-185e25beaf6a.jpg)

Should it work in this way ? Of course the test registration should only possible on one device, one time... but why it killed the old one with same id. It should not kill the same test registration. I heard from friend that it doesn't work for him. The result was uninstall the cwa app....

It's not the best behavior. 

---
Internal Tracking ID: [EXPOSUREAPP-7447](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7447)

Ein-Tim:
Thank you for reporting this!

Just that we all get you right, you scanned a QR-Code linked to a PCR test, afterwards you scanned a QR-Code linked to a rapid test. Then you got a SMS which contained the link to register the rapid test result which you'd already added to the Corona-Warn-App when you scanned the QR-Code. 
You clicked on the link in the SMS, the Corona-Warn-App opened, you stepped through the screens in CWA, you did see the screen "Hinweis - Sie haben bereits einen Schnelltest registriert", clicked on "Weiter" and did get the pop up "QR-Code ist ungültig". You clicked on "Abbrechen" and the quick test was removed and only the PCR test was left?

dsarkar:
@Armin83 Thanks for your report. 

Basically, as @Ein-Tim already asked:
I understand form your report that your rapid antigen test (RAT) disappeared. 
First you scanned RAT QR code, then you received as some point SMS where you pressed the link and with this the RAT was deleted. Please, confirm. I can confirm this behaviour on Android. iOS will not delete the test. Developers are informed about this issue.

Internal Tracking ID: [EXPOSUREAPP-7447](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7447)



---
Corona-Warn-App Open Source Team

dsarkar:
[...]

PR #3439, opened 2021-06-14
Prevent test replacement if new test is already redeemed (EXPOSUREAPP-7447)

d4rken:
Issue:
User scans QR-Code successfully, later receives a link for the same QR-Code and clicks it.
The app will show the warning that the current test will be replaced.
The user confirms it, the new test registration fails (because it was already consumed), but the old test was already removed.
The user is left with no test :frowning:.

This PR:
Introduces a way to register a test, but to abort registration and keep the previous test if desired, by offering a `precondition` and `postcondition` for `registerTest`.
This is implement via `SubmissionRepository` as `registerTest` and `tryReplaceTest`.

To align errorhandling a new common class `TestRegistrationStateProcessor` handles the registration and error mapping for multiple screens.

So when scanning a new test of an already registered type, and the new test registration fails because it's already redeemed, there should still be an error dialog, but the previous test should not be removed.

Testing:
* Scan the same test twice
* Scan a qrcode, then send yourself the link for that qrcode and click it
* Normal test registration for PCR and RA should still work as before.


MikeMcC399:
Link to #3354 for automatic PR status update in issue.

MikeMcC399:
There is a typo in the title
EXPOSURAPP-7447
should be
[EXPOSUREAPP-7447](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7447).
Would be good to get it corrected, because it will ultimately show up in the release log of fixes applied.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
```diff
@@ -76,56 +70,36 @@ class SubmissionConsentFragment : Fragment(R.layout.fragment_submission_consent)
...
-                        }
-                    }
+            when (state) {
+                State.Idle,
+                State.Working -> {
+                    // Handled above
+                }
+                is State.Error -> {
+                    state.getDialogBuilder(requireContext()).show()
+                    popBackStack()
```
mtwalli, line 90: is it intended to show a dialog and navigate back ?

mtwalli (commented):
Probably not related to the changes in here, but registering a PENDING RAT test and then turns into Negative does not have a DCC entry in the list.

mtwalli (approved):
The issue here is fixed , we need to check the DCC certificate for pending negative tests

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
mtwalli: I checked the old code and it had the same logic  🤷

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/coronatest/ui/CoronaTestTestFragmentViewModel.kt
```diff
@@ -28,17 +24,17 @@ class CoronaTestTestFragmentViewModel @AssistedInject constructor(
 ) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
 
     val errorEvents = SingleLiveEvent<Throwable>()
-    val pcrtState = coronaTestRepository.latestPCRT.map {
-        PCRTState(
-            coronaTest = it
-        )
-    }.asLiveData(context = dispatcherProvider.Default)
+    val pcrtState = coronaTestRepository.coronaTests
```
BMItter, line 27: ↓ call fitler directly?

BMItr (approved):
lgtm! 👌 
tested on oneplus 8 device

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/viewmodel/CWAViewModel.kt
```diff
@@ -27,6 +28,8 @@ abstract class CWAViewModel constructor(
         Timber.tag(tag).v("Initialized")
     }
 
+    fun <T> Flow<T>.asLiveData2() = asLiveData(context = dispatcherProvider.Default)
```
BMItter, line 31: maybe a more descriptive name here would be nice..

PR #3443, opened 2021-06-14
Travel Notice Custom View (EXPOSUREAPP-7844)

axelherbstreith:
Adding a custom view for the travel notice that can be used for every type of certificate


mtwalli (approved):

SamuraiKek (approved):
Nice one. LGTM

PR #3449, opened 2021-06-15
Fix test result code instead of alias printed in test certificates detail screen (EXPOSUREAPP-7781)

jurajkusnier:
- Should work on WRU and other endpoints
- May not work on local cli


BMItter:
@jurajkusnier 
general question: why did you delete the valueSetRepo from the other viewmodel?
Wasn't it intended to be in both viewmodels?

jurajkusnier:
> @jurajkusnier
> general question: why did you delete the valueSetRepo from the other viewmodel?
> Wasn't it intended to be in both viewmodels?

I moved just the trigger for updating value sets, to be updated earlier (in the certificates screen). Once the values are updated they are stored in shared preferences so it's not necessary to call this trigger from the second screen.

BMItter:
> > @jurajkusnier
> > general question: why did you delete the valueSetRepo from the other viewmodel?
> > Wasn't it intended to be in both viewmodels?
> 
> I moved just the trigger for updating value sets, to be updated earlier (in the certificates screen). Once the values are updated they are stored in shared preferences so it's not necessary to call this trigger from the second screen.

Yes I got it.
Could you please check whether the Viewmodel is called BEFORE the testCertificat && BEFORE the Vaccination Part?
If its not the case we will move the bug around.

jurajkusnier:
> > > @jurajkusnier
> > > general question: why did you delete the valueSetRepo from the other viewmodel?
> > > Wasn't it intended to be in both viewmodels?
> > 
> > 
> > I moved just the trigger for updating value sets, to be updated earlier (in the certificates screen). Once the values are updated they are stored in shared preferences so it's not necessary to call this trigger from the second screen.
> 
> Yes I got it.
> Could you please check whether the Viewmodel is called BEFORE the testCertificat && BEFORE the Vaccination Part?
> If its not the case we will move the bug around.

Yes, as we moved Vaccinations to the Certificates tab, I don't see any way how to access Vaccinations without opening the Certificates tab at least once.

BMItr (dismissed):

jurajkusnier:
> > > > @jurajkusnier
> > > > general question: why did you delete the valueSetRepo from the other viewmodel?
> > > > Wasn't it intended to be in both viewmodels?
> > > 
> > > 
> > > I moved just the trigger for updating value sets, to be updated earlier (in the certificates screen). Once the values are updated they are stored in shared preferences so it's not necessary to call this trigger from the second screen.
> > 
> > 
> > Yes I got it.
> > Could you please check whether the Viewmodel is called BEFORE the testCertificat && BEFORE the Vaccination Part?
> > If its not the case we will move the bug around.
> 
> Yes, as we moved Vaccinations to the Certificates tab, I don't see any way how to access Vaccinations without opening the Certificates tab at least once.

I reverted changes in `VaccinationListViewModel` after discussion regarding concerns for potential future changes in screen flow

BMItr (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/CertificatesViewModel.kt
```diff
@@ -19,17 +21,25 @@ import de.rki.coronawarnapp.covidcertificate.vaccination.ui.cards.ImmuneVaccinat
...
 class CertificatesViewModel @AssistedInject constructor(
     vaccinationRepository: VaccinationRepository,
+    valueSetsRepository: ValueSetsRepository,
+    @AppContext context: Context,
     private val vaccinationSettings: VaccinationSettings,
     private val testCertificateRepository: TestCertificateRepository
 ) : CWAViewModel() {
 
+    init {
+        valueSetsRepository.triggerUpdateValueSet(languageCode = context.getLocale())
```
mtwalli, line 40: does it have to be `context.getLocale()` , can we use `Locale.getDefault()` as everywhere else in the App?

jurajkusnier: I guess we can use both variants. I used the same as we are using in the VaccinationListViewModel in the same method call.

mtwalli (approved):

PR #3451, opened 2021-06-15
Fix Test Certificate with wrong timestamp (EXPOSUREAPP-7822)

jurajkusnier:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/cards/CovidTestCertificateCard.kt
```diff
@@ -26,8 +27,8 @@ class CovidTestCertificateCard(parent: ViewGroup) :
         val curItem = payloads.filterIsInstance<Item>().singleOrNull() ?: item
         testTime.text = context.getString(
             R.string.test_certificate_time,
-            curItem.testDate.toShortDayFormat(),
-            curItem.testDate.toShortTimeFormat(),
+            item.testDate.toString(DATE_FORMAT),
+            item.testDate.toString(shortTime)
```
mtwalli, line 31: is not toShortTimeFormat doing the same?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/cards/CovidTestCertificateCard.kt
```diff
@@ -26,8 +27,8 @@ class CovidTestCertificateCard(parent: ViewGroup) :
         val curItem = payloads.filterIsInstance<Item>().singleOrNull() ?: item
         testTime.text = context.getString(
             R.string.test_certificate_time,
-            curItem.testDate.toShortDayFormat(),
-            curItem.testDate.toShortTimeFormat(),
+            item.testDate.toString(DATE_FORMAT),
```
mtwalli, line 30: `toDayFromat()` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/cards/CovidTestCertificateErrorCard.kt
```diff
@@ -28,8 +29,8 @@ class CovidTestCertificateErrorCard(parent: ViewGroup) :
 
         testTime.text = context.getString(
             R.string.test_certificate_time,
-            item.testDate.toShortDayFormat(),
-            item.testDate.toShortTimeFormat(),
+            item.testDate.toString(DATE_FORMAT),
+            item.testDate.toString(shortTime)
```
mtwalli, line 33: same here ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/cards/CovidTestCertificateCard.kt
jurajkusnier: `toShortTimeFormat()` is implemented for `Instant` (which returns time in the wrong timezone) and `LocalDate` (hm... which actually doesn't make sense since LocalDate doesn't hold time info).
But maybe it makes sense to create an extension function for `DateTime` or `LocalDateTime`. I used implementation from `RATResultNegativeFragment` but I can change it also there.

mtwalli: Yes, please if you can unify this using an extension function that would be great to use it later in other places

mtwalli (approved):
Thanks!

BMItr (approved):

PR #3457, opened 2021-06-15
Fix missing DCC consent navigation for clicked RA test links (EXPOSUREAPP-7875) 

d4rken:
Fix clicked corona-test links, with DCC support, not navigating to DCC consent screen, so it behaves the same as it when it is scanned via camera.

Previously the navigation did not allow you to reach the consent screen for test certificates, when you came from a clicked link, instead of a qrcode.


mtwalli (approved):
Lgtm , Tested on S8

chiljamgossow (approved):

PR #3463, opened 2021-06-16
Recovery certificate repository (EXPOSUERAPP-7617)

chris-cwa:
This PR covers storage of recovery certificates. On review please check wether the correct identifiers are being compared.

- [ ] unit testing (in follow-up PR)
- [x] use container id from https://github.com/corona-warn-app/cwa-app-android/pull/3504


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+    @BaseGson val baseGson: Gson,
+) {
+
+    private val prefs by lazy {
+        context.getSharedPreferences("recovery_localdata", Context.MODE_PRIVATE)
+    }
+
+    private val gson by lazy {
+        baseGson.newBuilder().create()
+    }
```
mtwalli, line 25: why do we need this? can't we just use `baseGson` directly ?

mtwalli (commented):
Lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
```diff
@@ -0,0 +1,63 @@
...
+            prefs.edit {
+                prefs.all.keys.filter { it.startsWith(PKEY_RECOVERY_CERT_PREFIX) }.forEach {
+                    Timber.tag(TAG).v("Removing data for %s", it)
+                    remove(it)
+                }
+                value.forEach {
+                    val raw = gson.toJson(it)
+                    val identifier = it.identifier
+                    Timber.tag(TAG).v("Storing recovery certificate %s -> %s", identifier, raw)
+                    putString("$PKEY_RECOVERY_CERT_PREFIX$identifier", raw)
```
d4rken, line 54: What will `identifier` be? A UUID?
Alternative would be store them not individually but as a `Set<>`, I'm not quite sure what would be better in this case :thinking:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -20,23 +29,64 @@ class RecoveryCertificateRepository @Inject constructor(
...
+                    qrCodeExtractor = qrCodeExtractor
+                )
+            }
+            .toSet()
+            .also { Timber.tag(TAG).v("Restored recovery certificate data: %s", it) }
+    }
+
+    init {
+        internalData.data
+            .onStart { Timber.tag(TAG).d("Observing test data.") }
```
d4rken, line 53: `test` typo?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -20,23 +29,64 @@ class RecoveryCertificateRepository @Inject constructor(
...
+            .catch {
+                it.reportProblem(TAG, "Failed to snapshot recovery certificate data to storage.")
+                throw it
+            }
+            .launchIn(appScope + dispatcherProvider.IO)
+    }
+
+    val certificates: Flow<Set<RecoveryCertificateWrapper>> = combine(
+        internalData.data,
+        valueSetsRepository.latestVaccinationValueSets // TODO use corrects sets
```
d4rken, line 67: According to @BMItter there will be no valuesets for recovery certificates, at the moment.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -20,23 +29,64 @@ class RecoveryCertificateRepository @Inject constructor(
...
         Timber.tag(TAG).d("requestCertificate(qrCode=%s)", qrCode)
         throw NotImplementedError()
     }
 
     suspend fun deleteCertificate(identifier: RecoveryCertificateIdentifier) {
-        Timber.tag(TAG).d("deleteTestCertificate(identifier=%s)", identifier)
-        throw NotImplementedError()
+        Timber.tag(TAG).d("deleteCertificate(identifier=%s)", identifier)
+        internalData.updateBlocking {
+            mapNotNull { if (it.certificateId == identifier) null else it }.toSet()
```
d4rken, line 80: `certificateId` is different from `identifier`

* `certificateId` is the ID field in the scanned certificate JSON
* `identifier` is a unique ID we come up with

We could make them the same, but we didn't do this for the other certificates as we would have to rely on them being unique (which I don't think we should do, as we have seen, countries do all sorts of different things that are not according to schema specs...).

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
vaubaehn: They just don't care about any vaccine or test detail ... currently. Just stating who issued where, and how long valid:
https://github.com/ehn-dcc-development/ehn-dcc-schema/blob/a603410d760fefc9073931c8c807759d9714c136/DCC.combined-schema.json#L267

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
chris-cwa: thank you!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
chris-cwa: removed it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
chris-cwa: introduced uuid for client side identification

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,88 @@
...
+    }
+
+    val certificates: Flow<Set<RecoveryCertificateWrapper>> =
+        internalData.data.transform { set ->
+            set.map { RecoveryCertificateWrapper(null, it) }.toSet()
+        }
 
     suspend fun requestCertificate(qrCode: RecoveryCertificateQRCode): RecoveryCertificateContainer {
         Timber.tag(TAG).d("requestCertificate(qrCode=%s)", qrCode)
         throw NotImplementedError()
```
mtwalli, line 70: is that part of this PR ? or another one ?

chris-cwa: pr was draft only

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/StoredRecoveryCertificateData.kt
```diff
@@ -7,7 +7,18 @@ data class StoredRecoveryCertificateData(
...
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as StoredRecoveryCertificateData
+
+        if (identifier != other.identifier) return false
+
+        return true
+    }
+}
```
d4rken, line 21: Why the custom `equals` here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+        set(value) {
+            Timber.tag(TAG).d("recoveryCertificates - save(%s)", value)
+            prefs.edit()
+                .putStringSet(
+                    PKEY_RECOVERY_CERT,
+                    value.map { data ->
+                        gson.toJson(data).also {
+                            Timber.tag(TAG).v("Storing recovery certificate %s -> %s", data.identifier, it)
+                        }
+                    }.toSet()
```
d4rken, line 42: What do you think about storing the whole set as json vs using `putStringSet`? `putStringSet` has a spotty history, with a few footguns because it doesn't return a copy of the internal set when using `get`.

e.g.
```kotlin
      val type =  object : TypeToken<Set<StoredRecoveryCertificateData>>() {}.type
      gson.toJson<Set<StoredRecoveryCertificateData>>(data, type)
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+                )
+                .apply()
```
d4rken, line 44: Could use `edit { }` instead of `edit().....apply()`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
```diff
@@ -0,0 +1,51 @@
...
+                gson.fromJson<StoredRecoveryCertificateData>(it).also { data ->
+                    Timber.tag(TAG).v("recovery certificate loaded: %s", data)
+                    requireNotNull(data.identifier)
+                    requireNotNull(data.registeredAt)
+                }
+            }?.toSet() ?: emptySet()
+        }
+        set(value) {
+            Timber.tag(TAG).d("recoveryCertificates - save(%s)", value)
+            prefs.edit()
```
d4rken, line 35: Other storages we have remove the key when the set is empty, here we write an empty set, should we align?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/DuplicateRecoveryCertificateException.kt
```diff
@@ -0,0 +1,5 @@
+package de.rki.coronawarnapp.covidcertificate.recovery.core
+
+class DuplicateRecoveryCertificateException(
+    message: String
+) : IllegalArgumentException(message)
```
d4rken, line 5: Please use the existing:
```kotlin
InvalidRecoveryCertificateException(
    InvalidHealthCertificateException.ErrorCode.ALREADY_REGISTERED
)
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,108 @@
...
+        internalData.data.transform { set ->
+            set.map { RecoveryCertificateWrapper(null, it) }.toSet()
+        }
+
+    @Throws(DuplicateRecoveryCertificateException::class)
     suspend fun requestCertificate(qrCode: RecoveryCertificateQRCode): RecoveryCertificateContainer {
         Timber.tag(TAG).d("requestCertificate(qrCode=%s)", qrCode)
-        throw NotImplementedError()
+        qrCodeExtractor.extract(qrCode.qrCode).toContainer().apply {
+            if (internalData.data.last().contains(this)) {
```
d4rken, line 75: This hangs forever. `last()` waits for the terminal operator of a flow, but `internalData.data` never completes in normal operation. You likely want `first()` here to get the newest value.

d4rken: But then we are left with a racecondition. Between performing the check here, and updating the internal data with `updateBlocking`, the data may have changed and we could be adding a duplicate certificate. Unlikely, but still...

Place the check inside `updateBlocking { currentData -> .... }` and perform it against `currentData` to prevent the racecondition.

d4rken (dismissed):
Right approach, few bugs.

Please add some tests for the duplicate certificate exception and the json storage format.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/StoredRecoveryCertificateData.kt
chris-cwa: because there was no container id yet :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/DuplicateRecoveryCertificateException.kt
chris-cwa: using it now

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateStorage.kt
chris-cwa: created a dto for json

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,110 @@
...
+    val certificates: Flow<Set<RecoveryCertificateWrapper>> =
+        internalData.data.transform { set ->
+            set.map { RecoveryCertificateWrapper(null, it) }.toSet()
+        }
+
+    @Throws(InvalidRecoveryCertificateException::class)
     suspend fun registerCertificate(qrCode: RecoveryCertificateQRCode): RecoveryCertificateContainer {
         Timber.tag(TAG).d("registerCertificate(qrCode=%s)", qrCode)
-        throw NotImplementedError()
+        return qrCodeExtractor.extract(qrCode.qrCode).toContainer().apply {
```
d4rken, line 75: I don't think we need to extract the qrCode again here, `qrCodeExtractor.extract(qrCode.qrCode)` basically returns `qrCode: RecoveryCertificateQRCode` which we already have as function parameter.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,110 @@
...
+            internalData.updateBlocking {
+                if (any { it.containerId == containerId }) {
+                    throw InvalidRecoveryCertificateException(InvalidHealthCertificateException.ErrorCode.ALREADY_REGISTERED)
+                }
+                plus(this)
+            }
```
d4rken, line 81: I think it's the wrong `this`, `this` in that scope might be the current data snapshot from `updateBlocking`?

```kotlin
fun updateBlocking(
    action: suspend Set<RecoveryCertificateContainer>.() -> Set<RecoveryCertificateContainer>
): Set<RecoveryCertificateContainer>
```

`plus(this)` would be set + set, duplicating the original data and ignoring the new value.
So no compiler error because `SET.plus(SET)` is also valid.

You likely meant `this@apply`, but maybe we can write this a little bit different so future devs don't accidentally break this without noticing.

chris-cwa: put the new container into new var

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
chris-cwa: thank you

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,111 @@
...
+            plus(newContainer)
+        }
+        return newContainer
     }
 
-    suspend fun deleteCertificate(containerId: RecoveryCertificateContainerId): RecoveryCertificateContainer? {
+    private fun RecoveryCertificateQRCode.toContainer() = RecoveryCertificateContainer(
+        data = StoredRecoveryCertificateData(
+            identifier = uniqueCertificateIdentifier,
```
d4rken, line 88: If the identifier is the `uniqueCertificateIdentifier` from the certificate, then we don't need to store it extra.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,111 @@
...
+            set.map { RecoveryCertificateWrapper(null, it) }.toSet()
+        }
+
+    @Throws(InvalidRecoveryCertificateException::class)
     suspend fun registerCertificate(qrCode: RecoveryCertificateQRCode): RecoveryCertificateContainer {
         Timber.tag(TAG).d("registerCertificate(qrCode=%s)", qrCode)
-        throw NotImplementedError()
+        val newContainer = qrCode.toContainer()
+        internalData.updateBlocking {
+            if (any { it.containerId == newContainer.containerId }) {
```
d4rken, line 76: Prefer checking here for `uniqueCertificateIdentifier` explicitly, as what `containerId` is backed by may change over the time.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,111 @@
...
+            plus(newContainer)
+        }
+        return newContainer
     }
 
-    suspend fun deleteCertificate(containerId: RecoveryCertificateContainerId): RecoveryCertificateContainer? {
+    private fun RecoveryCertificateQRCode.toContainer() = RecoveryCertificateContainer(
+        data = StoredRecoveryCertificateData(
+            identifier = uniqueCertificateIdentifier,
+            registeredAt = data.header.issuedAt,
```
d4rken, line 89: `registeredAt` (in other places in the app) refers to the time in the app when the code was scanned. We should use a similar timestamp or rename the property (but if we rename the property, we don't need to store the headerfield twice, as we store the whole qrcode including header).

harambasicluka:
As @d4rken is on vacation someone else has to take over the review.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
chiljamgossow: What is the field registeredAt needed for? I cannot find any usage

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/DuplicateRecoveryCertificateException.kt
chiljamgossow: Please delete the exception

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -1,42 +1,109 @@
...
+import kotlinx.coroutines.flow.transform
+import kotlinx.coroutines.plus
 import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class RecoveryCertificateRepository @Inject constructor(
     @AppScope private val appScope: CoroutineScope,
-    private val dispatcherProvider: DispatcherProvider,
```
chiljamgossow, line 20: Why is it not private any more?

chris-cwa: i removed `val` also - so it's not a class member anymore - it is only used in `init`

chiljamgossow (dismissed):
Let's merge and test the whole flow

mtwalli:
@chris-cwa  I don't see any certificates in your branch.  Comparing to `2.5` where Test and vaccination certificates are showing up after scanning. Here I can not see them anymore , Could you please have a look ?

mtwalli (approved):
We can merge , I just pushed a fix for flow emission

chiljamgossow (approved):

PR #3466, opened 2021-06-16
Fix DCC request screen abort dialog actions (EXPOSUREAPP-7872)

d4rken:
Canceling the dialog (negative action) stays on screen.
Confirming the dialog (positive action) aborts the registration and sends the user back to the home screen.

Strings were changed after initial implementation. Will adjust the German ones here, depending on whether there will be an additional translation delivery, other languages will be delivered with 2.5.


NataliaLemmerth (approved):
UA approved

BMItr (approved):
lgtm

mtwalli (approved):

PR #3469, opened 2021-06-16
Add missing BottomNavigationView badge for retrieved test certificates (EXPOSUREAPP-7851)

d4rken:
Screenshot from 2021-06-16 13-41-58


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/CertificatesViewModel.kt
```diff
@@ -42,6 +45,18 @@ class CertificatesViewModel @AssistedInject constructor(
...
 
+    val markNewCertsAsSeen = testCertificateRepository.certificates
+        .onEach { wrappers ->
+            wrappers
+                .filter { !it.seenByUser && !it.isCertificateRetrievalPending }
+                .forEach {
+                    testCertificateRepository.markCertificateAsSeenByUser(it.identifier)
+                }
+        }
+        .map { }
```
BMItter, line 56: map { } -> { }

d4rken: What do you mean by that?
We just need to be subscribed, we don't want to return a specific value, to achieve that: `.map { Unit }` -> `.map { }` -> `Unit`, is there a better way?

BMItter: without the unit it seemed to me like a forgotten leftover

BMItr (approved):

chiljamgossow (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/CertificatesViewModel.kt
d4rken: I'll make it better in another PR :heart:

PR #3473, opened 2021-06-16
Partial PersonCertificatesProvider implementation (EXPOSUREAPP-7685)

d4rken:
Initial draft, combination should work and allow it to be integrated with other screens.
Actual sorting and implementation will follow in extra PRs.


SamuraiKek (approved):
Nice! Thanks for this!

mtwalli (approved):
Lgtm

PR #3474, opened 2021-06-16
Cannot obtain test certificate after deleting pending/failed test certificate (EXPOSUREAPP-7763)

kolyaopahle:
### Description
This PR changes a bit of logic during certifiacte creation where currently one failed certificate creation would result in the flow completing and no more certs being created untill the app is restarted.

### Steps to reproduce
1. this is kind of hard to reproduce as `de.rki.coronawarnapp.covidcertificate.test.core.TestCertificateRepository#requestCertificate` would have to throw an error
2. as this should not happen during normal operation i recommend forcing an exception throw in that function
3. register a test
4. check the certificate screen for a failed certificate
5. check if the flow is still active when scanning new certificates (note that they will never appear due to the thrown error)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalScheduler.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalScheduler.kt
@@ -14,7 +14,6 @@ import de.rki.coronawarnapp.util.device.ForegroundState
 import de.rki.coronawarnapp.util.flow.combine
 import de.rki.coronawarnapp.worker.BackgroundConstants
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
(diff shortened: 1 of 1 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalScheduler.kt
```diff
@@ -69,21 +68,27 @@ class TestCertificateRetrievalScheduler @Inject constructor(
...
             .launchIn(appScope)
 
         // For each change to the set of existing certificates, check if we need to refresh/load data
         refreshTrigger
             .onEach { checkCerts ->
-                Timber.tag(TAG).d("State change: checkCerts=$checkCerts")
-                if (checkCerts) scheduleWorker()
+                try {
+                    Timber.tag(TAG).d("State change: checkCerts=$checkCerts")
+                    if (checkCerts) scheduleWorker()
```
d4rken, line 87: I don't think this could reasonably throw an exception, but catching it doesn't hurt either.

d4rken (approved):
lgtm

jurajkusnier (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalScheduler.kt
kolyaopahle: i agree, i did it here for consistency with the other occasions, but we could remove it as well.

PR #3484, opened 2021-06-17
Register CoronaTest-Certificate directly (EXPOSUREAPP-7974)

d4rken:
Adds 

```kotlin
// TestCertificateRepository
suspend fun registerTestCertificate(qrCode: TestCertificateQRCode): TestCertificateContainer
```

Previously it was only possible to register "for a" test certificate by registering a corona test with DCC support.
Now it's possible to take a `TestCertificateQrCode` and add it directly.

Part of qrcode scanning refactoring, and support for "family certificates".
Allows @chiljamgossow to complete wiring for scanning all types of qrcodes.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/types/BaseTestCertificateData.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import de.rki.coronawarnapp.covidcertificate.test.core.storage.TestCertificateIdentifier
+import org.joda.time.Instant
+
+sealed class BaseTestCertificateData {
+    abstract val identifier: TestCertificateIdentifier
+    abstract val registeredAt: Instant
+    abstract val certificateReceivedAt: Instant?
+    abstract val testCertificateQrCode: String?
+}
```
d4rken, line 11: Shared/common properties for test certificates.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/types/GenericTestCertificateData.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    @SerializedName("registeredAt")
+    override val registeredAt: Instant,
+
+    @SerializedName("certificateReceivedAt")
+    override val certificateReceivedAt: Instant? = null,
+
+    @SerializedName("testCertificateQrCode")
+    override val testCertificateQrCode: String? = null,
+) : ScannedTestCertificate() {
```
d4rken, line 18: Test certificates that were added via QR-Code.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/types/RetrievedTestCertificate.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    abstract val encryptedDataEncryptionkey: ByteString?
+    abstract val encryptedDccCose: ByteString?
+
+    abstract val labId: String?
+    abstract val certificateSeenByUser: Boolean
+
+    val isPublicKeyRegistered: Boolean
+        get() = publicKeyRegisteredAt != null
+}
```
d4rken, line 24: TestCertificates that are sourced by registering a corona test in the CWA.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateRepository.kt
```diff
@@ -98,7 +100,13 @@ class TestCertificateRepository @Inject constructor(
...
-            if (values.any { it.registrationToken == test.registrationToken }) {
+
+            val matchesExisting = values
+                .map { it.data }
+                .filterIsInstance<RetrievedTestCertificate>()
+                .any { it.registrationToken == test.registrationToken }
+
+            if (matchesExisting) {
                 Timber.tag(TAG).e("Certificate entry already exists for %s", test.identifier)
                 throw IllegalArgumentException("A certificate was already created for this ${test.identifier}")
```
chiljamgossow, line 111: InvalidTestCertificateException should be used here

chiljamgossow: I tried registering the same test certificate twice and got 'Collection contains more than one matching element'. Not sure where it is thrown

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateRepository.kt
```diff
@@ -133,7 +141,38 @@ class TestCertificateRepository @Inject constructor(
...
+                qrCodeExtractor = qrCodeExtractor,
+            )
+            Timber.tag(TAG).d("Adding test certificate entry: %s", container)
+            mutate { this[container.identifier] = container }
+        }
+
+        // We just registered it, it MUST be available.
+        return updatedData.values
+            .single { it.certificateId == qrCode.uniqueCertificateIdentifier }
     }
```
chiljamgossow, line 176: Maybe here?

chiljamgossow (commented):
Thanks for adding the funktionality 🥰 
I tested by merging 2.5, and the test certificate was added.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateRepository.kt
d4rken: Ah, good catch. I'm not checking for existing certificates for scanned certs, only when retrieved.
Will fix.

d4rken: Fixed.

mtwalli:
There is inconsistency between certificate types and the repos,
Currently `CwaCovidCertificate` exposes `certificateId` which is fine in case of `VaccinationCertificate` and  its repo which is using certificateId internally to handle certificate logic such as deletion. 

But this is not the case in `TestCertificate` its `certificateId` is not what is called identifier of alias `TestCertificateIdentifier`

and the repo is using this identifier internally to handle the logic , I would prefer to stick to `certificateId` in all types and `RecoveryCertificate` as well

@chris-cwa are you doing the same in `RecoveryCertificateRepository` ?

chris-cwa:
> There is inconsistency between certificate types and the repos,
> Currently `CwaCovidCertificate` exposes `certificateId` which is fine in case of `VaccinationCertificate` and its repo which is using certificateId internally to handle certificate logic such as deletion.
> 
> But this is not the case in `TestCertificate` its `certificateId` is not what is called identifier of alias `TestCertificateIdentifier`
> 
> and the repo is using this identifier internally to handle the logic , I would prefer to stick to `certificateId` in all types and `RecoveryCertificate` as well
> 
> @chris-cwa are you doing the same in `RecoveryCertificateRepository` ?

in recovery certificate i check for duplicate `certificateId` when adding instances to the repo, so `certificateId` will suffice on delete.

d4rken:
@mtwalli Interesting point. It's not nice, I agree. The issue is that for test certificates, we don't have the `uniqueCertificateId` yet, when the certificate has not been retrieved yet, and the "pending card" still needs to be able to delete it.

I will look into better options and provide an extra PR for this.

mtwalli (approved):
Ok, let us merge it and improve it in another PR

PR #3490, opened 2021-06-17
Fix `labId` not being stored, when retrieved during polling (EXPOSUREAPP-7870)

d4rken:
`labId` may be null when retrieved on initial test registration.
If we then later receive a negative test result, the response may contain a non-null lab-ID, which we then ignored.
This lead to `TestCertificateRepository` not attempting to register a publickey and request a certificate.

For testing:
Test with and without this PR by scanning a pending test and changing it later to negative.
Check the logs for `labId=`. Without this PR the stored `labId` for this case is always null.


jurajkusnier (approved):
LGTM

kolyaopahle (approved):
LGTM 👍

Issue #3093, opened 2021-05-06
"Risiko-Ermittlung gestoppt" though it had not been turned off


kereng5:
## Describe the bug

![gestoppt_5168](https://user-images.githubusercontent.com/69790073/117271920-f4115b80-ae5a-11eb-9100-23badf84c35d.png)
I did not stop the exposure check, and it was active at 2:46 am. So who stopped it and why? 

One more small issue: When I click "Aktivieren" to reactivate the exposure check, switch the slider to the right and return to the previous screen by the "back triangle", it still shoes "Risiko-Ermittlung gestoppt", and the slider is left again. It works if I switch the slider to the right, close the CWA and start it new.

## Steps to reproduce the issue

I just start the phone in the morning and read "Risiko-Ermittlung gestoppt" in the CWA.
It happened today and on the day before yesterday, but not yesterday.

## Technical details

    Mobile device: Motorola G3
    Android version: 6.0.1
    CWA version 2.1.2
    ENF 18211213000
    no "Priorisierte Hintergrundaktivität"

---
Internal Tracking ID: [EXPOSUREAPP-7095](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7095)

dsarkar:
@kereng5 Thanks for the report. We getting since version 2.1 several of these reports. 
- Can you confirm that you have a stable internet connection?
- Can you confirm that this issue araised only after updating to CWA version 2.1

DerVogel2020:
This is similar to #2880 .

dsarkar:
Internal Tracking ID: [EXPOSUREAPP-7095](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7095)

vaubaehn:
Hi @DerVogel2020 and all, it's not similar to #2880. In my case, a different error occured ("Risiko-Überprüfung fehlgeschlagen" vs. "Risiko-Ermittlung gestoppt"), and contact tracing ("the switch") was still enabled.

I wonder, if this is a problem in ENS which is (correctly) reflected in CWA? (ENS got a hick up, and turned itself off, or something).
@kereng5 how is the memory situation of your internal storage? How much free space is in your internal memory?

MikeMcC399:
@kereng5 
> ENF 18211213000

It may be worth checking if there is an update to the Exposure Notifications System for your device, in case it is an issue with this part of the system.

On an Android 8 device I can do:
[...]

PR #3502, opened 2021-06-18
Increase resilience when checking whether ENF tracing is enabled (EXPOSUREAPP-7095)

d4rken:
As far as I can tell from the logs related to #3093, the underlying issue is that the ENF status check times out.
While we can't change the ENF behavior and it's unclear why the timeouts happen, we can improve the CWA's behavior and offer slightly better behavior.

Instead of a single check that waits for Google's timeout (2min), we do a quick 1-2 check. I benchmarked the check duration on our favorite low-end devices, and it usually takes around 25ms to check the status, outliers are around 500-600ms. We do a quick check with 1 second timeout that in 99% of cases should be the only check that is needed, and if that times-out, we perform another check with timeout up to the ENF 2min timeout. If the "request gets stuck" or something similar on the ENF side, the second request may go through.

We are also now catching the ENF timeout exception and prevent the polling flow from canceling. We will `emit(false)` but then just poll again.

The issue may look worse in some cases than it actually is. If the user never disabled tracing, tracing is still enabled, even if the `isEnabled` check has a timeout, but because the polling was canceled on timeout, the UI would have shown the last state in some cases (i.e. `isEnabled=false` due to timeout).

Besides this, I think we could only improve this on our end further if we introduce a third UI state:  ON/OFF + UNKNOWN. Not sure if that effort is warranted.


mtwalli (approved):
Lgtm

BMItr (approved):
👌

vaubaehn:
Hi, I hope it's ok, if I share some thoughts here, even I was some minutes too slow and PR has been merged in this second:

> As far as I can tell from the logs related to #3093, the underlying issue is that the ENF status check times out.

> Instead of a single check that waits for Google's timeout (2min), we do a quick 1-2 check.

I think it's a good idea to do this, but I would suggest to do it the other way round: first check 2min timeout, 2nd check 1s. Why?
It's very plausible that the system as a whole or just the ENF is too busy "with something" that it can't consume the API call or respond within timeout time. For the cases where this is going to happen, it would mean, that the now proposed first check (1s) fails due to system/ENS being busy. The second call will follow immediately after, and there is a high likelihood, that system/ENS is still busy and will also not answer the second call in time.
If you do it the other way round, (2min timeout, then 1s) the likelihood is hight, that system/ENS is already recovered when the 2nd call comes in, and can respond.
What do you think?

I was also wondering, what can make ENS that busy not to respond in time. For me the most plausible explanation would be heavy work load due to verifiying the signature of diagnosis keys, shortly after provision from CWA to ENS. For a different issue, I could make the oberservation in logcat (https://github.com/corona-warn-app/cwa-app-android/issues/2880#issuecomment-827846585), that on my old and slow phone the verification took well 70 seconds, see:
```
===> HERE ENS got the keys provided via API by CWA [...]
04-23 08:07:52.420  3606 26046 I ExposureNotification: Found flag-defined keys de.rki.coronawarnapp:262-v1,MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEc7DEstcUIRcyk35OYDJ95/hTg3UVhsaDXKT0zK7NhHPXoyzipEnOp3GyNXDVpaPi3cAfQmxeuFMZAIX2+6A5Xg== for package de.rki.coronawarnapp [CONTEXT service_id=236 ]
...
=> Now the ENS finished verification of the public key signature of the DK file! It took around 70 seconds!
04-23 08:09:02.034  3606 26046 I ExposureNotification: Signature verification succeeded [CONTEXT service_id=236 ]
```

But if the really signature verification stops ENS from answering API calls, then @d4rken should be able to find hints in the related debug logs, that diagnosis keys had been provided just recently/some time (max. 10 minutes, corresponding to the biggest differences I could find between CWA timestamp on risk card and Google ENS UI via system) before the check for tracing enabled failed. This would probably only occur if CWA tries to provide diagnosis keys more quickly one after the other than expected.

Last question: I guess the 2min timeout is set by CWA?

PR #3504, opened 2021-06-18
Introduce `CertificateContainerId` to reference certificates (DEV)

d4rken:
Re: https://github.com/corona-warn-app/cwa-app-android/pull/3484#issuecomment-864046040
where @mtwalli noticed the difference in referencing the certificates with regards to their repositories.

This PR refactores how certificates are referenced and aligns the behavior between all 3 repositores (vaccination, test, recovery) by introducing `CertificateContainerId` as common identifier when trying to load/modify/delete certificate data via repositories.

#### Problem:

We want a unique identifier that the UI can use to reference certificates, modify and delete them.

We don't want to use the `uniqueCertificateId` that is part of the certificate itself:
* It's only unique if every country follows the spec 1:1 (which sofar has not worked out well)
* Test certificates that are retrieved via CWA are part of the UI, but we don't know their certificate ID yet

#### Solution:

As each certificate is actually a `CertificateContainer` holding the QR Code data, we introduce `CertificateContainerId` to abstract the actual ID away. 

`CertificateContainerId` is actually a sealed class (type safety :partying_face: ):
Each type implementes this as it's viable:
* `VaccinationCertificateContainerId`: Vaccination certs don't have a custom unique ID, so we have to rely use the `uniqueCertificateId` due to legacy reasons (may migrate this in another PR).
* `TestCertificateContainerId`: Test certs use a stored UUID as the certificate may not have been retrieved yet.
* `RecoveryCertificateContainerId`: Recovery certs will also use a stored UUID as it will be more robust and we have no legacy concerns yet.

`CertificateContainerId` is also parcelable and can be passed around with almost zero chance of mixing it up (no more string IDs :partying_face:). In places where we need to do something with the certificate, we can use an exhaustive `when` for nice compiler checks.
Each repository also only accepts their specific subtype of `CertificateContainerId` (e.g. `VaccinationRepository` only takes `VaccinationCertificateContainerId`), so even more type checks :grin:.

/cc @chris-cwa

### DIFF
--- a/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/test/ui/CovidCertificateDetailsFragmentTest.kt
+++ b/Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/test/ui/CovidCertificateDetailsFragmentTest.kt
@@ -15,8 +15,8 @@ import dagger.android.ContributesAndroidInjector
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.covidcertificate.common.certificate.CertificatePersonIdentifier
 import de.rki.coronawarnapp.covidcertificate.common.qrcode.QrCodeString
+import de.rki.coronawarnapp.covidcertificate.common.repository.TestCertificateContainerId
 import de.rki.coronawarnapp.covidcertificate.test.core.TestCertificate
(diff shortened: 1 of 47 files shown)


mtwalli (approved):
Lgtm

chris-cwa (approved):
Thank you!

PR #3487, opened 2021-06-17
Person details screen (EXPOSUREAPP-7574, 7657, 7861)

SamuraiKek:
### _Should be merged after https://github.com/corona-warn-app/cwa-app-android/pull/3504_

🗡️  Combined Dagger UI Module for Covid certificates types 
🗑️ `CertificatesFragment` , `VaccinationListFragment`  and its cards , layouts  -> 🆕  `PersonDetailsFragment 🤺 `
🥅  Wire Certificates details Screens 
🖌️  Some refactoring and renaming to fit in `latest` structure
🔧  Bring consistency to main tabs Onboardings in Dark mode , Transitions, Toolbar 

### 🧪 Testing
- Scan Vaccination certificate  1,2 (Blue colour does not mean immune, it means current)  
- Scan Test certificates using Qr-Code
- Register a test and request certificate 

### ⏰ Note
- Recovery certificate screen is wired up ,but repo is not ready yet
- CWA user flag logic is not implemented yet
- Ordering is not implemented yet (Currently first certificate is high prio certificate)
- Proper navigation from `Qr Code scan -> Certificate details -> Person details`  will be addressed in another Ticket 🎟 

### `// ToDo` 
📸  Screenshots in another PR
🧪  Unit tests in another PR 
🦖  Clean up strings in another PR, maybe @axelherbstreith ?

PR #3509, opened 2021-06-21
Read/Write current CWA user for family certificates. (EXPOSUREAPP-7660)

d4rken:
Implement logic to set/monitor current CWA user with regards to certificates.

TODO: Wire up switch in person details. (likely after #3487)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificatePersonIdentifier.kt
```diff
@@ -8,28 +9,28 @@ import org.joda.time.LocalDate
...
 
     /**
      * Used internally to idenitify and store the data related to this person.
      */
-    internal val code: String by lazy {
-        val dob = dateOfBirth.toString()
-        val lastName = lastNameStandardized
-        val firstName = firstNameStandardized
-        "$dob#$lastName#$firstName"
-    }
```
d4rken, line 24: `lazy` does not play nice with `GSON`, on deserilzation it is null. We can still cache it by using a pattern like here:

https://github.com/corona-warn-app/cwa-app-android/blob/64fa93a4bd99b2084fb1e876c9532d5182362881/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1.kt#L39-L42

but I'm not sure if it's worth it in this case?

mtwalli (approved):
Works!

AlexanderAlferov (approved):
LGTM 🚀

PR #3513, opened 2021-06-22
New date parsing (EXPOSUREAPP-7856) 

chiljamgossow:
See tech spec https://github.com/corona-warn-app/cwa-app-tech-spec/pull/74/files


Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/test/ui/TestCertificateDetailsFragmentTest.kt
```diff
@@ -129,8 +130,8 @@ class TestCertificateDetailsFragmentTest : BaseUITest() {
                     get() = "Schneider"
                 override val fullName: String
                     get() = "Schneider, Andrea"
-                override val dateOfBirth: LocalDate
-                    get() = LocalDate.parse("18.04.1943 00:00", formatter)
+                override val dateOfBirthFormatted: String
+                    get() = "18.04.1943 00:00"
```
mtwalli, line 134: do we need time for birth date?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/vaccination/DccQrCodeCensor.kt
```diff
@@ -81,7 +79,7 @@ class DccQrCodeCensor @Inject constructor() : BugCensor {
             "vaccinationData/vaccineId"
         )
 
-        val vaccinatedAt = vaccinationData.vaccinatedAt.toString()
+        val vaccinatedAt = vaccinationData.vaccinatedOnFormatted.toString()
```
mtwalli, line 82: vaccinatedOnFormatted is a string ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificate.kt
```diff
@@ -7,7 +7,10 @@ import org.joda.time.LocalDate
 
 interface RecoveryCertificate : CwaCovidCertificate {
     override val containerId: RecoveryCertificateContainerId
-    val testedPositiveOn: LocalDate
+    val testedPositiveOnFormatted: String
+    val validFromFormatted: String
+    val validUntilFormatted: String
```
mtwalli, line 12: is it a general requirement that we need to add here and we should use ? can not we just format it based on the use case ?

chiljamgossow: Yes, tech spec defines how they must be formatted. To prevent mix up in formatting (which has been the case so far), I decided to expose the formatted String only. The local date parsed from that formatted String is only exposed in case there are calculations with the date. This also makes it easier in case further changes become necessary.

mtwalli: But is TestCertificate for example  has different formatting in case when it is the highest prio certificate (under qrcode)
and  when it is in the list below the qr code both are in the person details screen ,can we still access the `LocaleDate ` ?

chiljamgossow: For formatting on screens that do not need to follow the tech spec, local date is exposed

mtwalli: 👍

mtwalli (dismissed):

axelherbstreith (dismissed):

mtwalli (approved):

axelherbstreith (approved):

PR #3519, opened 2021-06-22
PPA test result (EXPOSUREAPP-7815)

chiljamgossow:
There were some calls missing/ at the wrong spot that prevented test result metadata from being assembled when initial test result was pending


mtwalli (approved):
LGMT

AlexanderAlferov (approved):

PR #3525, opened 2021-06-22
Implement backend communication for state local statistics (EXPOSUREAPP-7450)

kolyaopahle:
### Description
This PR implements the first draft of backend logic for the state local statistics on the homescreen,
I've also introduced a feature branch for the complete feature which is the base of this PR.

Will assign @jurajkusnier as he is working on other parts of this feature

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
@@ -25,7 +25,8 @@ sealed class StatsItem(val cardType: Type) {
         SEVEN_DAY_RVALUE(4),
         PERSONS_VACCINATED_ONCE(5),
         PERSONS_VACCINATED_COMPLETELY(6),
-        APPLIED_VACCINATION_RATES(7)
+        APPLIED_VACCINATION_RATES(7),
+        LOCAL_INCIDENCE(8)
     }
 
     abstract fun requireValidity()
@@ -63,6 +64,22 @@ data class IncidenceStats(
     override val updatedAt: Instant,
     override val keyFigures: List<KeyFigure>
 ) : StatsItem(cardType = Type.INCIDENCE) {
+
+    val sevenDayIncidence: KeyFigure
+        get() = keyFigures.single { it.rank == KeyFigure.Rank.PRIMARY }
+
+    override fun requireValidity() {
+        require(keyFigures.size == 1)
+        requireNotNull(keyFigures.singleOrNull { it.rank == KeyFigure.Rank.PRIMARY }) {
+            Timber.w("IncidenceStats is missing primary value")
+        }
+    }
+}
+
+data class LocalIncidenceStats(
+    override val updatedAt: Instant,
+    override val keyFigures: List<KeyFigure>
+) : StatsItem(cardType = Type.LOCAL_INCIDENCE) {
 
     val sevenDayIncidence: KeyFigure
         get() = keyFigures.single { it.rank == KeyFigure.Rank.PRIMARY }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
@@ -1,6 +1,7 @@
 package de.rki.coronawarnapp.statistics
 
 import de.rki.coronawarnapp.server.protocols.internal.stats.KeyFigureCardOuterClass.KeyFigure
+import de.rki.coronawarnapp.server.protocols.internal.stats.LocalStatisticsOuterClass
 import org.joda.time.Instant
 import timber.log.Timber
 
@@ -78,7 +79,8 @@ data class IncidenceStats(
 
 data class LocalIncidenceStats(
     override val updatedAt: Instant,
-    override val keyFigures: List<KeyFigure>
+    override val keyFigures: List<KeyFigure>,
+    val federalState: LocalStatisticsOuterClass.FederalStateData.FederalState
 ) : StatsItem(cardType = Type.LOCAL_INCIDENCE) {
 
     val sevenDayIncidence: KeyFigure
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsCache.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsCache.kt
@@ -1,29 +1,31 @@
 package de.rki.coronawarnapp.statistics.local.source
 
-import dagger.assisted.Assisted
-import dagger.assisted.AssistedFactory
-import dagger.assisted.AssistedInject
 import de.rki.coronawarnapp.statistics.Statistics
 import de.rki.coronawarnapp.statistics.local.FederalStateToPackageId
 import timber.log.Timber
 import java.io.File
+import javax.inject.Inject
+import javax.inject.Singleton
 
-class LocalStatisticsCache @AssistedInject constructor(
-    @Statistics cacheDir: File,
-    @Assisted private val stateForCache: FederalStateToPackageId,
+@Singleton
+class LocalStatisticsCache @Inject constructor(
+    @Statistics cacheDir: File
 ) {
 
     private val cacheFolder = File(cacheDir, "cache_raw_local")
-    private val cacheFile = File(cacheFolder, stateForCache.toString())
 
-    fun load(): ByteArray? = try {
+    fun load(stateForCache: FederalStateToPackageId): ByteArray? = try {
+        val cacheFile = File(cacheFolder, stateForCache.name)
+
         if (cacheFile.exists()) cacheFile.readBytes() else null
     } catch (e: Exception) {
         Timber.e(e, "Failed to load raw statistics from cache.")
         null
     }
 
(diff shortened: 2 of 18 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
```diff
@@ -75,6 +76,22 @@ data class IncidenceStats(
     }
 }
 
+data class LocalIncidenceStats(
```
jurajkusnier, line 79: I think this class should contain a federal-state id or translated name

kolyaopahle: done :)

jurajkusnier (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsCache.kt
```diff
@@ -0,0 +1,44 @@
...
+import timber.log.Timber
+import java.io.File
+
+class LocalStatisticsCache @AssistedInject constructor(
+    @Statistics cacheDir: File,
+    @Assisted private val stateForCache: FederalStateToPackageId,
+) {
+
+    private val cacheFolder = File(cacheDir, "cache_raw_local")
+    private val cacheFile = File(cacheFolder, stateForCache.toString())
```
d4rken, line 17: Please use something more explicit than `.toString()`, could be too easily accidentally affected by refactoring.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+            onError = { Timber.tag(TAG).e(it, "Failed to update statistics.") }
+        )
+    }
+
+    fun clear(forState: FederalStateToPackageId) {
+        Timber.d("clear()")
+
+        server.clear()
+        localStatisticsCacheFactory.create(forState).save(null)
+    }
```
d4rken, line 123: Do we need a `clear()` for all? To use with `DataReset`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsCache.kt
```diff
@@ -0,0 +1,44 @@
...
+import dagger.assisted.AssistedInject
+import de.rki.coronawarnapp.statistics.Statistics
+import de.rki.coronawarnapp.statistics.local.FederalStateToPackageId
+import timber.log.Timber
+import java.io.File
+
+class LocalStatisticsCache @AssistedInject constructor(
+    @Statistics cacheDir: File,
+    @Assisted private val stateForCache: FederalStateToPackageId,
+) {
```
d4rken, line 14: Why use one instance per federal state, instead of storing all of the states within one instance and a method that takes the ID as parameter?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+            .launchIn(scope)
+
+        localStatisticsConfigStorage.activeStates.flow
+            .onEach {
+                Timber.tag(TAG).d("The local statistics config was updated, triggering statistics update.")
+                triggerUpdate()
+            }
+            .catch { Timber.tag(TAG).e("Failed to trigger statistics update.") }
+            .launchIn(scope)
+    }
```
d4rken, line 69: What do you think about moving this into an extra class? Could make testing a bit easier.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
```diff
@@ -0,0 +1,128 @@
...
+
+    private val localStatisticsData = HotDataFlow(
+        loggingTag = TAG,
+        scope = scope,
+        coroutineContext = dispatcherProvider.IO,
+        sharingBehavior = SharingStarted.WhileSubscribed(
+            stopTimeoutMillis = Duration.standardSeconds(5).millis,
+            replayExpirationMillis = 0
+        )
+    ) {
```
d4rken, line 40: I think this is missing a flow subscription that monitors this and on change, writes all changes back to storage?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsParser.kt
```diff
@@ -0,0 +1,77 @@
...
+                null
+            }
+        }
+
+        val mappedAdministrativeUnit = parsed.administrativeUnitDataList.mapNotNull { rawState ->
+            try {
+                val updatedAt = Instant.ofEpochSecond(rawState.updatedAt)
+                val administrativeUnitIncidenceKeyFigure = rawState.sevenDayIncidence.toKeyFigure()
+
+                val federalStateId = rawState.administrativeUnitShortId.toString().dropLast(3).toInt()
```
d4rken, line 42: Where does the `dropLast(3)` come from?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
```diff
@@ -0,0 +1,118 @@
...
+    init {
+        localStatisticsRetrievalScheduler.updateStatsTrigger
+            .onEach {
+                if (it) {
+                    Timber.tag(TAG).d("Triggering local statistics update.")
+                    triggerUpdate()
+                }
+            }
+            .catch { Timber.tag(TAG).e("Failed to trigger local statistics update.") }
+            .launchIn(scope)
```
d4rken, line 60: I'd prefer if we reverse the dependency here.
The repo doesn't need to know about the scheduler, the repo can be passive.
The scheduler just needs to know about the repo an be able  to call a `triggerUpdate` update.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
```diff
@@ -0,0 +1,118 @@
...
+        loggingTag = TAG,
+        scope = scope,
+        coroutineContext = dispatcherProvider.IO,
+        sharingBehavior = SharingStarted.WhileSubscribed(
+            stopTimeoutMillis = Duration.standardSeconds(5).millis,
+            replayExpirationMillis = 0
+        )
+    ) {
+        try {
+            triggerUpdate()
```
d4rken, line 41: Can we prevent this from triggering a server request everytime `localStatisticsData` is subscribed to?
It only needs to trigger an update if the cache is `null`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
d4rken: e.g. `Scheduler.setup()`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsRetrievalScheduler.kt
```diff
@@ -0,0 +1,40 @@
...
+
+    val updateStatsTrigger = combine(
+        foregroundState.isInForeground,
+        localStatisticsConfigStorage.activeStates.flow
+    ) { isInForeground, statistics ->
+        val hasNewStats = statistics.any {
+            val isNew = !lastActiveStates.contains(it)
+            if (isNew) lastActiveStates.add(it)
+            isNew
+        }
```
d4rken, line 27: What happens if the user switches to `State A`, then to `State B` and then back to `State A`, does it still refresh data for `State A` again?

d4rken (commented):
Looks good, few comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsParser.kt
kolyaopahle: https://github.com/corona-warn-app/cwa-app-tech-spec/pull/61/files#diff-b934511166d6db28e5cc528bcfc6df5ac63dd27eb1a9cc9192a61f2b16f4b73eR666

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsRetrievalScheduler.kt
kolyaopahle: probably not, i've replaced the logic with a contains all check

jurajkusnier (approved):

d4rken (approved):
Looks good, additional testing happens when UI is done.

PR #3526, opened 2021-06-22
Fix Qr Code flickering (DEV)

mtwalli:
Steps to produce 
1- Open person details screen 
2- Switch  cwa user switch  ON/OFF
3- Qr-Code is flickering in `2.5 ` -> `Fixed here 🚀`


axelherbstreith (approved):

BMItr (approved):
good catch

PR #3528, opened 2021-06-23
Fix ppa test result (EXPOSUREAPP-7815)

chiljamgossow:
test result was not always persisted

testResultAtRegistration is actually just the test result and needs to be updated on each poll


AlexanderAlferov (approved):
Tested on Samsung S8: confirm two `test_result_metadata_set`

BMItr (approved):

Issue #3514, opened 2021-06-22
paper QR Code not readable


Timo1993:
Hi, I got a digital certificate. When I tried to scan a QR code, it is not detected. I used a Xiaomi Redmi Note 9. The most recent app from the play store is installed. I tried it about 50 times, but it is not detected. The code itself is detected with another QR code scanner.

---
Internal Tracking ID: [EXPOSUREAPP-8028](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8028)
Internal Tracking ID: [EXPOSUREAPP-10502](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10502)

Ein-Tim:
@Timo1993 Does the CovPass App recognize the code?

Timo1993:
> @Timo1993 Does the CovPass App recognize the code?

No, the code is also not recognized by the CovPass app. But when I use a QR code scanner app, I can read it.

Ein-Tim:
@Timo1993 

Are you sure that the QR-Code is a EU Digital Covid Certificate? 

A QR-Code for this is printed on such a sheet of paper:
![image](https://user-images.githubusercontent.com/67682506/122910803-ec4c3e80-d356-11eb-8758-bc83ccb8c31e.jpeg)

(Picture from https://www.coronawarn.app/en/faq/#vac_cert_invalid)

Timo1993:
Yes, I have exactly this document and it is printed in a good quality (in black and white, which short not cause any problems).

Ein-Tim:
@Timo1993 

Okay, thanks. You are explicit mentioning Xiaomi here. Do you have another device to try if the QR-Code can be scanned with another device?

vaubaehn:
What @Ein-Tim suggested would be very helpful: scanning with a different Android device, but also using CWA and CovPass.

Did you also try to scan from a bigger distance? >1m

Has the certificate been issued in Germany?

What does the certificate cover?
- Single or double vaccinating?
- if double vaccination, different vaccines or same?
- Was the vaccine Johnson & Johnson?
- Or is it a recovery certificate (tested positive before and vaccinated once later)?

MikeMcC399:
@Timo1993 
- You said that you are using the latest app version. Can you confirm that the App Information shows 2.3.4?
[...]

PR #3531, opened 2021-06-23
Qr code error level (EXPOSUREAPP-8073)

mtwalli:
- Set default Qr Code error correction level as `Medium`
- Version bump 2.4.2.0

### DIFF
--- a/Corona-Warn-App/src/main/res/navigation/nav_graph.xml
+++ b/Corona-Warn-App/src/main/res/navigation/nav_graph.xml
@@ -815,7 +815,7 @@
             app:argType="string" />
         <argument
             android:name="correctionLevel"
-            android:defaultValue="Q"
+            android:defaultValue="M"
             app:argType="com.google.zxing.qrcode.decoder.ErrorCorrectionLevel" />
     </fragment>
     <action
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -815,7 +815,7 @@
             app:argType="string" />
         <argument
             android:name="correctionLevel"
-            android:defaultValue="Q"
+            android:defaultValue="M"
```
harambasicluka, line 818: Why did you change this? The change should be about the generated QR Code, not the scanned one.

harambasicluka: @thomasaugsten, @mlenkeit  what do you think?

harambasicluka (commented):
not sure about the change on the scanning. But it might help with #3514.

Edit: misunderstood it, it does what it should do!

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
mtwalli: @harambasicluka  the qroode should not change when you got to full screen mode

chiljamgossow (approved):
LGTM

harambasicluka (approved):
Works 💯

axelherbstreith (approved):

PR #3533, opened 2021-06-23
Fix deadman notification delay (EXPOSUREAPP-7955)

chiljamgossow:
use createdAt timestamp to calculate delay of notification (not the package date)

Scenario: metered connection -> only day packages. In the morning of 06/23 it will download package 06/22. Before it would take the date at start of the day and plan a notification in 36h minus the age of the downladed package - which is already 24h hours old when it was downloaded. So 12h later the day the user gets the notification - despite having a daily download


AlexanderAlferov (approved):

jurajkusnier (approved):
LGTM

PR #3540, opened 2021-06-24
Test and recovery certificate censoring (EXPOSUREAPP-8083)

chiljamgossow:
Scan certificates, request test certificate and check log

Censor affects: dateOfBirth when requesting test certificate, qrCodes, any content extracted from the qrCode

- [X] unit test


AlexanderAlferov (dismissed):
Thank you, ████ all data could ████ and ███.

BMItter:
General Question
I'm not sure whether this is thread safe.
If I got it right its mainly addQRCodeStringToCensor()) -> checkLog()
If we call addQRCodeStringToCensor() from another part, there can be an inconsistant state because checkLog() trys to work with the data from the first and second call of addQRCodeStringToCensor()

for example:
_Thread1.addQRCodeStringToCensor(...)
Thread2.addQRCodeStringToCensor(...)_
....
_Thread2.checkLog()
Thread1.checkLog()_

chiljamgossow:
> General Question
> I'm not sure whether this is thread safe.
> If I got it right its mainly addQRCodeStringToCensor()) -> checkLog()
> If we call addQRCodeStringToCensor() from another part, there can be an inconsistant state because checkLog() trys to work with the data from the first and second call of addQRCodeStringToCensor()
> 
> for example:
> _Thread1.addQRCodeStringToCensor(...) Thread2.addQRCodeStringToCensor(...)_
> ....
> _Thread2.checkLog() Thread1.checkLog()_

@BMItter addQRCodeStringToCensor()) and checkLog() are both synchronised and should be thread safe. Or did you something else?

AlexanderAlferov (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/dcc/DccQrCodeCensor.kt
```diff
@@ -0,0 +1,270 @@
...
+        return newMessage
+    }
+
+    companion object {
+        private val qrCodeStringsToCensor = ArrayList<String>()
+
+        fun addQRCodeStringToCensor(rawString: String) = synchronized(qrCodeStringsToCensor) {
+            qrCodeStringsToCensor.apply {
+                if (contains(rawString)) return@apply
+                add(rawString)
```
d4rken, line 252: Not a blocker, but we could use `mutableSetOf` here, and just call `add` as the set will not allow the same object twice.

BMItter:
> > General Question
> > I'm not sure whether this is thread safe.
> > If I got it right its mainly addQRCodeStringToCensor()) -> checkLog()
> > If we call addQRCodeStringToCensor() from another part, there can be an inconsistant state because checkLog() trys to work with the data from the first and second call of addQRCodeStringToCensor()
> > for example:
> > _Thread1.addQRCodeStringToCensor(...) Thread2.addQRCodeStringToCensor(...)_
> > ....
> > _Thread2.checkLog() Thread1.checkLog()_
> 
> @BMItter addQRCodeStringToCensor()) and checkLog() are both synchronised and should be thread safe. Or did you mean something else?

its all right for me if you don't see an issue.

BMItr (approved):
tested on oneplus 8 device.
lgtm

PR #3554, opened 2021-06-25
No more default translation updates!!! (DEV)

harambasicluka:
With this the defaults will be delivered to `values` instead of `values-en` and no manual copy pasting is required 🎉


harambasicluka:
Okay, we are still not sure if this works, as we didn't specified a target for the english values. We are in contact with the support.

axelherbstreith (dismissed):

mtwalli (dismissed):

harambasicluka:
Aligned with the translation experts, we'll test it with his config.

PR #3563, opened 2021-06-28
Deadman for time travellers (DEV)

chiljamgossow:
taking the createdAt timestamp from the latest package - to make it work for users who have manipulated the system time in the past


d4rken (approved):

BMItr (approved):

vaubaehn:
@chiljamgossow 
Hi Chilja, can you confirm that this PR won't revert/affect #3533 ? Please see also https://github.com/corona-warn-app/cwa-app-android/issues/3625#issuecomment-873462246

I'm currently on mobile with low battery, can't peek into the code atm...

Have a nice weekend!

chiljamgossow:
@vaubaehn This PR is not reverting the fix as the createdAt timestamp is still used to calculate the delay. It just takes it from the most recent package from the server. This makes only a difference for those who play with the device time and is intended to make it work properly again when the device time is correct.

vaubaehn:
@chiljamgossow and here also: thanks for your reply!

Issue #3561, opened 2021-06-27
App shows "Full vaccination protection in 0 days" on day 14 after the last shot


Ein-Tim:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: not important
- Android version: not important
- App version: 2.4

## Describe the bug

On day 14 after the last vaccination shot was received, the app shows "Full vaccination protection in 0 days".
Here's a screenshot: 
<img src="https://pbs.twimg.com/media/E4yL01pWYAEpXsY?format=jpg&name=medium" width="330">

## Steps to reproduce the issue

1. Get your second vaccination shot
2. Wait 14 days or change the system time to 14 days in the future
3. See that the app shows "Full vaccination protection in 0 days"

## Expected behaviour

There shouldn't be the number "0" - this is not how we humans communicate 😁

## Additional context

This bug report is based on the information a user shared on Twitter: https://twitter.com/stejormur/status/1409031782448705537?s=20

---

Internal Tracking-ID: [EXPOSUREAPP-8121](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8121)

 In future it will show on the last day "Morgen/Tomorrow".

iOS PR corona-warn-app/cwa-app-ios#3029
Internal Tracking ID: EXPOSUREAPP-8157
R 2.5

Android PR #3575, https://github.com/corona-warn-app/cwa-app-android/pull/3567
Internal Tracking ID: EXPOSUREAPP-8158
R 2.4.3

heinezen:
@Ein-Tim 

Thanks for reporting. We are already investigating the issue internally.

----
Corona-Warn-App Open Source Team

DerVogel2020:
> @Ein-Tim @heinezen 

IMO under iOS the behaviour is correct. For me it shows 2 days while under Android the app shows 1 day.

guruguruguru:
![Screenshot_20210627-095323_Corona-Warn](https://user-images.githubusercontent.com/36473526/123553628-9fb09b00-d77c-11eb-9937-3ef69d4425a9.jpg)

This is what I got today.

MikeMcC399:
[...]

Issue #3562, opened 2021-06-28
Vaccination certificate validation time uses UTC+0 timezone


mawis:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Xiaomi Poco F2 Pro
- Android version: 11
- App version: 2.4.2
- Time zone of the device: Europe/Berlin

## Describe the bug

I was vaccinated the second time at June 13th. As it was confirmed in #3561 that means the certificate should get valid June 28th. Looking at the app at 1:08 a.m. the certificate was still in the state where it isn't valid yet. Looking at it again at 7:00 a.m. the certificate was valid.
Looking at the code, it seems to me that calculations were done in UTC, so probably the state changed at 2:00 a.m. Being in Germany under German jurisdiction the correct timezone to do the calculations probably should be Europe/Berlin.

## Steps to reproduce the issue

1. Get a certificate for a second vaccination, that is not valid yet
2. Import it to the CWA
3. Check the counter of days remaining for the certificate to become valid
4. Verify that the number of days does not change at midnight in Germany but at a later moment during the night.

See also the following screenshot taken at 2021-06-28 at 01:08+0200:
![](https://matthias-wimmer.de/CWA_Problem.jpeg)

## Expected behaviour

Calculated remaining days as well as the change of the certificate status should happen at midnight German time.

---
Internal Tracking ID: [EXPOSUREAPP-8121](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8121)

Ein-Tim:
Please check the behavior under iOS here. 

Maybe @DerVogel2020 could do this (by changing the system time)?

DerVogel2020:
> @Ein-Tim 
> Maybe @DerVogel2020 could do this (by changing the system time)?

No, that's not possible.

Ein-Tim:
@DerVogel2020 

Okay, then the devs have to check this.
Thank you anyways.

dsarkar:
@mawis Thanks for reporting. Issue is being analysed by the developers currently.

Ein-Tim:
FYI: PR: #3567

MikeMcC399:
[...]

PR #3567, opened 2021-06-28
Improve "time till immunity" calculation & display (EXPOSUREAPP-8121, EXPOSUREAPP-8158)

d4rken:
* Use local timezone to determine "midnight" which affects when the user's state switches from COMPLETE to IMMUNITY.
* Change calculation so that `IMMUNITY` is reached when `getTimeUntilImmunity` reaches 0. Previously we just compared days and had to check for `days < 0` which makes for confusing code due to rounding.
* UI now shows specific text when time until immunity is less than a day ("tomorrow")

See #3561 and #3562.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
@@ -3,8 +3,8 @@ package de.rki.coronawarnapp.covidcertificate.vaccination.core
 import de.rki.coronawarnapp.covidcertificate.common.certificate.CertificatePersonIdentifier
 import de.rki.coronawarnapp.covidcertificate.vaccination.core.repository.storage.VaccinatedPersonData
 import de.rki.coronawarnapp.covidcertificate.valueset.valuesets.VaccinationValueSets
-import de.rki.coronawarnapp.util.TimeAndDateExtensions.toUserTimeZone
-import org.joda.time.Duration
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toLocalDateUserTz
+import org.joda.time.Days
 import org.joda.time.Instant
 import org.joda.time.LocalDate
 
@@ -36,28 +36,24 @@ data class VaccinatedPerson(
         )
 
     fun getVaccinationStatus(nowUTC: Instant = Instant.now()): Status {
-        val timeToImmunity = getTimeUntilImmunity(nowUTC) ?: return Status.INCOMPLETE
+        val timeToImmunity = getDaysUntilImmunity(nowUTC) ?: return Status.INCOMPLETE
 
         return when {
-            timeToImmunity <= Duration.ZERO -> Status.IMMUNITY
+            timeToImmunity <= 0 -> Status.IMMUNITY
             else -> Status.COMPLETE
         }
     }
 
-    fun getTimeUntilImmunity(nowUTC: Instant = Instant.now()): Duration? {
+    fun getDaysUntilImmunity(nowUTC: Instant = Instant.now()): Int? {
         val newestFullDose = vaccinationCertificates
             .filter { it.doseNumber == it.totalSeriesOfDoses }
             .maxByOrNull { it.vaccinatedAt }
             ?: return null
 
-        val immunityAt = newestFullDose.vaccinatedAt
-            .toDateTimeAtStartOfDay()
-            .plus(IMMUNITY_WAITING_PERIOD)
+        val today = nowUTC
+            .toLocalDateUserTz()
 
-        val now = nowUTC
-            .toUserTimeZone()
-
-        return Duration(now, immunityAt)
+        return IMMUNITY_WAITING_DAYS - Days.daysBetween(newestFullDose.vaccinatedAt, today).days
     }
 
     enum class Status {
@@ -67,6 +63,6 @@ data class VaccinatedPerson(
     }
 
     companion object {
-        private val IMMUNITY_WAITING_PERIOD = Duration.standardDays(15)
+        private val IMMUNITY_WAITING_DAYS = 15
     }
 }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
@@ -3,10 +3,12 @@ package de.rki.coronawarnapp.covidcertificate.vaccination.core
 import de.rki.coronawarnapp.covidcertificate.common.certificate.CertificatePersonIdentifier
 import de.rki.coronawarnapp.covidcertificate.vaccination.core.repository.storage.VaccinatedPersonData
 import de.rki.coronawarnapp.covidcertificate.valueset.valuesets.VaccinationValueSets
-import de.rki.coronawarnapp.util.TimeAndDateExtensions.toLocalDateUtc
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toUserTimeZone
+import org.joda.time.DateTimeZone
 import org.joda.time.Duration
 import org.joda.time.Instant
 import org.joda.time.LocalDate
+import java.util.TimeZone
 
 data class VaccinatedPerson(
     internal val data: VaccinatedPersonData,
@@ -50,9 +52,14 @@ data class VaccinatedPerson(
             .maxByOrNull { it.vaccinatedAt }
             ?: return null
 
-        val immunityAt = newestFullDose.vaccinatedAt.toDateTimeAtStartOfDay().plus(IMMUNITY_WAITING_PERIOD)
+        val immunityAt = newestFullDose.vaccinatedAt
+            .toDateTimeAtStartOfDay(DateTimeZone.forTimeZone(TimeZone.getDefault()))
+            .plus(IMMUNITY_WAITING_PERIOD)
 
-        return Duration(nowUTC.toLocalDateUtc().toDateTimeAtStartOfDay(), immunityAt)
+        val now = nowUTC
+            .toUserTimeZone()
+
+        return Duration(now, immunityAt)
     }
 
     enum class Status {
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
@@ -63,6 +63,6 @@ data class VaccinatedPerson(
     }
 
     companion object {
-        private val IMMUNITY_WAITING_DAYS = 15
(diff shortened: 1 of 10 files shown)


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPersonTest.kt
```diff
@@ -117,15 +128,80 @@ class VaccinatedPersonTest : BaseTest() {
...
+                getVaccinationStatus(now) shouldBe VaccinatedPerson.Status.COMPLETE
+            }
+            // Immunity should be reached at midnight in the users timezone
+            Instant.parse("2021-06-27T22:00:00.000Z").let { now ->
+                getTimeUntilImmunity(now)!!.standardHours shouldBe -24
+                getTimeUntilImmunity(now)!!.standardDays shouldBe -1
+                getVaccinationStatus(now) shouldBe VaccinatedPerson.Status.IMMUNITY
+            }
+        }
+    }
```
d4rken, line 173: Case described by @mawis in #3562

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/ui/cards/VaccinationCard.kt
```diff
@@ -26,12 +27,17 @@ class VaccinationCard(parent: ViewGroup) :
...
+                vaccinationState.text = if (immunityIn < Duration.standardDays(1)) {
+                    resources.getString(R.string.vaccination_card_status_vaccination_complete_tomorrow)
+                } else {
+                    val days = immunityIn.standardDays.toInt()
+                    resources.getQuantityString(
+                        R.plurals.vaccination_card_status_vaccination_complete,
+                        days,
+                        days
+                    )
+                }
```
d4rken, line 40: Final shot at `2021-06-14`:

* `2021-06-27` -> `Vollständiger Impfschutz in 1 Tag`
* `2021-06-28` -> `Vollständiger Impfschutz ab morgen`
* `2021-06-29` -> IMMUNITY Card

Corona-Warn-App/src/main/res/values-de/vaccination_strings.xml
```diff
@@ -87,6 +87,7 @@
         <item quantity="few">Vollständiger Impfschutz in %1$d Tagen</item>
         <item quantity="many">Vollständiger Impfschutz in %1$d Tagen</item>
     </plurals>
+    <string name="vaccination_card_status_vaccination_complete_tomorrow">Vollständiger Impfschutz ab morgen</string>
```
d4rken, line 90: We can't use `plurals.zero` because the relation between quantity types is a grammatical one, e.g. in Germany there is no difference between `other` and `zero`, so `zero` is never used.

d4rken:
PR now also includes a string change, and some additional adjustments + tests to deal with daylight-savings-time.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/ui/cards/VaccinationCard.kt
Ein-Tim: Just that I get it right, this fixes #3561, right?

mawis: I think that for most people (especially non-programmers) the sentences “Vollständiger Impfschutz in 1 Tag” and “Vollständiger Impfschutz ab morgen” are equivalent. Therefore I think the message on 2021-06-27 is incorrect and should read “Vollständiger Impfschutz in 2 Tagen”.

d4rken: Yes, I think so.
Once `immunityIn < Duration.standardDays(1)` is `true`, `tomorrow` will be displayed.

d4rken:
Draft until confirmation with @mlenkeit  and a sync with iOS team.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/ui/cards/VaccinationCard.kt
d4rken: > I think that for most people (especially non-programmers) the sentences “Vollständiger Impfschutz in 1 Tag” and “Vollständiger Impfschutz ab morgen” are equivalent. Therefore I think the message on 2021-06-27 is incorrect and should read “Vollständiger Impfschutz in 2 Tagen”.

@mawis That would mean: Display "tomorrow" when `immunityIn < Duration.standardDays(2)`, I'm partial to both, would have to check with UX team in any case.

Ein-Tim: No, in 2 days stays in 2 days but in 1 day has to be tomorrow, if I understood @mawis correctly.

mawis: Yes. Probably that's the point, that non-programmers don't think the way integers work, but normally round up to the next full day. So at 23:59 “in 1 day” at least to me still means “the next day”, i.e. at that moment that's ”in one minute”.

So given that indeed the calculation of `immunityIn` could be changed. Anyway I think `immunityIn` could stay the same as it is (working the normal way with integers), but the formatting should be changed. Therefore I personally would append a `+ 1` at the end of line 34 in the diff above.

harambasicluka:
So we agreed on how to display the days:
> 15, 14, 13, 12, ..., 3, 2, tomorrow, immun

MikeMcC399:
@harambasicluka 
>So we agreed on how to display the days:
>
>> 15, 14, 13, 12, ..., 3, 2, tomorrow, immun

Hopefully spelled "immune" in English
https://dictionary.cambridge.org/dictionary/english/immune 🙂

harambasicluka:
> @harambasicluka
> 
> > So we agreed on how to display the days:
> > > 15, 14, 13, 12, ..., 3, 2, tomorrow, immun
> 
> Hopefully spelled "immune" in English
> https://dictionary.cambridge.org/dictionary/english/immune 🙂

No worries, I won't do the translation :D

jurajkusnier (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
```diff
@@ -50,9 +50,14 @@ data class VaccinatedPerson(
             .maxByOrNull { it.vaccinatedAt }
             ?: return null
 
-        val immunityAt = newestFullDose.vaccinatedAt.toDateTimeAtStartOfDay().plus(IMMUNITY_WAITING_PERIOD)
+        val immunityAt = newestFullDose.vaccinatedAt
```
chiljamgossow, line 53: Wouldn't it be simpler to just use the local date and get Days.daysBetween() as I did for days since installation? 
eg. val daysSinceInstallation: Int
        get() = Days.daysBetween(dayOfInstallation, today).days

chris-cwa (dismissed):
LGTM!

InkenTownsend (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
chiljamgossow: with val today: LocalDate
        get() = Instant.now().toLocalDateUserTz()

d4rken: Both should work.
I'm not sure if I find `daysBetween` simpler, just different?

Using `daysBetween` would also mean loosing resolution, currently we get back a `Duration` and ever consumer of the API can use a time resolution of their choice.

If this is not a blocker for you, I'd prefer to not change this.

chiljamgossow: A duration does not give you precise information about the how many date changes it contains. We are counting date changes here, not 24h intervals, I assume. The information needed for this is lost once converted to a duration.
I believe this solution is error prone.

d4rken: I think counting date changes is a UI detail, business logic wise, a `Duration` until immunity is reached (a countdown) makes sense to me. But I have no strong feelings about this if you want to change this.

mawis: @chiljamgossow is right here. Counting 24h intervals will never be completely right. At the start/end of daylight saving time a day is 23h/25h long.

So I think @d4rken's proposal is correct, that actually days should be counted.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPersonTest.kt
```diff
@@ -112,20 +125,119 @@ class VaccinatedPersonTest : BaseTest() {
...
+                getVaccinationStatus(now) shouldBe VaccinatedPerson.Status.COMPLETE
+            }
+            // Case Luka#1 happens on 01.15.21, this mean it's winter time!
+            // The users timezone is GMT+1 (winter-time) (UTC+1), not GMT+2 (summer-time) (UTC+2)
+            Instant.parse("2021-01-15T22:00:00.000Z").let { now ->
+                getTimeUntilImmunity(now)!!.standardMinutes shouldBe 60
+                getTimeUntilImmunity(now)!!.standardHours shouldBe 1
+                getTimeUntilImmunity(now) shouldBe Duration.standardHours(1)
+                getVaccinationStatus(now) shouldBe VaccinatedPerson.Status.COMPLETE
+            }
```
mlenkeit, line 221: @d4rken `"2021-01-15T23:00:00.000Z"` should then yield `IMMUNITY`, right? Can you add that case?

chiljamgossow: added

chiljamgossow (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
chiljamgossow: using Days.daysBetween now

InkenTownsend (approved):
UA approved

chris-cwa (approved):

PR #3573, opened 2021-06-29
Generate empty byte array without GMS dependency

fynngodau:
The `AesCryptography` class is currently generating an empty byte array using:

    Hex.stringToBytes("00000000000000000000000000000000")

The problem with this code is that `Hex` is contained in the package `com.google.android.gms.common.util`, which appears to be a Google Play services library that CCTG doesn't contain. For this reason, I propose generating the empty array as follows:

    ByteArray(16) { 0 }


d4rken (approved):
lgtm 

Screenshot from 2021-06-29 10-29-53

chiljamgossow (approved):

PR #3574, opened 2021-06-29
Don't ship debug binaries (EXPOSUREAPP-8027)

d4rken:
Coroutines dependency includes "DebugProbesKt.bin" by default.
See https://github.com/Kotlin/kotlinx.coroutines/issues/2274

Testing:
Build a release APK, extract it and check that it doesn't contain the mentioned file.


jurajkusnier (approved):
LGTM

BMItr (approved):
lgtm

PR #3556, opened 2021-06-25
Validation rules skeleton (DEV)

mtwalli:
- Some skeleton classes for Validation rules feature

#### Note 
- Some data structure should be provided by the library 📚 used for JOSN engine
- covid_certificate_strings.xml should be used for localisation 
- covid_certificate_graph.xml should be used to add the new screens

PR #3583, opened 2021-06-29
DCC validation skeletons v2 (DEV)

d4rken:
Follow up to #3556

Potentially allow splitting up the validation subtask.

Change validation result from sealed class to enum, the sealed class is not necessary at this level.
I thought it would help split up the different "view states" mentioned in the techspec, but this is something that needs to happen at UI level, and for that the enum is enough.


chiljamgossow (approved):
good enough is good enough 😄

chris-cwa (approved):

Issue #3571, opened 2021-06-28
Test result certificate can't be downloaded


simontb:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Google Pixel 4
- Android version: 11
- App version: 2.4.2

## Describe the bug

Today I used the rapid test profile for the first time. The registration went smoothly and I got my negative result. However, the download of the test certificate fails.

<img src="https://user-images.githubusercontent.com/6491719/123688945-f5577700-d852-11eb-99f3-250772d52742.png" height="500">
<img src="https://user-images.githubusercontent.com/6491719/123688948-f5f00d80-d852-11eb-80dd-2f88be44c972.png" height="500">
<img src="https://user-images.githubusercontent.com/6491719/123688951-f5f00d80-d852-11eb-97ba-b1715a43cdf2.png" height="500">
<img src="https://user-images.githubusercontent.com/6491719/123688954-f688a400-d852-11eb-9fe2-0d6d5f9ad28f.png" height="500">

## Expected behaviour

I can download the certificate.

## Additional context

Test center was SAP (WDF01).

Here is the stacktrace that is shown in the details of the error dialog:

```
de.rki.coronawarnapp.covidcertificate.exception.TestCertificateServerException: Private key request failed.

de.rki.coronawarnapp.covidcertificate.exception.TestCertificateServerException: Private key request failed.
at de.rki.coronawarnapp.covidcertificate.test.core.server.TestCertificateServer$registerPublicKeyForTest$2.invokeSuspend(TestCertificateServer.kt:27)
at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:18)
at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:1)
at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:11)
```

---
Internal Tracking ID: [EXPOSUREAPP-8162](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8162)

[...]

PR #3586, opened 2021-06-30
Fix environment setup configuration for DCC server (EXPOSUREAPP-8162)

d4rken:
* Fix DCC server configuration for prod environment
* Add sanity check for environment setup on app launch (fail early)
* Add unit tests that cover the actual json files (a bit hacky but works and catches this)

See #3571


BMItr (approved):
GOOD_JSON
👌

jurajkusnier (approved):
LGTM

PR #3588, opened 2021-06-30
Add Start Validation Button to Existing Screens (EXPOSUREAPP-8047)

axelherbstreith:
- Add start validation button to Person Details, Test Certificate, Recovery Certificate and Vaccination Certificate screens
- Wire them to Start validation screen
- Connect to repo
- Handle error cases 
- Use leading button in Start Validation screen

- [x] Added buttons
- [x] Adding custom button to match the figma design
- [x] Connect To Repo


NataliaLemmerth (dismissed):
UA approved

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsFragmentTest.kt
```diff
@@ -119,7 +119,7 @@ class PersonDetailsFragmentTest : BaseUITest() {
                 listOf(testCertificate, vaccinationCertificate1, vaccinationCertificate2), isCwaUser = isCwa
             )
 
-            add(PersonDetailsQrCard.Item(testCertificate, bitmap))
+            add(PersonDetailsQrCard.Item(testCertificate, bitmap) {})
```
d4rken, line 122: Should this `{}` do something later on? Maybe add a TODO?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsEvents.kt
```diff
@@ -8,4 +9,6 @@ sealed class PersonDetailsEvents
 data class OpenVaccinationCertificateDetails(val containerId: VaccinationCertificateContainerId) : PersonDetailsEvents()
 data class OpenTestCertificateDetails(val containerId: TestCertificateContainerId) : PersonDetailsEvents()
 data class OpenRecoveryCertificateDetails(val containerId: RecoveryCertificateContainerId) : PersonDetailsEvents()
+data class ValidationStart(val containerId: CertificateContainerId) : PersonDetailsEvents()
+data class ShowErrorDialog(val error: Throwable) : PersonDetailsEvents()
```
d4rken, line 13: In most cases prefer using `Exception` over `Throwable`, the difference would be that it can no longer accept `Error : Throwable`, but `Error` are usually not recoverable either and we should just crash.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -38,6 +39,12 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
     )
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) = with(binding) {
+
+        startValidationCheck.defaultButton.setOnClickListener {
+            startValidationCheck.isLoading = true
```
d4rken, line 44: Does the custom-view persist it's loading state via `SavedState`?
If not and the "loading operation" isn't canceled when the screen is left, then the user could press the home button, come back to the app, and the operation could  still be running, but the loading indicator would be gone.

We'd need to either make sure the view persists it's own state or save it in a instance that survives as long as the "loading operation", e.g. viewmodel?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -59,7 +66,7 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
 
         fullname.text = certificate.fullName
         dateOfBirth.text = certificate.dateOfBirthFormatted
-        recoveredFromDisease.text = certificate.targetDisease
+        recoveredFromDesease.text = certificate.targetDisease
```
d4rken, line 69: Previous spelling was correct?

Corona-Warn-App/src/main/res/layout/person_details_qr_card_item.xml
```diff
@@ -52,4 +52,18 @@
...
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginTop="4dp"
+        android:text="@string/start_validation_rule_check_button"
+        android:textColor="@color/colorSecondary"
+        app:layout_constraintBottom_toBottomOf="parent"
+        app:layout_constraintStart_toStartOf="@id/qr_subtitle"
+        app:layout_constraintTop_toBottomOf="@+id/qr_subtitle"
+        app:strokeColor="@color/colorSecondary" />
+
```
d4rken, line 68: Why is this not a `ProgressLoadingButton`?

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsFragmentTest.kt
mtwalli: it is a test , won't do anything

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsEvents.kt
mtwalli: I did not  get it , I don't what value could an` Exception` bring here. also 
`Throwable.toErrorDialogBuilder(context: Context) ` is actually handling throwable

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
mtwalli: @axelherbstreith  I think you renamed it ?

Corona-Warn-App/src/main/res/layout/person_details_qr_card_item.xml
mtwalli: this button has a different outlined style than the loading button, so far I have not found a way to pass the button `style` from a custom attribute

app:buttonStyle = "@style/buttonPrimary"
app:buttonStyle = "@style/OutlinedStyle"

any ideas are very welcome ? 🙏

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
axelherbstreith: @d4rken Yes _Disease_ is the right spelling, there was one attribute with was wrong e.g. _Desease_, which I wanted to fix, maybe I messed something up there, will be changed 👍

Corona-Warn-App/src/main/res/layout/person_details_qr_card_item.xml
axelherbstreith: I will look into that 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsEvents.kt
d4rken: It's not a blocker, just not good practice: Catching Throwable; as you shouldn't catch `Error` (error is `Error : Throwable`) (i.e. https://stackoverflow.com/questions/2274102/difference-between-using-throwable-and-exception-in-a-try-catch).

The error dialog builder works with throwable as there is no downside to making this possible (although it may be missleading if it encourages catching `Throwable`).

We already catch `Throwable`  in a few places and we should refactor it at some point. It's generally okay to "accept" Throwables in data classes or e.g. the `toErrorDialogBuilder` as it makes these APIs more powerful, but catching throwables, and thus catching `Error` maybes the app more vulnerable to undefined behavior if it doesn't crash in on `Error`.

Exceptions (:smile:) to that would be top-level error handling code, e.g. debug log, error logging etc. In all other cases, a Throwable that is not `Exception : Throwable` should just crash the app.

I don't think we've had this issue in practice anywhere yet, so this is all a theoretical exercise, but it would be good code quality if we do it like that :smiling_face_with_three_hearts:.

mtwalli: I don't handle throwable btw, I just use this data class for showing error dialog purpose only instead of creating a specific errorLiveData for that. it is an event for the UI to the dialog ,but the actual error handling is catching exceptions, that is why I did not really think that it would make any difference in this case 

https://github.com/corona-warn-app/cwa-app-android/pull/3588/files#diff-d8aecc75f9c23ee2a9d44a769962d8c734ba5735cb76923ece40fa0992a0eae2R122-R125

d4rken: Ah then we missunderstood each other. I mean that if we reduce the options to use `Throwable` where it shouldn't be used, we discourage other developers from catching `Throwable`.

Corona-Warn-App/src/main/res/layout/person_details_qr_card_item.xml
mtwalli: Addressed https://github.com/corona-warn-app/cwa-app-android/pull/3668

mtwalli (dismissed):
Seems good now

d4rken (approved):
Code lgtm, did not test all screens on device.

mtwalli (approved):
Thanks Axel!

NataliaLemmerth (approved):
UA approved

PR #3592, opened 2021-06-30
Implementation to retrieve countries supported by dcc validation (EXPOSUREAPP-8051)

d4rken:
* @AlexanderAlferov You can wire up the UI by using `DccValidationRepository.dccCountries`
* @axelherbstreith Call `DccValidationRepository.refresh()` and show the loading indicator until that `refresh` call returns. If there are no errors, continue navigation, if there are errors, catch them. You could use `Throwable.toErrorDialogBuilder` to show an error to the user.
* @chris-cwa We may be able to combine some logic later on?


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationRepository.kt
```diff
@@ -0,0 +1,32 @@
...
+    suspend fun refresh() {
+        dccCountryRepository.refresh()
+        validationRulesRepository.refresh()
+    }
+
+    suspend fun clear() {
+        dccCountryRepository.clear()
+        validationRulesRepository.clear()
+    }
+}
```
d4rken, line 32: Created this as common API for the UI to work with.
After @chiljamgossow  is further along with the validation logic and @chris-cwa with the rule retrieval, we can evaluate whether the current abstraction still makes sense. For now this allows more people to work in parallel.
Removal will depend a bit on where data is accessed by injecting the repos and where data is passed via function argument.

NataliaLemmerth (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/country/server/DccCountryServer.kt
```diff
@@ -1,10 +1,76 @@
...
+@Reusable
+class DccCountryServer @Inject constructor(
+    private val api: Lazy<DccCountryApi>,
+    @CertificateValidation private val cache: Cache,
+    private val signatureValidation: SignatureValidation,
 ) {
 
-    suspend fun dccCountries(): List<DccCountry> = listOf() // TODO
+    suspend fun dccCountryJson(): String {
+        Timber.tag(TAG).d("Fetching statistics.")
```
vaubaehn, line 26: Fetching statistics or onboarded countries?

chris-cwa (dismissed):

chris-cwa (approved):

AlexanderAlferov (approved):

NataliaLemmerth (approved):
UA approved

PR #3597, opened 2021-07-01
Filter emojis in RAT profile fields (EXPOSUREAPP-8108)

mtwalli:
🎥  

https://user-images.githubusercontent.com/25054729/124124766-1ef1e780-da79-11eb-9846-a457218ac486.mp4

### Testing 
1- Create a new RAT profile 
2- Try to enter emojis
3- Enter name or address in DE, EN, TR , ... it should be as before


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/EmojiFilter.kt
```diff
@@ -0,0 +1,32 @@
...
+        source: CharSequence,
+        start: Int,
+        end: Int,
+        dest: Spanned,
+        dstart: Int,
+        dend: Int
+    ): CharSequence? {
+        for (index in start until end) {
+            val type = Character.getType(source[index]).toByte()
+            if (type in listOf(SURROGATE, OTHER_SYMBOL)) return ""
```
harambasicluka, line 28: Could you please add some tests to it so that we can sure that it doesn't interfere with other alphabets, e.g. russian with cyrillic letters (@AlexanderAlferov may provide you some test data), danish with [Ø](https://en.wikipedia.org/wiki/%C3%98).

mtwalli: Tests https://github.com/corona-warn-app/cwa-app-android/pull/3597/commits/0f801ccae0bc949b55bffb4cdd2f5dcbce1baf49

d4rken (approved):
Nice :+1:

jurajkusnier (approved):
✌️✌️✌️✌️✌️✌️✌️✌️✌️️

PR #3600, opened 2021-07-01
Business rule validation (EXPOSUREAPP-8054)

chiljamgossow:
Wrapper that calls external implementation of CertLogicEngine

- [X] unit tests -> more tests will follow in separate PR with test data from Max


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/business/wrapper/Mappings.kt
```diff
@@ -0,0 +1,149 @@
...
+import java.time.ZonedDateTime
+
+internal fun assembleExternalParameter(
+    certificate: DccData<*>,
+    validationClock: Instant,
+    countryCode: String,
+    valueSets: Map<String, List<String>> = emptyMap(),
+): ExternalParameter {
+    return ExternalParameter(
+        kid = "", // leave empty
```
d4rken, line 27: Maybe add a TODO here?

d4rken (approved):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/business/wrapper/Mappings.kt
chiljamgossow: there is one in the next PR

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/business/BusinessValidator.kt
```diff
@@ -0,0 +1,61 @@
...
+) {
+    suspend fun validate(
+        arrivalCountry: DccCountry,
+        validationClock: Instant,
+        certificate: DccData<out DccV1.MetaData>,
+    ): BusinessValidation {
+
+        // TODO update value sets
+
+        ruleRepository.refresh()
```
d4rken, line 29: I think this can be removed.
The validator works with the data that is currently available.
The refresh of countries and rules should happen in the UI, before the user is allowed to reach the screen where date and country is selected, any error during refresh at that point will prevent the user from continuing.

See:
https://github.com/corona-warn-app/cwa-app-android/pull/3588/files#diff-d8aecc75f9c23ee2a9d44a769962d8c734ba5735cb76923ece40fa0992a0eae2R117-R126

chiljamgossow: Removed in next PR https://github.com/corona-warn-app/cwa-app-android/pull/3663/files#diff-7bb7162fbf68792f8a6ebdc2dfe685c21675e1702a085d7cbe39a6ec43161ebbL26

chris-cwa (approved):

PR #3601, opened 2021-07-01
Use dgc-certlogic-android "light" dependency (EXPOSUREAPP-8209)

d4rken:
An attempt at a better solution to include the Android specific certlogic engine `dgc-certlogic-android`.

The core idea is to remove everything from  `dgc-certlogic-android` that the CWA doesn't need (or causes issues), while keeping changes to a minimum to allow easy integration of upstream changes.

The included dependency is sourced from this branch
https://github.com/corona-warn-app/dgc-certlogic-android/compare/main...corona-warn-app:light-engine-only
Instead of a jar, we include an `aar` via local maven repository to gain better handling of transitive dependencies.

See build readme here:
https://github.com/corona-warn-app/dgc-certlogic-android/blob/light-engine-only/README.md

Encountered issues:
* In `dgc-certlogic-android`, `Rule` uses `ZonedDateTime`  which is `minAPI 26`, while CWA is at `minAPI 23`, currently fixed by introducing `coreLibraryDesugaring` to the CWA
* The room dependency in `dgc-certlogic-android` has a different version than we do and if we pull it in as transitive dependency, then we get compile issues with some of our `room` annotation.
    * Solved by setting dependencies in `dgc-certlogic-android` to `compileOnly` so they are not included
    * Removing the subpackage `dgca.verifier.app.engine.data` via `Proguard/R8`


chiljamgossow (approved):
Thanks!

mtwalli (approved):
Lgtm

PR #3603, opened 2021-07-02
Unite validation repos and server (EXPOSUREAPP-7618)

chris-cwa:
Generalized Repositories vor countries and rules into validation repository.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationCache.kt
```diff
@@ -9,11 +8,11 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class DccCountryLocalCache @Inject constructor(
+class DccValidationCache @Inject constructor(
     @CertificateValidation private val cacheDir: File,
 ) {
     private val mutex = Mutex()
-    private val cacheFile = File(cacheDir, "dcc_country_cache_raw")
+    private val cacheFile = File(cacheDir, "dcc_validation_cache_raw")
```
d4rken, line 15: Maybe `dcc_validation_cache_countries_raw` ?
We'll probably not write the rules and country into the same file.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationRepository.kt
```diff
@@ -1,33 +1,95 @@
...
+    private val dispatcherProvider: DispatcherProvider,
+    @BaseGson private val gson: Gson,
+    private val countryServer: DccValidationServer,
+    private val localCache: DccValidationCache,
+    private val rulesServer: DccValidationRulesServer,
 ) {
 
-    val dccCountries: Flow<List<DccCountry>>
-        get() = dccCountryRepository.dccCountries
+    private val internalCountries: HotDataFlow<List<DccCountry>> = HotDataFlow(
```
d4rken, line 41: Would it make sense to create a common container for both countries and rules?
Do we always need the latest version of both?

```kotlin
private val internalCountries: HotDataFlow<DccValidationData> 
```

We could still keep

```
val dccCountries: Flow<List<DccCountry>> = internalCountries.data.map { it.countries }
```
Just thinking out loud, what do you think? :thinking:

d4rken (commented):
Looks good, some thinking out loud comments :coffee:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationRepository.kt
chris-cwa: i was thinking the same thing

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationCache.kt
chris-cwa: changed it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationRepository.kt
chris-cwa: introduced new general data container

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationRepository.kt
```diff
@@ -1,33 +1,92 @@
...
+
+    private fun mapCountries(rawJson: String): List<DccCountry> {
+        val countryCodes = gson.fromJson<List<String>>(rawJson)
+
+        return countryCodes.map { cc ->
+            DccCountry(countryCode = cc)
+        }
+    }
+
+    suspend fun acceptanceRules(arrivalCountry: DccCountry): List<DccValidationRule> {
```
chiljamgossow, line 74: Can we change to countryCode: String here? Same for the method below. DccCountry does not provide any benefits here.

chiljamgossow (approved):
LGTM, comment not blocking

d4rken (approved):

PR #3613, opened 2021-07-02
Fix wrong certificate issuer value being displayed (EXPOSUREAPP-8214,EXPOSUREAPP-8163)

d4rken:
* Previously used the header field accidentally because they are named very similar
* Renamed the header fields to prevent future mistakes


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/TestCertificateContainer.kt
```diff
@@ -112,18 +112,18 @@ data class TestCertificateContainer(
                 get() = testCertificate.testCenter
 
             override val certificateIssuer: String
-                get() = header.issuer
+                get() = testCertificate.certificateIssuer
```
d4rken, line 115: Actual fix

axelherbstreith (approved):

harambasicluka (approved):

PR #3614, opened 2021-07-02
Updating-FAQ-Link(EXPOSUREAPP-8216)

NataliaLemmerth:
A new FAQ link target for EN.


harambasicluka:
Please only change values-de, everything has to be delivered via the translation system. so maybe give the translators a heads up to change it accordingly.

PR #3628, opened 2021-07-05
General error message handling for invalid certificates (DEV)

chiljamgossow:
move general certificate error handling to InvalidHealthCertificateException

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/exception/InvalidVaccinationCertificateException.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/exception/InvalidVaccinationCertificateException.kt
@@ -12,35 +12,12 @@ class InvalidVaccinationCertificateException(
 ) : InvalidHealthCertificateException(errorCode, cause) {
     override fun toHumanReadableError(context: Context): HumanReadableError {
         return HumanReadableError(
-            description = errorMessage.get(context) + " ($errorCode)"
+            description = errorMessage.get(context) + " ($PREFIX$errorCode)"
         )
     }
 
-    override val showFaqButton: Boolean
-        get() = errorCode in codesVcInvalid
-    override val faqButtonText: Int = R.string.error_button_vc_faq
-    override val faqLink: Int = R.string.error_button_vc_faq_link
-
-    private val codesVcInvalid = listOf(
-        ErrorCode.HC_BASE45_DECODING_FAILED,
-        ErrorCode.HC_CBOR_DECODING_FAILED,
-        ErrorCode.HC_COSE_MESSAGE_INVALID,
-        ErrorCode.HC_ZLIB_DECOMPRESSION_FAILED,
-        ErrorCode.HC_COSE_TAG_INVALID,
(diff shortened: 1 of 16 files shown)


NataliaLemmerth (dismissed):
UA approved

harambasicluka (changes requested):
Please change the id only in german and add it to `values` (do not delete the old ones), everything else will be handled by the translation delivery.

Corona-Warn-App/src/main/res/values-de/vaccination_strings.xml
```diff
@@ -109,7 +111,12 @@
     <!-- XBUT: Text for invalid certificate error button, linking to FAQ-->
     <string name="error_button_vc_faq">FAQ zu Zertifikaten</string>
     <!-- XTXT: Explains user about vaccination certificate: URL, has to be "translated" into english (relevant for all languages except german) - https://www.coronawarn.app/en/faq/#vac_cert_invalid -->
-    <string name="error_button_vc_faq_link">https://www.coronawarn.app/de/faq/#hc_prefix_invalid</string>
+    <string name="error_button_vc_faq_link">https://www.coronawarn.app/en/faq/#vac_cert_invalid</string>
+
+    <!-- XBUT: Text for invalid certificate error button, linking to FAQ-->
+    <string name="error_button_dcc_faq">FAQ zu Zertifikaten</string>
+    <!-- XTXT: Explains user about vaccination certificate: URL, has to be "translated" into english (relevant for all languages except german) - https://www.coronawarn.app/en/faq/#vac_cert_invalid -->
+    <string name="error_button_dcc_faq_link">https://www.coronawarn.app/de/faq/#hc_prefix_invalid</string>
```
mtwalli, line 119: Please add the new strings in `covid_certificates_strings.xml` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/exception/InvalidVaccinationCertificateException.kt
```diff
@@ -57,6 +34,7 @@ class InvalidVaccinationCertificateException(
         }
 }
 
-private const val ERROR_MESSAGE_VC_INVALID = R.string.error_vc_invalid
+private const val PREFIX = "VC_"
```
mtwalli, line 37: is it for Vaccination certificates error codes or for all types of certificates? and is it required to add this prefix ?
I found very difficult to find the actual error code , if you look for `VC_PREFIX_INVALID` as the error states

chiljamgossow: The prefix is part of the tech spec for vaccination - it is only affecting error messages specific to vaccination. The idea is that InvalidVaccinationCertificateException handles specific errors/ specific error messages for vaccination, while InvalidHealthCertificateException has general errors for all certificate types.

harambasicluka (commented):
strings look fine! but somehow the linter complains O.o

mtwalli (dismissed):
LGTM

harambasicluka (approved):

PR #3634, opened 2021-07-05
Add JsonSchema based validation to DCC data extraction (EXPOSUREAPP-8257)

d4rken:
* Add JsonSchema based DCC validation
* Will be applied to all newly scanned certificates (mode = `STRICT`)
* Will not be applied to already stored certificates (mode = `LENIENT`) due to being potentially more strict that our previous checks
* Will be used for certificates during the new validation feature
   * /cc @chiljamgossow
   * `DccJsonSchema.rawSchema` will give you the raw schema string to pass into the rule engine
   * `DccJsonSchemaValidator.isValid` to set `jsonSchemaCheckPassed`
* This PR adds the latest (`06fe7ed`) from the [tech-spec](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/multi-user-dcc/json-schemas/schemas/digital-green-certificate.json) though for compatibility i had to change `schema` to `"$schema": "https://json-schema.org/draft/2019-09/schema",`
   *  /cc @mlenkeit 
   * I just changed the attribute and it seems to work but we need to check it explicitly


Corona-Warn-App/src/main/assets/jsonschema-dcc-06fe7ed.json
```diff
@@ -0,0 +1,339 @@
+{
+  "$schema": "https://json-schema.org/draft/2019-09/schema",
```
d4rken, line 2: I just downgraded this :thinking:

Corona-Warn-App/build.gradle
```diff
@@ -252,6 +252,7 @@ android {
         }
         test {
             java.srcDirs += "$projectDir/src/testShared/java"
+            resources.srcDirs += "src/main/assets"
```
d4rken, line 255: So we do some mocking :mage: and load the actual schema in unit tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
```diff
@@ -124,11 +126,32 @@ class DccV1Parser @Inject constructor(
...
+            )
+        }
+        Mode.CERT_REC_LENIENT,
+        Mode.CERT_TEST_LENIENT,
+        Mode.CERT_VAC_LENIENT -> {
+            // We don't check schema in lenient mode, it may affect already stored certificates.
+            this
+        }
+    }
+
```
d4rken, line 147: This is the new check.

We don't need to return `this` and could just use `apply`, but by using the return value, we get a compiler check forcing us to make `when` exhaustive.

Corona-Warn-App/build.gradle
```diff
@@ -440,6 +441,8 @@ dependencies {
     implementation 'com.google.guava:guava:30.0-android'
     implementation 'joda-time:joda-time:2.10.6'
 
+    implementation 'com.networknt:json-schema-validator:1.0.56'
```
d4rken, line 444: I picked this library as it seems well maintained and  internally uses jackson which we will get as dependency via https://github.com/eu-digital-green-certificates/dgc-certlogic-android anyways.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
```diff
@@ -13,18 +13,21 @@ import javax.inject.Inject
...
-            this[keyEuDgcV1]?.run {
-                val (rawBody, dcc) = this.toCertificate()
-                Body(
-                    raw = rawBody,
-                    parsed = dcc.toValidated(mode)
-                )
-            } ?: throw InvalidVaccinationCertificateException(ErrorCode.HC_CWT_NO_DGC)
-        } ?: throw InvalidVaccinationCertificateException(ErrorCode.HC_CWT_NO_HCERT)
+        val dgcCbor = map[keyHCert].let {
+            if (it == null) throw InvalidVaccinationCertificateException(ErrorCode.HC_CWT_NO_HCERT)
```
chiljamgossow, line 21: It should be InvalidHealthCertificateException throughout this class. I thought I changed that, but it is still there

d4rken: Changed it. Good catch.

chiljamgossow (approved):
LGTM

mtwalli (approved):

PR #3642, opened 2021-07-06
One dose info (EXPOSUREAPP-8264)

mtwalli:
Screenshot 2021-07-06 at 10 37 14


BMItr:
@mtwalli is the new card only shon if its a one-dosis-shot?

mtwalli:
> @mtwalli is the new card only shon if its a one-dosis-shot?

Yes, as I understand

BMItr (approved):
nothing to complain

harambasicluka (approved):

PR #3644, opened 2021-07-06
New text for tile

ChristinaTsigganos:
IMPORTANT: Please take over this button text asap so that translation can consider it today and Android and iOS are the same.


mtwalli (approved):

NataliaLemmerth (approved):
UA approved

Issue #3638, opened 2021-07-05
"Test result certificate can't be downloaded #3571" is again here


TanCari:
I could not download the certificate for rapid tests, precisely as described in issue #3571. After the update to version 2.4.3 everything worked fine. Over the week end I could download two certificates correctly. Now I have exactly the same problem.
I beg for help, as I have to fly very early on wednesday: After seeing certificates on my mobile I thought the problem was definitely over. Thanx!

---
Internal Tracking ID: [EXPOSUREAPP-8162](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8162)

thomasaugsten:
Can you post an screenshot of the detail error

TanCari:
> Can you post an screenshot of the detail error

Yeah, just the time to prepare it now and I post it. The exception is extremely similar to the one in in #3571, I would say.

TanCari:
> Can you post an screenshot of the detail error

![Screenshot_20210706-001221](https://user-images.githubusercontent.com/35867729/124521314-5a7d1080-ddef-11eb-9e47-b978295f3189.jpg)

TanCari:
Possibly android was automatically updated between the last achieved certificate and this failed attempt. I'm at disposal for all further infomation needed.

vaubaehn:
@TanCari 
Can you try again? There is always a low chance for a short temporary server maintenance...
As I can reach the server now, you're probably able to get it now?

@dsarkar it's past midnight! Are you never sleeping? 😉

TanCari:
@vaubaehn I have tried it again and I astill get the same error. Also restarted the device and the app: it did not help.

vaubaehn:
@TanCari 
Thanks for trying. Since the weekend there was no update of CWA. I still believe this is most likely a server issue.
@thomasaugsten will handle it best way, I'm sure

vaubaehn:
@TanCari 
Now I read that you think your whole Android was updated? Full version upgrade?

@thomasaugsten is the key ring (private/public) mechanism for test data encryption on Android dependent on Android keystore? But in any case for every test a complete new key would be generated, there is no dependency on a master key, is there?

thomasaugsten:
The keys has no depending. You can close the app and try again on mobile network.

TanCari:
[...]

PR #3650, opened 2021-07-06
Generate test certificate rsa key-pair only once (EXPOSUREAPP-8162)

d4rken:
Generate RSA key-pair once on registration, instead "everytime until we have successfully registered a pair".

This prevents a stale-mate between client and server where the client tries to register a publickey but server returns HTTP409.
But if the client tried to register a new key pair, then the app doesn't have the already registered key that the server is referring to. See #3638

This PR changes the RSA key-pair generation to happen once when the test is registered, so we can't loose a registered key if a client/server errors during registration and we can just continue if the server responds with HTTP409.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt
@@ -140,7 +140,7 @@ class TestCertificateProcessor @Inject constructor(
             )
         } catch (e: Throwable) {
             Timber.tag(TAG).e(e, "RSA_DECRYPTION_FAILED")
-            throw InvalidTestCertificateException(InvalidHealthCertificateException.ErrorCode.RSA_DECRYPTION_FAILED)
+            throw InvalidTestCertificateException(InvalidHealthCertificateException.ErrorCode.RSA_DECRYPTION_FAILED, e)
         }
 
         val extractedData = qrCodeExtractor.extractEncrypted(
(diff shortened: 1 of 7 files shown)


BMItr (approved):
lgtm 👌

mtwalli (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt
```diff
@@ -51,30 +49,31 @@ class TestCertificateProcessor @Inject constructor(
             return data
         }
 
-        val rsaKeyPair = try {
```
chiljamgossow, line 54: How is this resolved to a human readable error?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt
```diff
@@ -51,30 +49,31 @@ class TestCertificateProcessor @Inject constructor(
...
-            rsaKeyPairGenerator.generate()
-        } catch (e: Throwable) {
-            throw InvalidTestCertificateException(InvalidHealthCertificateException.ErrorCode.RSA_KP_GENERATION_FAILED)
-        }
+        val publicKey = data.rsaPublicKey ?: throw IllegalArgumentException("rsaPublicKey was null")
```
chiljamgossow, line 52: How is this resolved to a human readable error?

d4rken: It's not resolved, but it is also not something that should be able to happen.

Every test with `publicKeyRegisteredAt == null` came from `TestCertificateRepository.requestCertificate` and has a keypair:

https://github.com/corona-warn-app/cwa-app-android/pull/3650/files/82e7e7bdb87457048027636a02e7472bdf47d668#diff-91615199099cc98a19cabc6f30424cabf97e9663712eb96a68d38b25efba977fR140-R142

The keypair can only be null, if the certificate was added via qrcode and then  `publicKeyRegisteredAt` is not null, 

https://github.com/corona-warn-app/cwa-app-android/pull/3650/files/82e7e7bdb87457048027636a02e7472bdf47d668#diff-91615199099cc98a19cabc6f30424cabf97e9663712eb96a68d38b25efba977fR183

and we should not get here due to:
https://github.com/corona-warn-app/cwa-app-android/pull/3650/files/82e7e7bdb87457048027636a02e7472bdf47d668#diff-3b73348d0d181c4b360ab8489616832b6f4e8f1dd5936f7f25eaa585e19ab048R47-R50

d4rken: You are right, already affected users would see the "not something that should be able to happen" error :thinking:

PR #3656, opened 2021-07-06
Additional edgecase handling for failed public key registration (EXPOSUREAPP-8162)

### LINKED ISSUE #3638: "Test result certificate can't be downloaded #3571" is again here

d4rken:
This is a follow up to PR #3650 which fixes #3638, by @mtwalli and me.

After further discussion about the error case handling here:
https://github.com/corona-warn-app/cwa-app-android/pull/3650#pullrequestreview-699953584
We noticed that the bugfix introduces new edge cases that could accidentally causes more users to be affected by #3638 due to the app updating :dizzy_face:.

After the update we have 3 types of users:
* `Normal user` who only registered a test after update to `2.5.x` :heavy_check_mark: 
* `3638 user` who was affected by the bug on `2.4.x` :frowning: 
* `unlucky user` who updated the the CWA to `2.5.x` after they have requested a certificate and before they have registered a public key. They are already past the creation phase before the new code to generate RSA keypairs existed. So they will attempt to register a public key without one. :fearful: 

Edgecase `unlucky user` requires:
* We generate a new RSA key pair if none exists, try it and see what the server responds
* If the server responds with `HTTP 409` it was actually a `3638 user`, but no harm done
* If the server responds with success, we helped the user past the edge case and all is well again
* If the server throws an error different from `HTTP 409` (connection reset, http 500 etc.) then we need to save the keypair we generated for the edge case handling, otherwise `unlucky user` could become `3638 user' if we retry again later, and the server responds with `HTTP 409`.     

Also users affected by #3638 will see a specific error message that retrying has no chance of success and the certificate request can be deleted.

Additional details:
https://github.com/corona-warn-app/cwa-app-android/blob/a79fd39edff8a587390a3b689d3b8e4e93300975/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt#L55-L122


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt
```diff
@@ -152,11 +155,10 @@ class PersonOverviewViewModel @AssistedInject constructor(
...
 
-    fun refreshCertificate(containerId: TestCertificateContainerId) =
-        launch {
-            val error = testCertificateRepository.refresh(containerId).mapNotNull { it.error }.singleOrNull()
-            error?.let { events.postValue(ShowRefreshErrorDialog(error)) }
-        }
+    fun refreshCertificate(containerId: TestCertificateContainerId) = launch(scope = appScope) {
+        val error = testCertificateRepository.refresh(containerId).mapNotNull { it.error }.singleOrNull()
+        error?.let { events.postValue(ShowRefreshErrorDialog(error)) }
+    }
```
d4rken, line 161: While the `refresh` call that happens automatically, happens on a worker, that should not be canceled mid process, the refresh call via the button in the UI runs the `viewModelScope` which is canceled when the screen is left.

But:
I don't think it's the case anymore, switching this to `AppScope` doesn't hurt.

The key registration itself happens via `HotData.updateBlocking` which passes the execution to the scope of `HotData` which already is `AppScope`:

```kotlin
    suspend fun updateBlocking(action: suspend T.() -> T): T {
        val update: Update<T> = Update(onModify = action)
        updateActions.tryEmit(update) // Passes the action to be executed on a different scope

        Timber.tag(tag).v("Waiting for update.")
        val ourUpdate = internalFlow.first { it.updatedBy == update } // Only waiting for this would be canceled

        ourUpdate.error?.let { throw it }

        return ourUpdate.value
    }
```

d4rken:
I'm still confused on how the initial bug can happen. :thinking: 
The only chance I see is to kill the app in a specific <10ms window after pressing the refresh button in the UI.
Sure we have many installs but not all installs register a new test at every launch and kill the app as fast as possible... :shrug:
With this and the previous PR, if it happens again, it should be recoverable though.

chris-cwa (dismissed):

d4rken:
I think error case handling in `TestCertificateServer.registerPublicKeyForTest` is not working correctly, but it's late today.
Setting this to draft so it doesn't get merged yet before we have a look tomorrow.

d4rken:
The error mapping issue mentioned here:
https://github.com/corona-warn-app/cwa-app-android/pull/3656#issuecomment-874955064
should be fixed by
https://github.com/corona-warn-app/cwa-app-android/pull/3656/commits/b5afae728d1f20ab70c7f926084a23a9ec49f0e8

I also noticed an additional issue where we could end up in a refresh loop, this is fixed by:
https://github.com/corona-warn-app/cwa-app-android/pull/3656/commits/732c6df736466d2c3b8041be518c270c13094bd8
this could actually have increased the chance for this error, as we did more server calls than necessary, which theoretically means more chances for bug #3638 to happen.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt
```diff
@@ -28,53 +29,103 @@ class TestCertificateProcessor @Inject constructor(
...
+
+        val (publicKey, privateKey) = when {
+            data.rsaPublicKey != null && data.rsaPrivateKey != null -> {
+                // Normal case, since ~2.5.1 (PR #3650)
+                Timber.tag(TAG).v("Using existing RSA KeyPair.")
+                data.rsaPublicKey!! to data.rsaPrivateKey!!
+            }
+            !isMissingKeyPairEdgeCase -> {
+                // Shouldn't happen
+                throw IllegalArgumentException("Missing keypair")
```
mtwalli, line 77: I think the message is wrong here , right? the keyPair is not missing ,but we should be here anyway

mtwalli (approved):
I think Illegal exception message is wrong, other than that I did a few test rounds and it looks good to me,
In 2.5 Tested:
1- Test certificate for RAT (Negative + pending -> Negative )
2- Test certificate for PCR (Negative + pending -> Negative )
3- Tested deletion 

In 2.4 Tested:
1- Test certificate is pending -> upgrade to 2.5 -> retrieve Test certificate

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/server/TestCertificateServer.kt
```diff
@@ -43,63 +44,63 @@ class TestCertificateServer @Inject constructor(
...
-                409 -> throw TestCertificateServerException(PKR_409)
-                500 -> throw TestCertificateServerException(PKR_500)
-            }
         } catch (e: Exception) {
             Timber.tag(TAG).w(e, "registerPublicKeyForTest failed")
             throw when (e) {
-                is TestCertificateServerException -> e
-                else -> TestCertificateServerException(PKR_FAILED, e)
+                is CwaWebException -> when (e.statusCode) {
+                    400 -> TestCertificateException(PKR_400, e)
```
BMItr, line 67: ⬇️  imho it would be a little bit nice to use a networkInterceptor. Benefit would be to see the original response without okhttp interfering..
But optional. Will work this way too of course

BMItr (approved):
code lgtm. (one optional remark)
Tested various upgrade scenarios (2.4 -> 2.5) on oneplus 8 device. 
👌

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/server/TestCertificateServer.kt
d4rken: I contemplated this, but decided against it as it would make an already complex PR more complex: Interceptor vs just try+catch and the interceptor would "hide" within the module that creates the retrofit API, outside of the server class.

Overall I actually prefer the NetworkIntercepter approach, but we also already have a NetworkInterceptor for our `CwaWebExceptions`, introducing another one without more significant refactoring feels like there could be pitfalls not something I'd want to tackle now with the time pressure for 2.5.x.

My suggestion would be to keep this, and then refactor the whole handling independently. We use this kind of error mapping in other places too, that could then also get an interceptor, or we come up with a more elaborate pattern that can be reused.

@harambasicluka JIRA ticket for the refactoring epic? "Refactor HTTP error handling & mapping"?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateProcessor.kt
d4rken: You are right the message doesn't fit, but unless we find something that needs to change, I'd keep it for now and not retrigger a review. It's not something that should be possible to happen anyways.

I'll fix it for 2.6.x

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/server/TestCertificateServer.kt
BMItr: agreed.

PR #3658, opened 2021-07-07
Validation repo (EXPOSUREAPP-7618)

chris-cwa:
Implementation of the Validation rule API, rule Server and caching.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationCache.kt
```diff
@@ -13,31 +14,45 @@ class DccValidationCache @Inject constructor(
...
+        invalidationRulesCacheFile.save(data, "raw invalidation rules")
+
+    @VisibleForTesting
+    internal suspend fun File.load(descr: String): String? = mutex.withLock {
         try {
-            if (countryCacheFile.exists()) countryCacheFile.readText() else null
+            if (exists()) readText() else null
         } catch (e: Exception) {
-            Timber.tag(TAG).e(e, "Failed to load raw dcc countries from cache.")
+            Timber.tag(TAG).e(e, "Failed to load $descr from cache.")
```
d4rken, line 38: Alternatively, use the filename to differentiate log messages.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/rule/DccValidationRuleSet.kt
```diff
@@ -0,0 +1,12 @@
+package de.rki.coronawarnapp.covidcertificate.validation.core.rule
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+@Parcelize
+data class DccValidationRuleSet(
```
d4rken, line 7: Where is this used?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/server/DccValidationServer.kt
```diff
@@ -24,17 +30,27 @@ class DccValidationServer @Inject constructor(
...
 
     private val dccValidationRuleApi: DccValidationRuleApi
         get() = rulesApi.get()
 
-    suspend fun ruleSet(ruleTypeDcc: DccValidationRule.Type): Set<DccValidationRule> {
-        return when (ruleTypeDcc) {
-            DccValidationRule.Type.ACCEPTANCE -> dccValidationRuleApi.acceptanceRules()
-            DccValidationRule.Type.INVALIDATION -> dccValidationRuleApi.invalidationRules()
+    suspend fun ruleSetJson(ruleTypeDcc: DccValidationRule.Type): String =
+        withContext(dispatcherProvider.Default) {
```
d4rken, line 40: We shouldn't explicitly switch to the `Default`, either no switch (and let the caller decide) or switch to `IO` (probably best). Due to the amount of threads in `Default` being capped, a http request timing out could block a thread for 10-20 seconds potentially starving other coroutines.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/server/DccValidationServer.kt
```diff
@@ -81,6 +97,29 @@ class DccValidationServer @Inject constructor(
         }
     }
 
+    private fun ResponseBody.parseBody() = parseBody(byteStream())
+
+    @VisibleForTesting
+    internal fun parseBody(inputStream: InputStream): String {
```
d4rken, line 103: Can parseBody also be used for the countries?

d4rken (commented):
Thanks for the refactoring!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/rule/DccValidationRuleSet.kt
chris-cwa: it is not :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/server/DccValidationServer.kt
chris-cwa: there sure is a lot of duplicate code!

chiljamgossow (dismissed):
Merge and test?

d4rken (dismissed):

d4rken (approved):

chiljamgossow (approved):
Looking good

PR #3593, opened 2021-06-30
Validation Rules Result Screen - valid (EXPOSUREAPP-8031)

BMItr:
This PR addresses EXPOSUREAPP-8031 (validation Result Screen - valid)
PR will be set to **DRAFT** because Button & Date - Wiring needed after Repo and Screens before this one are completed/merged (see EXPOSUREAPP-7738 for the flow)

Screenshot (scrolled down a little):

screen

PR #3664, opened 2021-07-08
Add text for clarity

ChristinaTsigganos:
Adding "in der Corona-Warn-App" to the sentence for more clarity after we received feedback from the community here.


mtwalli (approved):

d4rken (approved):

vaubaehn:
Hi @ChristinaTsigganos ,
I think you would need to repeat the change, as the change here went into https://github.com/corona-warn-app/cwa-app-android/pull/3593, that was dismissed in favor of #3678 . So the change seems to be gone...

Current branach 2.6.x:
https://github.com/corona-warn-app/cwa-app-android/blob/913147b7918a737beef74cfdaf808bf568034f2a/Corona-Warn-App/src/main/res/values-de/covid_certificate_strings.xml#L172

Anyway, have a nice week-end!

/cc @harambasicluka

PR #3667, opened 2021-07-08
Validation state (EXPOSUREAPP-8053)

chiljamgossow:
- [X] unit test


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationTest.kt
```diff
@@ -0,0 +1,102 @@
...
+import io.kotest.matchers.shouldBe
+import org.joda.time.Instant
+import org.junit.jupiter.api.Test
+import testhelpers.BaseTest
+
+class DccValidationTest : BaseTest() {
+
+    private val userInput = ValidationUserInput(
+        DccCountry("PT"),
+        Instant.now()
```
d4rken, line 16: Please avoid `Instant.now()` (hardcode timestamps) as it could make tests flaky.

d4rken (commented):
Looks good, would prefer if we remove the variable timesource in the tests.

d4rken (approved):

mtwalli (approved):

PR #3669, opened 2021-07-08
Pick highest version of rule (DEV)

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/business/wrapper/Mappings.kt
```diff
@@ -140,9 +141,39 @@ internal fun List<DccValidationRule>.filterRelevantRules(
...
+                if (word1 > word2) {
+                    return 1
+                }
+            }
+        }
+        return version1.size - version2.size
     }
 }
 
+internal fun String.parseVersion(): List<String> = this.split('.')
```
d4rken, line 175: What do we do if the version scheme is incorrect?
Should we enforce that?
e.g. 
```kotlin
Regex("(\d+)\.(\d+)\.(\d+))")
```
if it does not match, assume version is `0` lower then all others?

/cc @mlenkeit

mlenkeit: It's enforced on the server by schema validation; should never happen ™️. I would suggest to handle this case in some way that does not crash the client and log an error message.

mlenkeit: Version `0` as fallback in case parsing fails seems like a good idea

d4rken (dismissed):

d4rken (dismissed):

d4rken (approved):

BMItr (approved):
lgtm

PR #3680, opened 2021-07-09
Improve `stableId` for person details list (EXPOSUREAPP-8325)

d4rken:
Certificates are stored "per type" in different repositories, and the `certificateId` uniqueness check is "per type".
So it's technically possible for a single person two have 3 certificates with the same `certificateId`.

If the amount of entries and position of certificates in the persons certificate details view aligns correctly :star2:, then the  RecyclerView can throw

```java
java.lang.IllegalStateException: Two different ViewHolders have the same stable ID. Stable IDs in your adapter MUST BE unique and SHOULD NOT change.
 ViewHolder 1:TestCertificateCard{f81f0df position=5 id=1507681495, oldPos=-1, pLpos:-1 not recyclable(1)} 
 View Holder 2:TestCertificateCard{d6ed6f7 position=4 id=1507681495, oldPos=-1, pLpos:-1 not recyclable(1)} androidx.recyclerview.widget.RecyclerView{765ba6b VFED..... ......ID 0,168-1080,2229 #7f0a059d app:id/recycler_view_certificates_list}, adapter:de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsAdapter@ae64dca, layout:androidx.recyclerview.widget.LinearLayoutManager@3fedb3b, context:de.rki.coronawarnapp.ui.main.MainActivity@c022ddf
```

because the `getItemId() = certificateId.hashCode()` collide.


d4rken:
Not certain if this still goes into `2.5.x` :thinking:

harambasicluka:
> Not certain if this still goes into `2.5.x` 🤔

You are right, 2.5.x is closed for now. This should go into 2.6.x.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/VaccinationInfoCard.kt
```diff
@@ -39,6 +39,6 @@ class VaccinationInfoCard(parent: ViewGroup) :
         val daysUntilImmunity: Int?
     ) : CertificateItem, HasPayloadDiffer {
         override fun diffPayload(old: Any, new: Any): Any? = if (old::class == new::class) new else null
-        override val stableId = daysUntilImmunity.hashCode().toLong()
+        override val stableId = Item::class.hashCode().toLong()
```
mtwalli, line 42: it was originally this way ,but we used the daysUntilImmunity to update the card when time changes

d4rken: Hm, the hashcode of INT is the INT itself, isn't it?

I think it might have been the cause for this seemingly random crash:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8230

if the time changes, the `stableId` should stay fixed though for the info card, as it is still the same item, just it's content changed.

d4rken: We do use the latest payload too:
```
        val daysUntilImmunity = curItem.daysUntilImmunity
```
are we sure that it doesn't work?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/CertificateItem.kt
```diff
@@ -1,5 +1,13 @@
...
 
-interface CertificateItem : HasStableId
+interface CertificateItem : HasStableId {
+
+    fun CwaCovidCertificate.toStableListId(): Long = run {
+        var result = this::class.hashCode()
+        result = 31 * result + certificateId.hashCode()
+        result.toLong()
+    }
+}
```
d4rken, line 13: @d4rken Check on monday whether it would make more sense to use the `containerId` here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/VaccinationInfoCard.kt
mtwalli: Works

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/CertificateItem.kt
```diff
@@ -1,5 +1,13 @@
...
 import de.rki.coronawarnapp.util.lists.HasStableId
 
-interface CertificateItem : HasStableId
+interface CertificateItem : HasStableId {
+
+    fun CwaCovidCertificate.toStableListId(): Long = run {
+        var result = this::class.hashCode()
+        result = 31 * result + certificateId.hashCode()
+        result.toLong()
+    }
```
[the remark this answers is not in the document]
d4rken: Yeah I agree, `containerId` would be better, I'll adjust it.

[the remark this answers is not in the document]
d4rken: I've changed it to containerId, but had to adjust containerID to not be a data class, otherwise it being a `data class` would override the equals/hashcode and only consider the argument string (which may be the same certificateIds), see:
https://github.com/corona-warn-app/cwa-app-android/pull/3680/commits/1a1e5ad49c8d019624c68f6879bdd33637555a4b

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/VaccinationInfoCard.kt
d4rken: I've tried it.
There is no timer on that screen that would cause a live update, but if there would be, the value would change if the  `data class Item` is updated.

mtwalli (approved):
LGTM

BMItr (approved):
👌

MikeMcC399:
Link back to #3684

PR #3683, opened 2021-07-10
#3629 :: Check for last update in future

### LINKED ISSUE #3629: Risiko-Ermittlung zeigt falsches Datum/aktualisiert nicht mehr nachdem Systemdatum verstellt wurde.
## Avoid duplicates
* [?] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)

## Technical details

- Device name: Samsung Galaxy S10e
- Android version: 11 (Kernel 4.14.113-21644994)
- App version: 2.4.3

## Describe the bug

Meine Frau und ich hatten am gleichen Tag die Impfung, aber unterschiedliche Anzeigen wie lange es noch bis zur vollständigen Impfung dauerte (ich hatte noch Version 2.4.2, das war wohl der Grund). Bei der Suche nach dem Grund stolperte ich über einen Bericht zu  #3532 und probierte aus, mein Systemdatum auf den 1.8. zu stellen. Seit dem zeigt die Risikoermittlung "Aktualisiert: 1. Aug., 9:21". Ich bin nicht sicher, ob dennoch Aktualisierungen stattfinden oder ob ich jetzt 4 Wochen lang keine Aktualisierungen mehr erhalte.

1. Change the system date to the future
2. Open the CWA
3. Wait for Risk assessment
4. Reset system data to update automatically
5. Open the CWA
6. Witness the future date still displayed as "Aktualisiert"
-->

## Expected behaviour

"Aktualisiert" should be updated even if current date is before a fictive last date from the future.

## Additional context

---
 Internal Tracking ID: [EXPOSUREAPP-10253](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10253)
Internal Tracking ID: [EXPOSUREAPP-9077](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9077)

hd42:
Proposed Code change to fix #3629 .

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
@@ -174,6 +174,12 @@ class DownloadDiagnosisKeysTask @Inject constructor(
 
         val hasRecentDetection = nextForcedDetectionAt != null && now.isBefore(nextForcedDetectionAt)
 
+        val lastSuccesfulDetectionInFuture = nextForcedDetectionAt?.isBefore(now)
+        if(lastSuccesfulDetectionInFuture!=null&&lastSuccesfulDetectionInFuture){
+            Timber.tag(TAG).w("Last detection is more than 24h in the future. Assuming system date was wrong during last update.")
+            return false;
+        }
+
         return (hasRecentDetection && keySyncResult.newKeys.isEmpty()).also {
             if (it) Timber.tag(TAG).w("Aborting. Last detection is recent (<24h) and no new keyfiles.")
         }


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -174,6 +174,12 @@ class DownloadDiagnosisKeysTask @Inject constructor(
 
         val hasRecentDetection = nextForcedDetectionAt != null && now.isBefore(nextForcedDetectionAt)
 
+        val lastSuccesfulDetectionInFuture = nextForcedDetectionAt?.isBefore(now)
+        if(lastSuccesfulDetectionInFuture!=null&&lastSuccesfulDetectionInFuture){
+            Timber.tag(TAG).w("Last detection is more than 24h in the future. Assuming system date was wrong during last update.")
+            return false;
+        }
+
         return (hasRecentDetection && keySyncResult.newKeys.isEmpty()).also {
```
d4rken, line 183: Even if there was timetraveling, the `hasRecentDetectionAndNoNewFiles` would NOT abort, if new files have been downloaded as `keySyncResult.newKeys` should not be empty.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DownloadDiagnosisKeysTask.kt
```diff
@@ -174,6 +174,12 @@ class DownloadDiagnosisKeysTask @Inject constructor(
 
         val hasRecentDetection = nextForcedDetectionAt != null && now.isBefore(nextForcedDetectionAt)
 
+        val lastSuccesfulDetectionInFuture = nextForcedDetectionAt?.isBefore(now)
```
d4rken, line 177: The check seems wrong, shouldn't it be:

```kotlin
val lastSuccesfulDetectionInFuture = lastSuccessfulDetection?.startedAt?.isAfter(now) ?: false
```

* If `lastSuccessfulDetection` is in the distant past (time travel backwards), then `hasRecentDetection` is false.
* If `lastSuccessfulDetection` is in the distant future (time travel forwards), then `now.isBefore(nextForcedDetectionAt)` is always true, and for more than the 1 day we add, causing this to be `true` until the present has caught up with the future.
* But `lastSuccessfulDetection?.startedAt` should never be after or equal to the current time, except if time travel into the future occurs.

d4rken:
I don't think this is the solution to #3629 and it doesn't pass CI, so I'm closing it until we have found a better solution (and you are back from vacation).

PR #3686, opened 2021-07-12
Dcc Validation Open Screen (EXPOSUREAPP-8061)

jurajkusnier:
This PR wires up the logic to put the right items in the RecyclerView.
There will be follow-up PRs for polishing/finishing the list items.

screen_screenshots

If you want to test it, see here for some test setup: https://github.com/corona-warn-app/cwa-app-android/commit/07a37a1069bde91f2c373069ac4fa6e3fc34fea1


NataliaLemmerth (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/RuleHeaderVH.kt
```diff
@@ -34,7 +35,12 @@ class RuleHeaderVH(
                 subtitleText.text = text
             }
             DccValidation.State.OPEN -> {
-                titleText.setText(R.string.validation_rules_open_header_title)
+                if (item.showTitle) {
+                    titleText.isVisible = true
+                    titleText.setText(R.string.validation_rules_open_header_title)
+                } else {
+                    titleText.isGone = true
+                }
```
mtwalli, line 43: what about ?
```kotlin
 titleText.apply{
   isVisible = item.showTitle
   setText(R.string.validation_rules_open_header_title)
}
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/open/DccValidationOpenFragment.kt
```diff
@@ -1,5 +1,61 @@
...
+            }
+
+            toolbar.setNavigationOnClickListener { popBackStack() }
+        }
+
+        viewModel.listItems.observe2(this) {
+            validationResultAdapter.update(it)
+        }
+
+        super.onViewCreated(view, savedInstanceState)
```
mtwalli, line 59: I think it is not necessary to call super method here , it is empty actually , right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/open/DccValidationOpenViewModel.kt
```diff
@@ -0,0 +1,63 @@
...
+            DccValidation.State.PASSED -> {
+                Timber.e("State PASSED but we are on screen 'OPEN', wrong navigation?")
+            }
+            DccValidation.State.TECHNICAL_FAILURE -> {
+                Timber.e("State TECHNICAL_FAILURE but we are on screen 'OPEN', wrong navigation?")
+            }
+            DccValidation.State.FAILURE -> {
+                Timber.e("State FAILURE but we are on screen 'OPEN', wrong navigation?")
+            }
+        }
```
mtwalli, line 52: I would prefer the same way as in `DccValidationPassedViewModel` 
something like:
```kotlin
  if (validation.state != DccValidation.State.OPEN) {
            throw IllegalStateException(
                "Expected validation state to be ${DccValidation.State.OPEN.name} " +
                    "but is ${validation.state.name}"
            )
        }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/passed/DccValidationPassedViewModel.kt
```diff
@@ -42,7 +42,7 @@ class DccValidationPassedViewModel @AssistedInject constructor(
         return listOf(
             ValidationInputVH.Item(validation),
             ValidationOverallResultVH.Item(validation.state),
-            RuleHeaderVH.Item(type = validation.state, validation.rules.size),
+            RuleHeaderVH.Item(type = validation.state, ruleCount = validation.rules.size),
```
mtwalli, line 45: can't we just pass `validation` and display what you is needed ? instead of de-structuring it in the args ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/open/DccValidationOpenFragment.kt
d4rken: I'd prefer if we keep calling super on the standard lifecycle methods from Fragments and Activities, while they may be empty now, if we extend fragments for something in the future, or AndroidX libraries change Fragment implementation details, we'd end up with weird bugs.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/passed/DccValidationPassedViewModel.kt
jurajkusnier: I don't think so. We could use the whole `DccValidation` object in `DccValidationPassedViewModel` but in other viewmodels for example `DccValidationOpenViewModel` we use state/result for individual rules. It's basically a header for a group of rules not for the whole fragment.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/open/DccValidationOpenViewModel.kt
jurajkusnier: Yeah, looks better. Updated

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/RuleHeaderVH.kt
jurajkusnier: updated

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/passed/DccValidationPassedViewModel.kt
mtwalli: Ok

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/open/DccValidationOpenFragment.kt
mtwalli: I don't mind as I said it is actually empty, but we don't do this btw anywhere else in the App  , does this mean that we will have bugs everywhere ?. I think the only reason why it is used here is `apply`

if we will use super methods , then I think it should be moved as first line in the method body , not a the end ,right?

jurajkusnier: yes, it makes sense, updated

mtwalli (approved):
Thanks!

BMItr (approved):
lgtm!!

NataliaLemmerth (approved):
UA approved

vaubaehn:
Please allow me one remark:
I'm very happy that the UI/UX is different from what was suggested here: https://ec.europa.eu/health/sites/default/files/ehealth/docs/eu-dcc_validation-rules_en.pdf#%5B%7B%22num%22%3A36%2C%22gen%22%3A0%7D%2C%7B%22name%22%3A%22XYZ%22%7D%2C69%2C771%2C0%5D
and how I also found it somewhere in a verifier demo...
Thanks for all your nice works!

PR #3688, opened 2021-07-12
Link to FAQ to find test centers (EXPOSUREAPP-8086)

harambasicluka:
Implementation

image

image

_Note:_
The image on the side is not perfectly aligned as in the mockups but this would require a refactoring of this cards, this also affects the already implemented cards. To bring it a little bit closer to the mockups I changed the navigation colors.


vaubaehn:
You prefer opening FAQ over directly jumping to the map? As a fallback reason in case map service is interrupted/terminated? But TSI could provide a message on their site then. Like this it takes one click more effort to get to the map.

But nice to implement it in general 👍

mtwalli (dismissed):
Lgtm, tested on Android 6 emulator 😃

harambasicluka:
> You prefer opening FAQ over directly jumping to the map? As a fallback reason in case map service is interrupted/terminated? But TSI could provide a message on their site then. Like this it takes one click more effort to get to the map.
> 
> But nice to implement it in general 👍

@vaubaehn the FAQ entry contains multiple links and we want to be able to adjust the links without releasing a new app version :)

vaubaehn:
@harambasicluka 
Thanks, I simply forgot about the PCR tests, which are not necessarily included in the map. Also, the link to @Ein-Tim 's list is also worth to jump to FAQ :)

harambasicluka:
> @harambasicluka
> Thanks, I simply forgot about the PCR tests, which are not necessarily included in the map. Also, the link to @Ein-Tim 's list is also worth to jump to FAQ :)
> 
> Edit: why are you laughing about Android 6 emulator? It's important!! 😉

It was more like an internal joke, because it's my first implementation in a while :)

Corona-Warn-App/src/main/res/values-de/antigen_strings.xml
```diff
@@ -5,9 +5,9 @@
         Homescreen cards - register test card
     ###################################### -->
     <!-- XHED: Register test homescreen card: title -->
-    <string name="ag_homescreen_card_test_register_title">"Nächste Schritte"</string>
+    <string name="ag_homescreen_card_test_register_title">"Weiter"</string>
     <!-- XTXT: Register test homescreen card: body -->
-    <string name="ag_homescreen_card_test_register_body">"Nutzen Sie die App, um Ihr Testergebnis abzurufen und andere schneller warnen zu können, oder um Ihr COVID-Testzertifikat anzufordern."</string>
+    <string name="ag_homescreen_card_test_register_body">"Finden sie Teststellen in ihrer Umgebung und registrieren sie einen Test, um andere schneller warnen zu können, oder ihr digitales COVID-Testzertifikat anzufordern."</string>
```
DerVogel2020, line 10: Sie, Ihrer und Ihr müssten groß geschrieben werden.

DerVogel2020 (commented):
This will fix https://github.com/corona-warn-app/cwa-wishlist/issues/498

DerVogel2020:
Will this fix Internal Tracking ID: EXPOSUREAPP-7204?

Corona-Warn-App/src/main/res/values-de/antigen_strings.xml
NataliaLemmerth, line 10:
```suggestion
    <string name="ag_homescreen_card_test_register_body">"Finden Sie Teststellen in Ihrer Umgebung und registrieren Sie einen Test, um andere schneller warnen zu können, oder Ihr digitales COVID-Testzertifikat anzufordern."</string>
```

NataliaLemmerth (changes requested):
Just these small text changes. Everything else is fine.

BMItr (dismissed):
nothing to complain 👌

NataliaLemmerth (approved):
UA approved

harambasicluka:
> Will this fix Internal Tracking ID: EXPOSUREAPP-7204?

@DerVogel2020 Yes! You can already check the URL: https://www.coronawarn.app/de/faq/#where_can_i_get_tested

PR #3691, opened 2021-07-12
Fix link in EN (EXPOSUREAPP-8269)

mtwalli:
- 🔧  🔗 🇺🇸


harambasicluka:
This needs to be communicated to UA and the translators, otherwise it will be changed with the next translation delivery. But as this string is fully linked I think it could be split after `:` so that the whole string which currently is in `statistics_explanation_seven_day_r_link_label` can be linked.´, so that it doesn't has to match a label.

mtwalli:
> This needs to be communicated to UA and the translators, otherwise it will be changed with the next translation delivery. But as this string is fully linked I think it could be split after `:` so that the whole string which currently is in `statistics_explanation_seven_day_r_link_label` can be linked.´, so that it doesn't has to match a label.

Done

harambasicluka (approved):

NataliaLemmerth (approved):
UA approved

PR #3700, opened 2021-07-13
Rat redeemed state handling (EXPOSUREAPP-7517)

SamuraiKek:
Updated handling of redeemed RAT QR codes. Now also works with PCR_REDEEMED states in cases where the Verification Server sends PCR_REDEEMED instead of RAT_REDEEMED. 

Use the QR's provided in the Jira ticket to test.


d4rken (changes requested):
Looks good.

Do we need to add `PCR_OR_RAT_REDEEMED` in `RATestProcessor` to these too?

Also a final state?
https://github.com/corona-warn-app/cwa-app-android/blob/ecd9613d3c36f3cd6d197bb01d15279c08b326db/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RATestProcessor.kt#L227-L230

Polling abort?
https://github.com/corona-warn-app/cwa-app-android/blob/ecd9613d3c36f3cd6d197bb01d15279c08b326db/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RATestProcessor.kt#L121-L124

SamuraiKek:
> Looks good.
> 
> Do we need to add `PCR_OR_RAT_REDEEMED` in `RATestProcessor` to these too?
> 
> Also a final state?
> https://github.com/corona-warn-app/cwa-app-android/blob/ecd9613d3c36f3cd6d197bb01d15279c08b326db/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RATestProcessor.kt#L227-L230
> 
> Polling abort?
> https://github.com/corona-warn-app/cwa-app-android/blob/ecd9613d3c36f3cd6d197bb01d15279c08b326db/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RATestProcessor.kt#L121-L124

Yes, good catch. I'll add it here as well.

d4rken (approved):

chiljamgossow (approved):

PR #3705, opened 2021-07-13
Update build.gradle quickBuild task (DEV)

SamuraiKek:
Changed ktlintDeviceReleaseCheck to ktlintDeviceReleaseSourceSetCheck to prevent quickBuild from failing locally.

Test by running ./gradlew quickBuild


build.gradle
```diff
@@ -73,7 +73,7 @@ task quickBuild {
     dependsOn ':Corona-Warn-App:assembleDeviceRelease'
     dependsOn ':Corona-Warn-App:testDeviceReleaseUnitTest'
     dependsOn ':Corona-Warn-App:lintDeviceRelease'
-    dependsOn ':Corona-Warn-App:ktlintDeviceReleaseCheck'
+    dependsOn ':Corona-Warn-App:ktlintDeviceReleaseSourceSetCheck'
```
mtwalli, line 76: can you please change also in `quickBuildWithFixes` task ?

harambasicluka: we could even think about removing the quick from the name :D
image

harambasicluka (dismissed):

build.gradle
SamuraiKek: > we could even think about removing the quick from the name :D
You're lucky, it takes me 28m on average...

SamuraiKek: > can you please change also in `quickBuildWithFixes` task ?

Done

d4rken:
> Changed ktlintDeviceReleaseCheck to ktlintDeviceReleaseSourceSetCheck to prevent quickBuild from failing locally.

Why does one fail and the other one doesn't and why does this happen now and not before? :thinking:

SamuraiKek:
> > Changed ktlintDeviceReleaseCheck to ktlintDeviceReleaseSourceSetCheck to prevent quickBuild from failing locally.
> 
> Why does one fail and the other one doesn't and why does this happen now and not before? 🤔

From my discussion with @mtwalli yesterday, I think the name changed.

mtwalli (approved):
Lgtm

mtwalli:
> > > Changed ktlintDeviceReleaseCheck to ktlintDeviceReleaseSourceSetCheck to prevent quickBuild from failing locally.
> > 
> > 
> > Why does one fail and the other one doesn't and why does this happen now and not before? 🤔
> 
> From my discussion with @mtwalli yesterday, I think the name changed.

We upgraded ktlint two times, apparently the task name changed in the meantime.

> why does this happen now and not before?
I don't think that anyone was using these tasks recently, I never used it. As Luka already mentioned it is not really quick and we should remove it

harambasicluka (approved):

PR #3706, opened 2021-07-13
Improve statistics card redraw behavior (DEV)

d4rken:
Use PayloadDiffer to prevent redraw of statistic cards.
Removes card flashing when adding/removing local incidences.

Testing:
Compare with and without this PR.


kolyaopahle (approved):
LGTM 👍  seems to be behaving better on my s9

jurajkusnier (approved):
LGTM

PR #3715, opened 2021-07-14
Validation date time refactoring (DEV)

chiljamgossow:
Forward raw user input to certlogic engine to prevent time zone issues
unit tests for time zone


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidation.kt
```diff
@@ -55,5 +57,6 @@ data class DccValidation(
 @Parcelize
 data class ValidationUserInput(
     val arrivalCountry: DccCountry,
-    val arrivalAt: Instant,
+    val arrivalDate: LocalDate,
+    val arrivalTime: LocalTime,
```
d4rken, line 61: Why individual instances vs e.g. `LocalDateTime`?

chiljamgossow: I thought individual instance makes it less likely that people think it can just be converted into millis.

d4rken: I would prefer to do this with a comment and unit tests (which you already have), the additional code necessary for conversion doesn't seem worth it.

e.g. these could be simplified then:

* https://github.com/corona-warn-app/cwa-app-android/pull/3715/files#diff-88d64b26682ce8d94d41decfd41ec8c673c4a9aba93fb223ea6ee790397f41aeR57-R59
* https://github.com/corona-warn-app/cwa-app-android/pull/3715/files#diff-ec441c15dc78671703ae7a9083735ab9348872e7f3a36817cb27dc42c36dab2dR161-R168

d4rken (commented):
Thank you! Looks good!

One small comment about a potential typo?

d4rken (dismissed):
lgtm

Would love a small test that would have failed for:
https://github.com/corona-warn-app/cwa-app-android/pull/3715#discussion_r672385485

Not a blocker, up to you.

jurajkusnier (dismissed):
LGTM

jurajkusnier (approved):

d4rken (approved):

PR #3723, opened 2021-07-15
Add more logging to dcc parsing error handling (EXPOSUREAPP-8402)

kolyaopahle:
### Description
EXPOSUREAPP-8402 describes an issue during parsing of a dcc, with the current logging in place we are not able to figure out the root case of the error by looking at log files as the exception is only shown to the user but never logged.
This PR adds logging so that we may be able to figure out the underlying issue.

### This is not a fix


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccQrCodeExtractor.kt
```diff
@@ -189,6 +189,7 @@ class DccQrCodeExtractor @Inject constructor(
             Timber.v("Parsed covid certificate for %s", it.certificate.nameData.familyNameStandardized)
         }
     } catch (e: InvalidHealthCertificateException) {
+        Timber.e(e)
```
mtwalli, line 192: Maybe some `message`  ?  ¯\_(ツ)_/¯

NataliaLemmerth (approved):
UA approved - no UI texts

chiljamgossow (approved):
Logging makes sense. You should be able to see the cause in the debugger.

mtwalli (approved):

PR #3725, opened 2021-07-15
Low exposure risk card displays low encounter text when there are none (EXPOSUREAPP-8310)

SamuraiKek:
Added an extra condition that would ensure the right text is being displayed if the app ever reahces the state described in the ticket.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
```diff
@@ -108,7 +108,7 @@ data class LowRisk(
         context.getString(R.string.risk_card_body_not_yet_fetched)
     }
 
-    fun getRiskContactBody(context: Context): String = if (daysWithEncounters == 0) {
+    fun getRiskContactBody(context: Context): String = if (daysWithEncounters == 0 || lastEncounterAt == null) {
```
d4rken, line 111: Not a fan of this solution. Fixes the symptom, not the cause.
We should look deeper at the data sources.
Looking at the log provided in the internal ticket, the last returned result is:

```kotlin
LastCombinedRiskResults(
	lastCalculated=CombinedEwPtRiskLevelResult(
		ptRiskLevelResult=PtRiskLevelResult(
			calculatedAt=2021-07-08T07:25:07.341Z,
			riskState=LOW_RISK, 
			presenceTracingDayRisk=[], 
			checkInWarningOverlaps=[]
		), 
		ewRiskLevelResult=RiskLevelTaskResult(
			calculatedAt=2021-07-08T07:25:08.026Z, 
			failureReason=null, 
			ewAggregatedRiskResult=EwAggregatedRiskResult(
				totalRiskLevel=LOW, 
				totalMinimumDistinctEncountersWithLowRisk=0, 
				totalMinimumDistinctEncountersWithHighRisk=0, 
				mostRecentDateWithLowRisk=null, 
				mostRecentDateWithHighRisk=null, 
				numberOfDaysWithLowRisk=0, 
				numberOfDaysWithHighRisk=0, 
				exposureWindowDayRisks=null
			), 
			exposureWindows.size=null
		), 
		exposureWindowDayRisks=[
			ExposureWindowDayRisk(dateMillisSinceEpoch=1625616000000, riskLevel=LOW, minimumDistinctEncountersWithLowRisk=2, minimumDistinctEncountersWithHighRisk=0), 
			ExposureWindowDayRisk(dateMillisSinceEpoch=1625529600000, riskLevel=HIGH, minimumDistinctEncountersWithLowRisk=2, minimumDistinctEncountersWithHighRisk=0)
		]
	), 
	lastSuccessfullyCalculated=CombinedEwPtRiskLevelResult(
		ptRiskLevelResult=PtRiskLevelResult(
			calculatedAt=2021-07-08T07:25:07.341Z, 
			riskState=LOW_RISK, 
			presenceTracingDayRisk=[], 
			checkInWarningOverlaps=[]), 
			ewRiskLevelResult=RiskLevelTaskResult(
			calculatedAt=2021-07-08T07:25:08.026Z, 
			failureReason=null, 
		ewAggregatedRiskResult=EwAggregatedRiskResult(
				totalRiskLevel=LOW, 
				totalMinimumDistinctEncountersWithLowRisk=0, 
				totalMinimumDistinctEncountersWithHighRisk=0, 
				mostRecentDateWithLowRisk=null, 
				mostRecentDateWithHighRisk=null, 
				numberOfDaysWithLowRisk=0, 
				numberOfDaysWithHighRisk=0, 
				exposureWindowDayRisks=null
			),
			exposureWindows.size=null
		),
		exposureWindowDayRisks=null
	)
)
```

`lastCalculated` having 
```kotlin
exposureWindowDayRisks=[
			ExposureWindowDayRisk(dateMillisSinceEpoch=1625616000000, riskLevel=LOW, minimumDistinctEncountersWithLowRisk=2, minimumDistinctEncountersWithHighRisk=0), 
			ExposureWindowDayRisk(dateMillisSinceEpoch=1625529600000, riskLevel=HIGH, minimumDistinctEncountersWithLowRisk=2, minimumDistinctEncountersWithHighRisk=0)
]
```

is likely the cause for the displayed encounters.
It's curious that `lastCalculated` contains these EWs while `lastSuccessfullyCalculated` doesn't. :thinking:

SamuraiKek: I totally agree, it fixes the symptom, not the cause but I thought it was worth it, being such a small change in the code, at least for now. Reproducing the issue is quite tricky and it's the first time I've seen this reported. Maybe we could dedicate more time reproducing and investigating this bug at a later time. We could merge it for the sake of improving the user experience with the promise to have another look at a later point or we could just close it. It's up to you.

d4rken:
The day display comes from:

https://github.com/corona-warn-app/cwa-app-android/blob/1f3f403d8b435258268a536fffe4ad82c21abdac/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/CombinedEwPtRisk.kt#L41-L45

which is populated by:
https://github.com/corona-warn-app/cwa-app-android/blob/1f3f403d8b435258268a536fffe4ad82c21abdac/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/BaseRiskLevelStorage.kt#L206-L216

which itself (i.e the `exposureWindowDayRisks`) is sourced from a database table:
https://github.com/corona-warn-app/cwa-app-android/blob/1f3f403d8b435258268a536fffe4ad82c21abdac/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/BaseRiskLevelStorage.kt#L138-L146

So in the test case, the EW data was deleted via ENS, but the CWA doesn't know about that, and it doesn't wipe our database table, so we have the data for the days. Other calculations use data more directly from the last retrieved ENS data though:

https://github.com/corona-warn-app/cwa-app-android/blob/1f3f403d8b435258268a536fffe4ad82c21abdac/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/BaseRiskLevelStorage.kt#L88-L93

So we end up with a combination of new data from the last result, and the old data from our database.

To fix the underlying cause we'd need to either:

* source more data from our storage, and less from the ENS directly, so we have the matching rest of the risk data for the "day encounter" data we still have and show. i.e. "be less affected by the deletion via ENS" (seems a bit complicated...)
* or wipe the stored risk data when it diverges from what we get from the ENS. i.e. "be more affected by the deletion via ENS" (not sure if we can detect this circumstance...)

Not sure about a solution, but so far for research...
@SamuraiKek do you think the above is plausible as cause?

SamuraiKek:
> Not sure about a solution, but so far for research...
> @SamuraiKek do you think the above is plausible as cause?

This sounds very plausible. 

> So in the test case, the EW data was deleted via ENS, but the CWA doesn't know about that, and it doesn't wipe our database table, so we have the data for the days. Other calculations use data more directly from the last retrieved ENS data though:
> https://github.com/corona-warn-app/cwa-app-android/blob/1f3f403d8b435258268a536fffe4ad82c21abdac/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/storage/BaseRiskLevelStorage.kt#L88-L93

If I'm looking at it right, I think it also explains why in my proposed fix, the `lastEncounterAt` check "fixes" the issue. Because it comes from the last retrieved ENS calc whyle the `daysWithEncounters` relies on the data we have stored in the db.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
chiljamgossow: lastSuccessfullyCalculated is only there for the risk state (in case the lastCalculated failed and we want to show the risk state before), not for getting any details. Details are only available for the lastCalculated

chiljamgossow:
For me, the test setup is invalid. If someone starts deleting data outside of the app, then one should expect inconsistencies.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/tracing/states/TracingState.kt
chiljamgossow: A change I thought about but never implemented is to just expose lastSuccessfullyCalculatedState, to prevent misinterpretation

d4rken:
Putting this into draft again after discussion with @SamuraiKek .

@chiljamgossow  is right that this is weird use-case test it's not urgent enough to warrant a UI workaround without fixing the underlying issue.

harambasicluka:
I'll close this ticket, document it in the ticket and then we'll look into it when we have the time.

PR #3729, opened 2021-07-15
Fix padding on person details fragment (COMMUNITY)

fynngodau:
There is currently an incorrect bottom padding for the whole person details fragment, leading to a gray space at the bottom of the layout. It blends in with the navbar on some devices, but on devices where Gesture navigation is enabled or which have hardware keys, it becomes noticable:

Incorrect | Correct
---|---
Screenshot_1626357616 | Screenshot_1626357687

I also found that the `12dp` additional space at the bottom was too much. I reduced it to `8dp` to align with the vaccination, recovery and test views.

The `RecyclerView` already had `clipToPadding` set to `false`, hence all that is needed to fix this issue is to move the padding there.


harambasicluka:
Thanks again for your contribution and 🦅  👁️ ! :)

harambasicluka (approved):

mtwalli (approved):

PR #3734, opened 2021-07-16
Fix: 'Reset App' does not reset custom incidence locations (EXPOSUREAPP-8439)

jurajkusnier:
### Testing
- add one or more local statics cards
- perform app reset
- check if cards disappear

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
@@ -110,7 +110,7 @@ class LocalStatisticsProvider @Inject constructor(
     fun clear() {
         Timber.d("clear()")
 
-        localStatisticsConfigStorage.activeDistricts.update { emptySet() }
+        localStatisticsConfigStorage.clear()
         server.clear()
         localStatisticsCache.clearAll()
     }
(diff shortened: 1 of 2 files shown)


harambasicluka (dismissed):
💯

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
```diff
@@ -110,6 +110,7 @@ class LocalStatisticsProvider @Inject constructor(
     fun clear() {
         Timber.d("clear()")
 
+        localStatisticsConfigStorage.activeDistricts.update { emptySet() }
```
kolyaopahle, line 113: Id rather follow the pattern of having a clear method on `localStatisticsConfigStorage` that deletes the whole shared prefs like in `CWASettings` and our other storage classes, the provider does not need to know how to specifically reset the settings

kolyaopahle (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/local/source/LocalStatisticsProvider.kt
jurajkusnier: updated

mtwalli (approved):

harambasicluka (approved):

PR #3736, opened 2021-07-16
UA Bugfix(EXPOSUREAPP-8453)

NataliaLemmerth:
This fixes EXPOSUREAPP-8453


BMItr (approved):

harambasicluka (changes requested):
I think the previous text is right.

harambasicluka (approved):

PR #3742, opened 2021-07-16
Increase certificate storage resilience (EXPOSUREAPP-8113, EXPOSUREAPP-8117)

d4rken:
There are a handful of user reports that stored vaccination certificates are lost after updating the app.

Initial suspicions that I ruled out:
* Racecondition between loading and storing the data. We store every new data set including the initially loaded one. It's not possible to store new data before the initial emission (see `HotDataFlow` init). So once data was stored, it is returned on re-init, and thus also saved again. If it fails to load the initial data for some reason the app crashes.
* Racecondition between multiple data changes and saving them, causing overlapping access to the shared preferences used for storage. The debugger shows that the `onEach` used to save the data is only entered once. There is no overlap.

My working theory:
Unknown filesystem issues (permissions, low storage, corrupt file system) prevent storage from being written. SharedPreferences does not really throw exceptions. User sees the certificates while the app session is running as the data is in memory. It could be more likely if features like "adopt sdcard as internal storage" are used. User reports were not clear on whether such setups are in use. I don't see how the CWA could reasonably handle that. :thinking: 

This PR is more of a shot in a dark that increases the overall resilience with little downside:
* Assure no overlapping load/saves by using a `Mutex` to guard access.
* Use `edit().commit` instead of apply, to force a write to storage. We are on a different thread and get no penalty for the blocking call.


mtwalli:
Could it be that we do multiple `reading` and `writing` from the certificates storages?  and that increases the probability to lose them at  any time when  saving them again. it is very unlikely that the Certificates once scanned in the App that its data will change and therefore there is no need to have such mechanism for reading and writing. if we save them once and just read multiple times that would reduce the potential to have such issues, what is your thoughts about that?

d4rken:
> Could it be that we do multiple reading and writing from the certificates storages?

We only read once, when HotData initializes.
But we write multiple times, every time hot data changes.

> and that increases the probability to lose them at any time when saving them again. it is very unlikely that the Certificates once scanned in the App that its data will change and therefore there is no need to have such mechanism for reading and writing. if we save them once and just read multiple times that would reduce the potential to have such issues, 

I have thought about reducing the amount of writes, by checking if it is already written. We can't trust SharedPreferences a 100% there though as it also has an in memory cache, so even if SharedPreferences returns the same value, it's not guaranteed that it's written to disk already. It should only be a problem if we have multiple reads+writes., if we only have multiple writes, I don't see how we could up with `""` (empty data), as old data could only get lost by overwriting it.

> what is your thoughts about that?

We could add some filtering to the storage writer (flow collection in the `init` blocks), a `drop(1)` for the initial re-emission, and a `distinctUntilChanged` to not rewrite data if we already wrote it. :thinking: I don't see any downsides.

Overall I'm still fairly confident in the original code. I don't see a bug there.
The user reports are a bit inconsistent.
Some reports that say that rescanning yields "Error: Already registered", and some report that certificates are visible again after a reboot.
That would point to storage working fine, and it just being a UI issue. 
I have not found a cause there yet, but I'm still looking. :dizzy_face:

d4rken:
@mtwalli I've added the `.drop(1)` and a dose of unit tests.

I have not found anything in the UI, but any changes will be part of a different PR.
I think this PR contains all we can do on the "storage" side of things to improve the situation.

vaubaehn:
I hope it is ok if I throw in my 2 cents:
I was reading reports in Play Store in the past days and trying to find some patterns.
> Some reports that say that rescanning yields "Error: Already registered", and some report that certificates are visible again after a reboot.

That were most of all reports.
Some people stated, that only after waiting 10 to 20 minutes the DCCs suddenly appeared.
Another person tried to start an automatic error analysis, and started the error report feature. Disappointed, that the mobile didn't spit out any solution by itself, he stopped the error report - and suddenly the DCC appeared.

> That would point to storage working fine, and it just being a UI issue.

From what I read, now everything points into that direction. (I'm a friend of `edit().commit` though, maybe Matthis remembers from discussions in the last year 😉, so doesn't hurt at all to have it implemented imho)

What I don't understand are the people where a reboot/waiting/rescanning does not change anything.

> We only read once, when HotData initializes.

Is there any way that `HotData` initialization fails (loud or silently)? What would UI display in these cases?

d4rken:
> I hope it is ok if I throw in my 2 cents:
> I was reading reports in Play Store in the past days and trying to find some patterns.
> 
> > Some reports that say that rescanning yields "Error: Already registered", and some report that certificates are visible again after a reboot.
> 
> That were most of all reports.
> Some people stated, that only after waiting 10 to 20 minutes the DCCs suddenly appeared.
> Another person tried to start an automatic error analysis, and started the error report feature. Disappointed, that the mobile didn't spit out any solution by itself, he stopped the error report - and suddenly the DCC appeared.

My gut feeling says it's a hanging flow somewhere, that needs to emit a value, for the operations to continue, but doesn't do so.
So a downstream flow, or a `combine` of multiple that includes this hanging flow is stuck.
If i manually block one of the Test/Recovery/Vaccination-Repository flows, then I get a behavior that looks similar to what the users describe.
I just can't figure out what operation could be hanging.

It's likely something upstream from here, if this blocks, the certificate view is empty, it's emission is the "person card" itself:
https://github.com/corona-warn-app/cwa-app-android/blob/2943eff83e919812f9dffd0df2d6644898fa59e1/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt#L55-L65


----


> What I don't understand are the people where a reboot/waiting/rescanning does not change anything.
> 
> > We only read once, when HotData initializes.
> 
> Is there any way that `HotData` initialization fails (loud or silently)? What would UI display in these cases?
> 
> By the way: CovPass seems to have sometimes similar issues, more rare though.

If the `initialValueProvider` throws an error, the app crashes. If it hangs, then no value will ever be emitted, this means no initial value, and also no updates (as any other operation would be stuck behind the initialization).

It just reads from storage though, and I don't see how it could hang for any amount of time doing that:
https://github.com/corona-warn-app/cwa-app-android/blob/2943eff83e919812f9dffd0df2d6644898fa59e1/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/VaccinationRepository.kt#L44-L60

----

> And another idea: what would happen if a dependency doesn't initialize correctly? E.g., if module for QR code generation hangs, what would happen?

If it fails to initialize due to an exception, the app crashes. If it hangs, it would cause everything that requires it to hang too. The only thing big dependency being initialized in that area is the data extractor for the qrcode string, it's all static code dependencies, nothing that should be able to hang.

My favorite scape goat would be some kind of network access, but I can't make it work. The valueset repository (translations for certificates)  initiliazes only from storage, refresh via server happens async.

----

> Last but not least question: Do the reports point to certain phones or OS? My old Android 6 phone often has timeout problems and slow storage, and I would expect such phones to be more vulnerable. But it feels that mostly newer phones are affected?

Based just on what feedback the community team gathered, no a mix of popular devices androi Android 8 to 11.
As the issue doesn't result in a crash, I can't gather any broader statistics, e.g. via Google Play Console.

vaubaehn:
> If i manually block one of the Test/Recovery/Vaccination-Repository flows, then I get a behavior that looks similar to what the users describe.

Needle in haystack, but this sounds like a good starting ground.
Maybe https://github.com/corona-warn-app/cwa-app-android/pull/3751 is very promising in this context.



> It's likely something upstream from here, if this blocks, the certificate view is empty, it's emission is the "person card" itself:
> 
> https://github.com/corona-warn-app/cwa-app-android/blob/2943eff83e919812f9dffd0df2d6644898fa59e1/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt#L55-L65

Is the `cameraPermissionProvider.deniedPermanently` a static stored value, or is it requested again in the moment this is called?
If this value is always requested dynamically, this could be a good candidate for a hang.
What is the `certificatesProvider.qrCodesFlow` for?

> The only thing big dependency being initialized in that area is the data extractor for the qrcode string, it's all static code dependencies, nothing that should be able to hang.

I was especially pointing to QR code generation as I was able to observe that it sometimes took insanely much time to end generation on my device. Last week I added another 2 DCCs for 2 persons, and generation was taking more then 10s (progress indicator was running, no bar code appearing). I then jumped through menus forth and back, and the back made it appear.
Wondering whether that could interfere with the flow, resulting in symptoms described in all the reports.

> My favorite scape goat would be some kind of network access, but I can't make it work.

Is the zxing module at some point involved in the flow? There were reports that it relies on Google Play Services (afair somewhere reported in the Android Lollipop issue). In that case I wouldn't be surprised if initialization causes a network request via GPS/GMS. But I think it's not a likely cause of this issue here...

> no a mix of popular devices androi Android 8 to 11

If Android 6 & 7 could be clearly ruled out, then it could also be an implementation issue starting with Android 8...

d4rken:
> > If i manually block one of the Test/Recovery/Vaccination-Repository flows, then I get a behavior that looks similar to what the users describe.
> 
> Needle in haystack, but this sounds like a good starting ground.
> Maybe #3751 is very promising in this context.

Doubt it, it would require `Int.MAX_VALUE` update calls on a data to take effect, somekind of endless loop like that would likely crash the app in some way or another as all update actions would have to stay in memory too.

> > It's likely something upstream from here, if this blocks, the certificate view is empty, it's emission is the "person card" itself:
> > https://github.com/corona-warn-app/cwa-app-android/blob/2943eff83e919812f9dffd0df2d6644898fa59e1/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt#L55-L65
> 
> Is the `cameraPermissionProvider.deniedPermanently` a static stored value, or is it requested again in the moment this is called?

Checked again, but this is not known to hang, afaik.

> If this value is always requested dynamically, this could be a good candidate for a hang.
> What is the `certificatesProvider.qrCodesFlow` for?

Generates the QR Codes, but should not be the  cause because it starts out with an empty data. The UI will show the loading indicator, once the qrcodes are generated, a new emission is combined again. `combine` can only block if one of the flows has no emission at all.

https://github.com/corona-warn-app/cwa-app-android/blob/2943eff83e919812f9dffd0df2d6644898fa59e1/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt#L135-L144

The first `emit(...)` prevents this from causing hanging.

> > The only thing big dependency being initialized in that area is the data extractor for the qrcode string, it's all static code dependencies, nothing that should be able to hang.
> 
> I was especially pointing to QR code generation as I was able to observe that it sometimes took insanely much time to end generation on my device. Last week I added another 2 DCCs for 2 persons, and generation was taking more then 10s (progress indicator was running, no bar code appearing). I then jumped through menus forth and back, and the back made it appear.
> Wondering whether that could interfere with the flow, resulting in symptoms described in all the reports.

There are some not so nice things, #3752 :wink:

> > My favorite scape goat would be some kind of network access, but I can't make it work.
> 
> Is the zxing module at some point involved in the flow? There were reports that it relies on Google Play Services (afair somewhere reported in the Android Lollipop issue). In that case I wouldn't be surprised if initialization causes a network request via GPS/GMS. But I think it's not a likely cause of this issue here...

No it's not involved. At least not where it can block, would have to check whether we use it for image generation, but as said above, the qrcode image data starts out with an empty emission to not cause any blocking.

> > no a mix of popular devices androi Android 8 to 11
> 
> If Android 6 & 7 could be clearly ruled out, then it could also be an implementation issue starting with Android 8...

So far I wouldn't know what... but it seems just rare enough so that none of the more rare < Android 8 devices did not experience this yet.

vaubaehn:
Everything before

> The first `emit(...)` prevents this from causing hanging.

=> hm. 🤔 that doesn't make it look easier. 😞 

> There are some not so nice things, #3752 😉

That's great! Thanks for fixing 👍 

> but as said above, the qrcode image data starts out with an empty emission to not cause any blocking.

=> Another: "hm." 😞 Now I ran out of ideas.

> but it seems just rare enough so that none of the more rare < Android 8 devices did not experience this yet.

 Trying to understand the double negation in a single sentence makes my brain hurt at this late time 😉 

Have a nice evening and a good rest! 👋

mtwalli (approved):
Thanks for investigating

chiljamgossow (approved):

chiljamgossow:
LGTM.  I would expect shared prefs to be thread safe 🤔

PR #3747, opened 2021-07-19
Pass rawCOSEObject (EXPOSUREAPP-8474)

mtwalli:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccData.kt
```diff
@@ -1,8 +1,35 @@
...
+import de.rki.coronawarnapp.covidcertificate.common.decoder.RawCOSEObject
+
 data class DccData<CertT : DccV1.MetaData>(
     val header: DccHeader,
     val certificate: CertT,
     val certificateJson: String,
-    val kid: String
-)
+    val kid: String,
+    val rawCOSEObject: RawCOSEObject
```
chiljamgossow, line 10: Close, but not quite. It is extracted from the COSE object and a byte array. Please call it signature

chiljamgossow: so either ByteArray or ByteString

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccData.kt
```diff
@@ -1,8 +1,35 @@
...
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
```
chiljamgossow, line 13: How come we need a custom implementation?

mtwalli: because we pass ByteArray now ,any suggestions ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccData.kt
mtwalli: just to get that right. `RawCOSEObject` is the right value to pass ? or what you called `signature` should be extracted from it ?

mtwalli: `RawCOSEObject` is a `ByteArray` as I can see

chiljamgossow: the signature is a byte array (but not a COSE object)

-> val signature: ByteArray

mtwalli: my question here : what is the value that should be passed in ?

```kotlin
DccDate(
signature = rawCodeObject
)

// or 

DccDate(
signature = rawCodeObject.slice(0,n) // for example 
)

// or some different source ?
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccQrCodeExtractor.kt
```diff
@@ -183,7 +186,8 @@ class DccQrCodeExtractor @Inject constructor(
             header = header,
             certificate = body.parsed.asCertificate,
             certificateJson = body.raw,
-            kid = message.kid
+            kid = message.kid,
+            signature = this
```
mtwalli, line 190: @chiljamgossow  my question is basically, is that right here ` signature = this` ? or it should be different value than `this` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccData.kt
chiljamgossow: It need to be extracted in class DccCoseDecoder as described here https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/signature-verification-for-dcc/docs/spec/dgc-overview-client.md#verifying-the-signature-of-a-dgc-against-a-list-of-dscs

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccData.kt
chiljamgossow: ok, haven't tested comparisons of ByteArray. You may be right

d4rken: Alternatively a `ByteString` could be used (if it makes sense), not sure atm what the other classes do with the signature.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccQrCodeExtractor.kt
chiljamgossow: You need to add this to DccCoseDecoder: 
`private fun CBORObject.extractSignature(): ByteArray = try {
        this[3].GetByteString()
    } catch (e: Exception) {
        throw InvalidHealthCertificateException(HC_COSE_NO_SIGN1)
    }`

chiljamgossow: And extend the Message data class with the signature

chiljamgossow: And add the error code to the exception

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccData.kt
mtwalli: I think we should pass the whole COSE object not only the signature, the validation procedure requires byte representation of `header`,  `signature ` `payload` and `alg` is extracted from the` header `

mtwalli:
Closing this PR, @chiljamgossow  and I discussed this PR  ,Validation would rely on certificate QrCode string directly

Issue #3549, opened 2021-06-24
Vaccination Certificate disappears from UI


vaubaehn:
## Avoid duplicates
* [x] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [x] Bug is not already reported in another issue

## Technical details

- Device name: unknown, probably several
- Android version: unknown
- App version: 2.3.4

## Describe the bug
Since vaccination DCCs can be imported into CWA, I found several reviews in Google Play Store, that the certificate "vanishes" by itself. According to the following report, it's still persisted, but not displayed:
> Ich muss mein impfzertifikat täglich neu hinzufügen, da es einfach verschwindet... Nehm ich den digitalen impfausweiss halt lieber doch analog mit... Update: ich bin kein Idiot. Das Zertifikat fehlt. Aber Plot Twist: wenn ich es nochmal hinzufügen will kommt das es schon da ist. Ist also ein reiner GUI Glitch...

## Steps to reproduce the issue

1. Import/scan vac DCC
2. Wait some time (till next day), find it vanished
3. Try to re-import DCC
4. See error: Already scanned.

## Expected behaviour

The DCC should be displayed always after it was persisted.

## Additional context

I hope for your understanding, that I create an issue for a bug that I don't experience myself.
But as I already found quite a few reports of this kind, I thought could be good to track it here. Please have a look to Play Store reviews yourself. Maybe that issue is already known to the dev team, but then we have a place to track progress from here.

Edit: More reports
![Screenshot_2021-06-25-10-07-32~2](https://user-images.githubusercontent.com/68376758/123392659-9d441a80-d59d-11eb-9d55-96c86d2c8acc.png)
![Screenshot_2021-06-25-10-14-19~2](https://user-images.githubusercontent.com/68376758/123393304-3ecb6c00-d59e-11eb-8fd2-3418f8b0de95.png)
![Screenshot_2021-06-25-10-19-33~2](https://user-images.githubusercontent.com/68376758/123394034-fb253200-d59e-11eb-8ff9-763032ae9caa.png)
![Screenshot_2021-06-25-11-11-44~2](https://user-images.githubusercontent.com/68376758/123401108-c0bf9300-d5a6-11eb-8ca3-58f6cf5ed1b1.png)

---
[...]

PR #3750, opened 2021-07-19
Increase ValueSet storage resilience (DEV,EXPOSUREAPP-8113)

d4rken:
Same treatment as #3742
* Mutex locking on storage
* Don't retwrite the initial dataset
* If we really fail to store the ValueSets (which uses the same storage mechanism as the certificates), then be loud and rethrow the exception.


d4rken:
Part of the hunt for #3549

jurajkusnier (approved):

BMItr (approved):
mtgl

jurajkusnier (changes requested):
DELETED: wrong PR :)

PR #3751, opened 2021-07-19
Better `HotData` queue handling on fast operations (DEV)

d4rken:
Not an issue we had, afaict, but the code is treacherous and should be adjusted (noticed while hunting #3549).

Prevent silent NOOPs, by throwing the action away without notice.
We are not checking return values in the app anywhere, and the updateAction queue size is `Int.MAX_VALUE`.
If we really run into `emit(...)` blocking as the queue is full, then just block.


jurajkusnier (approved):
LGTM

mtwalli (approved):

PR #3755, opened 2021-07-20
 Convert validation clock to utc (EXPOSUREAPP-8470)

chiljamgossow:
CertLogicEngine seems to only accept time zone utc, rule validation results in OPEN otherwise


mtwalli (approved):
Lgtm

d4rken (approved):
Looks good :+1:  Timezones :dizzy_face:

PR #3756, opened 2021-07-20
Wrong result on rule validation screen for expired certs (EXPOSUREAPP-8468)

BMItr:
This PR addresses Exposureapp-8468.

Hot to test:
-Scan the QR code (can be found inside the ticket)
-Check if the result after validation is the same as the expected one (screenshot)

Screenshot:
image


harambasicluka:
How does this fixes the wrong text? So your screenshot looks like the expected result, but I don't get it.

BMItr:
> How does this fixes the wrong text? So your screenshot looks like the expected result, but I don't get it.

the failed result view has basically two elements (wrong format & cert not valid anymore).
The logic to show one of these elements is basically "boolish" and was not correct. This was adjusted to show the correct failed result.

chiljamgossow (approved):
LGTM

mtwalli (approved):

PR #3760, opened 2021-07-20
QRCode Generation Performance Improvements (DEV,EXPOSUREAPP-8113)

d4rken:
The core idea is to leverage an image loading library (which has caching, sampling, efficient bitmap generation) and let it treat our qrcode strings like image sources.

* Use image loading library [`Coil`](https://github.com/coil-kt/coil) with custom pipeline to efficiently generate QrCode image data.
	* Custom `Fetcher` that turns the qrcode data into a size agnostic bitmatrix (i.e. 1 for black square, 0 for white square).
	* Custom bit matrix decoder that turns the size agnostic QrCode data into a perfectly scaled bitmap for the target `ImageView`.
* Margin can't be specified anymore for generated QrCodes.
	* Every generated qrcode has a margin of at least 1 square size, as the margin is necessary for it to be scanned on dark backgrounds.
	* If more margin is necessary for design purposes, that's up to the ImageView.
* "QrCodeLength" can't be specified anymore.
	* The generated qrcode data is size agnostic, just 1 or 0 for black or white scare.
	* The bitmap will be generated with perfect size for each `ImageView`, `Coil` will take care of caching, re-use and downsampling.
* You specify the error correction level and the charset, and set an image target and the rest is automatic.
     * Special size requirements, e.g. QrCode for posters can be passed via `Coil` size requirements and will automatically be considered. 
* Can be re-used for all future QrCode generation needs:

Local testing showed a performance improvement between from 2x to 20x depending on the device and screen.
Generating the right size bitmaps seems to have the biggest impact on initial load, while the details screens profit from caching.

Overall nice for everyone, but I hope this also has a positive impact on #3549.

Basic use:

```Kotlin
imageView.loadAny(CoilQrCode(content = certificate.qrCode)) {
    crossfade(true)
    placeHolderView(image, progressBar)
}
```

Advanced use, (e.g. poster stuff):
```Kotlin
val req = ImageRequest.Builder(context).apply {
    data(
        CoilQrCode(
            content = traceLocation.locationUrl,
            options = QrCodeOptions(correctionLevel = correctionLevel)
        )
    )
    size(template.qrCodeLength)
}.build()
context.imageLoader.execute(req).drawable
```

### Testing
View QR Codes in the app:
* Certificates
* Tracing
* RATProfile
* Posters

Why `Coil` over `Picasso, Glide, Fresco`?
Overall `Coil` and `Glide` are pretty close to what I'm looking for. Glide is more powerful (or rather flexible), but we don't need that extra power and Coil is more lightweight as a dependency, using mostly libraries we already include. This could have been done with any of these libraries though.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+
+        Timber.v("Bitmap generation took %dms", System.currentTimeMillis() - bitmapStart)
+
+        return DecodeResult(
+            drawable = BitmapDrawable(resources, bitmap),
+            isSampled = false,
+        )
+    }
```
d4rken, line 94: This turns a size agnostic qrcode matrix into a bitmap with a target size. It handles sizing and margin.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/QrCodeBitMatrixFetcher.kt
```diff
@@ -0,0 +1,53 @@
...
+            val rawMatrix: Array<ByteArray> = qrCode.matrix.array
+            it.writeObject(rawMatrix)
+        }
+
+        return SourceResult(
+            source = buffer,
+            mimeType = BitMatrixDecoder.MIME_TYPE,
+            dataSource = DataSource.MEMORY,
+        )
+    }
```
d4rken, line 50: This uses the zxring library to turn a qrcode string into a size agnostic `ByteMatrix`.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/presencetracing/ui/poster/QrCodePosterTestViewModel.kt
```diff
@@ -84,13 +89,21 @@ class QrCodePosterTestViewModel @AssistedInject constructor(
...
+                            content = traceLocation.locationUrl,
+                            options = QrCodeOptions(correctionLevel = correctionLevel)
+                        )
+                    )
+                    size(length)
+                }.build()
+                context.imageLoader.execute(req).drawable
+            }
+
+            qrCodeImage.postValue(drawable)
```
mtwalli, line 106: This block seems to be repeated , maybe we cloud extract into a reusable function such as:
```kotlin
fun Context.loadQrcode(coilQrCode: CoilQrCode, size:Int) = run {
                 val req = ImageRequest.Builder(applicationContext).apply {
                     data(coilQrCode) 
                     size(size)
                 }.build()
                 applicationContext.imageLoader.execute(req).drawable
             }

   qrCodeImage.postValue(drawable)
}
```

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/presencetracing/ui/poster/QrCodePosterTestViewModel.kt
```diff
@@ -84,13 +89,21 @@ class QrCodePosterTestViewModel @AssistedInject constructor(
...
+
+            val drawable = run {
+                val req = ImageRequest.Builder(context).apply {
+                    data(
+                        CoilQrCode(
+                            content = traceLocation.locationUrl,
+                            options = QrCodeOptions(correctionLevel = correctionLevel)
+                        )
+                    )
+                    size(length)
```
mtwalli, line 101: size could be also specified in the `options`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/PersonDetailsQrCard.kt
```diff
@@ -29,9 +31,15 @@ class PersonDetailsQrCard(parent: ViewGroup) :
...
         val curItem = payloads.filterIsInstance<Item>().singleOrNull() ?: item
-        image.setImageBitmap(curItem.qrCodeBitmap)
+
         curItem.apply {
-            qrCodeBitmap?.let { progressBar.hide() }
+            image.loadAny(
+                CoilQrCode(content = curItem.certificate.qrCode)
+            ) {
+                crossfade(true)
+                placeHolderView(image, progressBar)
```
mtwalli, line 40: Looks unusual for me , why do we have to specify the` image` and `progressBar` as placeHolder ?
should not be only the progress bar?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/coil/CoilExtensions.kt
```diff
@@ -0,0 +1,21 @@
+package de.rki.coronawarnapp.util.coil
+
+import android.view.View
+import androidx.core.view.isInvisible
+import coil.request.ImageRequest
+
+fun ImageRequest.Builder.placeHolderView(
```
mtwalli, line 7: I think the name here is confusing , it is not actually a place holder displayed by image itself, maybe 🤔 `loadingView` and we cloud pass the progressBar only?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/QrCodeOptions.kt
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @param correctionLevel [ErrorCorrectionLevel]
+ * @param characterSet [String]
+ */
+data class QrCodeOptions(
+    val correctionLevel: ErrorCorrectionLevel = ErrorCorrectionLevel.M,
+    val characterSet: Charset = Charsets.UTF_8,
+) {
+    val optionsKey by lazy {
+        "#$correctionLevel#$characterSet".toSHA1()
```
mtwalli, line 16: maybe just the `hashCode` It is not really necessary to get `SHA1` right ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
```diff
@@ -0,0 +1,100 @@
...
+
+    companion object {
+        private val BLACK_SQUARE = 1.toByte()
+        const val MIME_TYPE = "image/qrcode-bitmatrix"
```
mtwalli, line 98: is that a standard MIME type ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
```diff
@@ -0,0 +1,100 @@
...
+    override suspend fun decode(pool: BitmapPool, source: BufferedSource, size: Size, options: Options): DecodeResult {
+        val bitmapStart = System.currentTimeMillis()
+
+        val matrix = ObjectInputStream(source.inputStream()).use {
+            it.readObject() as Array<ByteArray>
+        }
+        val (matrixHeight, matrixWidth) = run {
+            //  The matrix is array[y][x], e.g. a list of columns
+            matrix.size to matrix.first().size
+        }
```
mtwalli, line 44: `run` block is redundant

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
```diff
@@ -0,0 +1,100 @@
...
+
+        val marginFactor = 1
+        val squareSize = min(
+            targetWidth.toFloat() / (matrixWidth + marginFactor * 2),
+            targetHeight.toFloat() / (matrixHeight + marginFactor * 2)
+        ).nextUp()
+
+        // The rest of the space if the target size is not a multiple of the matrix size
+        val padding = (targetHeight - squareSize * matrixWidth) / 2f
```
mtwalli, line 58: `matrixHeight` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
```diff
@@ -0,0 +1,100 @@
...
+        val (targetWidth, targetHeight) = when (size) {
+            OriginalSize -> matrixWidth to matrixWidth
+            is PixelSize -> size.width to size.height
+        }
+
+        val marginFactor = 1
+        val squareSize = min(
+            targetWidth.toFloat() / (matrixWidth + marginFactor * 2),
+            targetHeight.toFloat() / (matrixHeight + marginFactor * 2)
+        ).nextUp()
```
mtwalli, line 55: what the margin here for ?
for specific ⬛  or for whole QrCode?
I think the margin can be handled by axing already

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/QrCodeBitMatrixFetcher.kt
```diff
@@ -0,0 +1,53 @@
...
+            /**
+             * We cannot use Charsets.UTF_8 as zxing calls toString internally
+             * and some android version return the class name and not the charset name
+             */
+            EncodeHintType.CHARACTER_SET to data.options.characterSet.name()
+        )
+
+        val qrCode = Encoder.encode(
+            data.content,
+            data.options.correctionLevel,
```
mtwalli, line 34: correction level is already passed by `hints`

Server-Protocol-Buffer/build.gradle
```diff
@@ -5,9 +5,6 @@ sourceSets {
     main.java.srcDirs += "${protobuf.generatedFilesBaseDir}/main/javalite"
 }
 
-sourceCompatibility = JavaVersion.VERSION_1_8
-targetCompatibility = JavaVersion.VERSION_1_8
```
mtwalli, line 9: I added this while working on screenshots , please double check? run the screenshots task if necessary

mtwalli:
`Coil` is also Kotlin friendly library and uses Coroutines. it is best choice for a Kotlin project.

mtwalli:
Tested on device looks very good 👍 with caching. Transition to qrcode full screen seems to be broken now , it does not have the smooth transition like before

SamuraiKek (commented):
Very nice, from a UI/performance perpsective this works really well. Much snappier on my Mate 8 with Android 6 and I can also see the difference on my Galaxy S10 👍

SamuraiKek:
> Tested on device looks very good 👍 with caching. Transition to qrcode full screen seems to be broken now , it does not have the smooth transition like before

I can still see the smooth transition. Which certificate did you see this happen with?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
d4rken: No, but it has the right format :wink:.
It's not exposed outside this internal pipeline, we basically just need something unique to identify the data, could also have been `aj1j123k123`.

Server-Protocol-Buffer/build.gradle
d4rken: Since adding Coil, it would not work like this.

I found this via:
https://stackoverflow.com/questions/31870426/how-to-set-source-version-of-java-library-in-android-studio-project

which seems to indicate that there is a bug when using pure java libraries as submodule in Android projects.

I'll check the screenshots again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/QrCodeBitMatrixFetcher.kt
d4rken: Correct, it can't be passed via the hints, will remove.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
d4rken: We no longer use zxing for generating the visual representation. We only use it to get the matrix data.
The margin is the size of a qrcode square.

So a margin factor of 1 means that there is at least white space equal to the size of 1 qrcode "pixel" around the whole qrcode.

I played around with it a bit, and if the qrcode is on dark background, we need a small border around the qrcode for it to be scanned reliably. Likely because the dark background touching the black orientation squares in the corners of the qrcode.

So the produced image has the minimum margin for it to be able to be scanned. Any extra margin should be added via `android:margin/padding` in the UI.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/coil/BitMatrixDecoder.kt
d4rken: Hm, actually even better, let's do:

```kotlin
        val paddingHeight = (targetHeight - squareSize * matrixHeight) / 2f
        val paddingWidth = (targetWidth - squareSize * matrixWidth) / 2f
```

then we also covered imageviews that are not square.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/qrcode/QrCodeOptions.kt
d4rken: The `hashCode` is not guaranteed to be the same across different app sessions (jvm runs).
`Coil` expects a key that is consistent, e.g. to use during disk caching.

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/presencetracing/ui/poster/QrCodePosterTestViewModel.kt
d4rken: The options are qrcode **data**  specific (like error correction level and charset), while the size is a UI detail.
The options go towards the data generation part (data matrix) and do not change depending on the size of the target.

* Step 1 generate data (doesn't need size)
* Step 2 draw data into bitmap (can re-use the data from step 1 for different bitmap sizes)

If we introduce the size as part of the options, the generated data matrix could no longer be re-used for images of different sizes as the `QrCodeOptions` would be different.

We only need to specify the size at all here, because the way the poster generation currently works, we don't have a target to give to `Coil`, so the library can't figure out the size itself.

Server-Protocol-Buffer/build.gradle
d4rken: Screenshots fail, but because of something else:
https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/18215/workflows/b290dbf6-db4d-47cd-9341-3e844f078091/jobs/155003

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/PersonDetailsQrCard.kt
d4rken: From inside the Coil listener, afaict, we don't have access to the target view without guessing/casting).
So we need the reference to to show/hide it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/coil/CoilExtensions.kt
d4rken: I've renamed it to `loadingView`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/PersonDetailsQrCard.kt
d4rken: The best solution would probably be to no longer have a "progressbar" but use an animated drawable that we set.
Outside of this PR's scope thought.

Server-Protocol-Buffer/build.gradle
d4rken: https://github.com/corona-warn-app/cwa-app-android/pull/3760/commits/cdc99618768a26b8795e4c12c50edb94f755b454

Should fix screenshots: Missing mocks and qrcode generation.

`FakeImageLoader.kt` offers methods to get the same image loading as in the live app, or override to always get the same. I added two examples, one for qrcodes and one for `R.drawable.*`

d4rken: New  screenshot test is running:
https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/18235/workflows/97ed8c1c-a451-43c3-b4da-d8744ebeca0b/jobs/155229

d4rken:
Currently still fixing screenshots:
https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android?branch=chore%2Fqrcode-screenshot-test

d4rken:
Screenshots passed:
https://app.circleci.com/pipelines/github/corona-warn-app/cwa-app-android/18241/workflows/84dc4c2b-0324-4860-b43d-31c0604f618c/jobs/155302/steps

QRCodes seem to have been generated on all screenshots too :partying_face: 
[Screenshot from 2021-07-21 16-13-58](https://user-images.githubusercontent.com/1439229/126503540-f384071c-cd25-456b-8cdc-9a45b2caf8b6.png)

mtwalli:
> > Tested on device looks very good 👍 with caching. Transition to qrcode full screen seems to be broken now , it does not have the smooth transition like before
> 
> I can still see the smooth transition. Which certificate did you see this happen with?

It is broken for me, it is not a blocker we can improve it in another PR. this one is already big enough

### PR:

https://user-images.githubusercontent.com/25054729/126509157-7e4e057c-4934-4827-82cb-bc5d8d6f665a.mp4

### 2.6:

https://user-images.githubusercontent.com/25054729/126509193-6e68cad6-5c33-4b99-a8ad-6fa3e986eee6.mp4

mtwalli (approved):
LGTM, nice improvement!

SamuraiKek (approved):

PR #3761, opened 2021-07-20
Retrieve DSCs from server (EXPOSUREAPP-8362)

jurajkusnier:
implement `DscServer` response for downloading List of DSCs from server


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/server/DscServer.kt
```diff
@@ -1,12 +1,70 @@
...
+) {
 
-    suspend fun getDscList(): ByteArray {
-        throw NotImplementedError()
+    suspend fun getDscList(): DscList {
+        return try {
+            dscApi.dscList().parseAndValidate(
+                ErrorCode.FILE_MISSING,
+                ErrorCode.SIGNATURE_INVALID,
+                ErrorCode.EXTRACTION_FAILED
```
d4rken, line 28: The error codes can be hardcoded.
This was passed in a different implementation as the parsing logic was re-used.
I think here the `dscList()` will be the only use, right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/server/DscServerModule.kt
```diff
@@ -0,0 +1,34 @@
...
+
+@Module
+class DscServerModule {
+
+    @Reusable
+    @Provides
+    fun apiV1(
+        @DownloadCDNHttpClient httpClient: OkHttpClient,
+        @DownloadCDNServerUrl url: String,
+        @CertificateValueSet cache: Cache,
```
d4rken, line 22: Is this the right cache?

d4rken: Did you mean `@CertificateValidation`?

d4rken (commented):
Nice tests :heart:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/server/DscServer.kt
jurajkusnier: updated

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/server/DscServerModule.kt
jurajkusnier: updated

d4rken (approved):
lgtm

mtwalli (approved):
LGTM

PR #3767, opened 2021-07-21
Fix networkstate update behavior (DEV)

d4rken:
The network callback for onUnavailable() is not reliable.
Solution: Just force a refresh on any state change.

Also see:
https://issuetracker.google.com/issues/144891976
According to that ticket, it's fixed in a "future release", but on my Pixel 3a@Android11, it's not :shrug:.

@BMItr noticed this during debugging.


BMItr:
@d4rken its much better than before.
Unfortunally its not as reliable..
e.g. Wifi OFF->ON is detected
Wifi ON->OFF is not detected.
Maybe you have an idea..

d4rken:
> e.g. Wifi OFF->ON is detected

This was also working for me before.

> Wifi ON->OFF is not detected.

This should be fixed by this PR, on which devices does it happen, and how did you test it?

BMItr:
> > e.g. Wifi OFF->ON is detected
> 
> This was also working for me before.
> 
> > Wifi ON->OFF is not detected.
> 
> This should be fixed by this PR, on which devices does it happen, and how did you test it?

Tested on Pixel 3xl.
There might be no 100%reliable solution...

chiljamgossow (approved):
Seems to be an improvement

BMItr (approved):
LGTM!

BMItr (approved):

PR #3773, opened 2021-07-21
Bouncy castle (DEV)

jurajkusnier:
- introduce a few methods helping with signature checks of certificates

PR #3779, opened 2021-07-21
Signature validation (EXPOSUREAPP-8365)

mtwalli:
Review [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/8a7f326acee2e640b2a23f93acb2476942e2e88e/docs/spec/dgc-overview-client.md#verifying-the-signature-of-a-dgc-against-a-list-of-dscs) 

Note : 
There is a ticket to add bouncy castle license , if not done by then

### TODO 
- [x] Add unit tests


Corona-Warn-App/build.gradle
```diff
@@ -318,6 +318,7 @@ dependencies {
     // Required for rdgca.verifier.app:dgc-certlogic-android-light API 26 Time API
     coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.1.5'
 
+    implementation 'org.bouncycastle:bcpkix-jdk15on:1.56'
```
d4rken, line 321: Wondering why this specific version :thinking: 
https://github.com/corona-warn-app/cwa-app-android/pull/3773#discussion_r674680548

mtwalli: @jurajkusnier  do you know?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationTestData.kt
```diff
@@ -174,11 +191,18 @@ class VaccinationTestData @Inject constructor(
         expiresAt = Instant.parse("2022-05-11T09:23:03.000Z"),
     )
 
-    val personBVac1CertificateData = DccData(
+    private val personBVac1CertificateData = DccData(
```
d4rken, line 194: Why make these private? Can't we no longer access them in tests then?

d4rken (commented):
Code looks good :+1:

Corona-Warn-App/build.gradle
jurajkusnier: hm.. let's try the latest version `1.69`
it could be just mistake

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidatorTest.kt
```diff
@@ -0,0 +1,24 @@
+package de.rki.coronawarnapp.covidcertificate.signature.core
```
d4rken, line 1: We should also test this against our minAPI. Encryption stuff often has API related issues.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationTestData.kt
mtwalli: IDE: could be private ?
Me: Yes, why not 😀

Corona-Warn-App/build.gradle
mtwalli: updated

d4rken:
Please also check my additional comments from #3773

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -1,19 +1,184 @@
...
+        "1.3.6.1.4.1.0.1847.2021.1.3"
+    )
+
+    private val oidSet = vcOids + tcOids + rcOids
+
+    /**
+     * @throws InvalidHealthCertificateException if validation fail, otherwise it is OK!
+     */
+    suspend fun validateSignature(dccData: DccData<*>) {
+        val dscData = dscRepository.dscData.first()
```
mtwalli, line 67: @jurajkusnier  I have added `DscItem` which offers data a convenient types such as:

```kotlin
dscListItem -> DscItem(
  dscListItem.kid.toOkioByteString().base64(),
  dscListItem.data.toOkioByteString()
)
```

Please map `dscListItems` into it ?

jurajkusnier: sure, it will be changed in my following PR

jurajkusnier (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/CborEncoding.kt
```diff
@@ -0,0 +1,40 @@
...
+import java.security.spec.RSAPublicKeySpec
+
+fun DscMessage.signedPayload(): ByteArray = CBORObject.NewArray().apply {
+    Add("Signature1")
+    Add(protectedHeader.toByteArray())
+    Add(ByteArray(0))
+    Add(payload.toByteArray())
+}.EncodeToBytes()
+
+fun PublicKey.toRsaPublicKey(): PublicKey {
```
d4rken, line 35: We have an internal abstraction of this, see `RSAKey` which offers automatic conversion & serialization options.

Would it make sense to use that here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+    private val dscRepository: DscRepository
+) {
 
-    suspend fun isSignatureValid(dscData: DscData, certificateData: DccData<*>): Boolean {
-        Timber.tag(TAG).d("isSignatureValid(certificateData=%s)", certificateData.header)
-        // TODO
-        return true
+    init {
+        Security.addProvider(BouncyCastleProvider()) // For SHA256withRSA/PSS
+    }
```
d4rken, line 37: I would prefer if we add this somewhere global, close to other access of the `Security.*` singleton.

Inside `CoronaWarnApplication` we have:
```kotlin
Security.insertProviderAt(Conscrypt.newProvider(), 1)
```

If we spread these calls over the app, then it could lead to weird issues.

We could either move this also to `CoronaWarnApplication` or maybe even create a `SecuritySetup.setup() ...` class that is called from `CoronaWarnApplication` and does these things.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+class DscSignatureValidator @Inject constructor(
+    private val dscRepository: DscRepository
+) {
 
-    suspend fun isSignatureValid(dscData: DscData, certificateData: DccData<*>): Boolean {
-        Timber.tag(TAG).d("isSignatureValid(certificateData=%s)", certificateData.header)
-        // TODO
-        return true
+    init {
+        Security.addProvider(BouncyCastleProvider()) // For SHA256withRSA/PSS
```
d4rken, line 36: Could the `addProvider` call lead to an exception being thrown that we could catch?
e.g. could `addProvider` fail on a device that natively has `SHA256withRSA/PSS` available so we could still continue?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+        "1.3.6.1.4.1.1847.2021.1.3",
+        "1.3.6.1.4.1.0.1847.2021.1.3"
+    )
+
+    private val oidSet = vcOids + tcOids + rcOids
+
+    /**
+     * @throws InvalidHealthCertificateException if validation fail, otherwise it is OK!
+     */
+    suspend fun validateSignature(dccData: DccData<*>) {
```
d4rken, line 57: I would like it if we could rename this `requireValidSignature` like the other apicalls that either do nothing or throw an exception, e.g. `requireNotNull(....)`.

What do you think?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+        }
+    }
+
+    private fun findDscCertificate(
+        dscData: DscData,
+        dscMessage: DscMessage
+    ): X509Certificate {
+        val toVerify = dscMessage.signedPayload()
+        val filteredDscSet = dscData.dscList.filter { it.kid == dscMessage.kid }
+        Timber.d("filteredDscSetSize=${filteredDscSet.size}")
```
d4rken, line 73: Please also add `tag(TAG)` to the other `Timber` calls.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+            try {
+                val valid = Signature.getInstance(dscMessage.algorithm.algName).verify(publicKey, toVerify, signature)
+                Timber.d("Dsc certificate (${dsc.kid}) is valid=$valid")
+
+                if (valid) {
+                    x509Certificate = dscCertificate
+                    break
+                }
+            } catch (ignored: Exception) {
+                // Ignore errors / continue
```
d4rken, line 98: Could we maybe add a small  `.v` log message here? 
This seems like a critical part that could fails for various reasons that we would be happy about to see in a debug log.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+            }
+        }
+
+        return x509Certificate ?: throw InvalidHealthCertificateException(HC_DSC_NO_MATCH)
+    }
+
+    private fun DscItem.toX509certificate(): X509Certificate = CertificateFactory
+        .getInstance("X.509")
```
d4rken, line 106: I don't know how large `matchedDscSet` can become, but maybe we should create the `CertificateFactory` for `X.509` only once. Either once before the  `for` in L:82 or maybe even once per class instance with a `lazy`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+        .generateCertificate(data.toByteArray().inputStream())
```
d4rken, line 107: According to the java docs of `generateCertificate` if `mark` and `reset` is supported by the inputstream, the method will only read one certificate from it and return.

That would mean that `generateCertificate` does not necessarily close the input stream, do we need to close it then? 

e.g. `data.toByteArray().inputStream().use { /*...*/ }`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -2,16 +2,144 @@ package de.rki.coronawarnapp.covidcertificate.signature.core
...
+        } catch (e: CertificateNotYetValidException) {
+            throw InvalidHealthCertificateException(HC_DSC_NOT_YET_VALID)
+        }
+    }
+
+    private fun X509Certificate.checkOidsIntersect(dccData: DccData<*>) {
+        val extendedKeysIntersect = extendedKeyUsage.orEmpty().toSet() intersect oidSet
+        if (extendedKeysIntersect.isEmpty()) return // OK!
+        when (dccData.certificate) {
+            is VaccinationDccV1 -> if (vcOids.intersect(extendedKeysIntersect).isEmpty())
```
d4rken, line 124: Would prefer to have brackets on multiline `if`s.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/encoding/Base45Decoder.kt
```diff
@@ -72,13 +72,12 @@ object Base45Decoder {
...
-            0L,
-            { index, acc: Long, element ->
-                if (!alphabet.contains(element))
-                    throw IllegalArgumentException(element.toString())
-                pow(int45, index) * alphabet.indexOf(element) + acc
-            }
-        )
+            0L
+        ) { index, acc: Long, element ->
+            if (!alphabet.contains(element))
```
d4rken, line 77: Same line or brackets please.
Errors are difficult to spot in code review when lines are added.

d4rken (commented):
Looks well. :+1:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
mtwalli: Please check the specs , it states on no logs are required

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
mtwalli: I don't think `require*` is a suitable name

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/CborEncoding.kt
mtwalli: I don't think so, plus we don't need serialised key here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
mtwalli: brakets will make it noisy , both are valid syntax for one  `if`   statement I prefer it that way

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
d4rken: Pretty common though? `requireContext/require()/requireNotNull`

Not a blocker for me though.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
d4rken: @mlenkeit Why should nothing be logged here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
d4rken: Hm, subjective.

Not a blocker for me.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/SecurityProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+import javax.inject.Inject
+
+class SecurityProvider @Inject constructor() {
+
+    fun setup() {
+        try {
+            Timber.d("Setting up security provider")
+            // Enable Conscrypt for TLS1.3 Support below API Level 29
+            Security.insertProviderAt(Conscrypt.newProvider(), 1)
+            Security.addProvider(BouncyCastleProvider()) // For SHA256withRSA/PSS
```
d4rken, line 16: Could you add a class name or PR comment here for what this is used for?
And a comment to the class using it too, so that when either is removed/changed, the new dev knows what to update or check?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/SecurityProvider.kt
```diff
@@ -0,0 +1,21 @@
...
+        } catch (e: Exception) {
+            Timber.d(e, "setting up Security provider failed")
```
d4rken, line 18: If you edit this file again, please capitalize align capitalization with the other log messages :bow:

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -143,6 +139,7 @@ class DscSignatureValidator @Inject constructor(
     }
 
     companion object {
+        private val certificateFactory = CertificateFactory.getInstance("X.509")
```
d4rken, line 140: We should make this a `lazy`.
It's tricky when `companion object` is initialized, and this call may access the provider, docs:

```
     * <p> Note that the list of registered providers may be retrieved via
     * the {@link Security#getProviders() Security.getProviders()} method.
```

We don't want this to accidentally initilize before the provider is setup (i.e. `Security.add` was called).

Sorry I only thought now of it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
mlenkeit: signature verification can naturally fail, so the idea was not to pollute the log in that case, especially when there's a larger set of DSCs to check. But in reality, the 99% case is 1 DSC and pass, so proper logging for the 1% makes sense. See spec update.

d4rken (approved):

jurajkusnier (approved):

PR #3788, opened 2021-07-22
Follow up: Local Statistics Unittests (DEV)

kolyaopahle:
### Description
This PR adds a few unittests to the local statistics serialization code and also reduces the size of the serialized json data by removing the uniqueID from the serialized package


mtwalli (approved):
Thank you Harry Potter !

BMItr (approved):
lgtm

PR #3793, opened 2021-07-23
Map error codes for NO_NETWORK correctly during vertification (EXPOSUREAPP-8546)

d4rken:
Turn off internet and try to verify a certificate, you should now see an error message about your internet connection.

Screenshot from 2021-07-23 11-08-50


jurajkusnier (approved):
Hm... dialogs on the certificated details screen and on the "Gültigkeit des Zertifikates" screen are different for the same connection issue. But it's not a merge blocker
image

mtwalli (approved):
LGTM

PR #3794, opened 2021-07-23
Vaccine details are reversed (EXPOSUREAPP-8547)

kolyaopahle:
### Description
During some refactoring of the vaccine details screen the Vaccine type and Vaccine labels got mixed up again, this PR fixes that mixup

### Steps to reproduce
1. Scan a test certificate
2. check that vaccine type is something like SARS-CoV-2 mRNA vaccine
3. check that vaccine is something like Comirnaty


marcauberer (approved):

BMItr (approved):
tested on oneplus 8.
image

PR #3799, opened 2021-07-23
Background check of certificate states (EXPOSUREAPP-8364)

d4rken:
Wires up most  of the state check background logic.

~~- [ ] Implement actually showing notifications (may be done in a follow up PR)~~
- [x] Fill out unit tests (may be done in a follow up PR)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscData.kt
```diff
@@ -1,7 +1,9 @@
 package de.rki.coronawarnapp.covidcertificate.signature.core
 
 import okio.ByteString
+import org.joda.time.Instant
 
 data class DscData(
-    val dscList: List<Pair<ByteString, ByteString>>
+    val dscList: List<Pair<ByteString, ByteString>>,
+    val updatedAt: Instant,
```
d4rken, line 8: @jurajkusnier You are doing the repo right? Could you provide such an `updatedAt` field?

jurajkusnier: sure, I can add it
btw I will change `dscList` type to `List<DscListItem>` instead of `Pair<>`

jurajkusnier: what should be in `updatedAt` when we load data from `DefaultDscSource` ?

d4rken: `Instant.EPOCH`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
```diff
@@ -18,7 +18,9 @@ class CovidCertificateConfigMapper @Inject constructor() : CovidCertificateConfi
 
         return CovidCertificateConfigContainer(
             testCertificate = rawConfig.dgcParameters.mapCovidCertificateConfig(),
-            expirationThresholdInDays = rawConfig.dgcParameters.expirationThresholdInDays
+            expirationThresholdInDays = Duration.standardDays(
```
chiljamgossow, line 21: Max said this is the number of date changes, not a multiple of 24h intervals. That's why I think this conversion is not accurate, however, it won't have any meaningful consequences for users.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
```diff
@@ -51,7 +53,7 @@ class CovidCertificateConfigMapper @Inject constructor() : CovidCertificateConfi
 
     data class CovidCertificateConfigContainer(
         override val testCertificate: CovidCertificateConfig.TestCertificate = TestCertificateConfigContainer(),
-        override val expirationThresholdInDays: Int = 1
```
chiljamgossow, line 54: I put that in here deliberately, not sure what the fallback should be.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/statecheck/DccStateChecker.kt
```diff
@@ -0,0 +1,46 @@
...
+    ): Flow<CwaCovidCertificate.State> = combine(
+        appConfigProvider.currentConfig,
+        dscRepository.dscData
+    ) { appConfig, dscData ->
+        val isSignatureValid = dscSignatureValidator.isSignatureValid(dscData, dccData)
+
+        val nowUtc = timeStamper.nowUTC
+
+        // expiration check
+        val expirationState = expirationChecker.getExpirationState(
```
chiljamgossow, line 35: you could move that down and only check expiration for valid signatures

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
```diff
@@ -51,7 +53,7 @@ class CovidCertificateConfigMapper @Inject constructor() : CovidCertificateConfi
...
+        override val expirationThresholdInDays: Duration = Duration.standardDays(1),
```
mtwalli, line 56: should not be `14` ? https://github.com/corona-warn-app/cwa-app-tech-spec/blob/78009f7e3aa6aef56892caedcfe478bfb9d4d691/docs/spec/handling-of-app-config.md#parameters-for-digital-green-certificate

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/expiration/DccExpirationChecker.kt
```diff
@@ -0,0 +1,31 @@
...
+        dccData: DccData<*>,
+        expirationThreshold: Duration,
+        now: Instant,
+        timeZone: DateTimeZone = DateTimeZone.getDefault()
+    ): CwaCovidCertificate.State = with(dccData) {
+        val expiresAt = header.expiresAt
+        val daysUntilExpiration = now.daysUntil(expiresAt, timeZone)
+        val diff = daysUntilExpiration - expirationThreshold.standardDays
+        return when {
+            daysUntilExpiration <= 0 -> CwaCovidCertificate.State.Expired(expiresAt)
```
mtwalli, line 25: Something to consider in unit tests?
https://github.com/corona-warn-app/cwa-app-android/pull/3804

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
d4rken: Seems like a duration to me though, whether we treat it as day changes would be up to the configs consumer. I'd prefer to model `expirationThresholdInDays` as a Duration.
I'll rename it `expirationThreshold` as whether it is hours/days/weeks is encoded in `Duration` and not something the config consumer has to care about.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
d4rken: I changed it to 14 days based on:
https://github.com/corona-warn-app/cwa-app-android/pull/3799#discussion_r676476499
and added some unit tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CovidCertificateConfigMapper.kt
d4rken: Done and added tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/statecheck/DccStateChecker.kt
d4rken: Thanks, good point!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/statecheck/DccStateChecker.kt
```diff
@@ -0,0 +1,49 @@
...
+        dccData: DccData<*>
+    ): Flow<CwaCovidCertificate.State> = combine(
+        appConfigProvider.currentConfig,
+        dscRepository.dscData
+    ) { appConfig, dscData ->
+        val isSignatureValid = dscSignatureValidator.isSignatureValid(dscData, dccData)
+        if (!isSignatureValid) {
+            Timber.tag(TAG).w("Certificate had invalid signature: %s (kid)", dccData)
+            return@combine CwaCovidCertificate.State.Invalid
+        }
```
mtwalli, line 35: `DscSignatureValidator` does not return boolean anymore , you should wrap`dscSignatureValidator.isSignatureValid(dscData, dccData)`  in a` try {} catch {}` 
if it throw an `InvalidHealthCertificateException` -> `Invalid` ,otherwise it should be `Valid`

d4rken: Done
https://github.com/corona-warn-app/cwa-app-android/pull/3799/commits/256de94211174800a2c7a04ac3218e26a1e55535

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/expiration/DccExpirationChecker.kt
```diff
@@ -0,0 +1,38 @@
...
+        now: Instant,
+        timeZone: DateTimeZone = DateTimeZone.getDefault()
+    ): CwaCovidCertificate.State = with(dccData) {
+        val expiresAt = header.expiresAt
+        val daysUntilExpiration = now.daysUntil(expiresAt, timeZone)
+
+        return when {
+            daysUntilExpiration < 0 -> CwaCovidCertificate.State.Expired(expiresAt)
+            daysUntilExpiration == 0 -> {
+                if (now.toDateTime(timeZone).isAfter(expiresAt)) {
```
chiljamgossow, line 27: Here you can just compare millis, both are Instants. time zone should not affect this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
```diff
@@ -8,10 +8,9 @@ import javax.inject.Inject
 @Reusable
 class DscSignatureValidator @Inject constructor() {
 
-    suspend fun isSignatureValid(dscData: DscData, certificateData: DccData<*>): Boolean {
+    suspend fun isSignatureValid(dscData: DscData, certificateData: DccData<*>) {
```
chiljamgossow, line 11: As it does not return a Boolean any more, is... is misleading. Maybe just checkSignature(...)?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
```diff
@@ -112,6 +112,10 @@ object TimeAndDateExtensions {
 
     fun Instant.toLocalDateUserTz(): LocalDate = this.toUserTimeZone().toLocalDate()
 
+    fun LocalDate.isDifferentDay(other: LocalDate): Boolean {
+        return this.dayOfYear != other.dayOfYear || this.year != other.year
```
chiljamgossow, line 116: Equals does not work?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidator.kt
d4rken: Was changed to `validateSignature` by merging 2.7 into my branch.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/TimeAndDateExtensions.kt
d4rken: I thought this wouldn't work, but you are right, it does work. Removed the extension function again.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/expiration/DccExpirationChecker.kt
d4rken: Thanks, will adjust.

chiljamgossow (approved):
Thanks!

mtwalli (approved):

PR #3775, opened 2021-07-21
Add Invalid QrCode State + Grey Card (EXPOSUREAPP-8373, 8376)

axelherbstreith:
Add invalid state to

- [x] certificate detail screen
- [x] person detail screen
- [x] person overview screen
- [x] Change underlaying qr code when invalid or expired
- [x] Add Grey Card to all certificate detail screens when invalid or expired
- [x] Add Grey state to Person Detail Card

To test: 
- check if the QR code gets greyed + color scheme changes to grey in PersonOverview, PersonDetail and all CertificateDetail Screens if a certificate is invalid or expired

PR #3803, opened 2021-07-26
Display certificate state (EXPOSUREAPP-8243)

SamuraiKek:
Adds the "Expiring soon/Expired/Invalid" states to the certificates in the certificates tab, person details screen and certificate details screen. Can be seen under the QR code of each screen and also on the individual certificate cards on person details screen.

Scan recovery/test/vaccination certificates. Travel to the future and check displayed status.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/PersonDetailsQrCard.kt
```diff
@@ -68,6 +70,48 @@ class PersonDetailsQrCard(parent: ViewGroup) :
                     )
                 }
             }
+
+            when (certificate.getState()) {
+                is CwaCovidCertificate.State.ExpiringSoon -> {
+                    expirationStatusIcon.visibility = View.VISIBLE
+                    expirationStatusIcon.setImageDrawable(ContextCompat.getDrawable(context, R.drawable.ic_av_timer))
```
mtwalli, line 77: there is a handy extension for that `context.getDrawableCompat( R.drawable.ic)` , please use it instead of 
`ContextCompat.getDrawable(context, R.drawable.ic)`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/PersonDetailsQrCard.kt
```diff
@@ -68,6 +70,48 @@ class PersonDetailsQrCard(parent: ViewGroup) :
...
+                            R.drawable.ic_error_outline
+                        )
+                    )
+                    expirationStatusText.visibility = View.VISIBLE
+                    expirationStatusText.text = context.getText(R.string.certificate_qr_invalid_signature)
+                    qrSubtitle.visibility = View.GONE
+                }
+
+                else -> {
+                }
```
mtwalli, line 113: should we handle `Valid` case also to hide validity  views and show `qrSubtitle`?
In normal cases state should not change from `Invalid`, `Expired`, `ExpiringSoon` to `Valid` again , but the device time is a factor here , some (users/ testers) would act as time travellers which could lead to having some views visible for old state

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -65,7 +68,65 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
...
+                    expirationStatusText.text = getText(R.string.certificate_qr_invalid_signature)
+                    expirationStatusBody.visibility = View.VISIBLE
+                    expirationStatusBody.text = getText(R.string.invalid_certificate_signature_info)
+                    qrSubtitle.visibility = View.GONE
+                }
+
+                else -> {
+                }
+            }
+        }
```
mtwalli, line 129: this code block seems to be common for all Certificates, should we create an extension function for that and reuse it for all?
```kotlin
private fun IncludeCertificateQrcodeCardBinding.bindValidityViews(certificate: CwaCovidCertificate) {
 // Add logic here 
}
```

mtwalli (commented):
1- Please use `getDrawableCompat` whenever needed?
2- Should we handle `Valid` state also?
3- Please create a common function to handle validity cases for all certificates?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
SamuraiKek: Yeah, it'd be nicer. I'm on it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/PersonDetailsQrCard.kt
SamuraiKek: Haven't thought about this but you're right, I'll handle it as well.

NataliaLemmerth (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+                expirationStatusText.visibility = View.GONE
+                expirationStatusBody.visibility = View.GONE
+                qrTitle.visibility = setVisibility(isPersonDetails)
+                qrSubtitle.visibility = setVisibility(isPersonDetails)
+                startValidationCheckButton.visibility = setVisibility(isPersonDetails)
+            }
+        }
+    }
+
+    private fun setVisibility(visibility: Boolean) =
```
mtwalli, line 103: there is `isVisible` from `core-ktx` , ex: `view.isVisible = true`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+        if (visibility) {
+            View.VISIBLE
+        } else {
+            View.GONE
+        }
+
+    fun displayIndividualCardsExpirationState(
```
mtwalli, line 110: Can we make it extension function of TextView`.displayIndividualCardsExpirationState`()

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+        certificateExpiration: TextView,
+        context: Context,
```
mtwalli, line 112: you can access the context from the view  😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+import de.rki.coronawarnapp.databinding.IncludeCertificateQrcodeCardBinding
+import de.rki.coronawarnapp.util.ContextExtensions.getDrawableCompat
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toShortDayFormat
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toShortTimeFormat
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toUserTimeZone
+
+object CertificateStateHelper {
+
+    fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
+        context: Context,
```
mtwalli, line 20: you can access the context from the view 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -0,0 +1,140 @@
...
+import de.rki.coronawarnapp.covidcertificate.recovery.core.RecoveryCertificate
+import de.rki.coronawarnapp.covidcertificate.test.core.TestCertificate
+import de.rki.coronawarnapp.covidcertificate.vaccination.core.VaccinationCertificate
+import de.rki.coronawarnapp.databinding.IncludeCertificateQrcodeCardBinding
+import de.rki.coronawarnapp.util.ContextExtensions.getDrawableCompat
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toShortDayFormat
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toShortTimeFormat
+import de.rki.coronawarnapp.util.TimeAndDateExtensions.toUserTimeZone
+
+object CertificateStateHelper {
```
mtwalli, line 17: I think the object is not required if you make all functions extensions?

mtwalli (commented):
Looks good 👍,Will test it once conflicts are fixed

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Good point :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Sure

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: I've tried this but I think it only works when it's an extension function over an actual View like `TextView`. This one is an extension function over the binding of this `include` and I don't see any context if I remove it explicitly.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: When turning this into an extensiuon function, I indeed have access to the context without having to pass it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Still required due to the extension function for the `include`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
mtwalli: maybe `val context = root.context` ?>

mtwalli:
The following are still missing for Invalid certificate:

1- Overlay 
Screenshot 2021-07-28 at 10 37 08


2- Icon
Screenshot 2021-07-28 at 10 37 19

SamuraiKek:
> The following are still missing for Invalid certificate:
> 
> 1- Overlay
> Screenshot 2021-07-28 at 10 37 08
> 
> 2- Icon
> Screenshot 2021-07-28 at 10 37 19

They're being added in #3775

mtwalli (approved):
Lgtm, Nice work!

chiljamgossow (approved):
LGTM

NataliaLemmerth (approved):

PR #3804, opened 2021-07-26
Fix certficate status last day expiration (DEV)

SamuraiKek:
In an edge case where the current date would be the same as the expiration date of the certificate, the status would be set to "Expired" even though the certificate should still be valid on that day as well. This small change sets the status to "ExpiringSoon".

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DccStateChecker.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DccStateChecker.kt
@@ -45,7 +45,7 @@ internal fun DccData<*>.getExpirationState(
     val daysUntilExpiration = now.daysUntil(expiresAt, timeZone)
     val diff = daysUntilExpiration - expirationThresholdInDays
     return when {
-        daysUntilExpiration <= 0 -> CwaCovidCertificate.State.Expired(expiresAt)
+        daysUntilExpiration < 0 -> CwaCovidCertificate.State.Expired(expiresAt)
         diff > 0 -> CwaCovidCertificate.State.Valid(expiresAt)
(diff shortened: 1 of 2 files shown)


mtwalli (approved):
Lgtm

chiljamgossow:
@mlenkeit I understand expiration as the first date of invalidity. How is it meant o be?

chiljamgossow:
If we want to make it more precise, we can change it to check the timestamp, but we should not show it as valid for the whole day.

SamuraiKek:
> If we want to make it more precise, we can change it to check the timestamp, but we should not show it as valid for the whole day.

I like this option the most. The expiresAt has a date and time: "2021-06-03T10:12:48.000+02:00". If I'm on 03.06.2021, and it's 9AM, I should still see the certificate as "ExpiringSoon".

SamuraiKek:
I've put it to draft until we get a clear answer about this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DccStateChecker.kt
```diff
@@ -45,7 +45,7 @@ internal fun DccData<*>.getExpirationState(
     val daysUntilExpiration = now.daysUntil(expiresAt, timeZone)
     val diff = daysUntilExpiration - expirationThresholdInDays
     return when {
-        daysUntilExpiration <= 0 -> CwaCovidCertificate.State.Expired(expiresAt)
+        daysUntilExpiration < 0 -> CwaCovidCertificate.State.Expired(expiresAt)
```
chiljamgossow, line 48: So you want to add the case == 0 and check the timestamp to decide whether it is already expired

mtwalli:
Addressed by https://github.com/corona-warn-app/cwa-app-android/commit/f5e6b716d0273d0f062082c643a80859a7b90e2d

PR #3811, opened 2021-07-27
Account for certificate state when determining priority (EXPOSUREAPP-8559)

d4rken:
When finding the highest priority certificate, take the certificate state (valid/expired/invalid) into account.

First group the certificates by state, then process them in batches.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
```diff
@@ -184,61 +183,75 @@ private fun Collection<CwaCovidCertificate>.rule9FindOldRaTest(
...
-        return it
+): CwaCovidCertificate = this
+    .also { Timber.v("findHighestPriorityCertificate(nowUtc=%s): %s", nowUtc, this) }
+    .run {
+        val valid = filter {
+            it.getState() is CwaCovidCertificate.State.Valid || it.getState() is CwaCovidCertificate.State.ExpiringSoon
+        }
+        val expired = filter { it.getState() is CwaCovidCertificate.State.Expired }
+        val invalid = filter { it.getState() is CwaCovidCertificate.State.Invalid }
+        if (this.size != (valid.size + expired.size + invalid.size)) {
```
chiljamgossow, line 197: Wouldn't this be something for a unit test?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
```diff
@@ -184,61 +183,75 @@ private fun Collection<CwaCovidCertificate>.rule9FindOldRaTest(
...
+            throw IllegalStateException("State grouping failure, certificate count does not match.")
+        }
+        listOf(valid, expired, invalid)
     }
+    .map { certsForState ->
+        certsForState.rule1FindRecentPcrCertificate(nowUtc)?.let {
```
chiljamgossow, line 203: Is this relying on valid being the first in the list?

chiljamgossow: Maybe add a comment...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
d4rken: I'm not sure how I could test this.
The idea is to have something that causes a crash if someone adds new states in the future.
I will check again if I can make it an exhaustive `when`.

chiljamgossow (approved):
🚀 🏄‍♂️

mtwalli (approved):

PR #3813, opened 2021-07-27
Extend technical validation (EXPOSUREAPP-8705)

mtwalli:
- Extend technical validation to check signature validation 
Screenshot 2021-07-27 at 17 52 54 Screenshot 2021-07-27 at 17 54 01


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidator.kt
```diff
@@ -39,13 +42,23 @@ class DccValidator @Inject constructor(
         return DccValidation(
             userInput = userInput,
             validatedAt = timeStamper.nowUTC,
+            signatureCheckPassed = signatureCheckPassed,
+            certificateExpiredAt = certificate.header.expiresAt.toLocalDateTime(DateTimeZone.UTC),
```
chiljamgossow, line 46: Shouldn't this be converted using user tz as it is shown to the user? Or is there a requirement to use utc?

mtwalli: Make sense

NataliaLemmerth (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidation.kt
```diff
@@ -13,6 +13,8 @@ import org.joda.time.LocalDateTime
 data class DccValidation(
     val userInput: ValidationUserInput,
     val validatedAt: Instant,
+    val signatureCheckPassed: Boolean,
+    val certificateExpiredAt: LocalDateTime,
```
chiljamgossow, line 17: I would have passed the Instant here and converted in the UI. But that's up to you

chiljamgossow (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidation.kt
chiljamgossow: Actually, I would have taken the date from the certificate, and not have added it to the validation result.

mtwalli: You are right, totally forgot that, it should be aligned with other related certificate information in other view holders

NataliaLemmerth (approved):
UA approved

chiljamgossow (approved):
LGTM

axelherbstreith (approved):

PR #3815, opened 2021-07-27
Local statistics connectivity (EXPOSUREAPP-8436)

BMItr:
This PR addresses Exposureapp-8436.
[Android] Statistics behavior when offline during onboarding.

How to test:
- visit statistic cards on the MainFragment
- check that its prevented to add a new card without internet connection (by notification or addView hiding)
- check if the addCard Element reappears after switching Internet/Wifi on
(if there is no cached data or internet connection, statistic data should not be shown)

Note: MERGE AFTER #3767 (done)

Screenshots:

No Internet Case:
- regular AddCard Item should not be visible (greyed out with explaining text)
image

with internet:
image


NataliaLemmerth (dismissed):
UA approved.

vaubaehn:
Hi @BMItr and @fynngodau ,
I think this PR is related to https://github.com/corona-warn-app/cwa-app-android/issues/3823, isn't it?

axelherbstreith (dismissed):
lgtm 👍 Tested on Samsung S8

marcauberer:
@BMItr Test scenario on a Pixel 3, Android 11:
- Added one local statistics card
- Disabled WiFi (or any other internet connection)

The result was the following:
device-2021-07-28-174550

I think we should introduce another error message to put on the disabled "add-card". Users might get confused when they've linked < 5 tiles, but the "add-card" shows "Maximale Anzahl der lokalen 7-Tage-Inzidenzen erreicht". What do you think?

BMItr:
> @BMItr Test scenario on a Pixel 3, Android 11:
> 
> * Added one local statistics card
> * Disabled WiFi (or any other internet connection)
> 
> The result was the following:
> 
> Screen 1
> 
> I think we should introduce another error message to put on the disabled "add-card". Users might get confused when they've linked < 5 tiles, but the "add-card" shows "Maximale Anzahl der lokalen 7-Tage-Inzidenzen erreicht". What do you think?

totally agree.
Check the new screenshot in the description.
Greetings.

marcauberer:
@BMItr perfect, thanks for fixing!

marcauberer (dismissed):

axelherbstreith (dismissed):

ChristinaTsigganos (changes requested):
please take over UA text

BMItr:
> please take over UA text

all right. 
LayoutAdjustment needed for longer texts. in progress...

fynngodau:
Two things are not entirely addressed by this PR so far:

* When the internet connection is disabled while the selection is open, the user can still finish adding the card with the same behavior as in https://github.com/corona-warn-app/cwa-documentation/issues/672.
  This could happen for example in a situation where the system thinks that the network is available, but requests are never answered and time out.
* After toggling a network connection, the statistics cards jump to the default position (first card after the "add card card").

BMItr:
> Two things are not entirely addressed by this PR so far:
> 
> * When the internet connection is disabled while the selection is open, the user can still finish adding the card with the same behavior as in [Bad behavior when adding local incidence without network access cwa-documentation#672](https://github.com/corona-warn-app/cwa-documentation/issues/672).
>   This could happen for example in a situation where the system thinks that the network is available, but requests are never answered and time out.
> * After toggling a network connection, the statistics cards jump to the default position (first card after the "add card card").

it could also be that an asteriod hits the user during adding a card. This case is also not handled.

axelherbstreith (approved):

marcauberer (approved):

vaubaehn:
I didn't have problems with asteroids so far, but
> This could happen for example in a situation where the system thinks that the network is available, but requests are never answered and time out.

is a situation that I'm often experiencing. And probably not me alone.
Sure, not everything needs to be pressed into one PR, but at least it's worth to think about.

ChristinaTsigganos (approved):
UA approved

BMItr:
> I didn't have problems with asteroids so far, but
> 
> > This could happen for example in a situation where the system thinks that the network is available, but requests are never answered and time out.
> 
> is a situation that I'm often experiencing. And probably not me alone.
> Sure, not everything needs to be pressed into one PR, but at least it's worth to think about.

@vaubaehn @fynngodau 
Yes. You are right on this.
Follow Up PR will come (I already prepared something about the jumping card issue, and maybe I have somehting for the in-between-no-internet-case).
For now I cannot pack more cases in this ticket (as requested) regarding other tickets...
Greetz 👌

PR #3818, opened 2021-07-27
Add FAQ link for error "already registered" (EXPOSUREAPP-8568)

d4rken:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/scan/DccQrCodeScanFragment.kt
```diff
@@ -110,6 +110,11 @@ class DccQrCodeScanFragment :
                             }
                             setNeutralButton(R.string.dcc_signature_validation_dialog_ok_button) { _, _ -> }
                         }
+                        it.errorCode == InvalidHealthCertificateException.ErrorCode.ALREADY_REGISTERED -> {
+                            setNeutralButton(R.string.error_button_dcc_faq) { _, _ ->
+                                openUrl(R.string.error_button_dcc_faq_link)
```
marcauberer, line 115: Maybe it could be a bit misleading if the app shows a dialog with the error code `VC_ALREADY_REGISTERED` and the [linked faq article](https://www.coronawarn.app/de/faq/#hc_prefix_invalid) deals with `HC_PREFIX_INVALID` 🤷‍♂️
But better than before, I guess ...

mtwalli (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/scan/DccQrCodeScanFragment.kt
chiljamgossow: Unfortunately, the ticket does not say where link should lead to. The FAQ was introduced in connection with the prefix error.

axelherbstreith (approved):

PR #3821, opened 2021-07-28
RSA Algo unit-test (DEV)

mtwalli:
https://github.com/corona-warn-app/cwa-app-tech-spec/commit/4ff61c569e15cb52615c1635c8c795674af322a1


axelherbstreith (approved):

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidatorTest.kt
```diff
@@ -44,6 +46,9 @@ class DscSignatureValidatorTest : BaseTest() {
         every { dscRepository.dscData } returns flowOf(dscData)
         every { securityProvider.setup() } just Runs
 
+        // Add RSA algorithm in unit tests context.
+        Security.addProvider(BouncyCastleProvider())
```
chiljamgossow, line 50: Does this need to removed again?

BMItr (approved):
lgtm

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscSignatureValidatorTest.kt
mtwalli: No, it will stay. SecurityProvider.setup() should NOT be mocked ,but the crypto but `org.conscrypt.*` is not available for unit testing. we could use an openKJDK implementation ,but it is not really worth it for one unit test, more information here. https://github.com/google/conscrypt/issues/649

mtwalli: For the actual App `SecurityProvider` will be doing that ,but it is a problem for the real app

PR #3827, opened 2021-07-28
Remove incorrect rounded corners in certificate valid asset (COMMUNITY)

fynngodau:
The Certificate Is Valid screen can currently show an incorrect blue corner like seen here:

Corner screenshot

This occurs because the asset that is displayed inside the toolbar has rounded corners. Those likely should not be there, as it causes this bug and as the open and the failure results also don't have such rounded corners.

Hence, this PR removes the rounded corners from the asset.

Previous | Now
---|---
image | image


mtwalli (approved):

marcauberer (approved):
lgtm

PR #3835, opened 2021-07-29
Improve scrolling behavior in validation result screen (COMMUNITY)

fynngodau:
In the validation result screen, the header image currently disappears so late that it can overlap with the text too much, which doesn't look great (see video demonstration below).

To improve the result screens, this PR contains the following changes:

* Separate gradient from background header vector asset (it is not quite a 1:1 reproduction but good enough).
* Attach `OnOffsetChangedListener` to bar layout through pre-existing `AppBarLayout.OnOffsetChangedListener` extension function to hide the header image at the same rate as e.g. the Europe background image on the vaccination details screen so that it disappears quicker and overlaps less with the toolbar text for a smoother experience.
* Show gradient in toolbar even if it is collapsed, like is the case for RAT, event organizer, event attendee, vaccination, …, screens.

#### Previously

https://user-images.githubusercontent.com/16943720/127478166-78397a2c-e638-4f9d-b597-d062598197a2.mp4

#### After

https://user-images.githubusercontent.com/16943720/127477560-a71f2582-9f9b-4ed6-8465-a0c6e39f1e70.mp4


mtwalli (approved):
LGTM

SamuraiKek (approved):
LGTM. Nice attention to details <3

PR #3840, opened 2021-07-30
List of DSCs test menu (EXPOSUREAPP-8752)

jurajkusnier:
- Add test menu
image

### DIFF
--- a/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/dsc/ui/DscViewModel.kt
+++ b/Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/dsc/ui/DscViewModel.kt
@@ -1,0 +1,42 @@
+package de.rki.coronawarnapp.test.dsc.ui
+
+import androidx.lifecycle.LiveData
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import de.rki.coronawarnapp.covidcertificate.signature.core.DscRepository
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
+import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
+import kotlinx.coroutines.flow.map
+import org.joda.time.Instant
+
+class DscViewModel @AssistedInject constructor(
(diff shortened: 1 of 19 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscRepository.kt
```diff
@@ -58,6 +58,9 @@ class DscRepository @Inject constructor(
     suspend fun clear() {
         Timber.d("clear()")
         localStorage.clear()
+        internalData.updateBlocking {
+            defaultDscData.getDscData()
+        }
```
mtwalli, line 63: why are you adding the default value again ?

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/dsc/ui/DscViewModel.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.lifecycle.LiveData
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+import de.rki.coronawarnapp.covidcertificate.signature.core.DscRepository
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
+import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
+import kotlinx.coroutines.flow.map
+import org.joda.time.Instant
+
+class DscViewModel @AssistedInject constructor(
```
mtwalli, line 12: Please add Test in the name ?  `DscTestViewModel`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/signature/core/DscRepository.kt
jurajkusnier: Because even if we clear the cache, `internalData` flow is not updated immediately by default

mtwalli (dismissed):

d4rken (approved):

mtwalli (approved):

PR #3843, opened 2021-07-30
TabBar badge for certificate state changes (EXPOSUREAPP-8377)

d4rken:
Store information about certificate state changes to display a TabBar badge.

* Check the current state and compare with the last acknowledged state
* If both are different (except for the happy path: `lastAcknowledgedState=null`, `currentState=Valid`), it's counted and displayed
* When visiting the certificate tab, the changed states are acknowledged, the repository is updated with the new acknowledged state and the badge will vanish.

Because this supports going Invalid -> Valid -> Invalid etc. you can test this by just time traveling into the future.


SamuraiKek (approved):
LGTM. Works as expected.

jurajkusnier (approved):

PR #3845, opened 2021-07-30
Fix certificate highest priority fallback behavior (DEV)

d4rken:
Fixes return the wrong certificate if we have no only expired/invalid ones.

If no certificates matches a rule, we return null.
Since the introduction of the states, we process them in the order Valid/Expiring>Expired>Invalid.
When we have `listOf(emptyList(), emptyList(), list(cert))` we would get `listOf(null,null,cert)`.
As our fallback behavior is `list.firstOrNull() ?: fallback` it would always trigger the fallback if the first list (valid/expiring) is empty.


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensionsTest.kt
```diff
@@ -174,4 +175,42 @@ class PersonCertificatesExtensionsTest : BaseTest() {
...
+     *
+     *
+     * Good: listOf(null,null,cert).mapNotNull {...}.firstOrNull() ?: fallback
+     *
+     * findHighestPriorityCertificate(nowUtc=2021-06-24T14:00:00.000Z): [VaccinationCertificate(#1), VaccinationCertificate(#7)]
+     * No certs with state Valid/ExpiringSoon
+     * Checking 2 certs with for Expired
+     * Rule 3 match (Series-completing Vaccination Certificate > 14 days): VaccinationCertificate(#7)
+     * No certs with state Invalid
+     */
```
d4rken, line 197: Example

BMItr (approved):
lgtm

mtwalli (approved):

PR #3868, opened 2021-08-03
ICAO name for certificates (EXPOSUREAPP-8686)

AlexanderAlferov:
Introducing ICAO (standardized) name in certificates (test, vaccination and recovery).

Screenshot 2021-08-03 at 15 58 18

How to test:
1) generate test, vaccination and recovery QR codes
2) go to certificates details


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -73,6 +73,8 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
     ) {
         qrCodeCard.bindValidityViews(certificate, isCertificateDetails = true)
         fullname.text = certificate.fullNameFormatted
+        icaoname.text = "${certificate.personIdentifier.lastNameStandardized}," +
+            " ${certificate.personIdentifier.firstNameStandardized}"
```
mtwalli, line 77: should we create something like `certificate.fullNameStanardizedFormatted` ? 
Similar to `certificate.fullNameFormatted`

Corona-Warn-App/src/main/res/values/covid_certificate_attribute_strings.xml
```diff
@@ -13,6 +13,8 @@
 
     <!-- XTXT: Vaccination Certificate Detail Screen Attribute: Name -->
     <string name="covid_certificate_attribute_name" translatable="false">"Name, Vorname / Name, First Name"</string>
+    <!-- XTXT: Vaccination Certificate Detail Screen Attribute: ICAO Name -->
+    <string name="covid_certificate_attribute_icao_name" translatable="false">"Standardisierter Name, Vorname / Standardized Name, First Name"</string>
```
mtwalli, line 17: Please add the key the German file too ?

AlexanderAlferov: We don't have a German version of this file: as you can see this file contains both German and English labels simultaneously: `Standardisierter Name, Vorname / Standardized Name, First Name`

Description in this file:
```
<!-- ####################################
        Special string files for the strings
        that should always keep the german / english
        version independent from the selected system language
```

jurajkusnier (commented):
Please update also screenshots

Corona-Warn-App/src/main/res/values/covid_certificate_attribute_strings.xml
mtwalli: Ah you are  right!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
```diff
@@ -86,6 +86,10 @@ data class RecoveryCertificateContainer(
             override val fullNameFormatted: String
                 get() = certificate.nameData.fullNameFormatted
 
+            override val fullNameStandardizedFormatted: String
+                get() = "${certificate.personIdentifier.lastNameStandardized}," +
```
jurajkusnier, line 90: for the name formatting we already have this logic:
```kotlin
val fullNameFormatted: String
            get() = when {
                firstName.isNullOrBlank() -> lastName
                else -> "$lastName, $firstName "
            }
```
we should probably consider also here that the `firstName` could be empty or null and adjust

mtwalli (approved):
lgtm

Screenshot 2021-08-04 at 14 44 45

jurajkusnier (approved):
LGTM

PR #3871, opened 2021-08-04
Adjust Checkin submission (EXPOSUREAPP-8792)

mtwalli:
- Provide unencrypted check-ins to submission payload if unencrypted flag is ON
- Provide encrypted check-ins to submission payload
- Adjust unit tests to cover both cases


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/cryptography/CheckInCryptography.kt
```diff
@@ -15,7 +16,9 @@ class CheckInCryptography @Inject constructor(
         transmissionRiskLevel: Int
     ): CheckInProtectedReport {
         // TODO
```
SamuraiKek, line 18: Do we still need the `TODO` here?

mtwalli: After Juraj's PR no 😉

jurajkusnier (commented):
looks good so far

PR #3875, opened 2021-08-04
Translation Delivery (DEV)

service-tip-git:
Translation Delivery

### DIFF
--- a/Corona-Warn-App/src/main/res/values/strings.xml
+++ b/Corona-Warn-App/src/main/res/values/strings.xml
@@ -291,13 +291,13 @@
     <!--    Dialog part 1-->
     <string name="risk_details_information_body_period_logged">"Your risk of infection can be calculated only for periods during which exposure logging was active. The logging feature should therefore remain active permanently. Exposure logging covers the last 14 days."</string>
     <!-- XTXT: risk details - infection period logged information body, under 14 days -->
-    <string name="risk_details_information_body_period_logged_assessment_under_14_days">"The Corona-Warn-App was installed %s days ago. Your risk of infection is calculated for the periods during which exposure logging was active. If you have encountered other people and exposure logging was active, your risk of infection is calculated."</string>
+    <string name="risk_details_information_body_period_logged_assessment_under_14_days">"The Corona-Warn-App was installed on %s. Your risk of infection is calculated for the periods during which exposure logging was active. If you have encountered other people and exposure logging was active, your risk of infection is calculated."</string>
     <!-- XTXT: risk details - infection period logged 0 days -->
-    <string name="risk_details_information_body_period_logged_today">"today"</string>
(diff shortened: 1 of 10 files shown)


service-tip-git:
This pull request is for translation delivery. Please review and approve them.

mtwalli (approved):

jurajkusnier (approved):

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -291,13 +291,13 @@
...
     <!-- XTXT: risk details - infection period logged information body, under 14 days -->
-    <string name="risk_details_information_body_period_logged_assessment_under_14_days">"The Corona-Warn-App was installed %s days ago. Your risk of infection is calculated for the periods during which exposure logging was active. If you have encountered other people and exposure logging was active, your risk of infection is calculated."</string>
+    <string name="risk_details_information_body_period_logged_assessment_under_14_days">"The Corona-Warn-App was installed on %s. Your risk of infection is calculated for the periods during which exposure logging was active. If you have encountered other people and exposure logging was active, your risk of infection is calculated."</string>
     <!-- XTXT: risk details - infection period logged 0 days -->
-    <string name="risk_details_information_body_period_logged_today">"today"</string>
+    <string name="risk_details_information_body_period_logged_today">"Today"</string>
```
marcauberer, line 296: This should not be capitalized, because it gets inserted to the string resource above.

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -291,13 +291,13 @@
...
     <!-- XTXT: risk details - infection period logged 1 day -->
-    <string name="risk_details_information_body_period_logged_yesterday">"yesterday"</string>
+    <string name="risk_details_information_body_period_logged_yesterday">"Yesterday"</string>
```
marcauberer, line 298: Here as well

marcauberer (changes requested):

Corona-Warn-App/src/main/res/values/strings.xml
ralfgehrer: Actually, we shall not use concatenated strings, because the translation system cannot handle these edge cases very well as strings are translated without the context of the placeholder. Spelling (capital or not) differs from language to language. Therefore, UA cannot/won't fix this.

I'd suggest we merge this PR nevertheless and I will create a follow-up bug ticket. What do you think @marcauberer and @mtwalli ?

mtwalli: I agree

marcauberer: Okay for me. I'll approve ...

marcauberer (approved):

PR #3882, opened 2021-08-05
Fix issue with text placeholders (EXPOSUREAPP-8870)

ralfgehrer:
@marcauberer this PR fixes the root cause of the text issues you raised in https://github.com/corona-warn-app/cwa-app-android/pull/3875

**Please test:**
App was installed

- Today
- Yesterday
-  between 2 and 14 days

and check the respective display text.


ChristinaTsigganos (dismissed):
UA approved

jurajkusnier (approved):

mtwalli (approved):

PR #3883, opened 2021-08-05
Download / Save Protected CheckIns (EXPOSUREAPP-8794)

mtwalli:
- Support encrypted APIs  
- Introduced api `mode` which is passed down from the Task into the sync tool. meaning if App config changes while the Task is running , the new value will be considered in the next run to avoid downloading/ saving the wrong data. 
- Both encrypted and decrypted packages are saved in binary files 

## TODO 
- [x] more unit tests


jurajkusnier (approved):
LGTM

BMItr (approved):
👌

PR #3886, opened 2021-08-06
Fix CWA crashes and cannot be restarted after the attempt to receive a faked-vaccination certificate (EXPOSUREAPP-8876)

jurajkusnier:
- DCC-Server response with fake cert couldn't crash the CWA
- correct RAT and PCR tests should work as usual


ralfgehrer (approved):

mtwalli (approved):
Lgtm, Tested PCR and RAT certificates and pending cases still work as before

PR #3887, opened 2021-08-06
"Aktuell verwendetes Zertifikat" unnecessary line break (EXPOSUREAPP-8871)

marcauberer:
Addresses [EXPOSUREAPP-8871](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8871)

- Layout components in the certificate cards are now aligned to the parent component and not to each other to avoid strange text cropping / line breaks
- Unified the usage of the regarding dimen resource instead of hardcoding the values


ralfgehrer:
Hi @marcauberer thanks for the PR. Can you please re-target the PR onto the branch `release/2.8.x` as we shall only fix high and very high bugs on 2.7 at this point.

Corona-Warn-App/src/main/res/layout/recovery_certificate_card.xml
```diff
@@ -12,8 +12,8 @@
         android:id="@+id/certificate_bg"
         android:layout_width="88dp"
         android:layout_height="95dp"
-        android:layout_marginStart="16dp"
```
ralfgehrer, line 15: If I recall correctly, we decided at some point within the android team to rather use explicit dimension - that is 16dp over @dimen/spacing_small for better tooling support within Android Studio and easier readability.

NataliaLemmerth (approved):
UA approved

PR #3888, opened 2021-08-06
"Aktuell verwendetes Zertifikat" unnecessary line break (EXPOSUREAPP-8871)

marcauberer:
Addresses [EXPOSUREAPP-8871](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8871)

- Layout components in the certificate cards are now aligned to the parent component and not to each other to avoid strange text cropping / line breaks
- Unified the usage of the hardcoding of dimension values


ralfgehrer (approved):
LGTM

mtwalli (approved):

PR #3889, opened 2021-08-06
Fix: correct dcc state check worker (EXPOSUREAPP-8849)

AlexanderAlferov:
Correct worker to check dcc state in background.


mtwalli (approved):
Lgtm

ralfgehrer (approved):

PR #3893, opened 2021-08-09
Certificates schema validation EU exceptions (EXPOSUREAPP-8687)

SamuraiKek:
Trims white spaces and removes null values from badly formed EU certificates that would otherwise not pass the schema validator but should be accepted by wallet apps.

Testing can be done through the CLI app on mock-cloud.

`cwa tc gen --preset 25`
`cwa vc gen --preset 27`


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
```diff
@@ -148,6 +153,29 @@ class DccV1Parser @Inject constructor(
         }
     }
 
+    private fun JsonElement.filterExceptions(): JsonElement =
```
mtwalli, line 156: Can you please add some unit tests ?

Corona-Warn-App/src/main/assets/jsonschema-dcc-ebb6720.json
```diff
@@ -122,8 +123,8 @@
         "gnt": {
           "title": "Standardised forename",
           "description": "The forename(s) of the person, transliterated ICAO 9303",
-          "type": "string",
-          "pattern": "^[A-Z<]*$",
+          "__comment": "Special case handling: original pattern check dropped",
+          "__pattern": "^[A-Z<]*$",
```
mtwalli, line 127: does this affect normal cases ?

SamuraiKek: I don't think it should. I've copied this from the tech spec https://github.com/corona-warn-app/cwa-app-tech-spec/pull/85/files#diff-2870955822e4772ae4e76684d41910ca95b6462cfcc2029016509bcde80701bd. Since the pattern check is simply dropped it shouldn't have any effect on existing certs and from my testing, the certs I've previously had were read just fine.

mtwalli (approved):
Lgtm, Tested `presets` mentioned in the description and other certificates too , looks fine

ralfgehrer (approved):

jurajkusnier (approved):

PR #3901, opened 2021-08-10
Improve logs for 8113 (DEV)

mtwalli:
Add more logs in the UI layer for EXPOSUREAPP-8113


jurajkusnier (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt
```diff
@@ -149,4 +154,8 @@ class PersonOverviewViewModel @AssistedInject constructor(
 
     @AssistedFactory
     interface Factory : SimpleCWAViewModelFactory<PersonOverviewViewModel>
+
+    companion object {
+        private const val TAG = "PersonOverviewViewModel"
```
BMItr, line 159: consider to use something like TAG = someObject::class.java.simpleName
(optional)

BMItr (approved):
lgtm

PR #3917, opened 2021-08-12
Warn info screen (EXPOSUREAPP-8947)

AlexanderAlferov:
Warn guests info screen.

Screenshot 2021-08-12 at 17 57 31 Screenshot 2021-08-12 at 17 57 07

How to test:
1) Home Screen -> "Planning an event?" card
2) Three point menu -> "In Vertretung warnen" 
3) ✨ Tadaaa ✨


NataliaLemmerth (approved):
UA approved

mtwalli (commented):
Very good 👍 , I don't think we need settings in this case

Corona-Warn-App/src/main/res/layout/trace_location_organizer_warn_info_fragment.xml
```diff
@@ -0,0 +1,136 @@
+<?xml version="1.0" encoding="utf-8"?>
```
jurajkusnier, line 1: could you call "Reformat Code" here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/warn/info/TraceLocationWarnInfoViewModel.kt
```diff
@@ -0,0 +1,18 @@
+package de.rki.coronawarnapp.ui.presencetracing.organizer.warn.info
```
jurajkusnier, line 1: Is this necessary? We have a static fragment with one button only for navigation to the next fragment

jurajkusnier (commented):
As @mtwalli mentioned we don't need settings.
IMHO in that case code could be much simpler just one static fragment no AutoInject, no ViewModel, no new modules

jurajkusnier (approved):
Nice 👍

mtwalli (approved):

PR #3928, opened 2021-08-16
First vaccination dose shows up as complete for recovered people (EXPOSUREAPP-8896)

SamuraiKek:
We already had the logic to show the vaccination as complete if the  `doseNumber == totalSeriesOfDoses`, what was left was to hide the Vaccination Info box that shows up if the vaccination is incomplete or it's complete but the number of days until immunity hasn't been reached. 

To test, scan one of the presets from [this PR](https://github.com/corona-warn-app/cwa-app-tech-spec/pull/91). The vaccination card should have the complete icon (the one with the chekmark) and it should say "Vaccination 1 of 1". Also the vaccination hint is not shown.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
@@ -102,17 +102,12 @@ class PersonDetailsViewModel @AssistedInject constructor(
             // Find any vaccination certificate to determine the vaccination information
             personCertificates.certificates.find { it is VaccinationCertificate }?.let { certificate ->
                 val vaccinatedPerson = vaccinatedPerson(certificate)
-                if (vaccinatedPerson != null && vaccinatedPerson.getVaccinationStatus(timeStamper.nowUTC) != IMMUNITY) {
-                    vaccinatedPerson.isFirstVaccinationDoseAfterRecovery()
-
-                    /**Don't show the vaccination hint section
-                     * if the person had their first dose of vaccine after recovering from covid.
-                     * Applies to ASTRA, BIONTECH and MODERNA vaccines only
-                     */
-                    if (!vaccinatedPerson.isFirstVaccinationDoseAfterRecovery()) {
-                        val timeUntilImmunity = vaccinatedPerson.getDaysUntilImmunity()
-                        add(VaccinationInfoCard.Item(timeUntilImmunity))
-                    }
+                if (vaccinatedPerson != null &&
+                    vaccinatedPerson.getVaccinationStatus(timeStamper.nowUTC) != IMMUNITY &&
+                    !vaccinatedPerson.isFirstVaccinationDoseAfterRecovery()
+                ) {
+                    val timeUntilImmunity = vaccinatedPerson.getDaysUntilImmunity()
+                    add(VaccinationInfoCard.Item(timeUntilImmunity))
                 }
             }
 
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
@@ -102,10 +102,7 @@ class PersonDetailsViewModel @AssistedInject constructor(
             // Find any vaccination certificate to determine the vaccination information
             personCertificates.certificates.find { it is VaccinationCertificate }?.let { certificate ->
                 val vaccinatedPerson = vaccinatedPerson(certificate)
-                if (vaccinatedPerson != null &&
-                    vaccinatedPerson.getVaccinationStatus(timeStamper.nowUTC) != IMMUNITY &&
-                    !vaccinatedPerson.isFirstVaccinationDoseAfterRecovery()
-                ) {
+                if (vaccinatedPerson != null && vaccinatedPerson.getVaccinationStatus(timeStamper.nowUTC) != IMMUNITY) {
                     val timeUntilImmunity = vaccinatedPerson.getDaysUntilImmunity()
                     add(VaccinationInfoCard.Item(timeUntilImmunity))
                 }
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
```diff
@@ -71,6 +71,16 @@ data class VaccinatedPerson(
         return IMMUNITY_WAITING_DAYS - Days.daysBetween(newestFullDose.vaccinatedOn, today).days
     }
 
+    private fun getNewestFullDose() = vaccinationCertificates.filter { it.doseNumber == it.totalSeriesOfDoses }
+
+    fun isFirstVaccinationDoseAfterRecovery(): Boolean {
```
mtwalli, line 76: Please write some unit tests

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
```diff
@@ -102,7 +102,10 @@ class PersonDetailsViewModel @AssistedInject constructor(
             // Find any vaccination certificate to determine the vaccination information
             personCertificates.certificates.find { it is VaccinationCertificate }?.let { certificate ->
                 val vaccinatedPerson = vaccinatedPerson(certificate)
-                if (vaccinatedPerson != null && vaccinatedPerson.getVaccinationStatus(timeStamper.nowUTC) != IMMUNITY) {
+                if (vaccinatedPerson != null &&
+                    vaccinatedPerson.getVaccinationStatus(timeStamper.nowUTC) != IMMUNITY &&
+                    !vaccinatedPerson.isFirstVaccinationDoseAfterRecovery()
```
mtwalli, line 107: Please write some unit tests?

mtwalli: I actually rather change the logic of `VaccinatedPerson.getVaccinationStatus()` to get the status without having to call `isFirstVaccinationDoseAfterRecovery()` here , so maybe we just need to change line   **(you need to check what is best here )**
https://github.com/corona-warn-app/cwa-app-android/blob/316bd648c326d5e97408d749952778165235bd48/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt#L57

To

```kotlin  
daysToImmunity <= 0 || isFirstVaccinationDoseAfterRecovery() -> Status.IMMUNITY
 ```
by doing that :
- Persons who got these kind of vaccine are Immune , right ? and  Immune state is used in details screen to display different colouring. 
- the state will be shared and changed in one place 

What are you thoughts on this?

File: Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt [hunk 3]
```diff
@@ -71,6 +71,16 @@ data class VaccinatedPerson(
...
+        val vaccinationDetails = getNewestFullDose()[0].rawCertificate.vaccination
```
mtwalli, line 77: - Please add tests? 
- are you sure that is not causing an exception for other vaccine doses 1,2 or maybe even 3 in future 
`getNewestFullDose` is looking only for certificate where  `doseNumber == totalSeriesOfDoses` if Certificate is 1 of 2, 1 of 3, or 2 of 3 `getNewestFullDose` will return an emptyList and `getNewestFullDose()[0]` might cause a crash ??

mtwalli (commented):
Looks good, please add unit tests covering different vaccine does?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
SamuraiKek: Yes, you're right, I'm working on it right now.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPersonTest.kt
```diff
@@ -264,4 +268,38 @@ class VaccinatedPersonTest : BaseTest() {
...
+
+        val personData = mockk<VaccinatedPersonData>().apply {
+            every { vaccinations } returns setOf(
+                mockk<VaccinationContainer>().apply {
+                    every { toVaccinationCertificate(any(), any()) } returns mockk<VaccinationCertificate>().apply {
+                        every { vaccinatedOn } returns LocalDate.parse("2021-01-01")
+                        every { doseNumber } returns 1
+                        every { totalSeriesOfDoses } returns 1
+                        every { rawCertificate.vaccination.doseNumber } returns doseNumber
+                        every { rawCertificate.vaccination.medicalProductId } returns "EU/1/20/1528" // BIONTECH
```
mtwalli, line 284: can you please add tests for other vaccine types too ?

mtwalli:
Presets old/new look good, but despite that 1:1 has full immunity when validating the rules it displays that it is not valid in the selected country and have to wait 14 days. Can you please check with Max if we need to change anything?

Screenshot 2021-08-18 at 11 36 11

SamuraiKek:
> Presets old/new look good, but despite that 1:1 has full immunity when validating the rules it displays that it is not valid in the selected country and have to wait 14 days. Can you please check with Max if we need to change anything?

Alright, I'll talk to him, I don't think the tech spec mentions this https://github.com/corona-warn-app/cwa-app-tech-spec/pull/91/files

mlenkeit:
@mtwalli @SamuraiKek thanks for checking and raising this. It's a problem with the rule definition and not directly related to this change. I'll bring it up with the responsible people to adjust the rule definition. But you can proceed and merge the PR (if that's the only thing that was preventing the merge) 👍

mtwalli (approved):
Lgtm

BMItr (approved):
good tests.
lgtm

PR #3937, opened 2021-08-18
Risk calculation with result CALCULATION_FAILED (EXPOSUREAPP-9020)

mtwalli:
How verify the fix?
Before:
1- build `deviceForTestersRelease` not `deviceForTestersDebug`
2- Scan Qrcode from ticket
3-From Test menu -> Presence Tracing -> Hit `Run Task` and `Calculate Risk` 
4- Risk Card will show failed state 

After:
Do the same, but the Risk will be LOW as it

### DIFF
--- a/Corona-Warn-App/proguard-rules.pro
+++ b/Corona-Warn-App/proguard-rules.pro
@@ -39,6 +39,10 @@
     @retrofit2.http.* <methods>;
 }
 
+-keepclasseswithmembers class * {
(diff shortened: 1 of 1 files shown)


SamuraiKek (approved):
Looks good to me. Tested on Samsung S10+.

BMItr (approved):
tested on Pixel 2 device.
lgtm.

image

PR #3947, opened 2021-08-19
Update Layout in Validation Result Screen (EXPOSUREAPP-8802)

axelherbstreith:
Aligned UI with figma according to the ticket

Screenshot from updated screen:

image


Corona-Warn-App/src/main/res/layout/covid_certificate_validation_result_fragments.xml
```diff
@@ -16,15 +16,15 @@
...
             android:background="@drawable/covid_certificate_validation_gradient"
+            android:nestedScrollingEnabled="true"
             app:layout_scrollFlags="scroll|exitUntilCollapsed|snap"
             app:titleEnabled="false">
 
             <ImageView
                 android:id="@+id/header_image"
                 android:layout_width="match_parent"
-                android:layout_height="match_parent"
+                android:layout_height="248dp"
```
mtwalli, line 27: does this fit all screen sizes ?

axelherbstreith: Good point, tested it and it worked fine! But after some more testing I figured out that the height is not different on various screen sizes when using match_parent compared to dp directly - so I reverted the changes since they do not add any value

mtwalli (approved):
lgtm

BMItr (approved):

jurajkusnier (approved):

PR #3948, opened 2021-08-19
Create screenshot for different certificate types (EXPOSUREAPP-8943)

jurajkusnier:
Add missing screenshots for:
- invalid/expired vaccination certificate as shown on the right screenshot 
- invalid/expired recovery certificate as shown on the right screenshot 
- invalid/expired test certificate as shown on the right screenshot


mtwalli:
should this PR go into 2.9 ?

jurajkusnier:
> should this PR go into 2.9 ?

these are screenshots for 2.7, so it was supposed to go even earlier

mtwalli:
> > should this PR go into 2.9 ?
> 
> these are screenshots for 2.7, so it was supposed to go even earlier

it is not necessary to put in 2.8. afaik when don't have graphic changes (store images) in 2.8. 
Please rebase to 2.9

PR #3958, opened 2021-08-20
Small UI Refactoring (DEV)

axelherbstreith:
Due to the change to material design, button height decreased and therefore margins increased which leads to some large gaps.
This PR fixes this by minimizing the margins between the buttons - goal is to save space for other content.

Feel free to test this PR with different phones. Let me know if the lowered margins look strange on your device :) 

**Here is an example:** 

Before: 

image

After:

image


marcauberer (approved):
lgtm

mtwalli (approved):

Issue #3670, opened 2021-07-09
Digital certificate for rapid test is issued with "Standardized surname/forename" with only first letter


jkrwdf:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [?] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Google Pixel 3XL
- Android version: 11
- App version: 2.4.3

## Describe the bug

The digital EU COVID-19 certificate which is issued after a rapid test contains in the CBOR elements "nam/fnt" and "nam/gnt" only the first letter of the respective content, for example "M" when the name is "Mustermann".

## Steps to reproduce the issue

I created a "Schnelltestprofil" with my full data and underwent a rapid test at a test location which is fully integrated (reads QR code of profile, sends results to CWA, supports creation of digital EU certificate for rapid tests). 

After receiving the test result and the certificate, I decoded the certificate content (strip "HC1:", Base45 decode, ZLIB deflate, CBOR playground). 

The data about the name of the testee are exemplary: 

"nam": {"fnt": "M", "fn": "Mustermann", "gnt": "E", "gn": "Erika"}

## Expected behaviour

Elements "fnt" and "gnt" are expected according to 
https://ec.europa.eu/health/sites/default/files/ehealth/docs/covid-certificate_json_specification_en.pdf
to contain the 

"Surname(s) of the holder transliterated using the same
convention as the one used in the holder’s machine
readable travel documents"

in the case above: MUSTERMANN and ERIKA. 

## Additional context

Remark: In my digital vaccination certificates, those elements are populated correctly, therefore I assume it is the CWA which transfers those data incorrectly to the RKI for signing.

vaubaehn:
@jkrwdf wrote:
> In my digital vaccination certificates, those elements are populated correctly, therefore I assume it is the CWA which transfers those data incorrectly to the RKI for signing.

For the vaccination certificates, another system is used: IBM/Ubirch for data entry (by medical staff) in web-frontend and processing in backend, next to 3rd-party-solutions.

For the RATs T-System's system is used before test data are forwarded for later creating dcc/signing: entry of data by testcenter staff in web-frontend and further processing in backend, next to 3rd-party-solutions that provide data to backend via API.

[...]

Issue #3547, opened 2021-06-24
Subsequent DCCs: Matching of persons - which fields are used how?


vaubaehn:
## Avoid duplicates
* [x] Question is not already answered in the [FAQ](https://www.coronawarn.app/en/faq/)
* [x] Question is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation) - iOS may handle it differently
* [x] Question has not already been asked in another issue -  I think so.

## Your Question
I found this in Google Play Store reviews today:

> Mein Nachname wurde auf dem Impfzertifikat 1.Impfung in kleinen Buchstaben und im Zertifikat der 2. Impfung mit grossem Anfangsbuchstaben geschrieben. Die App liefert eine Fehlermeldung beim Scan der 2.Impfung und akzeotiert dessen Zertifikat nicht. Was soll ich machen?

In case the error message results from a mismatch of different capitalization of the person's name throughout the issuing of the DCCs:
I was looking to a PR that was associated with person identifiers for certificate matching recently. For some reason I can't find it anymore. But when I was looking into the code, I was very certain that matching for the name is performed with standardized name, which means, the compared strings are all big caps and free of any non-Latin characters. I think @chiljamgossow was working on it, but I'm not sure. So, this user reported problem should not have happend.

**Is matching with standardized name introduced in a later release (>=2.4.x)? Or did something else went wrong?**

A kind of flexible matching of certifcates (with regard to persons' names) becomes crucial at the moment, when people take (further) vaccinations/tests in other countries as their home countries. Often medical staff in other countries can't use/type the special characters (like 'ä') used in persons' names on their keyboards when issuing certificates, so relying on the standardized name is important.

Edit: I found another, similar report:

> Leider erhalte ich beim Scan des elektr, 2. Impfzertifikates immer die Fehlermedung "Die persönlichen Daten des Impfzertifikates stimmen nicht mit denen der bereits registrierten Zertifikate überein. Sie können in der App nur Zertifikate einer Person registrieten". Die Zertifikate wurden offiziell durch d.Apotheke ausgegeben. Die App CovPass ordnet beide fehlerfrei für meine 1.und 2. Impfung zu. Nachtrag: Danke für.d.schnelle Hilfe; d.Namensdarstellung war tats. fehlerhaft. (Kleinschrift)

d4rken:
[...]

PR #3975, opened 2021-08-24
DCC empty string handling (EXPOSUREAPP-9017)

SamuraiKek:
Removes unnecessary required check on some dcc attributes. Relates to #3893.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
```diff
@@ -98,42 +97,6 @@ class DccV1Parser @Inject constructor(
         return (vaccinations?.size ?: 0) + (tests?.size ?: 0) + (recoveries?.size ?: 0) == 1
     }
 
-    private fun DccV1.checkFields() {
-        // check for non null (Gson does not enforce it) + not blank & force date parsing
-        require(version.isNotBlank())
-        require(nameData.familyNameStandardized.isNotBlank())
-        dateOfBirthFormatted
```
vaubaehn, line 105: Aren't these required to match certificates to persons? What will happen when a certificate is added, where one or both fields are empty? Will just a new container be created with empty DoB/fnt?

mtwalli: @vaubaehn  this was an additional check in Android side. we will rely only on json schema validation to achieve party with iOS where such cases are accepted

SamuraiKek: Honestly it's not clear to me. I would have left these checks in place as all certificates should contain at least this information. I'm putting this PR to draft until some of my comments in the Jira ticket are clarified and I'll also raise this concern.

vaubaehn: @mtwalli ok, thanks. So, I understand that `null ` fields for `dob` and `fnt` won't break creating/matching persons in general. 👍

vaubaehn (dismissed):
One question - was answered

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
mtwalli: @vaubaehn 
I think you are referring to the person identifier where `name` and `dob` are used to identify the person right?
Here are my thoughts about that:
1- Allowing missing or empty fields does not meant that all of them will be missing the same time , and it should not happen for real certificates, otherwise how someone could prove that the paper certificate belongs to them
2- In case such cases exist in reality and they are valid certificates , I don't think it is a good idea to allow this in iOS but in Android. Android user should be scan them too
3- Worst case, if all fields are missing then those certificates will fall under same person card , but that would be the case also when paper version does not have `name` or `dob` 
4- Point 3 can still happen in some situations too without allowing empty or missing fields for example if more than one person have the same last name , but no dob or first name .

Additionally I will quote what Max wrote about the json schema validation:

> Preliminary assessment: I've double-checked the JSON schema spec (what does required mean) and several implementations. It confirms my initial assumption that if a field is marked as required in the JSON schema, the attribute must be present. However, a required field does not make any implications about the value of the attribute (e.g. whether null or empty string are ok or now). This is something that has to be defined on the attribute specification (which makes sense, as required fields could be a string, an array, etc, for which there cannot be a common implication for what required means other than the attribute itself being present).

Regarding testing: this will definitely be tested. the issue here was reported by a tester that Android is being more restrict than iOS, I hope that helps , if you have more concerns please let us know?

AlexanderAlferov (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
vaubaehn: Short: everything is fine (after I had an even deeper look into DCC schema 1.3.0).

Long:
@mtwalli
> I think you are referring to the person identifier where `name` and `dob` are used to identify the person right?

Yes, I was referring to the person identifier, that also uses `dob`. But afaik, `CertificatePersonIdentifier` (and related) does not use `name` but `lastNameStandardized` and `firstNameStandardized` (see https://github.com/corona-warn-app/cwa-app-android/issues/3547#issuecomment-879208984). In case the standardized names are directly retrieved from the DCC, there is a higher chance of empty fields (just look to our issues regarding quality check for 3rd-party-provider implementations due to insufficient implementations/processes, e.g. by TSI or test providers in DE). And there is also a low chance that `dob` can be empty, as some countries started reliable birth registries for their citizens quite late in the 20th century. [DCC schema 1.3.0 explicitly allows empty](https://github.com/ehn-dcc-development/ehn-dcc-schema/blob/a603410d760fefc9073931c8c807759d9714c136/DCC.combined-schema.json#L48-L59) `dob`.

So my concern/question was a more technical one, how will CWA behave technically, when a DCC that shall be added comes with empty `lastNameStandardized` and `firstNameStandardized` as well as empty `dob`? Will the app crash when trying to create a person DCC container, or will that case be handled, i.e., deriving `lastNameStandardized` and `firstNameStandardized` from fields ... But I just realize, that field `fnt` (standardized family name) [is required according to schema 1.3.0](https://github.com/ehn-dcc-development/ehn-dcc-schema/blob/a603410d760fefc9073931c8c807759d9714c136/DCC.combined-schema.json#L99-L103), all other fields could be empty actually. So,
> 1- Allowing missing or empty fields does not mean that all of them will be missing the same time , and it should not happen for real certificates

is true, as `fnt` always must contain "something", even if it's only one letter... (https://github.com/corona-warn-app/cwa-app-android/issues/3670)

> 2- In case, such cases exist in reality and they are valid certificates , I don't think it is a good idea to allow this in iOS but in Android. Android users should be able to scan them too

With all said above, this _should_ not happen, and they won't be valid. However, it depends on implementations on 3rd party side. Afaik, in DE, IBM is not checking for validity in their signing process, and TSI is also not checking validity of contents (yet) before/during provider is online/in PROD.

> 3- Worst case, if all fields are missing then those certificates will fall under same person card

**And that was exactly the point, where I wondered how this person card could look like, when _all_ necessary fields for the `CertificatePersonIdentifier` were empty... Was it possible to create a card/container at all, or would CWA crash?**

> 4- Point 3 can still happen in some situations too without allowing empty or missing fields for example if more than one person have the same last name , but no dob or first name .

Yes, but then CWA would operate as expected and match the DCC accordingly (without crashing). It's just matching 'wrong' but the cause would be the quality of DCC contents.

Reading Max preleminary assessment:
> However, a required field does not make any implications about the value of the attribute (e.g. whether null or empty string are ok or now). This is something that has to be defined on the attribute specification (which makes sense, as required fields could be a string, an array, etc, for which there cannot be a common implication for what required means other than the attribute itself being present)

I have understood 'required' in a different way: that a value of a required field must not be `null`/empty, as it provides basic data to assess whether it is a valid certificate (e.g., if date of vaccination is missing, you can't judge at what point in time 'immunization' may be assumed).

And finally,
> @vaubaehn this was an additional check in Android side. we will rely only on json schema validation to achieve party with iOS where such cases are accepted

as JSON schema validation in this case means, that only certificates can be added that at least have a value in their `fnt` field, then I think there is no problem for CWA when creating/matching person, as at least one `val `for the `CertificatePersonIdentifier` should be valid.

> Regarding testing: this will definitely be tested

Great, as I am not trusting the quality of issued DCCs 100% 😉 . So, empty `fnt` should be rejected at first.

mtwalli (approved):

AlexanderAlferov (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/DccV1Parser.kt
vaubaehn: @mtwalli Thanks for all your responses and explanations!
What I didn't take into account before, was, that internal representation for a person where all fields are `null` is handled in `code`, which in any case would have a valid value (e.g., `'##'`). Thanks again and have a nice week!

PR #3976, opened 2021-08-24
Source app config locally when initializing `AppConfigProvider`

d4rken:
Source app config locally when initializing `AppConfigProvider` to prevent app components from hanging for a longer duration when network conditions are bad.

This addresses the certificates not being shown in adverse network conditions as reported by @FloSchwalm [here](https://github.com/corona-warn-app/cwa-app-android/issues/3549#issuecomment-903662403).

Since 2.7.x, `DccStateChecker`  is part of the loading certificate data and it requires config data. If the local config is outdated, and the network connection is bad retrieving app-config can take up to 20 seconds. If the app was freshly started and this is done as part of the `AppConfigProvider` initialization, then no other app data is available and all consumers of `AppConfigProvider.currentConfig` will hang.

See here for a more detailed explanation: https://github.com/corona-warn-app/cwa-app-android/issues/3549#issuecomment-903962421

To reproduce this firewall the CWA and drop the packets, or add a delay to the HttpModule, e.g.

Screenshot from 2021-08-24 21-08-35

While this overall changes app-config behavior to not perform refresh check on every cold-start, this was never intended AFAIK and rather a side-effect introduced through refactoring over the time. Any component that requires the absolut latest config is already using `getAppConfig` to force a refresh AFAICT.


BMItr (approved):
data-good

tested on OnePlus 8 without any issues.
thanks for you commitment 👌 

And my best Greetings!

mtwalli (approved):
LGTM

PR #3977, opened 2021-08-24
Proposal: Perform update check against `currentConfig`

d4rken:
tldr: Make the "time to pull CWA out of pocket and show someone the certificate"-path quicker.

@FloSchwalm has highlighted an important use-case [here](https://github.com/corona-warn-app/cwa-app-android/issues/3549#issuecomment-904081048):
Currently the user may look at the CWA splash screen for up to 5 seconds until they can use the app which may put the user in an awkward position, if they just pulled out the phone of their pocket to show their certificate to someone for verification and other people are waiting behind them in a hurry.

I propose to perform the update-check against the `currentConfig` (latest cached one) and to not force a new network check.

Upside:
* We make app launch up to 5 seconds faster in bad network conditions in a now very common use-case: Use the CWA to present the certificate to someone.

Downside:
* The app may be launched an additional time if an update is required
    * Counter: We would have let the user launch the app anyways if they had no internet at all.  
    * Counter: We force config refreshes in various other locations already, e.g. during the hourly pkg download, so the chance that this config update would yield a different result than the previous one is low.

I think the upside outweighs the downside here.

Do not merge this without #3976 as it would then actually prolong the delay due to the removal of the separate lower timeout.


mtwalli (approved):
Aligned with Thomas, We will give it a try. Thanks for your support!

BMItr (approved):

PR #3980, opened 2021-08-25
Add fake header (DEV)

mtwalli:
- This fixes the issue that submission is not working against `INT`

### Testing 
1- Ask me or David for a TAN 
2- Submit organizer Checkins to `INT`


AlexanderAlferov (approved):
Works great! Second submission with the same TAN leads to the expected error.
LGTM 🚀

BMItr (approved):
👌 
image

PR #3999, opened 2021-08-30
Booster notification api service (EXPOSUREAPP-9197)

BMItr:
This PR addresses Exposureapp-9197.

Draft because, some discussions regarding missing rule set.

Todo:

- [x] add relevant tests


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/core/DccValidationRepository.kt
```diff
@@ -107,11 +119,34 @@ class DccValidationRepository @Inject constructor(
             DccValidationData(
                 countries = newCountryData,
                 acceptanceRules = newAcceptanceData,
-                invalidationRules = newInvalidationData
+                invalidationRules = newInvalidationData,
+                boosterNotificationRules = this.boosterNotificationRules
```
mtwalli, line 123: Can we move it into it own repository? `Booster rules` and `Validation rules` are used in two different use cases, I would keep them independently to avoid any delay to load all kind of rules using the same repository. what do you think?

BMItr: yes we can! make sense

mtwalli (approved):
Thanks 🙏

AlexanderAlferov (approved):
Nice test coverage! 🚀

PR #4001, opened 2021-08-30
Apply booster rules (EXPOSUREAPP-9201)

mtwalli:
- Implement Booster rules appliance
- Inject CertLogic using Dagger 
- Cleaned up scheduler + worker to refresh Vaccinations which is not doing actually anything

### TODO 
- [x] Add more unit tests


vaubaehn:
@mtwalli 
Short question if I got it right: In `DccBoosterRulesValidator` you're creating a kind of 'artificial' (because it's not DCC schema compliant) JSON payload with combined vacc & rec cert, and then you check it against specially crafted 'internal' business rules (as no business rules for certlogic exist yet)?

mtwalli:
> @mtwalli
> Short question if I got it right: In `DccBoosterRulesValidator` you're creating a kind of 'artificial' (because it's not DCC schema compliant) JSON payload with combined vacc & rec cert, and then you check it against specially crafted 'internal' business rules (as no booster business rules for certlogic exist yet)?

@vaubaehn 
1- the `payload` is a DGC-like JSON data structure that has `ver`, `nam`, `v[0]` and may have `r[0]`.
As you can see in this test sample https://github.com/corona-warn-app/cwa-app-android/pull/4001/files#diff-4dff8e08943552d5a0b07a73e4957adca4e395a2feb486dd3eae05a2ff990f2bR82-R114

2- The rules are not internal, Server will provide them as other certificates validation rules , you can see more details in this PR https://github.com/corona-warn-app/cwa-app-android/pull/3999

vaubaehn:
@mtwalli 
That's brilliant!
From your response I see, I got it right.
When I wrote 'internal business rule' I meant a construction that is not (yet) officially supported through the IOP gateway. From your reference I could see, it's a Germany only solution for now.

A bit OT and not part of this PR: the approach that you are using, could/should be implemented to other wallet apps in EU countries, imho. To my best knowledge, IOP gateway could later support same booster rules without necessary technical adjustment to Certlogic.
Even verifier apps could implement it, but with downside on UX due to higher effort for users/gate keepers, as they would need to present/scan 2 QRs serially, and gate keepers would need to select to scan 2 codes in advance...

mtwalli:
> Is there a reason why you only test 1/2 doses?
> I guess 1/1 for vacc after recovery could be a common use case?

- We validate against most recent vaccination certificate, even if there are more than one, the recent one is the one to check. - Same apply for recovery certificates, but recovery certificate is optional. In other words it is enough that user has at least one Vaccination certificate registered to check booster rules.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/DigitalCovidCertificateModule.kt
```diff
@@ -1,23 +1,39 @@
...
         TestCertificateServerModule::class,
-        RecoveryCertificateModule::class,
         DccValidationModule::class,
-        DigitalSignatureCertificateModule::class,
         DscServerModule::class,
     ]
 )
-abstract class DigitalCovidCertificateModule
+
+class DigitalCovidCertificateModule {
```
BMItr, line 26: This cant be @Reusable, can't it?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/DccBoosterRulesValidator.kt
```diff
@@ -1,16 +1,136 @@
...
+    @BaseJackson private val objectMapper: ObjectMapper,
 ) {
-    suspend fun validateBoosterRules(dccList: List<DccData<*>>): EvaluatedDccRule? {
-        // TODO
-        throw NotImplementedError()
+    @Suppress("BlockingMethodInNonBlockingContext")
+    suspend fun validateBoosterRules(dccList: List<CwaCovidCertificate>): EvaluatedDccRule? {
+        Timber.tag(TAG).d("validateBoosterRules(dccList=%s)", dccList)
+
+        val boosterRules = boosterRulesRepository.rules.first()
```
BMItr, line 39: Use the method _updateBoosterNotificationRules_, once my pr ist merged. That will update the rules and return the most up-to-date rules

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/DccBoosterRulesValidator.kt
```diff
@@ -1,16 +1,136 @@
...
+import dgca.verifier.app.engine.data.CertificateType
+import dgca.verifier.app.engine.data.ExternalParameter
+import kotlinx.coroutines.flow.first
+import timber.log.Timber
+import java.time.ZonedDateTime
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class DccBoosterRulesValidator @Inject constructor(
```
BMItr, line 30: ⇓ According to TechSpec this should only run once a day ⇓

BMItr, line 30: ⇓ According to TechSpec this should only run once a day ⇓

mtwalli: this is part of the next pr

BMItr (approved):
👌 lvgtm!

PR #4002, opened 2021-08-31
Remove term "Auffrischimpfung" from booster certificate (EXPOSUREAPP-9224)

SamuraiKek:
Booster certificates will be treated and displayed as normal certificates. The term "Auffrischimpfung" is removed from the vaccination cards.

How to test: 
Preferably in this order:

`cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=2 v.0.dt='2021-08-24' -e wru`
`cwa vc gen --pii 42 -s v.0.dn=2 v.0.sd=2 v.0.dt='2021-08-25' -e wru`
`cwa vc gen --pii 42 -s v.0.dn=3 v.0.sd=3 v.0.dt='2021-08-26' -e wru`


NataliaLemmerth (dismissed):
UA approved

mtwalli (commented):
I think we need to change the priorities , according to 
AC 
>The exception, that booster vaccinations are not prioritised over the vaccination 2/2 or 1/1 is removed. Instead the latest vaccination has priority over all other vaccinations

jurajkusnier (approved):
LGTM

SamuraiKek:
> I think we need to change the priorities

I have pushed a change but please test with other types of certificates too and make sure the priority didn't get messed up with them.

mtwalli (approved):
LGTM

PR #4003, opened 2021-08-31
Updated priority rules (EXPOSUREAPP-9244)

SamuraiKek:
Small update to the priority rules that should have been part of #4002

Testing:
- scan every certificate from here: https://dgc.a-sit.at/ehn/
- in the end highest prio cert should be the recovery one.


AlexanderAlferov:
LGTM 🚀 
Screenshot 2021-08-31 at 17 01 13

AlexanderAlferov (approved):

mtwalli (approved):
Works, tested it with RC, 3-VC, TC

PR #4028, opened 2021-09-07
Fix behaviour after certificate scan (EXPOSUREAPP-9221)

mtwalli:
Fixes: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9221

**The issue:**
- Scan many certificates as many as you want about `6 `or more 
- At some point user will land on PersonOverviewFragment ,not the relevant certificate details screen.
- This is happening because in person details screen , `PersonCertificatesProvider.flow` is not providing the certificates immediately. When the certificates list gets longer , it gets slower and when looking for the person is not guaranteed to find it the latest set of persons provided by the flow and therefore navigating back is triggered.

Solution:
- Navigate directly to the respective details screen, because it is guaranteed to find the certificate itself 
  Scan screen does not navigate to the details unless the respective certificate repository already up to date.
- And by doing that we align navigation with iOS also


AlexanderAlferov (dismissed):
After scanning 15 certificates: navigation works great. 🚀

BMItr (approved):
lgtm! :ok_hand:

AlexanderAlferov (approved):

PR #4031, opened 2021-09-07
Test cert card status (EXPOSUREAPP-8718)

AlexanderAlferov:
Status for new test certificate and fix for red badge on details screen.
Screenshot 2021-09-07 at 15 31 05

How to test:
1) scan PCR test with DGC
2) wait some seconds and approve DGC
3) go to certificates tab
4) open newly added person 
5) check that test certificate card has "Neu" label
6) go to details
7) check that red dot is not visible if there is no "Test certificate" label right below QR code.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
@@ -27,36 +27,38 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
     invalidOverlay.isGone = valid
     image.isEnabled = isCertificateDetails && valid // Disable Qr-Code image from opening full-screen mode
 
+    val isNewTestCertificate = certificate is TestCertificate && certificate.isNewlyRetrieved
+    notificationBadge.isVisible = if (isNewTestCertificate) {
+        false
+    } else {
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -106,6 +106,7 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
             qrTitle.isVisible = isPersonDetails
             qrSubtitle.isVisible = isPersonDetails
             startValidationCheckButton.isVisible = isPersonDetails
+            notificationBadge.isVisible = isPersonDetails
```
mtwalli, line 109: notificationBadge.isVisible = `isPersonDetails && certificate.isNewlyRetrieved` ? 
Currently it is always visible in this screen

NataliaLemmerth (approved):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
AlexanderAlferov: Yes, sure 😅

mtwalli (approved):

BMItr (approved):

PR #4048, opened 2021-09-10
Unify scanners (EXPOSUREAPP-9330)

mtwalli:
### Changes
- All QrCode types  are of type  a `QrCode`
- Add  CheckIn's `QrCode` `QrCodeExtractor`
- Unify `QrCodeVerifier` that holds a setOf `QrCodeExtractor`s  to find and extract QrCode based on type
- After extracting QrCode it is handed to its relevant `Handler` to apply the required steps afterwards 
- `Extractor` functions are now `suspended` to suit all cases -> many changes in unit tests to address that  
- New Scanner screen (Not wired yet)
- Importing PDF, Images is considered in the new Scanner ViewModel

### TODO in next tickets
- Wire Scanner to Results screens
- Error handling 

### Note 
- App should function as before , new scanner  is not used  yet

 ### Testing
- Test menu -> QR Code Image scanning -> Open Scanner 
- Try importing and scanning ,but not that some qrcodes require subsequent steps in order to display the right result 
- You can verify if result is recognised by the `Toast`


SamuraiKek (approved):
LGTM. Code looks fine, although I haven't dived deep into it. Testing on device works nicely. I scanned VC, RC, TC, PCR, RAT and Checkin QR's. Also opened jpg equivalents from the file explorer and managed to scan them.

jurajkusnier (approved):
LGTM

PR #4057, opened 2021-09-13
Certificate display switches to expired-view (if expired) after opening the certificate view (EXPOSUREAPP-8837)

BMItr:
This PR addresses Exposureapp-[8337](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8837).
"Certificate display switches to "expired-view" only after app re-start"

**About:**
The check whether a certificate has expired or not usually runs every 24hours.
This PR adds a second check which is triggered opening the certificate-overview.

**How-to-test:**
- scan a Certificate which will expire in some minutes (can be found in the QR-Example Folder "Special QRs" or CLI Tool-> vc gen --exp 1minute)
- opening the CertView (MainScreen->BottomBar) will show a list of your certificates. 
- if the certificate is expired it should show up already in the certificate overview screen

**Screenshot**
image


mtwalli:
@BMItr  Is it acceptable that we do multiple validity checks now while the App in foreground ? can you please align with Max?

mtwalli (dismissed):

AlexanderAlferov (approved):
LGTM 🚀

mtwalli (approved):

PR #4061, opened 2021-09-14
UQS update submission flow screen order (EXPOSUREAPP-9332)

SamuraiKek:
Order of the screens in the PCR/RAT submission flow is updated to match flow of UQS. 

The PR addresses the second dot of the acceptance criteria from [Jira](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8903) but there's no thorough description provided for every scenario. I did my best to keep a logical flow but please test thorougly because  I might have missed something.

Testing flow:
- Register a new positive test by scanning QR code.
- First screen that shows up after pressing "Scan QR Code" is the QR scan screen.
- After scanning the QR, you are taken to the "Your Consent" screen.
   - opening the data privacy page works.
   - pressing the x button takes you back to QR scanner.
   - pressing the OS back button takes you back to the QR scanner.
   - accepting the consent takes you to the "Covid Test Certificate" screen where you can request a test cert.
- On the "Covid Test Certificate" screen, after selecting either option you get to "Your result is Available" screen.
- Pressing next takes you to the result screen.
 
- Register another test of same type.
- Flow is the same as the one above but after the consent screen, you first go to the deletion screen to delete the already registered test.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentViewModel.kt
```diff
@@ -14,13 +15,14 @@ import de.rki.coronawarnapp.ui.submission.viewmodel.SubmissionNavigationEvents
...
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
-import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
+import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactory
 import kotlinx.coroutines.flow.first
 import timber.log.Timber
 
 class SubmissionConsentViewModel @AssistedInject constructor(
     interoperabilityRepository: InteroperabilityRepository,
     dispatcherProvider: DispatcherProvider,
+    @Assisted private val qrCode: String,
```
mtwalli, line 25: Should not this be `CoronaTestQRCode` ? which is already parsed by the scanner

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
```diff
@@ -1,76 +1,41 @@
...
-                        allowReplacement = false
-                    )
-                } else {
-                    events.postValue(
-                        SubmissionNavigationEvents.NavigateToRequestDccFragment(ctQrCode, isConsentGiven)
-                    )
-                }
-            }
+            qrCodeValidator.validate(rawResult)
+            events.postValue(SubmissionNavigationEvents.NavigateToConsent(qrCode = rawResult))
```
mtwalli, line 27: this is scan fragment ,the test  registration should not be done here at all we should register after the user has given the consent and OS consent too ,  right ?

So I assume that registration should be done after the OS consent.
and we check for:
1- Duplicate test 
2- Request Dcc
3- Register and navigate to the right Test result screen

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
```diff
@@ -20,33 +20,28 @@ import de.rki.coronawarnapp.util.ui.observe2
...
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
         binding.viewModel = viewModel
-        navArgs.qrCode?.let {
-            viewModel.qrCode = it
-        }
-        binding.submissionConsentHeader.headerButtonBack.buttonIcon.setOnClickListener {
-            viewModel.onBackButtonClick()
-        }
+        binding.submissionConsentHeader.setOnClickListener { popBackStack() }
```
mtwalli, line 42: the whole is header is navigating back?

mtwalli:
We need to talk ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentViewModel.kt
SamuraiKek: Not necessarily. We can use the `rawString` from the scan result. And the consent fragment also had this `qrCode` attribute that was being passed from `MainActivity`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
SamuraiKek: Yes. That's why I've removed all registration related logic from the scanner.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentFragment.kt
SamuraiKek: Ah my bad. I changed the toolbar in the layout and put it to the whole header by mistake.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/scan/SubmissionQRCodeScanViewModel.kt
mtwalli: but where did you add it ? which screen now is registering the test ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/qrcode/consent/SubmissionConsentViewModel.kt
mtwalli: this is what get from the scanner , why you need to validate it again ?

mtwalli (approved):

AlexanderAlferov (approved):

PR #4062, opened 2021-09-14
Expose corrected json payload (EXPOSUREAPP-9435)

mtwalli:
- Issue parser was still returning the original json payload , not the corrected one
- Scan the certificate and run the booster rules from test menu before and after the fix 
- Before -> user is not eligible for booster 
- After -> user is eligible for booster 
## Fixes 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9435


jurajkusnier (approved):

SamuraiKek (approved):
LGTM.

PR #4084, opened 2021-09-17
No Notification to Warn Others after Positive PCR (EXPOSUREAPP-9122)

jurajkusnier:
## Testing
- scan positive PCR or RAT test, don't share the results, and wait for cca 2h for notification
You could also update `NotificationConstants` to get the notification earlier
```kotlin
val POSITIVE_RESULT_NOTIFICATION_INITIAL_OFFSET: Duration = Duration.standardHours(2)
val POSITIVE_RESULT_NOTIFICATION_INTERVAL: Duration = Duration.standardHours(2)
```
- then you could also scan another positive PCR or RAT test and check if the notifications are still working
- It looks like we save only a number of reminding positive test result reminders, and this count is reset when the test is removed, but not when the test is replaced by another one, so I have added also a test id 

## Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9122


mtwalli (approved):
LGTM

SamuraiKek (approved):
LGTM.

PR #4086, opened 2021-09-17
Update external qr scan flow (EXPOSUREAPP-9463)

SamuraiKek:
Scanning a RAT QR with an external scanner will bring up the deletion screen in case there's an already registered RAT test inside the app or the consent screen in case there isn't.  

Test the flow with & without any RATs saved in the app.


Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -474,6 +478,12 @@
             app:argType="boolean" />
     </fragment>
 
+    <action
+        android:id="@+id/action_submissionConsentFragment"
+        app:destination="@id/submissionConsentFragment"
+        app:popUpTo="@id/universalScanner"
+        app:popUpToInclusive="true" />
```
mtwalli, line 485: Universal scanner is not opened  in this case , does it ? should not that be  the `mainFragment`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -108,6 +111,20 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
             binding.mainBottomNavigation.updateCountBadge(R.id.covid_certificates_graph, count)
         }
 
+        vm.externalLinkEvents.observe(this) {
+            when (it) {
+                is HomeFragmentEvents.GoToCheckInsFragment -> navController.navigate(
```
mtwalli, line 116: I would rename/create new set of Events for `MainActivity` in this case it is not part of `HomeFragment` itself 
Maybe `DeepLinkDirections` ? what do you think ?

SamuraiKek: Yes, I wasn't sure wether to go this route or just throw them into an already existing Directions class. I will create a specific one just for them.

Corona-Warn-App/src/main/res/navigation/nav_graph.xml
SamuraiKek: Yes, you're right. Didn't catch that.

AlexanderAlferov (approved):
LGTM 🚀

mtwalli (approved):
LGTM

PR #4107, opened 2021-09-22
Improve QR code importing capabilities and prevent OOM crash

jurajkusnier:
## Testing
- scan some large images or pdf with QR codes
- app shouldn't crash
- you can use resources from https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8353 or your own


Corona-Warn-App/build.gradle
```diff
@@ -367,7 +367,7 @@ dependencies {
     // QR
     implementation('com.journeyapps:zxing-android-embedded:4.1.0') { transitive = false }
     // noinspection GradleDependency - needed for SDK 23 compatibility, in combination with com.journeyapps:zxing-android-embedded:4.1.0
-    implementation 'com.google.zxing:core:3.3.0'
+    implementation 'com.google.zxing:core:3.4.1'
```
mtwalli, line 370: Please revert ? newer versions has compatibility issues with Android 6

jurajkusnier: updated to `3.3.3`

BMItr:
what is in the case of low memory?

mtwalli (commented):
it does not crash, can we scale the image down? I tried with a Valid certificate qrode file size 5.7 MB ,but I got an error that file can not be read

[20210922_160020](https://user-images.githubusercontent.com/25054729/134361512-8e1c0c32-ca31-47bd-b0ae-2880d254aed7.jpg)

BMItr (approved):

mtwalli (approved):

PR #4109, opened 2021-09-22
Register DCC after info screen confirmation (EXPOSUREAPP-9562)

mtwalli:
Addresses: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9562

### Testing 
- If user is not onboarded the registration should happen after clicking `Continue`
- otherwise should happen on Scanner


SamuraiKek (dismissed):
LGTM. Tested on Samsung S10+

AlexanderAlferov (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerViewModel.kt
```diff
@@ -78,11 +78,16 @@ class QrCodeScannerViewModel @AssistedInject constructor(
         cameraSettings.isCameraDeniedPermanently.update { denied }
     }
 
-    private suspend fun onDccQrCode(qrCode: DccQrCode) {
+    private suspend fun onDccQrCode(qrCode: DccQrCode, rawQrCode: String) {
```
SamuraiKek, line 81: What are you using `rawQrCode` for?

SamuraiKek (dismissed):

AlexanderAlferov (approved):

SamuraiKek (approved):

PR #4123, opened 2021-09-26
Latest issued Vaccination DCC 2/2 does not have highest priority (EXPOSUREAPP-8756)

### LINKED ISSUE #3838: Latest issued Vaccination DCC 2/2 does not have highest priority
## Avoid duplicates
* [ ] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation) - I'm not sure how sorting is done in iOS. If the issue is similar, this issue could be transferred to documention repo.

## Technical details

- Device name: LG G4 (LGH-815)
- Android version: Android 6.0
- App version: 2.6.1

## Describe the bug

Most users will only import 1 or 2 vaccination certificates into CWA, with either dose 2/2, or first dose 1/2 or dose 2/2.
But there are some use cases, when users are importing additionally a re-issued VaccDCC (2/2) as a 3rd certificate at a later point in time. Either just because they can do it 🤷‍♂️ , or they need to: when the Netherlands decided to apply business rules schema v1.3.0 to their validation checks,  but German VaccDCCs were issued compliant to schema v1.0.0 before July 1st, and users want to be on safe side in border control and get there 2nd VacDCC re-issued (online, pharmacy). See https://github.com/corona-warn-app/cwa-documentation/issues/671 .
The 3rd VaccDCC (2/2) with a newer date of issue and later date of expiry can be imported into CWA without problems, but at least in my case, the 3rd VaccDCC is not taken into account as certificate with highest priority, but sorted as the 2nd VaccDCC in the list of all DCCs. This may lead to confusion upon presenting QR code to gate keepers, even more as there is currently no date of issue/expiry date displayed (you need to check certID by comparing with paper printouts/PDFs to find out...)

1. Import a vaccination DCC (2/2) that was issued before July 1st, 2021.
2. Import another vaccination DCC (2/2) that was issued somewhat after July 1st, 2021.
3. Check validity of both certificates against Netherland's business rules.
4. Find that the one with highest priority (automatically activated) in the list fails, while the second one passes - at least in my case. So, the first one is the old one, and the second one the new one.

## Expected behaviour

The (vacc)DCC (2/2) with latest date of issue/the longest expiry date should be the vaccDCC with the highest priority.

## Possible Fix

Apply another rule 3a that accounts for expiry date here: https://github.com/corona-warn-app/cwa-app-android/pull/3811

## Additional context

I don't know if this would also be necessary for recovery certificates...

### LINKED ISSUE #4000: Trouble with “Actually used Certificate” if more than one vaccination certificate is available…
## Avoid duplicates
* [ ] ?: Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)

## Technical details

- Device name:  Samsung Galaxy Note 10, SM-N970F/DS
- Android version:  Android 11 (One UI 3.1) with Android security update: 01. August 2021
- App version:  2.8.0

## Describe the bug

Problems after selecting the "Currently used Certificate" („Aktuell verwendetes Zertifikat“) because then another vaccination certificate is the current one...
If more than one vaccination certificate for a person has been scanned into the CWA, the following selection results, for example:

Screenshot_20210830-120444_Corona-Warn_2

If you now tap on the "Currently used Certificate" („Aktuell verwendetes Zertifikat“) you can see the details of this certificate; including the URN. For example, my URN ends with "5".
After tapping on the arrow at the top left [<-] you get to the previous screen (see picture above).

If you look closely, you can see that after tapping on [<-] the positions of the two 2/2 certificates are swapped. 
If you now tap on the "Currently used Certificate" („Aktuell verwendetes Zertifikat“) again and look at its details, the position swap of the two certificates is confirmed. In my example, the certificate with the URN "5" is no longer displayed, but my certificate with the URN "Q" at the end.

1. Start CWA.
2. Select "Certificates".
3. Select a person with multiple vaccination certificates.
4. Scroll down the screen; see screenshot above.
5. Select the "Currently used Certificate" („Aktuell verwendetes Zertifikat“) and pay attention e.g. to the URN under certificate identifier ("Unique Certificate Identifier").
6. Tap on [<-] (top left) - the two vaccination certificates 2 of 2 are visibly reversed in the order.
7. Select the "Currently used Certificate" („Aktuell verwendetes Zertifikat“) again and verfify the URN under Certificate ID ("Unique Certificate Identifier"). It is now the other certificate.

8. Steps 6 and 7 can now be repeated as required; the 'other' certificate is displayed in each case. 

Possibly incomplete Example:
Screenshot_20210830-120444_Corona-Warn_3

Notes: 
- If in steps 5 and 7 you do **not** select the "Currently used Certificate" („Aktuell verwendetes Zertifikat“) but rather the other one (_'Currently NOT used Certificate'_ 2/2), the order of the certificates is **not** changed; i.e. the same certificate is always displayed here. Depending on the initial situation, in my case either the one with the URN ending "5" or "Q".
 
- All of the above steps and results can be transferred 1:1 to certificates 1/2. When selecting the upper 1/2 certificate and taping on [<-] (top left), it is also visibly exchanged with the other (here second) 1/2 certificate in the sequence shown. 
When you tap on the lower 1 of 2 certificate (also several times) - as with the _'Currently NOT used Certificate'_ above - the order of the 1/2-certificates is not changed either, which means that always the same certificate is displayed. 

- Changing the order of certificates 2/2 does not affect the order of certificates 1/2 and vice versa.

(Hint: This means that the affiliations of the certificates (2/2 & 1/2) with regard to the displayed order are no longer given.)
 

## Expected behaviour

Is that so useful / desired? I don't think so.
This can become problematic and confusing if the 'country-specific checks of the vaccination certificates' are different for several existing valid certificates; e.g. because of different DCC schema versions or other.
(See "Additional Context" below.)

## Possible Fix

I suspect it would make sense to be able to set a certificate as the standard certificate, which is then also used for the 'country-specific check of the vaccination certificate' - until the standard certificate is explicitly changed, right? 
Of course, this assumes that a standard certificate can be set manually.

## Additional context

As a result for CWA v 2.8.0, the workaround for sorting the vaccination certificates no longer works; see: https://github.com/corona-warn-app/cwa-documentation/issues/679#issuecomment-896324471.

Cross links: 
- [Latest issued Vaccination DCC 2/2 does not have highest priority](https://github.com/corona-warn-app/cwa-app-android/issues/3838#issue-955824709).
- [Check for different schema versions for certificates stored for the same person](https://github.com/corona-warn-app/cwa-wishlist/issues/611#issue-966565025).

--------------

PS: I don't think we need to have another discussion here about whether or not it makes sense to have multiple valid certificates for one person.

PS2: A revision of the certificate handling is likely to be necessary if 3rd / 4th / ... vaccinations or their certificates have to be processed in the CWA in the future.  

Best regards, Joachim.

Internal Tracking ID: [EXPOSUREAPP-8756](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8756)
[Android] Latest issued Vaccination DCC 2/2 does not have highest priority

mtwalli:
- Find latest Vaccination certificate by `vaccinatedOn` and `issuedAt` dates 

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8756

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
@@ -140,7 +140,12 @@ private fun Collection<CwaCovidCertificate>.rule5findTooRecentFinalShot(
  */
 private fun Collection<CwaCovidCertificate>.rule6findOtherVaccinations(): CwaCovidCertificate? = this
     .filterIsInstance<VaccinationCertificate>()
-    .maxByOrNull { it.rawCertificate.vaccination.vaccinatedOn }
+    .maxWithOrNull(
+        compareBy(
+            { it.rawCertificate.vaccination.vaccinatedOn },
+            { it.headerIssuedAt }
+        )
+    )
 
 /**
  * 7
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
@@ -81,7 +81,12 @@ private fun Collection<CwaCovidCertificate>.rule3FindRecentLastShot(
     .filter {
         Days.daysBetween(it.rawCertificate.vaccination.vaccinatedOn, nowUtc.toLocalDateUtc()).days > 14
     }
-    .maxByOrNull { it.rawCertificate.vaccination.vaccinatedOn }
+    .maxWithOrNull(
+        compareBy(
+            { it.rawCertificate.vaccination.vaccinatedOn },
+            { it.headerIssuedAt }
+        )
+    )
 
 /**
  * 4
@@ -119,7 +124,12 @@ private fun Collection<CwaCovidCertificate>.rule5findTooRecentFinalShot(
     .filter {
         Days.daysBetween(it.rawCertificate.vaccination.vaccinatedOn, nowUtc.toLocalDateUtc()).days <= 14
     }
-    .maxByOrNull { it.rawCertificate.vaccination.vaccinatedOn }
+    .maxWithOrNull(
+        compareBy(
+            { it.rawCertificate.vaccination.vaccinatedOn },
+            { it.headerIssuedAt }
+        )
+    )
 
 /**
  * 6
--- a/Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensionsTest.kt
+++ b/Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensionsTest.kt
@@ -48,6 +48,7 @@ class PersonCertificatesExtensionsTest : BaseTest() {
         wrongOrder2 shouldNotBe expectedOrder
     }
 
+    @Suppress("LongMethod", "ComplexMethod")
     @Test
     fun `find Highest Priority Certificate`() {
         val first = mockk<TestCertificate>().apply {
@@ -75,6 +76,17 @@ class PersonCertificatesExtensionsTest : BaseTest() {
         }
 
         val third = mockk<VaccinationCertificate>().apply {
+            every { headerIssuedAt } returns Instant.parse("2021-06-24T14:00:00.000Z")
+            every { rawCertificate.vaccination.doseNumber } returns 2
+            every { rawCertificate.vaccination.totalSeriesOfDoses } returns 2
+            every {
+                rawCertificate.vaccination.vaccinatedOn
+            } returns time.minus(Duration.standardDays(15)).toLocalDateUtc()
+            every { getState() } returns mockk<State.Valid>()
+        }
+
+        val thirdIssuedAtAnotherDate = mockk<VaccinationCertificate>().apply {
+            every { headerIssuedAt } returns Instant.parse("2021-07-20T10:00:00.000Z")
             every { rawCertificate.vaccination.doseNumber } returns 2
             every { rawCertificate.vaccination.totalSeriesOfDoses } returns 2
             every {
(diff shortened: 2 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
```diff
@@ -81,7 +81,12 @@ private fun Collection<CwaCovidCertificate>.rule3FindRecentLastShot(
     .filter {
         Days.daysBetween(it.rawCertificate.vaccination.vaccinatedOn, nowUtc.toLocalDateUtc()).days > 14
     }
-    .maxByOrNull { it.rawCertificate.vaccination.vaccinatedOn }
+    .maxWithOrNull(
+        compareBy(
+            { it.rawCertificate.vaccination.vaccinatedOn },
+            { it.headerIssuedAt }
```
mtwalli, line 87: @mlenkeit  FYI ?

MikeMcC399:
@mtwalli 

I've verified the fix on Android 8 with an older and a new 2/2 vaccination certificate.

This looks good! 👍🏻 

** Steps to verify

1. Clean install CWA deviceForTestersDebug 2.11.0-RC2
2. Complete onboarding
3. Tap "Certificates" tab, tap "CONTINUE"
4. Tap "+ CERTIFICATE" and scan an older complete EU Digital COVID Vaccination Certificate e.g. Number in a series of vaccinations/doses 2/2
5. Scroll down to Zertifikatkennung / Unique Certificate Identifier and make a note of the last few characters of the identifier
6. Tap back arrow
7. Tap "+ CERTIFICATE" and scan an a re-issued complete EU Digital COVID Vaccination Certificate e.g. Number in a series of vaccinations/doses 2/2 for the same person and the same vaccination date as in step 4
8. Tap right arrow
9. Scroll down to "Currently used certificate" and tap on it
10. Note that the Unique Certificate Identifier is from the older certificate scanned in first. 🙁 

*** Now with the fix applied

1. Update to CWA from branch [fix/8756-order-by-issued-at](https://github.com/corona-warn-app/cwa-app-android/tree/fix/8756-order-by-issued-at)
2. Tap "Certificates" tab
3. Tap right arrow
4. Scroll down to "Currently used certificate" and tap on it
5. Note that the Unique Certificate Identifier is from the newer certificate scanned in second. 🚀 😄

SamuraiKek (approved):
LGTM.

jurajkusnier (approved):

PR #4128, opened 2021-09-27
Delete certificate pdf after exporting (EXPOSUREAPP-9556)

marcauberer:
Addresses [EXPOSUREAPP-9556](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9556).


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/poster/CertificatePosterFragment.kt
```diff
@@ -74,6 +74,13 @@ class CertificatePosterFragment : Fragment(R.layout.certificate_poster_fragment)
         }
     }
 
+    override fun onDestroyView() {
+        super.onDestroyView()
+        // Delete exported pdf file
+        Timber.i("CertificatePosterFragment closed. Deleting pdf export now.")
+        viewModel.sharingIntent.value?.file?.delete()
+    }
```
mtwalli, line 82: Delete can throw an exception, I would prefer if you can catch the exception just in case and delete on appScope 
so in ViewModel create 

```kotlin
    fun deleteFile() = launch(appScope) {
        try {
            sharingIntent.value?.file?.delete()
        } catch (e: Exception) {
            Timber.d(e, "deleteFile failed")
        }
    }
```

and call this method from `onDestroyView`


OR @jurajkusnier  can we use `temp`  folder I assume it should be deleted automatically

jurajkusnier (commented):
I think this implementation belongs to ViewModel itself. It could be called from `onCleared()` method.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/poster/CertificatePosterFragment.kt
jurajkusnier: I'm not sure whether it's really necessary to check `SecurityException` when deleting the internal file. But definitely, the implementation should be in ViewModel. It could be even called from ViewModel.onCleared() instead of from the Fragment.

marcauberer: Yes. I just committed the fix, that moves the code to the ViewModel.

marcauberer:
> I think this implementation belongs to ViewModel itself. It could be called from `onCleared()` method.

@jurajkusnier done 👍🏼

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/poster/CertificatePosterViewModel.kt
```diff
@@ -68,6 +68,21 @@ class CertificatePosterViewModel @AssistedInject constructor(
         }
     }
 
+    private fun deleteFile() = launch(context = dispatcher.IO) {
```
mtwalli, line 71: I think we still need to use the `appScope` ? launch is triggered from viewModel scope and onCleared  there is a high chance that it get cancelled 😞

mtwalli: launch(scope  = appScope, context = dispatcher.IO) look for `@AppScope` to see an example

mtwalli (dismissed):
Thanks

mtwalli (approved):

jurajkusnier (approved):
LGTM

PR #4129, opened 2021-09-27
Camera permission dialog (EXPOSUREAPP-9612)

mtwalli:
- Camera permission dialogs
Screenshot 2021-09-27 at 14 27 42

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9612


marcauberer (commented):
@mtwalli wouldn't it be better to check with `shouldShowRequestPermissionRationale()` if the user wants to see such a dialog, like it is recommended [here](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions)?

NataliaLemmerth (approved):
UA approved

mtwalli:
> @mtwalli wouldn't it be better to check with `shouldShowRequestPermissionRationale()` if the user wants to see such a dialog, like it is recommended [here](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions)?

`shouldShowRequestPermissionRationale` does not differentiate `when user can be asked again`  and `when permission is denied permanently`. `CameraPermissionProvider` was not introduced to workaround this , cause the dialog should be presented when user denied the permission permanently. Additionally `shouldShowRequestPermissionRationale()` is actually used:
https://github.com/corona-warn-app/cwa-app-android/blob/fa7780f1eafd587179f5b8d1e45d1647e731ff77/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt#L48-L59

marcauberer:
> > @mtwalli wouldn't it be better to check with `shouldShowRequestPermissionRationale()` if the user wants to see such a dialog, like it is recommended [here](https://developer.android.com/training/permissions/requesting#workflow_for_requesting_permissions)?
> 
> @marcauberer `shouldShowRequestPermissionRationale` does not differentiate `when user can be asked again` and `when permission is denied permanently`. `CameraPermissionProvider` was not introduced to workaround this , cause the dialog should be presented when user denied the permission permanently. Additionally `shouldShowRequestPermissionRationale()` is actually used:
> https://github.com/corona-warn-app/cwa-app-android/blob/fa7780f1eafd587179f5b8d1e45d1647e731ff77/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt#L48-L59

@mtwalli okay. Did not consider that. Then everything is fine 👍🏼

PR #4141, opened 2021-09-29
Update Notification Text for Invalid / Expired Certificates (EXPOSUREAPP-9620)

axelherbstreith:


marcauberer (approved):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/expiration/DccExpirationNotification.kt
```diff
@@ -31,7 +31,7 @@ class DccExpirationNotification @Inject constructor(
 ) {
     fun showNotification(containerId: CertificateContainerId): Boolean {
         Timber.d("showNotification(containerId=$containerId)")
-        showNotification(containerId, R.string.notification_body)
```
mtwalli, line 34: `notification_body` is still used? if not please delete it from `DE` and it shall be deleted by Translation tool from all

axelherbstreith: yes its still used in other places. According to figma this string should be displayed in different places like when receiving a test result

axelherbstreith: only the certificate notification should be updated

mtwalli (approved):

NataliaLemmerth (approved):
UA approved

PR #4143, opened 2021-09-29
Add content description to add qr code button (EXPOSUREAPP-9694)

marcauberer:
Addresses [EXPOSUREAPP-9694](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9694).

The "+ QR-Code" button now has the content description "Add QR code"

### DIFF
--- a/Corona-Warn-App/src/main/res/values-bg/event_registration_strings.xml
+++ b/Corona-Warn-App/src/main/res/values-bg/event_registration_strings.xml
@@ -164,8 +164,6 @@
     <string name="trace_location_organiser_list_title">"Моите QR кодове"</string>
     <!-- XBUT: Event organiser qr codes list: new qr code button  -->
     <string name="trace_location_organiser_list_new_button">"QR код"</string>
-    <!-- XBUT: Event organiser qr codes list: new qr code button description -->
-    <string name="trace_location_organiser_list_new_button_description"></string>
     <!-- XTXT: Event organiser qr codes list: new qr code button  -->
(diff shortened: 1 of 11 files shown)


Corona-Warn-App/src/main/res/values-bg/event_registration_strings.xml
```diff
@@ -164,6 +164,8 @@
     <string name="trace_location_organiser_list_title">"Моите QR кодове"</string>
     <!-- XBUT: Event organiser qr codes list: new qr code button  -->
     <string name="trace_location_organiser_list_new_button">"QR код"</string>
+    <!-- XBUT: Event organiser qr codes list: new qr code button description -->
+    <string name="trace_location_organiser_list_new_button_description"></string>
```
mtwalli, line 168: Please add DE/ default only ? Translation tool should add the rest

NataliaLemmerth (approved):
UA approved

Corona-Warn-App/src/main/res/values-bg/event_registration_strings.xml
mtwalli: @marcauberer  you need to keep key in the default `res/values/string.xml` otherwise it won't compile. Sorry I was no clear about that in my comment above. 

What we normally do:
1- Add new translation key to `res/values-de/strings.xml` (reference file for the. translation tool)
2- Add the same  translation key to `res/values/strings.xml` ,so it can build 

Other locales will be added by Translation delivery of the release

mtwalli (approved):

PR #4151, opened 2021-09-30
Improve CheckIn scanning using slow connection (EXPOSUREAPP-9732)

### LINKED ISSUE #4148: CWA unable to scan venue QR code on bad network connection
Take any venue QR code and select "Check-in" and "QR-Code scannen". Point the camera to the QR code. Do this with

a) a good Wifi internet connection
b) a good mobile internet connection (e.g. 4G)
c) any bad internet connection (e.g. limit the network standard in your device to 2G)
d) no internet connection at all

In cases a), b) and d), the CWA will recognize the QR code immediately. In case c), the CWA will scan forever with no result.

Expected: The CWA handles case c) the same way as case d), so the scan will return as fast as the QR code can be decoded.

(This situation occurs often inside large buildings. If scanning does not work, the easiest way to resolve the problem is to log on to the local Wifi network. Scanning will all of a sudden work with the same lighting and from all angles. I have seen many, many people having this problem - and the solution is simple, yet not obvious.)

My setup:
Samsung Galaxy A40
Android 11
Security patch level September 1st, 2021
CWA 2.10.1 ENF 18213314000 from Google Play

mtwalli:
Follow steps in the tickets to create a slow connection and then scan ChekIn QrCode
## Ticket 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9732


jurajkusnier (dismissed):
LGTM tested using android emulator

BMItr (approved):

jurajkusnier (approved):

MikeMcC399:
Link back to issue https://github.com/corona-warn-app/cwa-app-android/issues/4148.

PR #4152, opened 2021-09-30
Auto checkout fix (EXPOSUREAPP-8700)

AlexanderAlferov:
## Changes
Added non zero check for permanent locations, as it is not possible to create permanent location with zero duration.

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8700

## How to test
1) Create non permanent location with:
    a) zero duration
    b) start date differs from end date more than 15 minutes
    c) in future (so endDate is after Now()) 
2) Scan it
~~3. Check that auto checkout is bigger than 15 minutes (= end date - start date)~~
~~4. Use your time machine to teleport to future 15 minutes and check that checkout was not performed~~
3) **[Update]** According to new data from PO: duration should be the event duration (not larger than 23:45).


jurajkusnier (commented):
Tested according to the jira ticket

### Steps:

1. open the app (CWA)
1. select check-in in the tab-bar
1. scan the QR-Code of the event
1. check in

### Expected result:
- automatic check-out at the end of the event **didn't happen**

BMItr (approved):
lgtm 👌 

image
if no duration is provided, the end time of the event is taken.

jurajkusnier (approved):

PR #4166, opened 2021-10-04
Fix Toolbar is not visible in UQS screen (DEV)

mtwalli:
- Bring Toolbar to front , so it is not hidden by camera view


marcauberer (approved):
lgtm

AlexanderAlferov (approved):
LGTM 🚀

PR #4168, opened 2021-10-04
UQS Certificate Scan - "Neu" ist missing on Vacc und Rec Certificates (EXPOSUREAPP-9756) 

mtwalli:
- `isNew` flag is introduced to indicate that the certificate is added newly in the App whether it is TC, VC, or RC
- Renamed `isNewlyRetreived` of TC  to match the new way
- Now all certificates should have `Neu` label once they are newly scanned or requested from server for TC case

- [x] Update Instrumentation tests

## Ticket 
- https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9756


BMItr (approved):
code lgtm.
Tested on OnePlus 8 Device.

image

PR #4180, opened 2021-10-07
Fix crash due to mismatch of time types (EXPOSUREAPP-9880)

mtwalli:
Issue: Mismatch in time types led to class cast exception 

How to test:
- install 2.11
- scan qrcodes from the ticket
- App will crash 
- Update using the fix here it should not crash anymore

## Ticket 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9880


jurajkusnier (approved):
LGTM no crash

BMItr (approved):
tested on oneplus 8 device.
lgtm

PR #4187, opened 2021-10-08
RecycleBin - Return recycled certs (EXPOSUREAPP-9845, EXPOSUREAPP-9857)

BMItr:
This PR addresses Exposureapp-9845 & EXPOSUREAPP-9857.
returning recycled certs for feature Branch (RecycleBin)

ToDo:
- [ ] clean/adjust tests


mtwalli (commented):
@BMItr 
LGTM, I think we still need to filter out the recycled certificates from showing up in the current provided certificates.

mlenkeit:
> @BMItr I think we should `exclude` recycled certificates from `Validity Check` and `Booster Check`, so you will need to filter them out and not expose them to whatever is doing these checks.
> 
> /cc @mlenkeit

I agree 👍

mtwalli (approved):

PR #4205, opened 2021-10-12
Restore DGC from Reycle Bin on scanning confirmation (EXPOSUREAPP-9843)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9843


KathrinLuecke (approved):
UA approved

PR #4224, opened 2021-10-14
Fix new release back navigation (EXPOSUREAPP-9971, 10005)

SamuraiKek:
Fixes issue with navigation from new release screen back to info screen if user opened the screen from the info screen.

Test navigation by going to info screen -> new features and then go back.

Please also test by installing an updated version. You can increase the VERSION_MINOR from gradle.properties. Try pressing the X button. It should continue to the next screen.


AlexanderAlferov (approved):
LGTM 🚀

BMItr (approved):

PR #4231, opened 2021-10-15
RecycleBin - Clean Up Recycled Certificates older than 30 days (EXPOSUREAPP-9844)

BMItr:
This PR addresses Exposureapp-9844.
"CleanUp all recycled certificates that are older than 30 days on App start"


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/reyclebin/common/RecyclableExtensions.kt
```diff
@@ -0,0 +1,6 @@
+package de.rki.coronawarnapp.reyclebin.common
+
+import org.joda.time.Days
+import org.joda.time.Instant
+
+fun Recyclable.retentionDaysInRecycleBin(now: Instant): Int = recycledAt?.let { Days.daysBetween(it, now).days } ?: 0
```
mtwalli, line 6: should not the calculations here be `seconds` or `milliseconds` ? [SPECS](https://github.com/corona-warn-app/cwa-app-tech-spec/pull/111/files#diff-25c8b7188514d7cdfb430f7f8f963fed63bca279fd74753bc6cefe29dd599b55R1884)

BMItr: with "Days" only Full Days are consided. so imho its fully sufficent.

daysBetween(ReadablePartial start, ReadablePartial end)
_Creates a Days representing the number of whole days between the two specified partial datetimes._
https://www.joda.org/joda-time/apidocs/org/joda/time/Days.html

mtwalli: Please consider adding some unit tests for it ?

mtwalli (approved):

jurajkusnier (approved):

PR #4240, opened 2021-10-18
Fix vaccination certificates priority (EXPOSUREAPP-9982)

SamuraiKek:
Fixed the vaccination priorities for boosters. 

How to test:

- first scan a vacc cert with dose 1/2 that's < 14 days old: 
`cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=2 v.0.mp='EU/1/20/1528' v.0.dt='2021-10-12' -e wru`, this has prio as it's the only vacc cert.
- then scan a vacc cert with dose 2/2 that's < 14 days old: 
`cwa vc gen --pii 42 -s v.0.dn=2 v.0.sd=2 v.0.mp='EU/1/20/1528' v.0.dt='2021-10-12' -e wru`, this has prio now as dose is 2/2.
- then scan a vacc cert with dose 2/2 that's > 14 days old: 
`cwa vc gen --pii 42 -s v.0.dn=2 v.0.sd=2 v.0.mp='EU/1/20/1528' v.0.dt='2021-09-12' -e wru`, this has prio now as dose is 2/2 and immunity is reached.
- then scan a vacc cert with dose 3/3 that's < 14 days old: 
`cwa vc gen --pii 42 -s v.0.dn=3 v.0.sd=3 v.0.mp='EU/1/20/1528' v.0.dt='2021-10-12' -e wru`, this has prio now as dose is 3/3
- then scan a vacc cert with dose 3/3 that's > 14 days old: 
`cwa vc gen --pii 42 -s v.0.dn=3 v.0.sd=3 v.0.mp='EU/1/20/1528' v.0.dt='2021-09-12' -e wru`, the prio remains unchanged because for booster vaccinations, the latest booster has highest prio.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
@@ -74,24 +74,6 @@ private fun Collection<CwaCovidCertificate>.rule2FindRecentRaCertificate(
     .filter { Duration(it.rawCertificate.test.sampleCollectedAt, nowUtc) <= Duration.standardHours(24) }
     .maxByOrNull { it.rawCertificate.test.sampleCollectedAt }
 
-//private fun Collection<CwaCovidCertificate>.rule3FindBoosterShot(
-//    nowUtc: Instant
-//): CwaCovidCertificate? {
-//    val vaccinesWithTwoDoses = listOf("EU/1/20/1528", "EU/1/21/1529", "EU/1/20/1507")
-//    return this
-//        .filterIsInstance<VaccinationCertificate>()
-//        .filter {
-//            with(it.rawCertificate.vaccination) { doseNumber == totalSeriesOfDoses }
-//        }
-//        .filter {
-//            with(it.rawCertificate.vaccination) {
-//                if(totalSeriesOfDoses > 2 && vaccinesWithTwoDoses.contains(medicalProductId)) {
-//                    return it
-//                } else if (totalSeriesOfDoses <= 2 && !vaccinesWithTwoDoses.contains(medicalProductId))
-//            }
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
```diff
@@ -85,20 +85,31 @@ private fun Collection<CwaCovidCertificate>.rule2FindRecentRaCertificate(
...
-    }
-    .filter {
-        Days.daysBetween(it.rawCertificate.vaccination.vaccinatedOn, nowUtc.toLocalDateUtc()).days > 14
-    }
-    .maxWithOrNull(
-        compareBy(
-            { it.rawCertificate.vaccination.vaccinatedOn },
-            { it.headerIssuedAt }
+): CwaCovidCertificate? {
+    val vaccinesWithTwoDoses = listOf("EU/1/20/1528", "EU/1/21/1529", "EU/1/20/1507")
```
chiljamgossow, line 89: If this really needs to change, we should look for a more sustainable solution. We could just sort for dose number, so 3/3 would be selected over 2/2

chiljamgossow (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
SamuraiKek: What I'm trying to do here is differentiate between vaccines that require two doses to get full immunity and the ones that only require one dose like J&J. In the case of J&J, a vaccine whith dose 2/2 would already be considered a booster, right?

SamuraiKek:
Closing this PR as it's not actually a bug that needs to be fixed.

PR #4258, opened 2021-10-22
Clean CWA flag when certs are all recycled or deleted (DEV)

mtwalli:
#### Scenario 1
- scan some certificates 
- switch cwa flag on 
- recycle cwa user certificates
- restore them back 
- cwa flag should be disabled 

#### Scenario 1
- scan some certificates 
- switch cwa flag on 
- recycle cwa user certificates
- delete them all 
- scan certificates again
- cwa flag should be disabled


jurajkusnier (approved):
LGTM

BMItr (approved):
:ok_hand:

PR #4260, opened 2021-10-22
Fix booster notification FAQ link (EXPOSUREAPP-10143)

SamuraiKek:
Changes the link of the FAQ link from the booster notification card to point to the right section from the coronawarnapp website.

To test, register a test older than 150 days: `cwa vc gen -s v.0.dt=-150days` and then go to test settings -> booster rules -> click "run booster rules". The cli tool doesn't always generate a certificate with a booster rules so please keep generating one with the above command until you get a certificate that gets the booster notification. You will just have to repeat the cycle of generate cert -> go to test settings to run the booster rules until you get one that has the rules.

After that, just go to the certificate with the notification and press on the faq link. 
It should open https://www.coronawarn.app/de/faq/#vac_booster if you have you phone's language set to German and https://www.coronawarn.app/en/faq/#vac_booster if you have it set to any other language.


jurajkusnier (approved):

BMItr (approved):

mlenkeit (approved):
LGTM 👍

InkenTownsend (approved):
UA Approved

PR #4272, opened 2021-10-25
Use << in standardized name (EXPOSUREAPP-10140)

chiljamgossow:
uses << as separator in the standardized name of certificate


mtwalli (approved):
LGTM, Tested with RC, VC, TC

SamuraiKek (approved):

PR #4281, opened 2021-10-26
Nullable qrCodeHash (DEV)

mtwalli:
- Legacy saved Test does not consider the default value , therefore It is nullable now


jurajkusnier (approved):
LGTM

BMItr (approved):

PR #4283, opened 2021-10-26
Fixing Universal (File-)Scanner does not scan files when there is no default FileManager 

jurajkusnier:
## Testing

- install CWA on a non-Pixel phone
- install alternative Filemanager (e.g. Dateimanager+)
 - Setup no default Filemanager in device settings 
 - Open QR Code Scanner > Open File > select any build-in or 3rd party file manager and check if the selected file is propagated into the app
 - Check for side effects of changed `launchMode`

## Testing Side Effects
- Register positive test
- Share the keys
- while accessing the keys from ENF a notification will be shown from ENF
- Click it, it should open CWA on the same fragment as 2.13

## Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9805


SamuraiKek (approved):
LGTM. My Samsung chose the default file explorer for me, even though I didn't have one set, but other than that selecting the file works and pressing the notification takes you to the right place.

AlexanderAlferov (approved):

mtwalli (approved):

PR #4285, opened 2021-10-27
Error handling text differences (EXPOSUREAPP-9780)

KathrinLuecke:
Adapted the messages for the QR-Code scanner according to https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9780 (last comment) to be consistent with iOS.

However, there is still a difference: In iOS, we have twice the title "Kein geeigneter QR-Code", whereas in Android, this occurs only once, and the title of the second message is "Datei nicht lesbar". Could you please adapt this?


CV113 (approved):
UA approved

marcauberer (approved):

PR #4304, opened 2021-11-01
Bugfix: App Crash after vacc. deletion (EXPOSUREAPP-10153)

jurajkusnier:
## Description
- probably race condition, where the app crash when a user delete the certificate in `PersonDetailsFragment`
- hard to reproduce
- crash log
```
2021-10-22T07:20:26.415Z E/DefaultDispatcher-worker-8: Uncaught exception!
java.util.NoSuchElementException
	at de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsViewModel.access$createUiState(PersonDetailsViewModel.kt:35)
	at de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsViewModel$uiState$1.invokeSuspend(PersonDetailsViewModel.kt:4)
	at de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsViewModel$uiState$1.invoke(PersonDetailsViewModel.kt:1)
	at kotlinx.coroutines.flow.FlowKt__ZipKt$combine$1$1.invokeSuspend(Zip.kt:1)
	at kotlinx.coroutines.flow.FlowKt__ZipKt$combine$1$1.invoke(Zip.kt:1)
	at kotlinx.coroutines.flow.internal.CombineKt$combineInternal$2.invokeSuspend(Combine.kt:29)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:3)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:18)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:1)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:10)
```


vaubaehn:
@jurajkusnier 
I have a gut feeling that the crash you take care of here has the same reasons like the problems in
https://github.com/corona-warn-app/cwa-app-android/issues/3549 .
I know it sounds weird, but I was already suspecting an issue with Kotlin `flow`, likely in connection with `R8`. It could be that a few JVM implementations do not execute the code as expected (unrealiable flow that does not emit values when needed). The crash report of your error here points a bit into that direction.

@mtwalli recently updated Kotlin - has your crash been reported for an older Kotlin version or the recently updated one?

jurajkusnier:
> @jurajkusnier I have a gut feeling that the crash you take care of here has the same reasons like the problems in #3549 . I know it sounds weird, but I was already suspecting an issue with Kotlin `flow`, likely in connection with `R8`. It could be that a few JVM implementations do not execute the code as expected (unrealiable flow that does not emit values when needed; or maybe here: emits when it shouldn't?). The crash report of your error here points a bit into that direction.
> 
> @mtwalli recently updated Kotlin - has your crash been reported for an older Kotlin version or the recently updated one?
> 
> Anyway, even it was not related to flow, could be a similar race condition that Matthias was already suspecting for #3549 .

Thanks for the comment. I wasn't able to reproduce this issue. Even the tester who reported it wasn't able to reproduce it anymore so it's hard to investigate and find the cause of the problem.
The crash has been reported for the build using kotlin v1.5.30, so not the latest one, but again it's hard to confirm or rule out whether the problem was fixed by kotlin update or is there another issue, as the race condition

AlexanderAlferov (dismissed):
Not able to reproduce. Solution LGTM 🚀

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
```diff
@@ -70,7 +70,14 @@ class PersonDetailsViewModel @AssistedInject constructor(
...
-        val priorityCertificate = personCertificates.highestPriorityCertificate
+        val priorityCertificate = try {
+            personCertificates.highestPriorityCertificate
+        } catch (e: NoSuchElementException) {
+            Timber.e(e)
+            events.postValue(Back)
+            return UiState(name = "", emptyList())
+        }
+
         val certificateItems = mutableListOf<CertificateItem>().apply {
```
chiljamgossow, line 81: Not sure about this. We allow PersonCertificates to be created with an empty list of certificates, which will always result in a crash as personIdentifier and highestPriorityCertificate are defined as not nullable. We could fix this underlying issue.

jurajkusnier: - `PersonCertificates.personIdentifier` 
- `PersonCertificates.highestPriorityCertificate`

are now nullable

chiljamgossow (approved):
Looking good, also tested removing vaccination certificate without issues. Thanks!

mtwalli (approved):

vaubaehn:
> Thanks for the comment. I wasn't able to reproduce this issue. Even the tester who reported it wasn't able to reproduce it anymore so it's hard to investigate and find the cause of the problem.
> The crash has been reported for the build using kotlin v1.5.30, so not the latest one, but again it's hard to confirm or rule out whether the problem was fixed by kotlin update or is there another issue, as the race condition

@jurajkusnier Sorry, I forgot to thank for your reply in a timely manner!

PR #4329, opened 2021-11-04
Improve Person regrouping (EXPOSUREAPP-10154)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10154

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/VaccinationRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/VaccinationRepository.kt
@@ -28,7 +28,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.SharingStarted
 import kotlinx.coroutines.flow.catch
-import kotlinx.coroutines.flow.drop
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.launchIn
@@ -73,7 +72,6 @@ class VaccinationRepository @Inject constructor(
     init {
         internalData.data
             .onStart { Timber.tag(TAG).d("Observing VaccinationContainer data.") }
-            .drop(1) // Initial emission, restored from storage.
             .onEach { vaccinatedPersons ->
                 Timber.tag(TAG).v("Vaccination data changed: %s", vaccinatedPersons)
                 storage.save(vaccinatedPersons.map { it.data }.toSet())
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/VaccinationRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/VaccinationRepository.kt
@@ -326,7 +326,7 @@ class VaccinationRepository @Inject constructor(
 
             val updatedPerson = vaccinatedPerson.copy(
                 data = vaccinatedPerson.data.copy(
-                    lastSeenBoosterRuleIdentifier = vaccinatedPerson.data.boosterRule?.identifier
+                    lastSeenBoosterRuleIdentifier = vaccinatedPerson.data.boosterRuleIdentifier
                 )
             )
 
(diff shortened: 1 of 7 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/storage/VaccinationStorage.kt
```diff
@@ -56,23 +56,17 @@ class VaccinationStorage @Inject constructor(
...
                 remove(it)
             }
-            persons.forEach {
-                val raw = gson.toJson(it)
-                val identifier = it.identifier
-                Timber.tag(TAG).v("Storing vaccinatedPerson %s -> %s", identifier, raw)
-                putString("$PKEY_PERSON_PREFIX${identifier.groupingKey}", raw)
+            persons.groupDataByIdentifier().forEach {
+                if (it.vaccinations.isNotEmpty()) {
+                    val raw = gson.toJson(it)
```
chiljamgossow, line 61: checking for vaccinations as accessing identifier would crash if the are none

chiljamgossow (dismissed):
Tested migration + update from 2.12

BMItr (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/storage/VaccinationStorage.kt
```diff
@@ -56,45 +56,41 @@ class VaccinationStorage @Inject constructor(
...
-        save(load().groupDataByIdentifier())
-    }
-
     companion object {
         private const val TAG = "VaccinationStorage"
         private const val PKEY_PERSON_PREFIX = "vaccination.person."
     }
 }
 
 internal fun Set<VaccinatedPersonData>.groupDataByIdentifier(): Set<VaccinatedPersonData> {
```
BMItr, line 76: for this fun the following **suggestion**:

internal fun Set<VaccinatedPersonData>.groupDataByIdentifier(): Set<VaccinatedPersonData> =
    filterNot { it.vaccinations.isNullOrEmpty() }
        .groupBy { it.identifier }
        .mapNotNull { entry ->
            val personDataList = entry.value
            if (personDataList.isEmpty()) {
                Timber.v("Person data list was empty, returning early")
                return@mapNotNull null
            }

            val newestData = personDataList.maxByOrNull {
                it.lastBoosterNotifiedAt ?: Instant.EPOCH
            }
            val vaccinations = personDataList.flatMap { it.vaccinations }.toSet()
            newestData?.copy(vaccinations = vaccinations)
        }.toSet()

BMItr (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/VaccinationRepository.kt
```diff
@@ -73,7 +73,6 @@ class VaccinationRepository @Inject constructor(
     init {
         internalData.data
             .onStart { Timber.tag(TAG).d("Observing VaccinationContainer data.") }
-            .drop(1) // Initial emission, restored from storage.
             .onEach { vaccinatedPersons ->
                 Timber.tag(TAG).v("Vaccination data changed: %s", vaccinatedPersons)
                 storage.save(vaccinatedPersons.map { it.data }.toSet())
```
chiljamgossow, line 78: the regrouping is only applied to the storage, not the value in the flow

BMItr (dismissed):
image
badge tested successfully on oneplus 8
No crash could be spooted yet.

BMItr (dismissed):

chiljamgossow (dismissed):

BMItr (dismissed):

chiljamgossow (dismissed):

chiljamgossow (approved):

BMItr (approved):

vaubaehn:
🤞

vaubaehn:
According to Google Play Store Reviews the fix made many peolpe very happy 🎉 

Some community members and me are quite curious what actually happened.
As far as I understood the problem, it was a special case when applying booster notification rules on probably J&J jabs only?
And another community member wondered, how that influenced the amount of serialization that happened on start-up which is what the OOM stacktrace from the related issue points to?

Generally, we fully understand that the high time pressure likely made it impossible to document more on that hot fix. But we would be very happy, if you might return to the style some weeks/months ago, where PRs were documented with some more detail ❤️ 
Explaining PRs with "JIRA" or "TICKET" sometimes makes it hard to guess 😉 

Anyway, have a nice week and many greetings from all of us!

PR #4171, opened 2021-10-05
Uqs error handling / dialog clenaup (EXPOSUREAPP-9777)

BMItr:
This PR addresses ExpsoureApp-9777.
Dialog adjustments.

How to test:
- Open the Universal Scanner
- Scan an invalid QR code (can be found in the ticket)
- Check the dialog and dialog navigation

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9777

PR #4340, opened 2021-11-08
Update certificate validation logic (EXPOSUREAPP-10447)

jurajkusnier:
- updated validation
- error dialog
## Testing
The following presets can be used for testing the feature against the mock server:

- should be blocked:
    - `cwa vc gen --preset 60`
    - `cwa vc gen --preset 61`
    - `cwa vc gen --preset 63`

- should not be blocked:
    - `cwa vc gen --preset 62`

## Jira Link
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10238


CV113 (approved):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/ui/onboarding/CovidCertificateOnboardingViewModel.kt
```diff
@@ -26,7 +29,12 @@ class CovidCertificateOnboardingViewModel @AssistedInject constructor(
         covidCertificateSettings.isOnboarded.update { true }
         val event = if (dccQrCode != null) {
             try {
-                val containerId = dccQrCodeHandler.handleQrCode(dccQrCode)
+                val containerId = dccQrCodeHandler.handleQrCode(
```
mtwalli, line 32: what about injecting the AppConfig in `DccQrCodeHandler` itself, so you don't need to inject the AppConfigProvider everywhere you need to call the Handler

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CwaCovidCertificate.kt
```diff
@@ -105,6 +107,17 @@ interface CwaCovidCertificate : Recyclable {
                 .registerSubtype(ExpiringSoon::class.java, "ExpiringSoon")
                 .registerSubtype(Expired::class.java, "Expired")
                 .registerSubtype(Invalid::class.java, "Invalid")
+                .registerSubtype(Blocked::class.java, "Blocked")
+        }
+
+        override fun equals(other: Any?): Boolean {
+            if (this is Blocked && other is Blocked) return true
+            if (this is Recycled && other is Recycled) return true
```
mtwalli, line 115: why do you need special handling for these cases ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/ui/onboarding/CovidCertificateOnboardingViewModel.kt
mtwalli: I would rather do that in the `BlocklistValidator` itself

mtwalli:
All certificates in the description are `VC` do you have some `TC, RC` that can be blocked?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/ScannerExtensions.kt
```diff
@@ -37,6 +37,12 @@ fun Throwable.toQrCodeErrorDialogBuilder(context: Context): MaterialAlertDialogB
                         context.openUrl(R.string.error_dcc_already_registered_faq_link)
                     }
                 }
+                throwable.isBlocked -> {
+                    setTitle(R.string.error_dcc_in_blocklist_title)
+                    setNegativeButton(R.string.error_dcc_in_blocklist_faq) { _, _ ->
+                        context.openUrl(R.string.dcc_signature_validation_dialog_faq_link)
+                    }
+                }
```
mtwalli, line 45: Dialog buttons should be reversed in order
Screenshot 2021-11-09 at 15 37 14

Look at figma https://www.figma.com/file/KuDEcMSREg3VhZRhC3FXTY/COVID19_Master_2.11?node-id=37357%3A27922

jurajkusnier: AFIK that's iOS dialog, and we should rather stick to our material design guidelines where the button order is as follows: 
image

mtwalli: OK, btw I think there is an issue here from before , we should use `setNeutralButton` as other dialogs ,but apparently it is overridden by  ` .setNeutralButton(null, null) // Remove details` and this removes any link in the dialogs. can you please fix it? I think the following would fix it 
```kotlin
    return toErrorDialogBuilder(context).apply {
        if (throwable is InvalidHealthCertificateException) {
            when {
                throwable.isCertificateInvalid ->
                    setNeutralButton(R.string.error_button_dcc_faq) { _, _ ->
                        context.openUrl(R.string.error_button_dcc_faq_link)
                    }

                throwable.isSignatureInvalid -> {
                    setTitle(R.string.dcc_signature_validation_dialog_title)
                    setNeutralButton(R.string.dcc_signature_validation_dialog_faq_button) { _, _ ->
                        context.openUrl(R.string.dcc_signature_validation_dialog_faq_link)
                    }
                }
                throwable.isAlreadyRegistered -> {
                    setNeutralButton(R.string.error_button_dcc_faq) { _, _ ->
                        context.openUrl(R.string.error_dcc_already_registered_faq_link)
                    }
                }
                throwable.isBlocked -> {
                    setTitle(R.string.error_dcc_in_blocklist_title)
                    setNeutralButton(R.string.error_dcc_in_blocklist_faq) { _, _ ->
                        context.openUrl(R.string.dcc_signature_validation_dialog_faq_link)
                    }
                }
            }
        } else {
            setNeutralButton(null, null) // Remove details
        }
    }
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CwaCovidCertificate.kt
jurajkusnier: comparing state with deserialized state didn't work as expected in https://github.com/corona-warn-app/cwa-app-android/blob/747cf79192f218129e4d9fbdad8de8fb8d8ab9ca/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/storage/VaccinationContainer.kt#L175

even if `state` and `lastSeenStateChange` were objects of the same type `Blocked`, equals returned false. It works for data classes.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/ScannerExtensions.kt
jurajkusnier: other `InvalidHealthCertificateException` shouldn't have an FAQ button according to the Figma design. It has been changed by this PR https://github.com/corona-warn-app/cwa-app-android/pull/4171 by adding `setNeutralButton(null, null)` but I can probably remove `setNeutralButton()` directly from `toErrorDialogBuilder()` method when we don't need it anymore

mtwalli: > other InvalidHealthCertificateException shouldn't have an FAQ button according to the Figma design.
Where do you see that ?
#4171  was only to remove the `details` button not FAQ button ,I think this wasn't intended ,right?
/cc @BMItr

BMItr: yes, #4171 was only to remove the details button (and stay on the screen to scan again)

jurajkusnier: updated

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/ui/onboarding/CovidCertificateOnboardingViewModel.kt
jurajkusnier: it's not used there anymore

PR #4341, opened 2021-11-08
Mitigate base45 security vulnerability (EXPOSUREAPP-10454)

chiljamgossow:
covered by unit tests


vaubaehn:
😁 
Is this actually a technical security risk, or more a good way to detect a broader range of fake DCCs?

mtwalli (approved):

vaubaehn:
@chiljamgossow 
> This has been reported as a security risk and the EU recommends to fix this. Fake DCC will be covered by other means.

thanks for the clarification. The DCC that you use in your unit test was shared in a public discussion that possibly triggered the EU recommendation. From what I understood explained in other places (https://github.com/ministero-salute/it-dgc-verificaC19-android/issues/185#issuecomment-958802827 and following) some base45 decoders parse the valid beginning correctly, stop when reaching invalid paths but return back the valid (COSE) object that has already been decoded. The appended not decoded part may then include rubbish, in the case of the DCC that is also used in your unit test, some unpolite profanities... Even this may not be a security risk per se, it's probably a very good idea to adapt the decoder to be more strict...
Because that of protecting against such base45 tricking I thought, rejecting such QR codes will fight fake DCCs in parallel.
Enjoy your night!

jurajkusnier (approved):

floscher:
@chiljamgossow @vaubaehn These changes from the original implementation might also be of interest, they seem to also cover some edge cases in Base45 encoding/decoding: https://github.com/ehn-dcc-development/hcert-kotlin/commit/af2aa4da2288548258ee2fe8c0a987c288438ec8

floscher:
> These changes from the original implementation might also be of interest, they seem to also cover some edge cases in Base45 encoding/decoding: [ehn-dcc-development/hcert-kotlin@af2aa4d](https://github.com/ehn-dcc-development/hcert-kotlin/commit/af2aa4da2288548258ee2fe8c0a987c288438ec8)

nvm, your PR seems to have fixed the same issues addressed in that commit, so you can ignore my comment(s)

vaubaehn:
@chiljamgossow and @mtwalli 
To be honest, I do not feel good to have that specific DCC for the unit test in our CWA code that you're using currently. Call me suspicious, but for me it has a bad karma...

I'd propose to use this instead:
`HC1:NCFOXN%TSMAHN-H/RCMPQ5GE5I00H9GBH3QNAD6.LQLX85ZS GJTSJ4NKP1HCV4*XUA2PSGH.+H$NI4L6F$S-N1FYBRR1$Q1+GOF+P$HQPHQHTQ.SQ6$PUKRN95404.W7UX4795L*KDYPWGO+9AZDOHCRL35IWMSDOP7OQ+M70AK$8 96XY4SBLU96:/6N9R%EPL8RY9DOA60-K.IA.C8KRDL4O54O4IGUJKJGI0JAXD15IAXMFU*GSHGRKMXGG6DBYCBMQN:HG5PAHGG8KES/F-1JW-K%B3A9ENO4B-S-*O4-G1FD/U47HAE1MI4OE0G1:HHD4AB874MM-6B:HKJSQ.TAG3CR1638W9AV88G64PB4VHRY2EK03NFJL4M10KP3AT2VK LT5GGFV85I0*10W2ZXJSBTMFW*+KM2T8-CXR32BMF7RAEAYKMWHE/NH UP4SNGENEWUY97 -3YM0.HAM:D:00ZY35XRT1100MFKWEWYHKCZIZJ0CAQYIKOZIZJ0DAQCDQGAE62ORR7HL0POYQCGMGBBEQFCBCDCT4VYFQZEPYFXK69ZRV9T8IY6NWAYEJF0VXVKNFCG/X9UIG5HFWRPEDW2EPW00S1K83I3900:919GETVACCINATEDSOONDEARS9IKM24N21AW7ZUR2L013300WEWILLFIGHTCOVIDTOGETHERRS666I3AK96XYR01300PROTECTYOURSELFANDOTHERSANDBESAFELI5340XRT09SHUZVB33IR0FIGHTVIRUSHERE83H73KFI280I23KL0YRZT91Y0WITHLOVEFROMCWA*PWEOXCI/QXRLOZKN0LM*03I2RWT0Z`
 
It has the same DCC contents, and it can be imported into CWA 2.6.1 without problem (CovPass and CovPassCheck refuse that DCC). So it provides same functionality for testing, but the hidden message is by far more appropriate.

If you agree, I may open a PR - or you may replace it yourself.

Thanks for your consideration!

vaubaehn:
That's sad, indeed.
@chiljamgossow & @mtwalli But could you consider to replace the certificate for the unit test like suggested in https://github.com/corona-warn-app/cwa-app-android/pull/4341#issuecomment-965129908 , or is the cert/unit test also part of the EU reference implementation?

PR #4310, opened 2021-11-02
Kathrin luecke patch 1 (EXPOSUREAPP-10246)

KathrinLuecke:
Changed "submission_status_card_button_show_results" from "Test anzeigen" to "Testergebnis abrufen".
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10246

PR #4348, opened 2021-11-08
Revert "Kathrin luecke patch 1 (EXPOSUREAPP-10246)"

KathrinLuecke:
Reverts corona-warn-app/cwa-app-android#4310

The change also affected texts in other contexts - see comment in https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10246. Therefore it needs to be reverted and analyzed.


CV113 (approved):
UA approved

mtwalli (approved):

PR #4354, opened 2021-11-09
Fix: DCC parent color shade (EXPOSUREAPP-9950)

AlexanderAlferov:
## Whats new
DCC details fragment inherits color shades from icon in Person Detail Fragment.
Bookmark icon now has no dark mode and has 3 blue variants. 

## Ticket
[Jira](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9950)
⚠️ ATTENTION ⚠️ : Check the last two comments in the ticket before testing

## Testing
Scan different DCC (tests, vaccination, recovery) for several persons.
Check the background of DCC details fragment:
1) Navigating from notification and Universal Scanner background should be changed according to old logic
2) Navigating from Person Detail fragment background should inherit color from icon of DCC in Person Detail fragment.
3) Bookmark icons now should be same color as icon of DCC.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
```diff
@@ -162,19 +167,36 @@ class PersonDetailsViewModel @AssistedInject constructor(
...
+                    events.postValue(
+                        OpenRecoveryCertificateDetails(
+                            containerId = certificate.containerId,
+                            colorShade = getItemColorShade(certificate.isValid, isCurrentCertificate)
+                        )
+                    )
                 }
             )
         }
     }
```
BMItr, line 192: private fun getItemColorShade(isValid: Boolean, isCurrentCertificate: Boolean): PersonColorShade = when(..) {  }

chiljamgossow (approved):
Tested, works

BMItr (approved):
very nice 🚀

PR #4355, opened 2021-11-09
Json and extractor (EXPOSUREAPP-10477)

chiljamgossow:
- [X] unit tests

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingData.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingData.kt
@@ -1,0 +1,22 @@
+package de.rki.coronawarnapp.dccticketing.core.qrcode
+
+import com.google.gson.annotations.SerializedName
+
+data class DccTicketingData(
+    @SerializedName("protocol")
+    val protocol: String?,
+    @SerializedName("protocolVersion")
+    val protocolVersion: String?,
+    @SerializedName("serviceIdentity")
+    val serviceIdentity: String?,
+    @SerializedName("privacyUrl")
+    val privacyUrl: String?,
+    @SerializedName("token")
+    val token: String?,
+    @SerializedName("consent")
+    val consent: String?,
+    @SerializedName("subject")
+    val subject: String?,
+    @SerializedName("serviceProvider")
+    val serviceProvider: String?,
+)
--- a/Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeExtractorTest.kt
+++ b/Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeExtractorTest.kt
@@ -29,7 +29,7 @@ class DccTicketingQrCodeExtractorTest {
 
     @Test
     fun `missing service provider throws exception`() = runBlockingTest {
-        extractor.canHandle(validQrCode) shouldBe true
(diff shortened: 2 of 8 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingData.kt
```diff
@@ -0,0 +1,22 @@
...
+    val privacyUrl: String?,
+    @SerializedName("token")
+    val token: String?,
+    @SerializedName("consent")
+    val consent: String?,
+    @SerializedName("subject")
+    val subject: String?,
+    @SerializedName("serviceProvider")
+    val serviceProvider: String?,
+)
```
mtwalli, line 22: The fields don't like optional according to [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/749572abba54f77e670bdf384f52b0b6f747ebaa/docs/spec/dcc-validation-service.md#data-structure-of-the-initialization-data) 

We do `validate` some fields, but not all such `serviceIdentity` 

the following  json will pass
```json
    {
    "protocol": "DCCVALIDATION",
    "subject": "00241d11-27b4-41af-9e77-3418c3eccfd4",
    "serviceProvider": "Booking Demo"
    }
```

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeExtractorTest.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    @BeforeEach
+    fun setup() {
+        DaggerCovidCertificateTestComponent.factory().create().inject(this)
+    }
+
+    @Test
+    fun `happy path extraction`() = runBlockingTest {
+        extractor.canHandle(validQrCode) shouldBe true
+        extractor.extract(validQrCode)
```
mtwalli, line 27: `shouldNotThrowAny{}` ?

mtwalli (commented):
Lgtm, few questions

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingData.kt
chiljamgossow: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/749572abba54f77e670bdf384f52b0b6f747ebaa/docs/spec/dcc-validation-service.md#from-qr-code-to-consent-i Here it says that only those three need to be validated. If I make them nonnull, then it will fail during parsing if null and the expected error code cannot be provided @mlenkeit What fields are nullable?

mtwalli (approved):
LGTM, Thanks!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingData.kt
mlenkeit: @chiljamgossow @mtwalli for Initialization Data, all of them are mandatory in the sense that they are not null and at least an empty string. So I think the `?` can be removed. On top of that, there's the field-specific validation some fields that you referred to.

BMItr (approved):
👌

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingData.kt
chiljamgossow: I will adapt in a subsequent PR

PR #4368, opened 2021-11-10
Lenient mode for decoder (EXPOSUREAPP-10507)

chiljamgossow:
... prevents the app from crashing if a manipulated certificate has already been scanned


vaubaehn:
Thanks a lot for the extra way... I'm still on 2.6.1 with many DCCs for debugging. When the day of forced update comes, your commit will save my ... eh... app.

BMItr (approved):
code lgtm 👌 
tested with oneplus 8 device. 2.12->2.13

mtwalli (approved):
Works 
- `2.12.x` -> `2.13.3.2` (crash) -> branch (Recover)
- Scanning not possible (Branch)

floscher:
Even better would be if the contents of the manipulated QR code would be discarded completely. At the moment in lenient mode the manipulated ByteArray is still returned where strict mode would throw an exception. Maybe return just an empty ByteArray or something safe that won't crash the app but also does not return any value that results from manipulated codes.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/encoding/Base45Decoder.kt
```diff
@@ -69,29 +69,36 @@ object Base45Decoder {
...
-    fun decode(input: String): ByteArray =
+    fun decode(input: String, mode: Mode = Mode.STRICT): ByteArray =
         input.toByteArray().asSequence().map {
             decoding_charset[it.toInt()].also { index ->
                 if (index < 0) throw IllegalArgumentException("Invalid characters in input.")
             }
         }.chunked(3) { chunk ->
-            if (chunk.size < 2) throw IllegalArgumentException("Invalid input length.")
-            chunk.reversed().toInt(45).toBase(base = 256, count = chunk.size - 1).reversed()
+            if (mode.isStrict() && chunk.size < 2) throw IllegalArgumentException("Invalid input length.")
```
floscher, line 78: Kind of like this (not tested!)

```suggestion
            if (chunk.size < 2) { if (mode.isStrict()) throw IllegalArgumentException("Invalid input length.") else return ByteArray(0) }
```

floscher:
@chiljamgossow @mtwalli @vaubaehn @BMItr With this PR, the certificates with malformed Base45 that were addressed in #4341 will no longer crash the app, but they will still show as valid again in the certificate list, if they were imported into the CWA in a previous version. Those certificates should at least be marked as invalid, or even removed from the certificates list.

I tested this with the certificate mentioned in https://github.com/corona-warn-app/cwa-app-android/pull/4341#issuecomment-965129908 . The root cause is what I mentioned above: Lenient mode still returns the content of the QR code, instead of discarding the content of malformed codes.

PR #4374, opened 2021-11-11
Init transaction (EXPOSUREAPP-10484)

chiljamgossow:
initialisation and data structures


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeExtractor.kt
```diff
@@ -29,6 +29,7 @@ class DccTicketingQrCodeExtractor @Inject constructor(
         }
     }
 
+    @Suppress("UselessCallOnNotNull")
```
mtwalli, line 32: Can we remove it if `null` is not expected anymore ?

mtwalli (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeExtractor.kt
chiljamgossow: it will crash here if the value was null (not while parsing) and then expected error code is not showing

PR #4383, opened 2021-11-12
Cancel booster indication (EXPOSUREAPP-10232)

mtwalli:
- Booster indication is getting cancelled automatically if user is not eligible for booster anymore
- Booster rules runs 
  - after scanning new certificates
  - Recycle certificate
  - Restore certificate

### Testing 
- Scan certificates set from the attached one to the ticket where user is eligible (1,2)
- Scan the booster certificate (3) -> the booster indication message should be gone after the scan

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10232

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
@@ -34,7 +34,7 @@ class DccQrCodeHandler @Inject constructor(
             is TestCertificateQRCode -> testCertificateRepository.registerCertificate(dccQrCode).containerId
             else -> throw UnsupportedQrCodeException()
         }.also {
-            boosterCheckScheduler.runNow()
+            boosterCheckScheduler.scheduleNow()
         }
     }
 }
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
@@ -1,5 +1,6 @@
 package de.rki.coronawarnapp.qrcode.handler
 
+import de.rki.coronawarnapp.covidcertificate.booster.BoosterCheckScheduler
 import de.rki.coronawarnapp.covidcertificate.common.exception.InvalidHealthCertificateException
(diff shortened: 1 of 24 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
```diff
@@ -17,21 +18,23 @@ class DccQrCodeHandler @Inject constructor(
...
+    suspend fun handleQrCode(dccQrCode: DccQrCode): CertificateContainerId {
+        dscSignatureValidator.validateSignature(dccData = dccQrCode.data)
+        return when (dccQrCode) {
+            is RecoveryCertificateQRCode -> recoveryCertificateRepository.registerCertificate(dccQrCode).containerId
+            is VaccinationCertificateQRCode -> vaccinationRepository.registerCertificate(dccQrCode).containerId
+            is TestCertificateQRCode -> testCertificateRepository.registerCertificate(dccQrCode).containerId
+            else -> throw UnsupportedQrCodeException()
+        }.also {
+            boosterCheckScheduler.scheduleNow()
         }
```
chiljamgossow, line 38: Is it possible to move booster check into registerCertificate() so the workaround with the 3000 ms delay is not needed?

KathrinLuecke:
I am not sure why UA should review this as no text changes are included?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
mtwalli: Unfortunately it is not possible for the following reason:
- it creates a cyclic DI dependencies between BoosterNotificationService  <-> VaccinationRepository

mtwalli:
> I am not sure why UA should review this as no text changes are included?

No UA is required here

CV113 (dismissed):
UA approved

SamuraiKek (approved):
Tested with all certs from the ticket. Works nicely.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/handler/DccQrCodeHandler.kt
chiljamgossow: 🤷‍♀️

chiljamgossow (approved):

PR #4395, opened 2021-11-16
[Text] Adjust text "Find Testing Center"

CV113:
[Text] Adjust text "Find Testing Center" --> https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10585


SamuraiKek (dismissed):

marcauberer (changes requested):
@CV113 Shouldn't we here rather use `\n` instead of a line break

CV113:
> @CV113 Shouldn't we here rather use `\n` instead of a line break

@marcauberer I'm not familiar with the development format but if you consider that this is better, than please go ahead.  :-) I agree

marcauberer (approved):

KathrinLuecke (approved):
UA approved

PR #4396, opened 2021-11-16
Unit test certlogic French rule (DEV)

chiljamgossow:
added a unit test to ensure that the rule about market auth holders passes
see https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10455


mtwalli (dismissed):

AlexanderAlferov (dismissed):

AlexanderAlferov:
Its all green 🚀 
But before merge kindly check if we need it in 2.14. Mb it should go directly to 2.15?

mtwalli (approved):

PR #4399, opened 2021-11-16
Request Access Token (EXPOSUREAPP-10488)

jurajkusnier:
## Request Access Token

https://github.com/corona-warn-app/cwa-app-tech-spec/blob/cae7102b3109260f554e97142464b6fd25a8283e/docs/spec/dcc-validation-service.md#requesting-the-access-token

## Jira Link
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10488


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccticketing/core/common/JwtTokenValidatorKtTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import de.rki.coronawarnapp.dccticketing.core.transaction.DccTicketingAccessToken
+import io.kotest.assertions.throwables.shouldNotThrow
+import io.kotest.assertions.throwables.shouldThrow
+import io.mockk.MockKAnnotations
+import io.mockk.every
+import io.mockk.impl.annotations.MockK
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+
+internal class JwtTokenValidatorKtTest {
```
mtwalli, line 12: `: BaseTest` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccTicketingAccessTokenValidator.kt
```diff
@@ -0,0 +1,8 @@
+package de.rki.coronawarnapp.dccticketing.core.common
+
+import de.rki.coronawarnapp.dccticketing.core.transaction.DccTicketingAccessToken
+
+fun DccTicketingAccessToken.validate() {
+    if (t != 1 && t != 2) throw DccTicketingException(DccTicketingException.ErrorCode.ATR_PARSE_ERR)
```
mtwalli, line 6: `t !in 1..2` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/DccTicketingServer.kt
```diff
@@ -85,4 +85,19 @@ class DccTicketingServer @Inject constructor(
         val certificateChain = handshake?.peerCertificates ?: emptyList()
         serverCertificateChecker.checkCertificate(certificateChain, jwkSet)
     }
+
+    suspend fun getAccessToken(
```
BMItr, line 89: You should validate the raw response with cert pinning in here, not sure if this will work with an Retrofit Response because it strips of the server certs.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/transaction/DccJWK.kt
```diff
@@ -43,4 +48,18 @@ data class DccJWK(
         @SerializedName("enc")
         ENCRYPTION
     }
+
+    @IgnoredOnParcel
+    private val certificateFactory by lazy {
```
BMItr, line 53: maybe use a fun with cert factory as param

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccTicketingAccessTokenValidator.kt
```diff
@@ -0,0 +1,8 @@
...
+    if (aud.isBlank()) throw DccTicketingException(DccTicketingException.ErrorCode.ATR_AUD_INVALID)
+}
```
BMItr, line 8: some Logs would be nice!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/AccessTokenRequest.kt
```diff
@@ -0,0 +1,10 @@
+package de.rki.coronawarnapp.dccticketing.core.server
+
+import com.fasterxml.jackson.annotation.JsonProperty
```
BMItr, line 3: Why jackson?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/transaction/DccJWK.kt
jurajkusnier: I think `CertificateFactory` is a singleton, what benefit it would give us to use it as a parameter?
btw I'm using a similar approach as we already have in `DscSignatureValidator`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/AccessTokenRequest.kt
jurajkusnier: yes, it should be `@SerializedName`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccTicketingAccessTokenValidator.kt
jurajkusnier: ok

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/DccTicketingServer.kt
jurajkusnier: I guess cert pins shouldn't be checked in the retrofit response but rather using [CertificatePinner](https://square.github.io/okhttp/4.x/okhttp/okhttp3/-certificate-pinner/) in OkHttp client.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/transaction/DccJWK.kt
BMItr: No, it is not a Singleton. Check the implementation. And in DscSignatureValidator there is only one instance created and used for every check. You are creating a new Instance for every DccJWK which gets checked..

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/DccTicketingServer.kt
BMItr: https://tenor.com/Vm1U.gif

BMItr: Acutally, it should. The problem with their cert pinner check is, that it differs from the one described in our TechSpec. In their check it is also not possible to change the trusted certs dynamically, which we get from the jwks. Furthermore, their check happens on every call, while the check **should only happen on certain calls**.

jurajkusnier: yes, it only happens in the call where you include httpClient with the CertificatePinner and you can set up cert pinning before creating okhttpclient. but maybe let's have a call or discuss it in the chat

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/transaction/DccJWK.kt
jurajkusnier: updated

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/DccTicketingServer.kt
jurajkusnier: it could be resolved later after https://github.com/corona-warn-app/cwa-app-android/pull/4415 is finalized and merged

PR #4412, opened 2021-11-18
Statistics Turkish formatting issue (EXPOSUREAPP-10330)

SamuraiKek:
When changing language to Turkish, the percentage values on the statistics cards used to have the percent symbol on the right side of the value. Only after closing the app completely, the percent symbol showed up on the left side of the value.

This PR updates the percent sybmol position at runtime and doesn't requqire the app to be restarted. 

How to test:

- Have your app set in German/English.
- Change system language to Turkish and go back to the app. 
- Scroll through the statistics tyles and check the percentage value on the ones that have it. It should show up on the left side right away. 

Ticket: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10330


AlexanderAlferov (approved):
LGTM 
Screenshot 2021-11-19 at 13 34 01

BMItr (approved):
image

lgtm

PR #4415, opened 2021-11-19
Certificate pinning (EXPOSUREAPP-10635)

BMItr:
This PR addresses Exposureapp-10635. Subtask for the TicketValidation Story.
Implements server certificate check.
Test added.

PR #4422, opened 2021-11-22
Certificate Pinning (EXPOSUREAPP-10635)

BMItr:
This PR addresses Exposureapp-10635. Subtask for the TicketValidation Story.
Implements server certificate check.
Test added.

(This is a follow up PR for #4415)
@harambasicluka thx for your rocket and best Greetings 😄


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/check/DccTicketingServerCertificateCheckException.kt
```diff
@@ -9,6 +9,7 @@ class DccTicketingServerCertificateCheckException(
         val message: String
     ) {
         CERT_PIN_NO_JWK_FOR_KID("No matching jwk for required kid"),
-        CERT_PIN_MISMATCH("The SHA-256 fingerprint of leafCertificate is not included in requiredFingerprints")
+        CERT_PIN_MISMATCH("The SHA-256 fingerprint of leafCertificate is not included in requiredFingerprints"),
+        CERT_PIN_UNSPECIFIED_ERR("Unspecified error occurred. Needs further investigation")
```
mtwalli, line 13: where did you get this error code from ? [Specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/46af2e99868fb6a57d686fff066a5254758b6d13/docs/spec/dcc-validation-service.md#checking-the-server-certificate-against-a-set-of-jwks)

Please note that these specific error codes are mapped later on based on other use cases into different error codes , we can not just add new ones without aligning with @mlenkeit

mtwalli: what about returning `CERT_PIN_MISMATCH` ? if operation failed for any other reason

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/check/DccTicketingServerCertificateChecker.kt
```diff
@@ -2,42 +2,78 @@ package de.rki.coronawarnapp.dccticketing.core.check
...
-        // 7. Implement all steps
+        when (requiredFingerprints.contains(leafCertificate.createSha256Fingerprint())) {
+            true -> Timber.d("Certificate check was successful")
+            false -> throw DccTicketingServerCertificateCheckException(ErrorCode.CERT_PIN_MISMATCH)
+        }
+    } catch (e: Exception) {
+        throw when (e) {
+            is DccTicketingServerCertificateCheckException -> e
+            else ->
+                DccTicketingServerCertificateCheckException(errorCode = ErrorCode.CERT_PIN_UNSPECIFIED_ERR, cause = e)
```
jurajkusnier, line 54: could you log the unspecified error?
and then rather return known error code as @mtwalli mentioned

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/check/DccTicketingServerCertificateCheckException.kt
BMItr: spec.
adjusted.

mtwalli (approved):
Danke 🙏

jurajkusnier (approved):

PR #4426, opened 2021-11-23
Verifying the Signature of a JWT with a Public Key and JSON Web Token library (EXPOSUREAPP-10647)

jurajkusnier:
- Verifying the Signature of a JWT with a Set of JWKs
- Verifying the Signature of a JWT with a Public Key
- Updating `AccessTokenRequestProcessor` implementation
- Introducing new library https://bitbucket.org/connect2id/nimbus-jose-jwt/src/master/

## Testing
Everything should be covered by unit tests

## Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10647


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKVerification.kt
```diff
@@ -0,0 +1,80 @@
...
+import org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPublicKey
+import timber.log.Timber
+import java.security.PublicKey
+import java.security.interfaces.RSAPublicKey
+import javax.inject.Inject
+
+class DccJWKVerification @Inject constructor() {
+
+    init {
+        X509CertUtils.setProvider(BouncyCastleProviderSingleton.getInstance())
```
mtwalli, line 27: Can we move it to `SecurityProvider` ? See `DscSignatureValidator` for example for how it is used

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKVerification.kt
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    fun verify(jwt: String, jwkSet: Set<DccJWK>) {
+        // 1. Check for empty jwkSet
+        if (jwkSet.isEmpty()) throw DccTicketingException(JWT_VER_NO_JWKS)
```
mtwalli, line 32: should we create another exception like `DccTicketingServerCertificateCheckException` so we can map it to 
`DccTicketingException` in other places ?

With this we can just change the error code ,which I think can lead to issues , I prefer if create a totally different exception with these error codes of verification what do you think ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKVerification.kt
```diff
@@ -0,0 +1,80 @@
...
+            throw DccTicketingException(JWT_VER_ALG_NOT_SUPPORTED)
+
+        // 3. Extract kid from JWT
+        if (signedJWT.header.keyID.isNullOrEmpty()) throw DccTicketingException(JWT_VER_NO_KID)
+
+        // 4. Check for empty jwkSet
+        if (jwkSet.none { it.kid == signedJWT.header.keyID }) throw DccTicketingException(JWT_VER_NO_JWK_FOR_KID)
+
+        // 5. Filter jwkSet by kid
+        jwkSet.filter { it.kid == signedJWT.header.keyID }.forEach {
```
mtwalli, line 52: can we use `filter`  first  and then if the out put is empty to have one loop only over the jwk?
`none` and `filter` both loop over the list

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKVerification.kt
```diff
@@ -0,0 +1,80 @@
...
+            try {
+                // 6. Verify signature
+                val publicKey = X509CertUtils.parse(it.x5c.first().decodeBase64()?.toByteArray()).publicKey
```
mtwalli, line 55: can `dccJWKConverter.createX509Certificate(jwk = it)`   be used ?

Corona-Warn-App/build.gradle
```diff
@@ -452,6 +452,9 @@ dependencies {
     implementation 'com.networknt:json-schema-validator:1.0.58'
     implementation 'net.swiftzer.semver:semver:1.1.1'
 
+    //JWT
+    implementation 'com.nimbusds:nimbus-jose-jwt:9.15.2'
```
mtwalli, line 456: Please add license ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKVerification.kt
jurajkusnier: `DccTicketingJwtException` has been created

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKVerification.kt
jurajkusnier: `DccJWKConverter` has not been merged yet. I tested the code anyway, but unfortunately, it cannot satisfy the JWT library as it returns the public key in a different format. Eg. exception:

```
java.lang.ClassCastException: class sun.security.ec.ECPublicKeyImpl cannot be cast to class org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPublicKey (sun.security.ec.ECPublicKeyImpl is in module jdk.crypto.ec of loader 'platform'; org.bouncycastle.jcajce.provider.asymmetric.ec.BCECPublicKey is in unnamed module of loader 'app')
```

Corona-Warn-App/build.gradle
jurajkusnier: updated

mtwalli (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccTicketingJwtException.kt
```diff
@@ -0,0 +1,15 @@
+package de.rki.coronawarnapp.dccticketing.core.common
+
+class DccTicketingJwtException(
+    val errorCode: ErrorCode,
+    override val cause: Throwable? = null
+) : Exception(errorCode.message, cause) {
+
+    enum class ErrorCode(val message: String) {
+        JWT_VER_NO_JWKS("JWT_VER_NO_JWKS"),
```
chiljamgossow, line 9: Not sure why we need the same code in the message. I would prefer a short description of the error (as it was before)

jurajkusnier: I'm going to clean up errors in another task https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10648

chiljamgossow (approved):
LGTM

PR #4430, opened 2021-11-24
Loading Event - Consent I (EXPOSUREAPP-10649)

BMItr:
This PR addresses Exposureapp-10649 (Feature: ExposureApp-10194 (Ticketvalidation)).

**Contains**: 
Entry for the UI / LoadingButton Event Handling.

**How to test:**
- Use cwa cli command: cwa ti gen -o
- Scan QR Code
- Navigate to the Consent 1 Screen
- Press "Einverstanden" on the Consent Screen
(Loading Animation should appear, followed by succesful loading or failure dialog as the Ticketvalidation-Story is in dev state)

Press it
image


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/consent/one/DccTicketingConsentOneViewModel.kt
```diff
@@ -1,31 +1,138 @@
...
+
+    fun onUserConsent(): Unit = launch {
+        Timber.d("onUserConsent()")
+        currentIsLoading.compareAndSet(expect = false, update = true)
+        processUserConsent()
+        currentIsLoading.compareAndSet(expect = true, update = false)
+    }
+
+    private suspend fun processUserConsent() = try {
+        Timber.d("processUserConsent()")
```
chiljamgossow, line 56: Can you move all logic to a separate class, so we can easily setup integration tests without the UI?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/consent/one/DccTicketingConsentOneViewModel.kt
```diff
@@ -1,31 +1,138 @@
...
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModelFactory
 import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.map
+import timber.log.Timber
 
 class DccTicketingConsentOneViewModel @AssistedInject constructor(
-    @Assisted private val dccTicketingTransactionContext: DccTicketingTransactionContext,
-    dispatcherProvider: DispatcherProvider
+    @Assisted private val dccTicketingSharedViewModel: DccTicketingSharedViewModel,
```
mtwalli, line 23: Why passing `DccTicketingSharedViewModel` ?  Shared ViewModel is ONLY used to pass data from the main graph to sub-graph due to limitation of passing pracelable when navigating by URI. It is better just to pass the required `data` object. Let us say we found a way to pass the `data` using `args` this approach would be obsolete , but since `DccTicketingConsentOneViewModel` is using `DccTicketingSharedViewModel` it won't be straight foreward. Please keep it the old way and pass only the `data` object from the fragment

mtwalli: Another question why do you need  DccTicketingSharedViewModel , don't we have one already to share these data ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionFragment.kt
```diff
@@ -5,20 +5,22 @@ import androidx.fragment.app.Fragment
...
-import androidx.navigation.fragment.navArgs
+import androidx.navigation.navGraphViewModels
 import de.rki.coronawarnapp.R
+import de.rki.coronawarnapp.dccticketing.ui.shared.DccTicketingSharedViewModel
 
 class DccTicketingCertificateSelectionFragment : Fragment() {
 
-    private val args by navArgs<DccTicketingCertificateSelectionFragmentArgs>()
+    private val dccTicketingSharedViewModel:
+        DccTicketingSharedViewModel by navGraphViewModels(R.id.dcc_ticketing_nav_graph)
```
mtwalli, line 15: What DccTicketingSharedViewModel is doing ? `QrcodeSharedViewModel` is sharing the data already and it should be passed to `DccTicketingConsentOneViewModel` .

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccTicketingException.kt
```diff
@@ -24,86 +30,90 @@ class DccTicketingException(
...
+    private fun ErrorCode.errorMessageRes(serviceProvider: String): LazyString = when (textKey) {
+        TextKey.SERVICE_PROVIDER_ERROR_NO_NAME ->
+            R.string.dcc_ticketing_error_service_provider_error_no_name.toResolvingString()
+        TextKey.SERVICE_PROVIDER_ERROR ->
+            R.string.dcc_ticketing_error_service_provider_error.toResolvingString(
+                serviceProvider
+            )
+        TextKey.TRY_AGAIN ->
+            R.string.dcc_ticketing_error_try_again.toResolvingString()
+    }
```
mtwalli, line 107: @jurajkusnier  is this something you already doing?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/consent/one/DccTicketingConsentOneViewModel.kt
```diff
@@ -1,31 +1,138 @@
...
+                .requestAccessToken()
+                .also { updateTransactionContext(ctx = it) }
+        }
+        postEvent(NavigateToCertificateSelection)
+    } catch (e: DccTicketingException) {
+        Timber.d(e, "Error while processing user consent")
+        val serviceProvider = currentUiState.first().provider
+        val lazyErrorMessage = e.errorMessage(serviceProvider = serviceProvider)
+        postEvent(ShowErrorDialog(lazyErrorMessage = lazyErrorMessage))
+    }
```
mtwalli, line 70: I think we should handle `Exception` too, for example keyPair Generation can throw an exception in this case it is not DccTicketingException and it will crash the App. Better catch it and display error message

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccTicketingException.kt
jurajkusnier: yes, I put this my task as obsolete as this solution looks good enough

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/consent/one/DccTicketingConsentOneViewModel.kt
BMItr: adjusted

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionFragment.kt
BMItr: "What DccTicketingSharedViewModel is doing?"
image
(description)

It is used to share the current TransactionContext **between all DccTicketing Fragments**. This approach is easier & more comfortable than using Fragment Args.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/consent/one/DccTicketingConsentOneViewModel.kt
```diff
@@ -1,31 +1,81 @@
...
+            }
+            NavigateToCertificateSelection
+        } catch (e: Exception) {
+            Timber.e(e, "Error while processing user consent")
+            val lazyErrorMessage = when (e) {
+                is DccTicketingException -> {
+                    val serviceProvider = currentUiState.first().provider
+                    e.errorMessage(serviceProvider = serviceProvider)
+                }
+                else -> R.string.errors_generic_text_unknown_error_cause.toResolvingString()
```
chiljamgossow, line 64: Are the null checks going to be an unknown error?

BMItr: yes. They should only appear when a dev has done something "wrong"

BMItr: Like when someone forgot to set a value in the step before

mtwalli (approved):

jurajkusnier (approved):

PR #4438, opened 2021-11-25
Fix lists displaying old data, due to payload race-condition, if redraw is delayed (COMMUNITY)

d4rken:
### TL;DR

This PR fixes a UI race-condition that can cause lists to display outdated values until the next list update.

Cause:
If multiple list data updates are issued quickly, and the device is slow to redraw, the `DiffUtil` mechanism can end up issuing multiple payloads. Payload selection uses `singleOrNull() ?: oldData` causing old data to be displayed if multiple payloads are issued by `DiffUtil`.

Solution:
Because we don't do partial updates of single viewholders, and we always get the full data item as payload, we can fix this by just using `lastOrNull()`.

#### Bonus :cake: 

This also adds a default implementation to `HasPayloadDiffer`, because 41 classes implemented the same `diffPayload` code:

```Kotlin
interface HasPayloadDiffer {
    fun diffPayload(old: Any, new: Any): Any? = if (old::class == new::class) new else null
}
```

Only `DiaryLocationListItem` and `DiaryPersonListItem` have a custom `diffPayload` implementation and need to override it.

### Not long enough, want to read more

As is customary, all good CWA stories start with an innocent comment from @vaubaehn :beers::

> Hi Matthias, today on the first opening, my CWA's statistics bar did not refresh completely. Some (randomly mixed) values were old, some were the latest. After closing and restarting, all values were refreshed and latest. For my understanding, all values are retrieved via app config. So, when app config has already the latest numbers, what could be the reason for the partly refresh? Could it be a problem with flow?

The stats data is emitted as is, and not individually, so it's unlikely to be `Flow` related.
The statistics data on the home screen is a nested RecyclerView with both adapters using `DiffUtil` to prevent flickering and allow for efficient UI updates.
My first thought was issues with `stableId` but that looks correct for the statistic card items.

We don't do partial redraws of individual viewholders, [so in 50 places we just want the latest data item using](https://github.com/corona-warn-app/cwa-app-android/blob/e5d162ddf800d164e4de1078954eab444a4376fb/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/cards/SevenDayRValueCard.kt#L36):

```Kotlin
val curItem = payloads.filterIsInstance<LocalStatisticsCardItem>().singleOrNull() ?: item
```

The documentation for [`onBindViewHolder`](https://developer.android.com/reference/androidx/recyclerview/widget/RecyclerView.Adapter#onBindViewHolder(VH,%20int,%20java.util.List%3Cjava.lang.Object%3E)) states:

> The payloads parameter is a merge list from notifyItemChanged(int, Object) or notifyItemRangeChanged(int, int, Object). 

So if quick updates happen without redraw, `singleOrNull` will return null and by `?: item` we use the original (stale) data.

Statistics also have a good chance for rapid updates, [see here](https://github.com/corona-warn-app/cwa-app-android/blob/e5d162ddf800d164e4de1078954eab444a4376fb/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt#L149-L153)

Solution seems straightforward with little downside:
The payloads itself are [an ordered list](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:recyclerview/recyclerview/src/main/java/androidx/recyclerview/widget/RecyclerView.java;l=11489?q=RecyclerView) so we can assume that the latest payload is always the last and fix this race condition by using `lastOrNull` instead of `singleOrNull`.

### Testing
Play with lists that update quickly, especially nested lists such as the statistics.


chiljamgossow (approved):
Tested and looked fine to me
Thanks for refactoring

mtwalli (approved):
LGTM, Next PR we should migrate to Compose `LazyColumn` 😄

BMItr (commented):
@d4rken first off all my best greetings to you and yourself.
I have to apologize for out late response, as we had to many urgents, features, hot topics and so on :rabbit: 

We highly appreciate your commitment and its really helpful!
If you dont mind, I would recommend to use an extension fun basically replacing all the
 `filterIsInstance<Item>().lastOrNull() ?: item` statements

it could look like _lastInstanceOrDefault(item)_ or similar 

Feel free to change it as you have your own mind ;-) (optional)

Best! :smile_cat:

d4rken:
@mtwalli @chiljamgossow Thanks for your reviews.

@BMItr 

Unless it would be a blocker for you I would have preferred it to stay as is. Motivation declines with each day a PR stays open :wink: 

> If you dont mind, I would recommend to use an extension fun basically replacing all the
filterIsInstance<Item>().lastOrNull() ?: item statements

Replacing the `filterIsInstance` with an extension function would not reduce duplicate code by much imho. In it's current form it is a known set of Kotlin extension functions that most devs should be able to understand. The tradeof here would be "multiple known functions vs single unknown function", you could argue for both.

A cleaner solution that reduces the duplicate code here could be to change the ViewHolder to support alternative `val onBindData` implementations, such that most VH could switch to an implementation that hides the item+payload mechanism and just returns the latest item.

That would probably have caused this PR to stay open even longer though and I'd leave that for someone else :sweat_smile:.

PR #4484, opened 2021-12-06
Trigger value set download (EXPOSUREAPP-10903)

chiljamgossow:
Moves the trigger to download value sets to the repository as the overview screen may be skipped using the universal scanner
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10903

@mlenkeit Can you confirm that this is fine with you as it is a change from the tech spec that requested the check to be on the certificate overview?


mlenkeit:
@chiljamgossow yes, I think that's fine 👍

mtwalli (approved):
LGTM, Tested on S8

BMItr (approved):
lgtm

PR #4478, opened 2021-12-06
Filter jwk against allowlist (EXPOSUREAPP-10572)

mtwalli:
### Changes
1- Filter `JWK`s against `allowList`
2- Filter validation service in request `decorator` by `allowlist`

### Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10572

PR #4486, opened 2021-12-06
Check service provider (EXPOSUREAPP-10865)

chiljamgossow:
after https://github.com/corona-warn-app/cwa-app-android/pull/4478


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeHandler.kt
```diff
@@ -34,6 +47,12 @@ class DccTicketingQrCodeHandler @Inject constructor(
         )
     }
 
+    private fun Set<DccTicketingServiceProviderAllowListEntry>.validateServiceIdentity(serviceIdentity: String) {
+        if (!qrCodeSettings.checkServiceIdentity.value) return
+        val hash = serviceIdentity.toSHA256().encodeUtf8()
```
mtwalli, line 52: `enodeUtf8()` isn't working as expected , please use `serviceIdentity.encode().sha256()`?
```suggestion
        val hash = serviceIdentity.encode().sha256()
```

Currently  the comparison is failing 
```diff
- serviceIdentityHash=[text=8646c015ffec92cfec73220f602c4892267c054f0eca4aff4d1b40238d394561]
+ ServiceProviderAllowList=[DccTicketingServiceProviderAllowListEntry(serviceIdentityHash=[hex=8646c015ffec92cfec73220f602c4892267c054f0eca4aff4d1b40238d394561])]
find=null
```
hashes are the same ,but the key is different `text=..` ,`hash=..`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeHandler.kt
```diff
@@ -34,6 +47,12 @@ class DccTicketingQrCodeHandler @Inject constructor(
...
+        find { it.serviceIdentityHash == hash } ?: throw DccTicketingException(SP_ALLOWLIST_NO_MATCH)
```
mtwalli, line 53: Please some logs would be helpful ?

mtwalli (dismissed):
LGTM, Tested with my suggestion and works fine

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/qrcode/DccTicketingQrCodeHandler.kt
chiljamgossow: now it is chatty

mtwalli (approved):
LGTM

jurajkusnier (approved):
LGTM

PR #4491, opened 2021-12-07
Check in for G events person overview UI changes (EXPOSUREAPP-10888)

SamuraiKek:
https://user-images.githubusercontent.com/19816790/145071143-6323393a-dda0-4e15-8d58-b877c4ac594f.mp4

EDIT: Just realized the inactive button should have black text. Currently working on it.
EDIT 2: Fixed the inactive button color.

How to test:
- for 2G certificate, scan a  complete vc or rc:
`cwa vc gen --pii 42 -s v.0.dn=2 v.0.sd=2 v.0.mp='EU/1/20/1528' v.0.dt='2021-11-04'`
-for 2G+ certificate, scan a tc after scanning a vc for the same person:
`cwa tc gen --pii 42`
- for 3G certificate, scan a RAT certificate
- for 3G+ certificate, scan a PCR certificate
- scanning an expired test certificate will display the test as valid in the overview screen
`cwa tc gen --exp 1minute`
- scanning an expired vc will display the vs as invalid
`cwa vc gen --exp 1minute`


vaubaehn:
Nice UI 👏 ❤️

KathrinLuecke (approved):
UA approved

LukasLechnerDev:
Some font colors are white in dark mode: 

[screenshot](https://user-images.githubusercontent.com/10398034/145195740-65937492-1423-4994-8138-2e4b0ec1946f.png)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -130,11 +136,106 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
...
+        }
+        statusText.isVisible = valid
+        statusBadge.isVisible = valid
+        statusBadge.text = context.getString(statusBadgeText)
+        covpassInfoTitle.isVisible = valid
+        covpassInfoButton.isVisible = valid
+        covpassInfoButton.setOnClickListener { onCovPassInfoAction() }
+        invalidOverlay.isGone = valid
+        image.isEnabled = valid
+        certificateToggleGroup.isVisible = isCombinedCertificate && valid
```
LukasLechnerDev, line 174: The admission State `Other` is the only one that can contain invalid certificates (e.g. an outdated RAT test). Therefore, we don't need the `&& valid` here.

SamuraiKek:
> Some font colors are white in dark mode:
> 
> [screenshot](https://user-images.githubusercontent.com/10398034/145195740-65937492-1423-4994-8138-2e4b0ec1946f.png)

Ah, it's the new background I added, I need to add a dark one as well.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/items/PersonCertificateCard.kt
```diff
@@ -25,48 +25,88 @@ class PersonCertificateCard(parent: ViewGroup) :
...
+                    secondaryCertificate = curItem.admissionState.testCertificate,
+                    colorShade = curItem.colorShade,
+                    isCombinedCertificate = true,
+                    statusBadgeText = R.string.confirmed_status_2g_plus_badge,
+                    badgeCount = curItem.badgeCount,
+                    onCovPassInfoAction = curItem.onCovPassInfoAction
+                )
+            }
+
+            is TwoGPlusRAT -> {
```
LukasLechnerDev, line 57: This calls the same arguments as the block above. We can do 
```kotlin
is TwoGPlusPCR, is TwoGPlusRAT -> { ...}
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Yep, you're right. I think  i can ditch the `isCombinedCertificate` boolean alltogether and just check if `secondaryCertificate` is not null instead.

chiljamgossow (commented):
> Looks weird in dark mode, can you check?

Yes, I forgot to add a dark background for dark mode.
EDIT: Sorry, I edited this comment instead of replying and deleted the photo 😄

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/items/PersonCertificateCard.kt
SamuraiKek: I tried that and but then I get Unresolved reference for twoGCertificate & testCertificate.

Corona-Warn-App/src/main/res/layout/include_certificate_overview_qr_card.xml
```diff
@@ -0,0 +1,150 @@
...
+    android:layout_height="wrap_content"
+    android:layout_marginHorizontal="24dp"
+    android:padding="13dp">
+
+    <TextView
+        android:id="@+id/status_text"
+        style="@style/subtitleBoldSixteen"
+        android:layout_width="wrap_content"
+        android:layout_height="0dp"
+
```
LukasLechnerDev, line 16: Empty line ;)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/items/PersonCertificateCard.kt
LukasLechnerDev: Ah I see, the smart cast isn't working in this case anymore. Then we should leave it as is.

SamuraiKek:
Darkmode is now fixed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -130,11 +136,107 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
...
+@Suppress("LongParameterList", "ComplexMethod")
+fun PersonOverviewItemBinding.setUIState(
+    primaryCertificate: CwaCovidCertificate,
+    secondaryCertificate: CwaCovidCertificate? = null,
+    colorShade: PersonColorShade,
+    statusBadgeText: Int = 0,
+    badgeCount: Int = 0,
+    onCovPassInfoAction: () -> Unit
+) {
+    val valid = primaryCertificate.isValid
```
LukasLechnerDev, line 150: `isValid` doesn't return `true` for Tests with state `Expiring_soon`, see https://github.com/corona-warn-app/cwa-app-android/pull/4493#discussion_r764766447. (mentioned by @mtwalli )

Therefore, tests that are expiring soon would be displayed as invalid...

I think we have this bug in other places as well, e.g. in `PersonDetailsViewModel`...

Corona-Warn-App/src/androidTest/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewFragmentTest.kt
```diff
@@ -93,6 +96,20 @@ class PersonOverviewFragmentTest : BaseUITest() {
...
+    @Screenshot
+    fun capture_fragment_two_g_plus_certificate() {
+        every { viewModel.personCertificates } returns MutableLiveData(twoGPlusCertificate())
+        takeSelfie("two_g_plus")
+    }
+
+    @Test
+    @Screenshot
+    fun capture_fragment_two_g_plus_certificate_with_badge() {
+        every { viewModel.personCertificates } returns MutableLiveData(twoGPlusCertificate())
```
LukasLechnerDev, line 109: `twoGPlusCertificateWithBadge()` must be returned here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -130,11 +136,107 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
...
+            expirationStatusIcon.setImageDrawable(context.getDrawableCompat(R.drawable.ic_error_outline))
+            expirationStatusText.isVisible = badgeCount == 0
+            expirationStatusText.text = context.getText(R.string.certificate_qr_invalid_signature)
+        }
+        is CwaCovidCertificate.State.Blocked -> {
+            expirationStatusIcon.isVisible = badgeCount == 0
+            (expirationStatusIcon.layoutParams as ConstraintLayout.LayoutParams).verticalBias = 0f
+            expirationStatusIcon.setImageDrawable(context.getDrawableCompat(R.drawable.ic_error_outline))
+            expirationStatusText.isVisible = badgeCount == 0
+            expirationStatusText.text = context.getText(R.string.error_dcc_in_blocklist_title)
```
mtwalli, line 224: The code slightly changes between these three states, maybe extract to a function and parameterize what changes ?
such as:
```kotlin
fun Binding.updateExpirationViews(verticalBias:Float=0f@StringRes statusTextRes:Int){
    expirationStatusIcon.isVisible = badgeCount == 0
    (expirationStatusIcon.layoutParams as ConstraintLayout.LayoutParams).verticalBias = verticalBias
    expirationStatusIcon.setImageDrawable(context.getDrawableCompat(R.drawable.ic_error_outline))
    expirationStatusText.isVisible = badgeCount == 0
    expirationStatusText.text = context.getText(statusTextRes)
}
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -130,11 +136,107 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
...
+                    image.loadAny(secondaryCertificate?.getValidQrCode(Locale.getDefault().language)) {
+                        crossfade(true)
+                        loadingView(qrCodeCard.image, qrCodeCard.progressBar)
+                    }
+                    primaryCertificateButton.typeface = Typeface.DEFAULT
+                    secondaryCertificateButton.typeface = Typeface.DEFAULT_BOLD
+                }
+            }
+        }
+    }
```
mtwalli, line 198: Please crack it into small chunks to make  it readable ? maybe related view logic in one function each ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Actually `isValid` returns `true` for expired/expiring soon test certificates. The only time it returns false for a test certificate is when the state of the certificate is `State.Invalid` (I know, it's confusing). For vaccination and recovery certificates we don't make this exception & I don't remember why we need to display expired test certificates as valid but we had a whole discussion on this when we first introduced the certificates screen.
I remember writing that `isValid` value in the `CwaCovidCertificate` interface with @mtwalli in a pair programming session when we first introduced the certificates screen.

I don't think it should change for the UI here. 
@mtwalli's comment on your PR made sense because we need to have only the actual valid certificates to calculate the admission state, but for the UI, we need to display expired/expiring soon test certificates as valid. 

@mtwalli, you were there with me when we implemented this feature, please give your input here. Am I making sense? 😰

mtwalli: @SamuraiKek  with this change , how expired TC is suppose to be displayed ?

vaubaehn: @SamuraiKek 
I was not involved but I could imagine that `isValid` and `State.Invalid` differ by technical expiration (COSE header timestamps; technical validity) and test expiration (DCC timestamps; effective test validity). It can be useful to keep expired tests (effective validity) as valid (technical validity) as restrictions may differ between EU countries (e.g., RAT test valid 24hrs vs. 48hrs).
Was it this why the destinction was made?

SamuraiKek: > @SamuraiKek with this change , how expired TC is suppose to be displayed ?

An expired TC will still be displayed as valid. 
I haven't changed any logic, I've only introduced a new layout just for the overview screen. Until now, we had the same layout shared between the overview screen QR card and the person details QR card. This new feature that lets you select different certs on the overview screen would have introduced even more bloat to the shared UI, that's why I did the separation.

In terms of behaviour, it's the same as before, only now we have these two additional buttons in the case of 2G/2G+ certs.

SamuraiKek: > @SamuraiKek I was not involved but I could imagine that `isValid` and `State.Invalid` differ by technical expiration (COSE header timestamps; technical validity) and test expiration (DCC timestamps; effective test validity). It can be useful to keep expired tests (effective validity) as valid (technical validity) as restrictions may differ between EU countries (e.g., RAT test valid 24hrs vs. 48hrs). Was it this why the destinction was made?
> 
> Edit: Wasn't `expiring_soon` used with regard to technical validity (to notify user to get a new (vaccination) DCC issued)?

@vaubaehn, this might be the case, and yes, `expiring_soon` is used mostly to notify the user to get a new vaccine.

mtwalli: @SamuraiKek , @LukasLechnerDev , @vaubaehn , we have differentiate between two things:
1- How certificates are displayed in the App?
2- What certificates are eligible for G events? 

Lukas did work on determining the state based on VALID certificates (Valid,Expiring_Soon) which is correct ,since expired TC should not be considered in the admission state.

Alexandru is working on how to display the certificates in the cards (UI) , Expired TC should be displayed as Valid (without the whitish overlay).

I think `isValid` flag was used mainly for this purpose only, I suggest to rename it to `isDisplayValid` 

@SamuraiKek  Please make sure that Expired TC has to displayed as pre this change ?

SamuraiKek: @mtwalli, @LukasLechnerDev, I've renamed `isValid` to `isDisplayValid` and rechecked that the expired test certificates are displayed as valid. All comments should be addressed now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Addressed with latest changes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: Addressed with latest changes

mtwalli:
Should not the second vaccination certificate has a `full` shield icon?
Screenshot 2021-12-13 at 13 37 18

SamuraiKek:
> Should not the second vaccination certificate has a `full` shield icon? Screenshot 2021-12-13 at 13 37 18

There are two versions from what I've seen. One with a full shield when the vaccination is complete and one with this empty shield when the vaccination is complete and 14 days have passed. Not sure if this needs to change or not, but either way it's not part of this 2G events task and should be addressed in a different PR.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -130,11 +137,115 @@ fun IncludeCertificateQrcodeCardBinding.bindValidityViews(
...
+        }
+        covpassInfoTitle.isVisible = valid
+        covpassInfoButton.isVisible = valid
+        covpassInfoButton.setOnClickListener { onCovPassInfoAction() }
+        invalidOverlay.isGone = valid
+        image.isEnabled = valid
+        certificateToggleGroup.isVisible = secondaryCertificate != null
+        primaryCertificateButton.typeface = Typeface.DEFAULT_BOLD
+        secondaryCertificateButton.typeface = Typeface.DEFAULT
+        certificateToggleGroup.setOnCheckedChangeListener { checkedId ->
```
mtwalli, line 179: Currently the switch effect is working upon user click, which means:
1- When app is set in background and fragment is restored again the user selection is gone 
2- In case many people in the list of person scrolling up and down will also reset user selection.

I suggest that you set user selection per person as part of the ViewModel state as long as the viewModel is alive you will be able to reflect that in the `UI`

mtwalli:
> > Should not the second vaccination certificate has a `full` shield icon? Screenshot 2021-12-13 at 13 37 18
> 
> There are two versions from what I've seen. One with a full shield when the vaccination is complete and one with this empty shield when the vaccination is complete and 14 days have passed. Not sure if this needs to change or not, but either way it's not part of this 2G events task and should be addressed in a different PR.

I checked in 2.16 it has the same behavior

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
SamuraiKek: I was wondering if we should save the selection state for each person but the tech spec doesn't require it yet. 
Saving it just for the lifecycle of the ViewModel is and edge case and probably a handful of people will stumble upon this. If you were to switch screens and come back to the certificates list, the state would be lost anyway. And since we are now limiting the persons list to only 20, I doubt scrolling will reset the selection. I've just tested it with 30 certificates and it still retains the selection.

mtwalli (approved):
Tested Looks good

chiljamgossow (approved):

LukasLechnerDev (approved):
Checked again, looks good! Good work!

PR #4493, opened 2021-12-07
Determine admission state (EXPOSUREAPP-10887)

LukasLechnerDev:
`PersonCertificate` now has a new property `admissionState` (2G+RAT, 2G+PCR, 2G, 3GwithRAT, 3GwithPCR, other)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificates.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificates.kt
@@ -15,10 +15,11 @@ data class PersonCertificates(
         certificates.findHighestPriorityCertificate()
     }
 
-    // TODO: would be implemented in admission state logic task
-    // null when certificates list is empty
+    val validCertificates: List<CwaCovidCertificate>
+        get() = certificates.filter { it.isValid }
+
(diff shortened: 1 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificates.kt
```diff
@@ -15,10 +15,11 @@ data class PersonCertificates(
         certificates.findHighestPriorityCertificate()
     }
 
-    // TODO: would be implemented in admission state logic task
-    // null when certificates list is empty
+    val validCertificates: List<CwaCovidCertificate>
+        get() = certificates.filter { it.isValid }
```
mtwalli, line 19: `isValid` is dealing with expired Test Certificate as Valid, I think this is not what is required here, All certs should be either `Valid or Expiring Soon` to be eligible for G rules

LukasLechnerDev: You are absolutely right. Good catch! Adjusted it so that test certificates with state `ExpiringSoon` are also considered for the admission state. Added tests for this too.

jurajkusnier (approved):

mtwalli (approved):

PR #4502, opened 2021-12-09
Restrict number of persons (EXPOSUREAPP-10613)

mtwalli:
## Test against LOCAL mock server (`Max=5 ,threshold=2`)
- Scan certificates for different persons
- When threshold is `numberOfDistinctPersons + 1` > `threshold` -> warning should appear with every new person added 
   - Certificate is still imported 
- When `numberOfDistinctPersons` >=  `Max ` Block dialog should appear 
   - Certificate is NOT imported 

### Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10613

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/CWAConfig.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/CWAConfig.kt
@@ -16,5 +16,9 @@ interface CWAConfig {
 
     val validationServiceMinVersion: Int
 
+    val dccPersonWarnThreshold: Int
+
+    val dccPersonCountMax: Int
+
     interface Mapper : ConfigMapper<CWAConfig>
 }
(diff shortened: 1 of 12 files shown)


Ein-Tim:
Hi @mtwalli 

Could you elaborate a bit what is planned here?

Thank you!

mtwalli:
@Ein-Tim  The idea is to restrict the number of persons to whom certificates can imported in the App. To avoid importing certificates when it is not intended (such as Validating certificates using CWA) in this case user will be warned to use CovPassCheck for that and if the number exceeds the maximum allowance importing new certificates will be blocked

Ein-Tim:
@mtwalli

Thanks for the answer! This makes sense!

AlexanderAlferov (approved):

chiljamgossow (approved):

KathrinLuecke (commented):
UA approved

chstdu:
It would be even better than fixed thresholds to use an additional "Do Not Import" flag in the QR code to prevent accidentally importing certificates "again". This was already proposed on the CWA wishlist months ago: https://github.com/corona-warn-app/cwa-wishlist/issues/601

Ein-Tim:
@mtwalli JFYI: I'm still waiting for it :D

PR #4505, opened 2021-12-09
Max person checker (EXPOREAPP-10613) 

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerViewModel.kt
```diff
@@ -147,7 +149,13 @@ class QrCodeScannerViewModel @AssistedInject constructor(
...
     private suspend fun onDccQrCode(dccQrCode: DccQrCode) {
         Timber.tag(TAG).d("onDccQrCode()")
-        // TODO: Check number of persons
+
+        when (dccMaxPersonChecker.checkForMaxPersons(dccQrCode)) {
+            DccMaxPersonChecker.Result.PASSED -> { /*continue*/
+            }
+            DccMaxPersonChecker.Result.EXCEEDS_THRESHOLD -> TODO()
+            DccMaxPersonChecker.Result.EXCEEDS_MAX -> TODO()
+        }
```
mtwalli, line 158: I think we should do the check after 
         `val recycledContainerId = recycledCertificatesProvider.findCertificate(dccQrCode.qrCode)` otherwise it will prevent the restoring from scanner

mtwalli (approved):
LGTM

PR #4530, opened 2021-12-13
Fix: Ticketvalidation - missing cancelation popup on screen (EXPOSUREAPP-10914)

jurajkusnier:
### Testing
- dialog should appear only when at least one valid certificate is in the list
device-2021-12-13-112322

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10914

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionViewModel.kt
@@ -30,7 +30,7 @@ class DccTicketingCertificateSelectionViewModel @AssistedInject constructor(
     @Assisted private val dccTicketingSharedViewModel: DccTicketingSharedViewModel,
 ) : CWAViewModel(dispatcherProvider) {
 
-    var hasValidCertificate = false
+    private var hasValidCertificate = false
     val events = SingleLiveEvent<DccTicketingCertificateSelectionEvents>()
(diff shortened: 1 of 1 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionViewModel.kt
```diff
@@ -72,7 +74,11 @@ class DccTicketingCertificateSelectionViewModel @AssistedInject constructor(
         }
 
     fun onUserCancel() {
-        events.postValue(ShowCancelConfirmationDialog)
+        if (hasValidCertificate) {
```
mtwalli, line 77: `items.value?.isNotEmpty()` ? and no need for `hasValidCertificate`

jurajkusnier: I'm not checking `items` but `certificates`

mtwalli: Ah Ok, please make hasValidCertificate `private` ? it is only used in the VM

jurajkusnier: updated

mtwalli (approved):

BMItr (approved):

PR #4531, opened 2021-12-13
Delete duplicate ids (DEV)

chiljamgossow:
For discussion:

We allow the scanning of certificates with the same identifier, but the internal logic relies on identifiers to be unique. So once they are in the app, updates, including deletion, are not possible any more.


mtwalli:
I thought you had the same issue as 
[EXPOSUREAPP-11050](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11050).

I tested this PR fixes it , one certificate details was empty (just labels)

chiljamgossow:
I think @mtwalli PR is addressing the issue, then this can be closed

PR #4540, opened 2021-12-13
Increase max length to 255 for check ins (EXPOSUREAPP-10535)

chiljamgossow:
To support Check In QR codes from Luca, the max length for description and address is increased to 255


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionViewModel.kt
```diff
@@ -30,12 +30,14 @@ class DccTicketingCertificateSelectionViewModel @AssistedInject constructor(
     @Assisted private val dccTicketingSharedViewModel: DccTicketingSharedViewModel,
 ) : CWAViewModel(dispatcherProvider) {
 
+    private var hasValidCertificate = false
```
chiljamgossow, line 33: some unmerged changes from 2.15

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/create/TraceLocationCreateViewModel.kt
```diff
@@ -111,7 +111,7 @@ class TraceLocationCreateViewModel @AssistedInject constructor(
         )
     }
 
-    private fun String.isTextFormattedCorrectly() = trim().length in 1..100 && !contains('\n')
+    private fun String.isTextFormattedCorrectly() = trim().length in 1..MAX_LENGTH_DESCRIPTION && !contains('\n')
```
mtwalli, line 114: if we have two different constants, then we need to consider them both while validations. this function is used to validate `description` and `address`. if both constants are not the same the other will fail

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/create/TraceLocationCreateViewModel.kt
```diff
@@ -111,7 +111,7 @@ class TraceLocationCreateViewModel @AssistedInject constructor(
         )
     }
 
-    private fun String.isTextFormattedCorrectly(max: Int) = trim().length in 1..max && !contains('\n')
+    private fun String.isTextFormattedCorrectly(max: Int) = isNotBlank() && length <= max && !contains('\n')
```
mtwalli, line 114: `"            This is not blank               "` would pass ?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/qrcode/ValidQRCodeProvider.kt
```diff
@@ -91,6 +92,25 @@ class ValidQRCodeProvider : ArgumentsProvider {
...
+                        TraceLocationOuterClass.CrowdNotifierData.newBuilder()
+                            .setCryptographicSeed(CRYPTOGRAPHIC_SEED.decodeBase64()!!.toProtoByteString())
+                            .setPublicKey(PUB_KEY.decodeBase64()!!.toProtoByteString())
+                            .setVersion(1)
+                    )
+                    .setVendorData("CAEQARgK".decodeBase64()!!.toProtoByteString())
+                    .setLocationData(
+                        TraceLocationOuterClass.TraceLocation.newBuilder()
+                            .setDescription("Ice Cream Shop with a very long description that has more than one hundred characters as allowed by Luca app.")
+                            .setAddress("Main Street 1")
```
LukasLechnerDev, line 109: Maybe also make the address > 100 characters long

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/create/TraceLocationCreateViewModel.kt
chiljamgossow: yes, why shouldn't it?

LukasLechnerDev:
I was a bit confused because `InvalidQrCodeProvider` has an entry with `.setDescription((0..101).joinToString { "a" })`, which still gets classifyed as `invalid` by `CheckInQrCodeHandlerTest`. This is because `(0..101).joinToString { "a" })` has 304 characters]((0..101).joinToString { "a" } -> [play.kotlinlang.org](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiIvKipcbiAqIFlvdSBjYW4gZWRpdCwgcnVuLCBhbmQgc2hhcmUgdGhpcyBjb2RlLiBcbiAqIHBsYXkua290bGlubGFuZy5vcmcgXG4gKi9cblxuZnVuIG1haW4oKSB7XG4gICAgdmFsIHJlc3VsdCA9ICgwLi4xMDEpLmpvaW5Ub1N0cmluZyB7IFwiYVwiIH1cbiAgICBwcmludGxuKHJlc3VsdC5sZW5ndGgpXG4gICAgcHJpbnRsbihyZXN1bHQpXG59In0=)

Maybe we should make it a bit less confusing by using `a.repeat(256)` as in one of the other tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/create/TraceLocationCreateViewModel.kt
mtwalli: it should not be valid , See [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/main/docs/spec/event-registration-client.md#organizer-tracelocation-creation)

chiljamgossow: blanks are not allowed and the text has max length of 255, anything else goes.
see https://github.com/corona-warn-app/cwa-app-tech-spec/pull/127/files#diff-d50804d49c54cd0ea1d31678c893711ac77c8125fd596f5dcf65e429fb0e7b85

mtwalli: `and may not have more than 255 characters after removing any leading or trailing whitespace` `isNotBlank` is not trimming is not it ?

mtwalli: This applies to Event Creation https://github.com/corona-warn-app/cwa-app-tech-spec/pull/127/files#diff-d50804d49c54cd0ea1d31678c893711ac77c8125fd596f5dcf65e429fb0e7b85R1140

chiljamgossow: The sentence above is less than 255, or what am I missing?

chiljamgossow: I put the trimming in although it is pointless if the UI restricts to 255 characters

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/presencetracing/checkins/qrcode/ValidQRCodeProvider.kt
chiljamgossow: done

chiljamgossow:
> I was a bit confused because `InvalidQrCodeProvider` has an entry with `.setDescription((0..101).joinToString { "a" })`, which still gets classifyed as `invalid` by `CheckInQrCodeHandlerTest`. This is because `(0..101).joinToString { "a" })` has 304 characters]((0..101).joinToString { "a" } -> [play.kotlinlang.org](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMS42LjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiIvKipcbiAqIFlvdSBjYW4gZWRpdCwgcnVuLCBhbmQgc2hhcmUgdGhpcyBjb2RlLiBcbiAqIHBsYXkua290bGlubGFuZy5vcmcgXG4gKi9cblxuZnVuIG1haW4oKSB7XG4gICAgdmFsIHJlc3VsdCA9ICgwLi4xMDEpLmpvaW5Ub1N0cmluZyB7IFwiYVwiIH1cbiAgICBwcmludGxuKHJlc3VsdC5sZW5ndGgpXG4gICAgcHJpbnRsbihyZXN1bHQpXG59In0=)
> 
> Maybe we should make it a bit less confusing by using `a.repeat(256)` as in one of the other tests.

I adjusted the test

mtwalli (approved):
Thanks

LukasLechnerDev (approved):
LGTM

Issue #4200, opened 2021-10-12
Vaccination certificate confusing color use


MikeMcC399:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [ ] Bug is specific for Android only - TBD
* [ ] Bug is not already reported in another issue
Issue is mentioned also in https://github.com/corona-warn-app/cwa-app-android/issues/3927#issuecomment-940199373

## Technical details

- Device name: Samsung Galaxy A5 (2017) SM-A520F
- Android version: 8.0.0
- First reported on App version: 2.12.0-RC0
- Also present on App version: 2.14.1

## Describe the bug

In the certificate overview the color blue is used to denote the currently used certificate. Other certificates are grey.

In the certificate detailed view the color blue is used to denote full vaccination series, whereas grey means incomplete vaccination series.

Overview blue means selected. Detailed view blue means complete. Color is therefore used with different meanings in different, but closely related contexts. The use of color is not documented and because of its different meanings it is also not intuitive.

## Steps to reproduce the issue

Scan in 1/2 and 2/2 certificates and experiment with how they are shown in the overview and detailed view.

### Overview

Incomplete | Complete
--- | ---
![Vaccert12overview](https://user-images.githubusercontent.com/66998419/136908450-e30e4ddd-f283-4959-956a-e18507569716.jpg) | ![Vaccert22overview](https://user-images.githubusercontent.com/66998419/136908488-6b123f54-d7bb-4e6f-8209-a5a6ec222123.jpg)

### Detailed View

Incomplete | Complete
--- | ---
![Vaccert12detail](https://user-images.githubusercontent.com/66998419/136908231-9ffca4a7-40f5-4ef7-9194-68efb04619ed.jpg) | ![Vaccert22detail](https://user-images.githubusercontent.com/66998419/136908278-693afdcc-6064-46bc-985a-917db0b63dd3.jpg)

## Expected behavior

The use of color and symbols for vaccination certificates should be consistent and intuitively understandable.

## Possible Fix

Use a different color for a selected incomplete (1/2) vaccination certificate in the overview. This is the shield symbol half filled. (The shield symbol with a tick is used for a complete vaccination certificate.)

This is a screenshot of the currently displayed color:

![Vaccert12overview_selected](https://user-images.githubusercontent.com/66998419/136909386-25c45a48-8526-46bd-aa94-ca8f15034cff.jpg)

## Additional information

[...]

PR #4544, opened 2021-12-14
FIX: DCC parent color shade (EXPOSUREAPP-9950)

AlexanderAlferov:
## Whats new
DCC details fragment inherits color shades from icon in Person Detail Fragment.
Bookmark icon now has no dark mode and has 3 blue variants. 

## Ticket
[Jira](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9950)

## Testing
Scan different DCC (tests, vaccination, recovery) for several persons.
Check the background of DCC details fragment:
1) Navigating from notification and Universal Scanner background should be changed according to old logic
2) Navigating from Person Detail fragment background should inherit color from icon of DCC in Person Detail fragment.
3) Bookmark icons now should be same color as icon of DCC.


SamuraiKek (commented):
LGTM, but I would like to merge #4469 first as we would have conflicts for the 2G task if we merge this PR first.

SamuraiKek (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonColorShade.kt
```diff
@@ -8,31 +8,43 @@ enum class PersonColorShade(
...
         R.drawable.ic_admission_badge_3,
     ),
     COLOR_INVALID(
         R.color.starsColorInvalid,
         R.drawable.bg_person_overview_invalid,
         R.drawable.bg_certificate_grey,
+        R.drawable.ic_bookmark,
+        R.drawable.ic_admission_badge_1,
+    ),
+    COLOR_UNDEFINED(
```
chiljamgossow, line 43: For which situation do we need undefined?

chiljamgossow (commented):
Tests with device look fine.
Little confused about the code. Maybe some comments?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonColorShade.kt
AlexanderAlferov: As a default value: because we use the same fragment when navigating from UQS. Not using this COLOR_UNDEFINED would lead to additional navigation flag (because we need to define if colour was defined or not) and I don't want use it.

chiljamgossow: Could you also use null instead of undefined? If I see it correctly, the attributes of undefined are never used.

chiljamgossow (approved):

MikeMcC399:
- Mentioned in https://github.com/corona-warn-app/cwa-app-android/issues/4200

PR #4549, opened 2021-12-14
Fix EN link (EXPOSUREAPP-10927)

mtwalli:
### Ticket 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10927


marcauberer (approved):
And again ;)

jurajkusnier:
I don't think we can fix it this way, because it could be overwritten by translation delivery again.
Maybe contact UA and ask them to correct their translations, then it could be fixed in the next translation delivery.

mtwalli:
> I don't think we can fix it this way, because it could be overwritten by translation delivery again.
> 
> Maybe contact UA and ask them to correct their translations, then it could be fixed in the next translation delivery. 

the german translation is right though ,but other languages are not, so I don't see any reason why it wasn't in the first place

chiljamgossow (approved):

PR #4550, opened 2021-12-14
Advanced root warning (EXPOSUREAPP-10698)

BMItr:
This PR adresses Expsoureapp-10698.

_As a CWA user using a rooted device I want to be able to hide the message about root detection to the next CWA upgrade, so that I don't get bothered every time by this message_

How to test:
- Take a rooted device
- check the message on app start
- disable the message by checkbox
- install app update

toDos:
- [x]  clean source
- [ ] add relevant testcases (maybe in follow up PR)

Screenshot:
image


MyIgel:
Is there a reason why the option is not "hide it indefinitely"/forever? I mean, using Magisk Hide to never show the message is still a feasible workaround but its will be still annoying anyway.

KathrinLuecke (dismissed):
UA approved

mlenkeit:
> Is there a reason why the option is not "hide it indefinitely"/forever? I mean, using Magisk Hide to never show the message is still a feasible workaround but its will be still annoying anyway.

@MyIgel it's a compromise between continuing to remind users whose device was rooted without malicious intent and without them really knowing about it on the hand, and sparing tech-savvy users from seeing this all the time on the other hand.

jurajkusnier (dismissed):
LGTM, works as expected

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/launcher/LauncherActivityViewModel.kt
```diff
@@ -83,9 +83,11 @@ class LauncherActivityViewModel @AssistedInject constructor(
 
     private fun checkForRoot() = launch {
         Timber.tag(TAG).d("checkForRoot()")
-        when (rootDetectionCheck.isRooted()) {
-            true -> events.postValue(LauncherEvent.ShowRootedDialog)
-            false -> checkForUpdate()
+        with(rootDetectionCheck) {
+            when (shouldShowRootInfo && isRooted()) {
```
LukasLechnerDev, line 87: Minor: IMO it is nicer to have the `isRooted()` check within the `shouldShowRootInfo` property. Then the `ViewModel` only needs to ask the `RootDetectionChecker` if it should show the info and doesn't need to additionally check the root status itself.

However we can also leave it as is due to time pressure.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/launcher/LauncherActivityViewModelTest.kt
```diff
@@ -249,6 +251,47 @@ class LauncherActivityViewModelTest : BaseTest() {
...
+        coEvery { rootDetectionCheck.shouldShowRootInfo } returns false
+
+        coEvery { updateChecker.checkForUpdate() } returns UpdateChecker.Result(isUpdateNeeded = true)
+        coEvery { appUpdateManager.getUpdateInfo() } returns
+            mockk<AppUpdateInfo>().apply {
+                every { updateAvailability() } returns UpdateAvailability.UPDATE_AVAILABLE
+            }
+
+        createViewModel().run {
+            onRootedDialogDismiss()
```
LukasLechnerDev, line 286: I think we shouldn't call this method, as it also won't be called in production in this case. The test case is still successful after deleting it.

LukasLechnerDev (dismissed):
LGTM. The tests on my device were successfull as well.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/launcher/LauncherActivityViewModelTest.kt
BMItr: dne

jurajkusnier (approved):

LukasLechnerDev (approved):

PR #4555, opened 2021-12-15
Fix: Unify allow list error message (DEV)

mtwalli:
Against mock cloud, scan:
`cwa ti gen -e ngrok --preset SP_ALLOWLIST_NO_MATCH -o`
Screenshot 2021-12-15 at 11 23 56

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/allowlist/internal/DccTicketingAllowListException.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/allowlist/internal/DccTicketingAllowListException.kt
@@ -21,12 +21,13 @@ class DccTicketingAllowListException(
         CLIENT_ERROR("Update of Allow list failed with client error."),
 
         ALLOWLIST_NO_MATCH("AllowList no match."),
+        SP_ALLOWLIST_NO_MATCH("Service Provider allowlist mismatch "),
     }
 
     override fun toHumanReadableError(context: Context): HumanReadableError {
         return HumanReadableError(
             title = TITLE_ID.toResolvingString().get(context),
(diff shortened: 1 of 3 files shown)


BMItr (approved):
lgtm

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/allowlist/internal/DccTicketingAllowListException.kt
```diff
@@ -21,12 +21,13 @@ class DccTicketingAllowListException(
...
         ALLOWLIST_NO_MATCH("AllowList no match."),
+        SP_ALLOWLIST_NO_MATCH("Service Provider allowlist mismatch "),
     }
 
     override fun toHumanReadableError(context: Context): HumanReadableError {
         return HumanReadableError(
             title = TITLE_ID.toResolvingString().get(context),
-            description = MESSAGE_ID.toResolvingString().get(context)
+            description = MESSAGE_ID.toResolvingString().get(context) + " ($errorCode)"
         )
```
chiljamgossow, line 31: What happens to CLIENT_ERROR?

chiljamgossow: Did you fix two bugs here? Wrong message and the missing append of error code?

mtwalli: they are not reported , the allowlist is simply empty , so technically it will result in ALLOWLIST_MISMATCH

AlexanderAlferov (approved):
🚀

PR #4564, opened 2021-12-15
Fix duplicate certificate list item (EXPOSUREAPP-11052)

marcauberer:
Addresses [EXPOSUREAPP-11052](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11052).

This fixes the edge case that a RA test certificate and a PCR test certificate is required, but the certificate type list contains `PCR_TEST`, `RA_TEST` and `TEST`.

The previous result was:
```
PCR-Testzertifikat, Schnelltest-Testzertifikat,
PCR-Testzertifikat, Schnelltest-Testzertifikat
```

The new result should be:
```
PCR-Testzertifikat, Schnelltest-Testzertifikat
```
because `PCR_TEST` and `RA_TEST` are filtered as soon as `TEST` is present.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionHelperFunctions.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionHelperFunctions.kt
@@ -3,26 +3,24 @@ package de.rki.coronawarnapp.dccticketing.ui.certificateselection
 import android.content.Context
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingCertificatesFilterType
-import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingCertificatesFilterType.PCR_TEST
-import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingCertificatesFilterType.RA_TEST
-import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingCertificatesFilterType.RECOVERY
-import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingCertificatesFilterType.TEST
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionHelperFunctions.kt
```diff
@@ -10,19 +10,21 @@ import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingC
 import de.rki.coronawarnapp.dccticketing.core.certificateselection.DccTicketingCertificatesFilterType.VACCINATION
 
 fun Context.certificateTypesText(certificateTypes: List<String>, separator: String = ", "): String =
```
LukasLechnerDev, line 12: Hi Marc! I am Lukas, and I started on Dec 1st in the CWA Android Team 👋.

A unit test for this extension function that fails before this fix and is successful after this fix would be very useful here I think.

mtwalli (dismissed):
Thanks @marcauberer!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionHelperFunctions.kt
mtwalli: Added

mtwalli (approved):

LukasLechnerDev (approved):
LGTM!

marcauberer:
@mtwalli thanks for the tests 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/certificateselection/DccTicketingCertificateSelectionHelperFunctions.kt
marcauberer: Hi @LukasLechnerDev! I am Marc, helping the CWA Android Team here and there with fixing bugs ...

PR #4571, opened 2021-12-16
Wrong vacc icon when eligible for booster (EXPOSUREAPP-10479)

SamuraiKek:
If a user with a complete vaccination scheme recieved the "eligible for booster" notification, the icon on the vaccination cards would no longer show up as a full shield but instead were displaying the outline of the shield. 

How to test:

Scan a 1/1 or 2/2 vaccination that is eligible for booster and look at the vaccination icon on the vaccination card:

`cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=1 v.0.mp='EU/1/20/1525' v.0.dt='2021-10-12' -e wru`


CV113 (approved):
UA approved

BMItr (approved):

LukasLechnerDev (approved):
👍

PR #4574, opened 2021-12-16
Ticketvalidation screens sometimes empty (EXPOSUREAPP-11054)

LukasLechnerDev:
The `transactionContext` property of the ViewModel that is shared within the whole ticketing flow `DccTicketingSharedViewModel` doesn't survive process death, and therefore the testers are sometimes seeing empty screens (certificate selection and result screen) after bringing the app in background and foreground again. 

This PR adds the `savedStateHandle` to this ViewModel and stores the `transactionContext` there. Now the screens get populated correclty again even after process death. 

How to test:
- the cli command `cwa ti gen -e wru -o` currently generates a qr code with: Service Provider Allowlist: FAIL - therefore you have to currently disable the "Check against allow list" switch in the DCC Ticketing Test Menu in ordere to be able to navigate though the ticketing flow
- Go to certificate selection screen / result screen
- Put app in background
- Kill app process (e.g. Android Studio -> Logcat -> Terminate Application
- Bring app in foreground
- UI should be the same as before


chiljamgossow (approved):
Works nicely

mtwalli (approved):

PR #4584, opened 2021-12-17
Fix : save state the right way (DEV)

BMItr:
This PR fixes a possible race condition in _DccTicketingSharedViewModel_.kt
Its possible that tryEmit (fun _updateTransactionContext_) doesn't push the update.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/ui/shared/DccTicketingSharedViewModel.kt
```diff
@@ -3,45 +3,51 @@ package de.rki.coronawarnapp.dccticketing.ui.shared
...
+            .launchIn(viewModelScope)
+    }
+
     fun updateTransactionContext(ctx: DccTicketingTransactionContext) {
-        Timber.d("updateTransactionContext(ctx=%s)", ctx)
+        Timber.tag(TAG).d("updateTransactionContext(ctx=%s)", ctx)
         currentTransactionContext.tryEmit(ctx).also {
-            Timber.d("Update was successful? %s", it)
-            saveTransactionContext(ctx)
+            Timber.tag(TAG).d("Update was successful? %s", it)
```
mtwalli, line 36: Please, help me understand why the previous code was problematic ? how this fixes it. `tryEmit` for this state flow always returns `true` 
```kotlin
    override fun tryEmit(value: T): Boolean {
        this.value = value
        return true
    }
```
and despite that previous code was saving the `context` anyway regardless of the output

would not it be just easier if we just use `stateFlow.value = ctx`?

BMItr: The code was not really thread-safe and this could have lead to an inconsistent state, where the current context and the saved context could have differed.
I reworked updating the context to make it fully thread-safe. better safe than sorry

LukasLechnerDev: Hmm, this is getting a bit complicated. What about getting rid of the `StateFlow` and use `LiveData` instead 🤔.

`SavedStateHandle` can provide LiveData: `savedState.getLiveData<DccTicketingTransactionContext>(TRANSACTION_CONTEXT_SAVED_STATE_KEY)` so after saving something to the `savedState`, the changes are then directly delivered via `LiveData`...

BMItr: That's not really complicated.. I don't see much reason using a LiveData in this case. The transaction context is only called in other ViewModels and never directly exposed to any ui elements without modification. We also would lose the thread safety.
I know that my approach semms a little bit over engineered for the current case. But this could easily change in the furture like when we adding a cancellation service. So like i said, better safe than sorry

mtwalli (approved):
LGTM

LukasLechnerDev (approved):

PR #4593, opened 2021-12-20
FIx: Pop Up in Contact Journal (EXPOSUREAPP-11079)

jurajkusnier:
### Testing
- Go to Contact Journal
- Create two Persons and choose both
- Click on time, mask notes etc. for both persons
- uncheck one person (you have to have fast fingers, this error is hard to reproduce)
- same behavior also in "places" 

### Jira Link
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11079


LukasLechnerDev:
Hi! I can reproduce it reliably on my device. It happens when the focus is on the circumstances textbox with some text filled in and then deselect the person/location. This is because the deselection triggers the deletion of the person/location before the `afterTextChangedListener` of the textfield fires which tries to update the person/location, but it got deleted already.

Your fix works fine.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/repo/DefaultContactDiaryRepository.kt
```diff
@@ -126,8 +126,10 @@ class DefaultContactDiaryRepository @Inject constructor(
     ) {
         executeWhenIdNotDefault(visitId) {
             val original = contactDiaryLocationVisitDao.entityForId(visitId)
-            val updatedVisit = update(original.toContactDiaryLocationVisit())
-            contactDiaryLocationVisitDao.update(updatedVisit.toContactDiaryLocationVisitEntity())
+            original?.let {
```
LukasLechnerDev, line 129: Maybe add a comment in which case `original` is actually `null` would be helpful.

mtwalli (dismissed):
LGTM, Please write a comment as Lukas suggested

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/storage/repo/DefaultContactDiaryRepository.kt
jurajkusnier: updated

mtwalli (approved):

LukasLechnerDev (approved):
LGTM! Thanks

PR #4594, opened 2021-12-20
Fix redeemed state (EXPOSUREAPP-10035)

chiljamgossow:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10035

Likely cause: Our calculation for 21-days-old was not correct. So between day 21 and 22, it is already removed from the server (HTTP error 400 not found), but isOlderThan21Days returned false, so we show the error 400.

I also removed isFinal, as it is identical with isRedeemed and causes confusion, as in other parts of the app isFinal also includes POSITIVE and NEGATIVE results.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/SubmissionRepository.kt
```diff
@@ -124,23 +122,6 @@ class SubmissionRepository @Inject constructor(
         submissionSettings.clear()
     }
 
-    fun removeTestFromDevice(type: CoronaTest.Type) {
-        Timber.tag(TAG).v("removeTestFromDevice(type=%s)", type)
```
chiljamgossow, line 128: unused code

mtwalli (approved):
LGTM

LukasLechnerDev (approved):
LGTM.

Also thanks for cleaning stuff up 👍

PR #4600, opened 2021-12-21
Privacy (EXPOSUREAPP-10995)

chiljamgossow:
To get it right, I copy-pasted the attached privacy files from the ticket. Seems like there were some German words in the Turkish file.


LukasLechnerDev (approved):
👍

mtwalli (approved):

PR #4560, opened 2021-12-15
Fix: Duplicate `ci` (EXPOSUREAPP-11050)

mtwalli:
## How to test?
- Scan certificates from the ticket
- Try recycling, restoring, deleting, dismissing the badge and restoring from Scanner 
- Use https://dgc.a-sit.at/ehn/ to generate `VC, RC` with the same conditions (Different certificates with same `ci`)
- Try recycling, restoring, deleting, dismissing the badge  and restoring from Scanner 
- Request TC certificate for PCR and RAT and try different scenarios such (Successful Requests - Failures - Refreshing)

Note: This PR changes the certificate's container to user `QRCode` hash representation  as identifier due to the error cases appeared from using `ci`

## TODO
- [x] Fix unit tests
- [x]  Update pending TC `ci`

### Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11050

PR #4604, opened 2021-12-22
Display Unique Vaccination Certificate/Assertion Identifier (UVCI) (EXPOSUREAPP-11108)

mtwalli:
### Issue 
Because of the changes made by #4560 to rely on the qrcode hash to identify the certificate , hash is displayed instead of `ci` in :
- Certificates details screen 
- Exported PDF
- Validation results 

### Fix
- Renamed certificareId (legacy field) to `qrcodeHash` to avoid misusing it in the future
- Expose `uniqueCertificateIdentifier` and display where it is needed 
 
###  Background on the `why`
- These kind of certificates that have same `ci` are possible in Testing tools we have got some issues regarding it 
- It was possible to import two vaccination certificates with the same `ci` as long as they belong to two different persons
- Sine in recycle bin we rely on QrCode Hash to restore the certificate and not `ci`, it was also possible to use it as a workaround to scan many certificates of the same `ci` 
- with the changes made by #4560  QrCode hash is used to identify the certificate

## Testing
- Make sure that `ci` is displayed right in the required places 
- Certificates with same `ci` can be deleted / restored / recycled properly which was the initial cause of such a change  

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11108

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
@@ -40,9 +40,9 @@ data class RecoveryCertificateContainer(
     }
 
     override val containerId: RecoveryCertificateContainerId
-        get() = RecoveryCertificateContainerId(certificateId)
+        get() = RecoveryCertificateContainerId(qrcodeHash)
 
-    val certificateId: String
+    val qrcodeHash: String
         get() = data.recoveryCertificateQrCode.toSHA256()
 
     val personIdentifier: CertificatePersonIdentifier
@@ -132,8 +132,11 @@ data class RecoveryCertificateContainer(
                 get() = Locale(userLocale.language, recoveryCertificate.certificateCountry.uppercase())
                     .getDisplayCountry(userLocale)
 
-            override val certificateId: String
-                get() = this@RecoveryCertificateContainer.certificateId
+            override val qrcodeHash: String
+                get() = this@RecoveryCertificateContainer.qrcodeHash
+
+            override val uniqueCertificateIdentifier: String
+                get() = recoveryCertificate.uniqueCertificateIdentifier
 
             override val headerIssuer: String
                 get() = header.issuer
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
@@ -39,11 +39,11 @@ data class RecoveryCertificateContainer(
         }
     }
 
-    override val containerId: RecoveryCertificateContainerId
-        get() = RecoveryCertificateContainerId(qrcodeHash)
-
-    val qrcodeHash: String
+    override val qrCodeHash: String
         get() = data.recoveryCertificateQrCode.toSHA256()
+
+    override val containerId: RecoveryCertificateContainerId
+        get() = RecoveryCertificateContainerId(qrCodeHash)
 
     val personIdentifier: CertificatePersonIdentifier
         get() = certificateData.certificate.personIdentifier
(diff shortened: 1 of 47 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
```diff
@@ -40,9 +40,9 @@ data class RecoveryCertificateContainer(
     }
 
     override val containerId: RecoveryCertificateContainerId
-        get() = RecoveryCertificateContainerId(certificateId)
+        get() = RecoveryCertificateContainerId(qrcodeHash)
 
-    val certificateId: String
+    val qrcodeHash: String
         get() = data.recoveryCertificateQrCode.toSHA256()
```
chiljamgossow, line 47: Can you use get() = containerId.identifier here? Or remove this attribute all together ? so we have the hash logic only once

chiljamgossow: Maybe call it qrCodeHash everywhere to avoid further confusion

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/TestCertificateContainer.kt
```diff
@@ -63,7 +63,7 @@ data class TestCertificateContainer(
     val isCertificateRetrievalPending: Boolean
         get() = data.certificateReceivedAt == null
 
-    val certificateId: String
+    val qrcodeHash: String
         get() = data.testCertificateQrCode?.toSHA256() ?: data.identifier
```
chiljamgossow, line 68: qrCodeHash with upper case C

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/TestCertificateContainer.kt
```diff
@@ -129,8 +129,11 @@ data class TestCertificateContainer(
             override val certificateCountry: String
                 get() = Locale(userLocale.language, testCertificate.certificateCountry.uppercase())
                     .getDisplayCountry(userLocale)
-            override val certificateId: String
-                get() = this@TestCertificateContainer.certificateId
+            override val qrcodeHash: String
+                get() = this@TestCertificateContainer.qrcodeHash
+
+            override val uniqueCertificateIdentifier: String
+                get() = testCertificateQRCode!!.hash
```
chiljamgossow, line 137: this seems wrong, uniqueCertificateIdentifier should be the ci element

BMItr (dismissed):
I have nothing to complain. 👌 
consider the naming adjustments from @chiljamgossow and ping in case of reapproval needed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/storage/VaccinationContainer.kt
```diff
@@ -70,7 +70,7 @@ data class VaccinationContainer internal constructor(
     val vaccination: DccV1.VaccinationData
         get() = certificate.vaccination
 
-    val certificateId: String
+    val qrcodeHash: String
         get() = vaccinationQrCode.toSHA256()
```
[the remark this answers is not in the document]
mtwalli: I added `qrCodeHash` in Container interface ,but it has to overridden anyway

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/storage/RecoveryCertificateContainer.kt
mtwalli: Deleted it totally

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/TestCertificateContainer.kt
mtwalli: done

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/storage/TestCertificateContainer.kt
mtwalli: Thanks! , Fixed

BMItr (approved):

chiljamgossow (approved):
I could not confirm that test certificates from RAT/PCR still work. We should make sure that testers check.

PR #4605, opened 2021-12-22
FIX: NoSuchElementException in PersonDetailsViewModel (DEV)

jurajkusnier:
I have only this stack trace from the google play console

```
java.util.NoSuchElementException: 
  at de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsViewModel.access$createUiState (PersonDetailsViewModel.kt:39)
  at de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsViewModel$uiState$1.invokeSuspend (PersonDetailsViewModel.kt:4)
  at de.rki.coronawarnapp.covidcertificate.person.ui.details.PersonDetailsViewModel$uiState$1.invoke (PersonDetailsViewModel.kt:1)
  at kotlinx.coroutines.flow.FlowKt__ZipKt$combine$1$1.invokeSuspend (Zip.kt:1)
  at kotlinx.coroutines.flow.FlowKt__ZipKt$combine$1$1.invoke (Zip.kt:1)
  at kotlinx.coroutines.flow.internal.CombineKt$combineInternal$2.invokeSuspend (Combine.kt:29)
  at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith (ContinuationImpl.kt:3)
  at kotlinx.coroutines.DispatchedTask.run (DispatchedTask.kt:18)
  at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely (CoroutineScheduler.kt:1)
  at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run (CoroutineScheduler.kt:10)
```

According to the mapping, it looks like the crash happened in `getNewestDoseVaccinatedOn()` but I wasn't able to reproduce it. A similar crash https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10153 has been fixed before, even though nobody was able to reproduce it. Here is a short test scenario:

- 2 Personal wallets, each has 4 different types of certificates
- delete from each personal wallet one certificate
- repeat 3
- --> target: have a recycle bin with 4 deleted certificates of person 1/2/1/2
- I did this on 4 devices in parallel - 2 iPhones, 2 Androids
- one Android crashed during 4.)
- after reopening the app: the deletion has happened and everything is fine again

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
@@ -36,7 +36,7 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.combine
-import kotlinx.coroutines.flow.first
+import kotlinx.coroutines.flow.firstOrNull
 import kotlinx.coroutines.flow.mapNotNull
 import timber.log.Timber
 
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
```diff
@@ -107,19 +108,23 @@ class PersonDetailsViewModel @AssistedInject constructor(
...
+                                vaccinationStatus = vaccinationStatus,
+                                daysUntilImmunity = daysUntilImmunity,
+                                boosterRule = boosterRule,
+                                daysSinceLastVaccination = daysSinceLastVaccination,
+                                hasBoosterNotification = vaccinatedPerson.hasBoosterNotification
+                            )
                         )
-                    )
+                    } catch (e: Exception) {
+                        Timber.e(e, "creating VaccinationInfoCard.Item failed")
```
mtwalli, line 126: what happens in UI in this case ?

jurajkusnier: `VaccinationInfoCard` is not visible in UI. But if the trigger for this crash is the same as for the bug mentioned in the PR description then this should happen only when the user is deleting certificates so he/she will be forwarded to a list of persons immediately and this shouldn't be a concern.

BMItr (approved):

mtwalli (approved):

PR #4612, opened 2021-12-23
Store RATProfile with Jetpack Datastore (EXPOSUREAPP-10912)

LukasLechnerDev:
Jira Ticket: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10912

Our custom `FlowPreference` seems to have a glitch that it sometimes still emitts values even though the underlying `SharedPreference` value got already deleted. 

We thought that this fix is a good opportunity to try out [Jetpack DataStore](https://developer.android.com/topic/libraries/architecture/datastore). So this PR now stores the `RatProfile` in the `DataStore` instead of in the `SharedPreferences`.

Testing:
- Please play around with the RATProfile - Add it, modify it, delete it, reset the app ...
- Test if migration from a previous version that used `SharedPreferences` works fine

Open Tasks:
- [x] Address `TODO` comments in the code
- [x] Adjust / Extend Tests

Open Question:
- keep SharedPrefs implementation and create common interface so that we can easily roll back if issues arise during testing?


chiljamgossow (approved):
Tested migration, modification, deletion, reset
Works nicely 👍
code LGTM

mtwalli (approved):

PR #4629, opened 2021-12-29
Issuer country in DE or EN in print version (EXPOSUREAPP-9863)

### LINKED ISSUE #4175: Print version of vaccination certificate: Member state of vaccination rendered in phone language (in my case Japanese)
## Avoid duplicates
* [ ] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)  
    → I don't own a iOS device with which I could verify this

## Technical details

- Device name: Google Pixel (first generation)
- Android version: 9
- App version: 2.10.1

## Describe the bug

When I generate a print version of vaccination certificates, the resulting PDF document is almost completely in English and German, but the member state of vaccination is rendered in Japanese (the language my phone is set to).

The affected section of the document therefore reads:

Mitgliedstaat der Impfung:　‌　‌　ドイツ
Member State of vaccination:

Generate a print version of a vaccination certificate on a phone that's set to Japanese.

## Expected behaviour

I would expect the member state to be rendered in English. Given that “ドイツ” is probably not recognizable to mean anything for most people.

## Possible Fix

Don't rely on the phone's language settings to render the state name in the PDF.

## Additional context

\-

---
Internal Tracking ID: [EXPOSUREAPP-9863](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9863)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9863

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/repository/CertificateRepoContainer.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/repository/CertificateRepoContainer.kt
@@ -2,7 +2,6 @@ package de.rki.coronawarnapp.covidcertificate.common.repository
 
 import de.rki.coronawarnapp.reyclebin.common.Recyclable
 import de.rki.coronawarnapp.util.qrcode.coil.CoilQrCode
-import java.util.Locale
 
 interface CertificateRepoContainer : Recyclable {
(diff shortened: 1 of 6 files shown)


chiljamgossow (approved):
Tested
LGTM

PR #4644, opened 2022-01-04
VC issued by italy can be printed (EXPOSUREAPP-11155)

mtwalli:
- Check based on `iss` not `co` 
## Testing 
- Scan certificate from the ticket and verify that it can not be exported as PDF
- Scan DE issued certificate and verify that it can exported 
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11155


LukasLechnerDev (approved):
👍 Certificate from ticket can *not* be printed but generated certificate with Issuer `DE` *can* be printed

SamuraiKek (approved):
Tested on device with the QR from the ticket. Works.

PR #4649, opened 2022-01-04
Restore statistics state (EXPOSUREAPP-11082)

chiljamgossow:
Please test the behavior of the statistics cards and click on the info button and when the statistics are updated.

What still does not work: When the app is moved to background, the lifecycle callback that stores the states is gone, so the first time you navigate to the info screen afterwards, restoring is not correct, but second time it works again.


LukasLechnerDev (approved):
LGTM. State is preserved after closing the info screen. As you described, it doesn't work the first time after the app was in background, but still this PR is an improvement!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/lists/modular/mods/SavedStateMod.kt
```diff
@@ -15,26 +15,28 @@ class SavedStateMod<T : ModularAdapter.VH> :
 
     private val savedStates = mutableMapOf<String, Parcelable>()
 
+    private var initial = true
+
     override fun onPostBind(adapter: ModularAdapter<T>, vh: T, pos: Int) {
```
BMItr, line 20: if I got it right, the change of SavedStateMod does not only affect the Statistic Cards but all Lists in the App...
Under normal circumstances.
as this Mod is currently only used by the statistics, maybe you could add a comment or something

BMItr: @chiljamgossow 
I will approve it, but please add a comment in some PR later

BMItr (approved):

PR #4650, opened 2022-01-04
Easy language link (DE) only (EXPOSUREAPP-10220)

mtwalli:
- Remove old keys from DE strings to clean them up 
- Add news keys to `non_translatable.xml` as easy language is only available in German 

Screenshot 2022-01-04 at 15 15 34
Screenshot 2022-01-04 at 15 16 16

## Testing 
- Check First Onboarding Screen and About the App screen in `DE` -> easy language should be visible and clickable 
- other language -> no link
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10220


KathrinLuecke (approved):
UA approved

ndegendogo:
oh .. why this?
Why hide this link from all users except those that have chosen German as primary language in cwa?

Actually I liked the Android solution very much, because:
- the audience for easy language are not only German natives, but also users with limited German 
- especially bi-lingual users might appreciate the easy language
- last not least, the linked content is not only easy language but also **sign language**

So, please consider to let the users decide themselves if they want to open that link

chiljamgossow:
> oh .. why this? Why hide this link from all users except those that have chosen German as primary language in cwa?
> 
> Actually I liked the Android solution very much, because:
> 
> * the audience for easy language are not only German natives, but also users with limited German
> * especially bi-lingual users might appreciate the easy language
> * last not least, the linked content is not only easy language but also **sign language**
> 
> So, please consider to let the users decide themselves if they want to open that link

@thomasaugsten FYI

> oh .. why this? Why hide this link from all users except those that have chosen German as primary language in cwa?
> 
> Actually I liked the Android solution very much, because:
> 
> * the audience for easy language are not only German natives, but also users with limited German
> * especially bi-lingual users might appreciate the easy language
> * last not least, the linked content is not only easy language but also **sign language**
> 
> So, please consider to let the users decide themselves if they want to open that link

I think this is a valid point.

mtwalli:
> oh .. why this? Why hide this link from all users except those that have chosen German as primary language in cwa?
> 
> Actually I liked the Android solution very much, because:
> 
>     * the audience for easy language are not only German natives, but also users with limited German
> 
>     * especially bi-lingual users might appreciate the easy language
> 
>     * last not least, the linked content is not only easy language but also **sign language**
> 
> 
> So, please consider to let the users decide themselves if they want to open that link

Hi @ndegendogo , Thanks for your feedback. The easy language link is appearing in two places in the App 
- Onboarding Flow -> was only visible in `DE`
- App Information screen -> is visible in all languages ,but led to the German explanation and it is considered as a bug in Android side. 

I see your point. if it has to visible in the App information regardless of user's locale. it has to be visible everywhere for consistency. I will transfer your feedback internally and come back to you

SamuraiKek (approved):

mtwalli:
We got a feedback that it should be fine to have it visible in other locales too

PR #4654, opened 2022-01-05
Trigger value set update after app reset (EXPOSUREAPP-11141)

LukasLechnerDev:
[Jira Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11141)

The update of the value set was only triggered on app start (When the `ValueSetRepository` gets initialised). 
When the user resets the app, the value set is cleared, but never freshly downloaded unless he manually restarts the app. 

This PR moves the triggering of the value set update to the init block of `MainViewModel`. In this case the update also happens after an app reset. It now happens with the same frequency as the environment toast shows up. 

How to test:
- Scan certificate (generate one by eg `cwa vc gen -e wru -o`)
- Go to detail screen and check if the labels have readable names (like "Vaccine Type") and not cryptic numbers
- grafik

- Reset the app
- Perform Onboarding
- Refresh the list of DSCs in DCC Signature Verification Screen of Test Menu - otherwise the signature check will fail
- Scan certificate again
- labels should still have readable names


chiljamgossow (approved):
Tested

mtwalli (approved):

PR #4665, opened 2022-01-07
Fix: Unknown statistics type, enhance logs

jurajkusnier:
Enhancing logs since the root cause is not obvious from the current logs we have.

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-10680


BMItr (dismissed):
lgtm!

mtwalli (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/source/StatisticsProvider.kt
```diff
@@ -78,6 +83,9 @@ class StatisticsProvider @Inject constructor(
         val rawData = server.getRawStatistics()
         return parser.parse(rawData).also {
             Timber.tag(TAG).d("Parsed from server: %s", it)
+            if (!it.isDataAvailable) {
+                Timber.tag(TAG).e("RawData: ${rawData.toHexString()}")
```
BMItr, line 87: Create an func for that, so you don't have to repeat yourself
Prefer to use %s instead of string templates when using Timber

BMItr, line 87: and also maybe LogLevel W would be a better match.
Its not really an Error

jurajkusnier: updated

BMItr (approved):

mtwalli (approved):

PR #4680, opened 2022-01-11
Statistics for boosters (EXPOSUREAPP-10765)

SamuraiKek:
Added a new statistics card that shows the number of people with boosters.

To test it, switch to `MOCK_CLOUD` and the card should be second to last. Don't mind the values, they're inflated from the mock server.

Also, make sure it works in night mode and that all the people on the cards are now wearing a mask in both light/dark mode.


AlexanderAlferov (changes requested):
I think new text is missing in the Info Screen ℹ️ Kindly check the Figma 😄

SamuraiKek:
> I think new text is missing in the Info Screen ℹ️ Kindly check the Figma 😄

Thanks, I always forget about it 😄

AlexanderAlferov (dismissed):
LGTM 🚀

CV113 (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
```diff
@@ -238,6 +239,28 @@ data class PersonsVaccinatedCompletelyStats(
...
+    val boosterDoses: KeyFigure
+        get() = keyFigures.single { it.rank == KeyFigure.Rank.PRIMARY }
+
+    val total: KeyFigure
+        get() = keyFigures.single { it.rank == KeyFigure.Rank.TERTIARY }
+
+    override fun requireValidity() {
+        require(keyFigures.size == 2)
+        requireNotNull(keyFigures.singleOrNull { it.rank == KeyFigure.Rank.PRIMARY }) {
+            Timber.w("PersonsVaccinatesWithBoostersStats is missing primary value")
```
jurajkusnier, line 256: PersonsVaccinatesWithBoostersStats > PersonsVaccinatedWithBoosterStats
typo?

jurajkusnier (commented):
Could you double-check that Android texts are updated? There is a mismatch between iOS and Android design
out

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/StatsItem.kt
SamuraiKek: Yes, good catch!

SamuraiKek:
> Could you double-check that Android texts are updated? There is a mismatch between iOS and Android design 

Turns out the iOS one is correct. I will update it.

CV113 (approved):
UA approved

jurajkusnier (approved):

AlexanderAlferov (approved):

PR #4343, opened 2021-11-08
Most relevant DCC, update vaccination priority (EXPOSUREAPP-9443)

SamuraiKek:
It's now a feature, not a bug 😄 
The logic for vaccinations has been updated to accomodate boosters.

What has changed from the previous logic: 
1. A BioNTech/Moderna/AstraZeneca vaccine that is taken after a recovery (that means it will have dose 1/1), will now get priority right away. If it's a J&J vaccine, it will get priority after 14 days.
2. A BioNTech/Moderna/AstraZeneca vaccine with dose 3/3 will get priority right away.
3. a J&J vaccine with dose 2/2 will get priority right away.
3. All other cases are covered by the old logic.

How to test:
Start from a clean slate. Delete all other certificates if you have any. Switch to WRU env. I'm using BioNTech and J&J vaccines to test, but you can substitute BioNTech with Moderna or AstaZeneca. 
Moderna: `EU/1/20/1507`
AstraZeneca: `EU/1/21/1529`

1. Scan a recovery certificate < 180 days old: 
 `cwa rc gen --pii 42 -s r.0.fr='2021-08-01' r.0.df='2021-10-10' -e wru`. This will have highest prio as it's the only one.
2. Scan a J&J vaccine < 14 days old:
`cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=1 v.0.mp='EU/1/20/1525' v.0.dt='2021-11-04' -e wru`. The recovery certificate should still have highest prio as this vaccine is < 14 days old. Vaccination Status info box should say you have received all planned vaccinations but protection won't be complete for x days.
3. Scan a J&J vaccine > 14 days old:
`cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=1 v.0.mp='EU/1/20/1525' v.0.dt='2021-10-04' -e wru`. This vaccine will now get the highest prio as the 14 days have passed. Vaccination Status info box should say you have received all planned vaccinations and protection is complete.
4. Remove the two J&J vaccines and leave the recovery certificate in place.
5. Scan a BioNTech vaccine with dose 1/1 that's < 14 days old:
`cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=1 v.0.mp='EU/1/20/1528' v.0.dt='2021-11-04' -e wru`. This vaccine should get highest priority right away. Vaccination Status info box should say you have received all planned vaccinations and protection is complete.
6. Scan a BioNTech vaccine with dose 1/1 that's > 14 days old:
[...]

PR #4687, opened 2022-01-14
Support new booster notation (EXPOSUREAPP-11272)

jurajkusnier:
### Testing
- scan as many DCC QR codes as you can and carefully watch which certificate has been selected as the certificate with the highest priority.
- focus on vaccination certificates where `doseNumber` > `totalSeriesOfDoses` or `v[0].dn is > than v[0].sd`
- you can generate such certificates using cli tool, eg:
```
cwa vc gen -o --pii 32 -s v.0.dt=2021-08-01T08:57:25Z -s v.0.dn=2 -s v.0.sd=1
```

### Jira Link
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11272


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
```diff
@@ -54,14 +54,14 @@ fun Collection<CwaCovidCertificate>.toCertificateSortOrder(): List<CwaCovidCerti
...
- * Find Vaccination Certificates where total number of doses == number of administered doses and
- * 3.1 For vaccines with dose 3/3, priority will be received right away
- * 3.2 For BioNTech/Moderna/AstraZeneca vaccines that are taken after a recovery, priority will be received right away
- * 3.3 For J&J vaccines with dose 2/2, priority will be received right away
- * 3.4 If none of the criteria above is met, priority will be received after a 14 day period
- * If there is one or more certificates matching these requirements,
- * the first one is returned as a result of the operation.
+ * find Vaccination Certificates (i.e. DGC with v[0]) where v[0].dn is >= than v[0].sd and one of
+ * the time difference between the time represented by v[0].dt and the current device time is > 14 days, or
+ * v[0].dn is > 1 and v[0].mp equals EU/1/20/1525 (Johnson & Johnson) (booster vaccination), or
```
vaubaehn, line 59: Hi @jurajkusnier ,
> v[0].dn is > 1 and v[0].mp equals EU/1/20/1525 (Johnson & Johnson) (booster vaccination)

For my understanding that won't work, when v[0] for both conditions is the _same_ DCC. I assume the array v[0] refers to the vaccination entry of _any_ DCC.
Think, a user was J&J-jabbed and enters that certificate into CWA first. Then v[0][1].dn is == 1 and v[0][1].mp is == EU/1/20/1525. Then user gets the booster with mRNA and enters the certificate as v[0][2]. Now v[0][2].dn is > 1and v[0][2].mp is different than EU/1/20/1525.
So, the check/filter would need to process 2 different certificates for that single result.

vaubaehn (changes requested):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
vaubaehn: Sorry, please ignore my above comment.
I now understood that above condition refers to a booster _with_ J&J, not a booster _after_ J&J.
Then `v[0].dn is > than v[0].sd` will do the trick for the new EU notation and boosters _after_ J&J or recovery.
Sorry again

SamuraiKek (approved):
LGTM. Tested with the scenarios from #4343 and adjusted the 2/2 J&J vaccines to 2/1. Everything looks alright.

chiljamgossow (approved):
LGTM

PR #4691, opened 2022-01-17
Fix: CameraX native crash + Transition (DEV)

mtwalli:
For more details see: https://issuetracker.google.com/issues/214062063


BMItr (approved):
implementation is done according to the suggestion.
We should keep an eye on further findings by google (SurfaceView)

jurajkusnier (approved):
LGTM, tested on Pixel 4, Android 11

PR #4700, opened 2022-01-19
Translation Delivery

service-tip-git:
Translation Delivery

PR #4704, opened 2022-01-19
[INTERNAL] Translation delivery: commit by LX Lab (#4700)

mtwalli:
Fix translations


Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -1,30 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?><resources xmlns:tools="http://schemas.android.com/tools" xmlns:xliff="urn:oasis:names:tc:xliff:document:1.2" tools:ignore="MissingTranslation">
-
```
BMItr, line 2: could this comment stay inside the file ?!
it helps the developers to distinguish which label to choose

mtwalli: if it is not in DE I think it will be removed , maybe we need to add it to `DE` file instead

PR #4716, opened 2022-01-21
CCL Skeleton (EXPOSUREAPP-11478)

mtwalli:
Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11478

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
@@ -1,6 +1,6 @@
 package de.rki.coronawarnapp.ccl.dccwalletinfo.model
 
-// TODO delete file
+// DO delete file
 private val admissionState = AdmissionState(
     visible = true,
     badgeText = SingleText(
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
@@ -1,0 +1,105 @@
+package de.rki.coronawarnapp.ccl.dccwalletinfo.model
+
+// TODO delete file
+private val admissionState = AdmissionState(
+    visible = true,
+    badgeText = SingleText(
+        type = "string",
+        localizedText = mapOf("de" to "2G+"),
+        parameters = listOf()
+    ),
+    titleText = SingleText(
+        type = "string",
+        localizedText = mapOf("de" to "Status-Nachweis"),
+        parameters = listOf()
+    ),
+    subtitleText = SingleText(
+        type = "string",
+        localizedText = mapOf("de" to "2G+ PCR-Test"),
+        parameters = listOf()
+    ),
+    longText = SingleText(
+        type = "string",
+        localizedText = mapOf("de" to "Ihre Zertifikate erfüllen [...]"),
+        parameters = listOf()
+    ),
+    faqAnchor = "dcc_admission_state"
+)
+private val vaccinationState = VaccinationState(
+    visible = true,
+    titleText = SingleText(
+        type = "string",
+        localizedText = mapOf("de" to "Impfstatus"),
+        parameters = listOf()
+    ),
+    subtitleText = PluralText(
+        type = "plural",
+        quantity = 25,
+        localizedText = mapOf(
+            "de" to QuantityText(
+                zero = "Letzte Impfung heute",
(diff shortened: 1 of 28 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
```diff
@@ -0,0 +1,105 @@
+package de.rki.coronawarnapp.ccl.dccwalletinfo.model
+
+// DO delete file
```
LukasLechnerDev, line 3: USE TODO instead of DO? Then Android Studio can show all open TODOs.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DccWalletInfoParserTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        admissionState = admissionState,
+        vaccinationState = vaccinationState,
+        verification = verification,
+        boosterNotification = boosterNotification,
+        mostRelevantCertificate = mostRelevantCertificate,
+        validUntil = "2022-01-14T18:43:00Z"
+    )
+
+    @Test
+    fun `Serialize DCCWalletInfo`() {
```
LukasLechnerDev, line 122: I think here we are testing the *de*serialization so I think the test name is wrong? If I am correct then also the function name below needs to be changed.

LukasLechnerDev (dismissed):
2 optional comments, otherwise I think this is a good foundation 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
mtwalli: Yes :D , sonar does not like many `TODO` s which led to `CI` fail  🤷‍♂️

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DccWalletInfoParserTest.kt
mtwalli: done

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/model/DummyWalletInfo.kt
LukasLechnerDev: Ah, I suspected something like this ;)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/text/TextResource.kt
```diff
@@ -0,0 +1,33 @@
...
+        text!!.format(cclText.parameters.format())
+    }
+}
+
+internal fun List<Parameters>.format(): List<String> = map { parameters ->
+    parameters.toString() // DO: Formatting a UI Text - Format Parameters
+}
+
+private const val EN = "en"
+private const val DE = "en"
```
chiljamgossow, line 33: this should be "de"

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/text/TextResource.kt
```diff
@@ -0,0 +1,33 @@
...
+fun textResource(
+    cclText: CCLText,
+    locale: Locale = Locale.getDefault()
+) = lazy { formatCCLText(cclText, locale.language) }
+
+internal fun formatCCLText(cclText: CCLText, lang: String): String = when (cclText) {
+    is PluralText -> "TODO: Format plural text"
+    is SingleText -> {
+        // DO: provide fallback
+        val text = cclText.localizedText[lang]
```
chiljamgossow, line 21: We should consider the list of languages from Android, otherwise we might get a mix of languages, eg, when the first language is Spanish, the second is Bulgarian

chiljamgossow (dismissed):
We should keep the language resolution in mind if several languages are used, but not a blocker

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/text/TextResource.kt
chiljamgossow: never mind, that will be handled in the dcc layer

chiljamgossow (approved):
Thanks

LukasLechnerDev (approved):

PR #4750, opened 2022-01-28
CCL wallet info binding for certificates overview screen (EXPOSUREAPP-11498)

SamuraiKek:
Change won't be visible until we have some real data to work with. For now, the certificate displayed per person is the one from the fallback mechanism, meaning that even though we could have a 2G+ status with a vacc cert & a test cert, the QR code card will only contain the vacc cert with no status details and no option to switch to the test cert.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificates.kt
```diff
@@ -23,16 +25,18 @@ data class PersonCertificates(
...
-            certificates.firstOrNull { it.qrCodeHash == certRef.certificateRef.barcodeData.toSHA256() }
-        }
-
-        listOf(VerificationCertificate(certificate = certificates.findFallbackDcc()))
+            certificates.firstOrNull { it.qrCodeHash == certRef.certificateRef.barcodeData.toSHA256() }?.let {
+                VerificationCertificate(
+                    certificate = it,
+                    buttonText = formatCCLText(certRef.buttonText, Locale.getDefault().language)
+                )
+            }
```
mtwalli, line 35: You  should `take(2)` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt
```diff
@@ -63,13 +65,15 @@ class PersonOverviewViewModel @AssistedInject constructor(
     ) {
         persons.filterNotPending()
             .forEachIndexed { index, person ->
-                val admissionState = person.admissionState
+                val admissionState = person.dccWalletInfo?.admissionState
+                val certificatesForOverviewScreen = person.certificatesForOverviewScreen
+                Timber.d("VerificationCertificates ${person.certificatesForOverviewScreen}")
                 val color = PersonColorShade.shadeFor(index)
-                if (admissionState != null) {
+                if (certificatesForOverviewScreen.isNotEmpty()) {
```
mtwalli, line 72: Please check the ViewModel Tests? which is not currently disabled

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificates.kt
SamuraiKek: Ah yes, forgot to add it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModel.kt
SamuraiKek: View model tests pass after enabling the disabled tests.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificates.kt
```diff
@@ -23,16 +25,23 @@ data class PersonCertificates(
...
+                null -> emptyList()
+                else -> listOf(VerificationCertificate(cert))
+            }
         }
-
-        listOf(VerificationCertificate(certificate = certificates.findFallbackDcc()))
     }
 
+    // Obsolete, remove in cleanup PR
     val admissionState: AdmissionState? get() = certificates.determineAdmissionState()
```
jurajkusnier, line 45: as it's not used anymore feel free to remove it now

SamuraiKek: I thought about it, but I would rather have it removed in the cleanup PR because if I start with this, then I should also remove the `determineAdmissionState` function from `PersonCertificateExtensions`, and if I'm there, I should probably clean the rules, and so on 😄  I will add a `TODO` to the comment for better visibility.

jurajkusnier (approved):
LGTM

mtwalli (approved):

PR #4764, opened 2022-02-01
Booster vaccination info card shows full immunity right away (EXPOSUREAPP-11634)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11634)

A 2/1 vaccination < 14 days was still showing the days counter until full immunity, even though full immunity should be granted right away. 

To test, follow the steps from the ticket.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationCertificate.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationCertificate.kt
@@ -25,6 +25,7 @@ interface VaccinationCertificate : CwaCovidCertificate {
     override val rawCertificate: VaccinationDccV1
 
     val isSeriesCompletingShot get() = doseNumber >= totalSeriesOfDoses
+    val isBooster get() = doseNumber > totalSeriesOfDoses
 
     companion object {
         const val BIONTECH = "EU/1/20/1528"
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationCertificate.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationCertificate.kt
@@ -25,7 +25,6 @@ interface VaccinationCertificate : CwaCovidCertificate {
     override val rawCertificate: VaccinationDccV1
 
     val isSeriesCompletingShot get() = doseNumber >= totalSeriesOfDoses
-    val isBooster get() = doseNumber > totalSeriesOfDoses
 
     companion object {
         const val BIONTECH = "EU/1/20/1528"
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
@@ -97,7 +97,7 @@ data class VaccinatedPerson(
     private fun isSeriesCompletingOverTwoWeeks(today: LocalDate): Boolean {
         return when {
             vaccinationCertificates.isEmpty() -> false
-            vaccinationCertificates.any { it.isBooster } -> true
+            vaccinationCertificates.any { it.doseNumber > it.totalSeriesOfDoses } -> true
             else -> vaccinationCertificates.any {
                 it.isSeriesCompletingShot && Days.daysBetween(
                     it.rawCertificate.vaccination.vaccinatedOn,
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinatedPerson.kt
@@ -95,10 +95,16 @@ data class VaccinatedPerson(
         vaccinationCertificates.maxOf { it.vaccinatedOn }
 
     private fun isSeriesCompletingOverTwoWeeks(today: LocalDate): Boolean {
-        val certificate = vaccinationCertificates
-            .filter { it.isSeriesCompletingShot }
-            .firstOrNull { Days.daysBetween(it.rawCertificate.vaccination.vaccinatedOn, today).days > 14 }
-        return certificate != null
+        return when {
+            vaccinationCertificates.isEmpty() -> false
+            vaccinationCertificates.any { it.isBooster } -> true
+            else -> vaccinationCertificates.any {
+                it.isSeriesCompletingShot && Days.daysBetween(
+                    it.rawCertificate.vaccination.vaccinatedOn,
+                    today
+                ).days > 14
+            }
+        }
     }
 
     private fun getNewestFullDose(): VaccinationCertificate? = vaccinationCertificates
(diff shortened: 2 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/VaccinationCertificate.kt
```diff
@@ -25,6 +25,7 @@ interface VaccinationCertificate : CwaCovidCertificate {
     override val rawCertificate: VaccinationDccV1
 
     val isSeriesCompletingShot get() = doseNumber >= totalSeriesOfDoses
+    val isBooster get() = doseNumber > totalSeriesOfDoses
```
jurajkusnier, line 28: This could be misleading when we apply it to all VC. According to the EU spec
>Boosters will be recorded as follows:
>- 3/3 for a booster dose following a primary 2-dose vaccination series.
>- 2/1 for a booster dose following a single-dose vaccination or a one dose of a 2-dose vaccine administered to a recovered person.

or did something change?

SamuraiKek: Alright, I will remove it from here and just add a comparison in `VaccinatedPerson`

vaubaehn:
Did you take into account that in Germany now a 2/1 is 'Grundimmunisierung'? Same would be valid for a single dose vaccination after a single recovery, afaict

jurajkusnier (approved):

SamuraiKek:
> Grundimmunisierung



> Did you take into account that in Germany now a 2/1 is 'Grundimmunisierung'? Same would be valid for a single dose vaccination after a single recovery, afaict
> 
> @Ein-Tim , can you confirm?

Not sure I understand. This fix will let the user know that with a 2/1 vaccine, they are fully immunised right away. Not waiting for 14 days to get full immunisation.

AlexanderAlferov (approved):

Ein-Tim:
@SamuraiKek 

And exactly this is the question here, is this the case?

We have two cases here:
Case 1: Second vaccination after recovery. This should be valid immediately, I think. 

Case 2: One additional vaccination after one vaccination with J&J. This is only the end of the "Grundimmunisierung". Is this vaccination also valid immediately or does one have to wait 14 days, how it would be if you would be vaccinated with e.g. Biontech.

cc @mlenkeit

vaubaehn:
> > Grundimmunisierung
> 
> > Did you take into account that in Germany now a 2/1 is 'Grundimmunisierung'? Same would be valid for a single dose vaccination after a single recovery, afaict
> > @Ein-Tim , can you confirm?
> 
> Not sure I understand. This fix will let the user know that with a 2/1 vaccine, they are fully immunised right away. Not waiting for 14 days to get full immunisation.

@SamuraiKek 
For my understanding that would be wrong since a couple of days, since PEI states you'll need another rMNA vaccination after a single shot J&J, and another rMNA shot as a booster.

Afaik, for recoveries it's handled in a similar way.
That's Germany only, other EU nations may differ, but that's checked via business rules and CCL from 2.18

vaubaehn:
@Ein-Tim 
According to new EU notation and PEI standards, a booster vaccination would be noted as 3/1 (2/1 is first vaccination after recovery)

Ein-Tim:
@vaubaehn Please, see my DM in Slack.

Ein-Tim:
https://www.pei.de/DE/newsroom/dossier/coronavirus/coronavirus-inhalt.html?nn=169730&cms_pos=3 says:

**Zeitintervall nach Impfung für einen vollständigen Impfschutz**

Seit der letzten erforderlichen Einzelimpfung müssen 14 Tage vergangen sein, um einen vollständigen Impfschutz zu erreichen.

If I interpret this correctly, this is also the case here. After the second vaccination after J&J 14 days need to have passed until the one has the "vollständige Impfschutz". 

This means that 2/1 can be valid immediately (if it is the second booster after recovery) or that it is only valid after 14 days (if it is the "Abschluss der Grundimmunisierung").

Would be good to get this confirmed by someone.

Ein-Tim (changes requested):
This PR should be reviewed for correctness (whether 2/1 is valid immediately in all cases).

SamuraiKek:
> This PR should be reviewed for correctness (whether 2/1 is valid immediately in all cases).

I've set it to draft so nobody rushes to merge it. I'm asking around to get a deffinitive answer.

Ein-Tim:
Thank you @SamuraiKek!

mlenkeit:
@Ein-Tim @vaubaehn you're right, a 2/1 vaccination can have two meanings:

- a second shot (aka booster) after a 1/1 vaccination (i.e. recovery vaccination) - in that case, 2/1 is valid immediately
- a second shot after a 1/1 vaccination with Janssen - in that case, 2/1 is valid only after 14 days

However, it's been agreed with RKI that CWA will currently only support the first case, so from that perspective I think the PR is fine and can be merged/reviewed.

With CCL, this will be properly represented. Until then, we'll need to live with this compromise.

SamuraiKek:
> @Ein-Tim @vaubaehn you're right, a 2/1 vaccination can have two meanings:
> 
> * a second shot (aka booster) after a 1/1 vaccination (i.e. recovery vaccination) - in that case, 2/1 is valid immediately
> * a second shot after a 1/1 vaccination with Janssen - in that case, 2/1 is valid only after 14 days
> 
> However, it's been agreed with RKI that CWA will currently only support the first case, so from that perspective I think the PR is fine and can be merged/reviewed.
> 
> With CCL, this will be properly represented. Until then, we'll need to live with this compromise.

Thanks for the clarification @mlenkeit! In this case I will reopen the PR.

vaubaehn:
Hi all, I'd like to share the results of an intensive discussion with @Ein-Tim :

We agreed on, 
- a single J&J vaccination is/was noted as **1/1**. According to the new regulations published by PEI and referenced from @Ein-Tim above, this is not sufficient for a basic immunization in Germany anymore.
- the next vaccination after the first vaccination 1/1 J&J (be it another J&J shot or a mRNA vaccination) is necessary to complete the basic immunization ("Grundimmunisierung"). Before Feb 2022 it was noted in Germany as 2/2 for the DCC. According to the changes by EU, notation would now be **2/**1 on EU level. According to PEI this is not a booster shot and a 14-day waiting period applies before the immunization state is valid as "Grundimmunisierung".
- therefore, the next vaccination after 1/1 J&J and 2/1 mRNA/J&J would be the booster shot according to national regulations and be valid immediately without waiting period. According to the new EU notation, it would be noted as **3/**1.

Where we had problems to find an agreement:
- one could interpret the EU statement for vaccinations after a recovery either in the way, that 2/1 means it's the second dose of a vaccination after recovery (booster), and 1/1 would be the first dose after recovery (basic immunization) - so then 2/1 would count as a booster, or
- one could interpret the EU statement for vaccination after a recovery in the way, that 2/1 means it's the first dose of a vaccination after recovery (the recovery itself would invisibly be counted as 1/1) which would be basic immunization, and 3/1 would be the second vaccination after recovery what would be the booster shot.

However, PEI (referenced by @Ein-Tim ) states, that after (any) vaccination after recovery there is no waiting period for validity, so 2/1 for recovery would be valid immediately.

But we also agreed on, that confusion with regulation and notation everywhere is a mess.

vaubaehn:
Ah, I was a bit late with my reply, didn't read your comments before... Was a bit destracted by my daugther... thanks to covid-suspected-home-office...

Fine, everything is clarified!

Have a great night everybody!

PR #4773, opened 2022-02-02
Ccl Json Functions (DEV)

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculation.kt
```diff
@@ -16,38 +15,33 @@ import de.rki.coronawarnapp.covidcertificate.common.certificate.CwaCovidCertific
...
     ): DccWalletInfo {
 
-        val input = getDccWalletInfoInput(dccList = dccList).toJsonNode()
-        val output = jsonFunctions.evaluateFunction(
-            FUNCTION_NAME,
-            input
-        )
-
+        val output: JsonNode
+        runBlocking(Dispatchers.IO) {
```
LukasLechnerDev, line 39: I think we should switch to the IO Dispatcher in `JsonFunctionsWrapper.evalateFunction()` so that callers of that function can call it on any dispatcher.

LukasLechnerDev: Using `Dispatchers.Default` is even better here I think. And we should use our `DispatcherProvider`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/JsonFunctionsWrapper.kt
```diff
@@ -0,0 +1,56 @@
...
+import de.rki.coronawarnapp.util.serialization.BaseJackson
+import de.rki.jfn.JsonFunctions
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class JsonFunctionsWrapper @Inject constructor(
```
LukasLechnerDev, line 16: Maybe `JsonFunctionsEngine` or `JsonFunctionsEvalutaor` is a better name?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculation.kt
```diff
@@ -16,38 +15,33 @@ import de.rki.coronawarnapp.covidcertificate.common.certificate.CwaCovidCertific
...
-    ) {
-        jsonFunctions = JsonFunctions()
+    fun init(boosterRules: List<DccValidationRule>) {
         boosterRulesNode = gson.toJson(boosterRules).toJsonNode()
-        cclConfiguration.logic.jfnDescriptors.forEach {
-            jsonFunctions.registerFunction(it.name, it.definition.toJsonNode())
-        }
     }
 
     fun getDccWalletInfo(
```
LukasLechnerDev, line 34: Maybe make this function a `suspend` function so that we don't need to call `runBlocking{}`

chiljamgossow: if it is a suspend function, the compiler complains about mapper.treeToValue being blocking

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/JsonFunctionsWrapper.kt
chiljamgossow: I renamed it to CclJsonFunctions, as it is specific now

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculation.kt
chiljamgossow: I removed it. Running on the same thread is fine.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculation.kt
LukasLechnerDev: Is it a warning? I don't see anything on my machine...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    fun update(cclConfigurations: List<CCLConfiguration>) {
+        jsonFunctions = create(cclConfigurations)
+    }
+
+    fun evaluateFunction(
+        functionName: String,
+        parameters: JsonNode
+    ) = jsonFunctions.evaluateFunction(functionName, parameters)
```
LukasLechnerDev, line 38: No mutex necessary here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
```diff
@@ -0,0 +1,53 @@
...
+class CclJsonFunctions @Inject constructor(
+    @BaseJackson private val mapper: ObjectMapper,
+    @AppScope private val appScope: CoroutineScope,
+    private val configurationRepository: CCLConfigurationRepository,
+) {
+
+    private lateinit var jsonFunctions: JsonFunctions
+
+    init {
+        appScope.launch {
```
LukasLechnerDev, line 26: Why is the `appScope` necessary here?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculationTest.kt
```diff
@@ -1,35 +1,127 @@
...
+        type = CCLConfiguration.Type.CCLConfiguration,
+        country = "DE",
+        version = "1.0.0",
+        schemaVersion = "1.0.0",
+        engine = "JsonFunctions",
+        engineVersion = "1.0.0",
+        _validFrom = "2021-10-07T00:00:00Z",
+        _validTo = "2030-06-01T00:00:00Z",
+        logic = CCLConfiguration.Logic(jfnDescriptors = listOf(jfnDescriptor))
+    )
```
LukasLechnerDev, line 74: This is not needed for this test, I think? Also the `cclConfigurationRepository`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
mtwalli: to get the configuration you need a coroutine

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculationTest.kt
mtwalli: why ? I need to provide a list of configurations

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
mtwalli: `evaluateFunction` itself has a mutex it is supposed to be thread save

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculationTest.kt
mtwalli: Ah, I see this is another test :D

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
mtwalli: would you prefer using `runBlocking` here ?

LukasLechnerDev: Dagger is creating its object graph on the Main Thread, right - so with the `appScope` the performance would be a bit better since the Main Thread wouldn't need to wait for data from the repo.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
LukasLechnerDev: I think we need one here and another one in the `create()` function. Otherwise, if `evaluateFunction()` and `update()` are called from different threads, something could tried to be evaluated although not all functions have been registered yet.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
mtwalli: So, I leave as is

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclJsonFunctions.kt
mtwalli: mutex added

LukasLechnerDev (dismissed):
LGTM!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/DccWalletInfoCalculation.kt
```diff
@@ -31,10 +34,11 @@ class DccWalletInfoCalculation @Inject constructor(
         boosterRulesNode = gson.toJson(boosterRules).toJsonNode()
     }
 
-    fun getDccWalletInfo(
+    @Suppress("BlockingMethodInNonBlockingContext")
+    suspend fun getDccWalletInfo(
         dccList: List<CwaCovidCertificate>,
         dateTime: DateTime = DateTime.now()
-    ): DccWalletInfo {
+    ): DccWalletInfo = withContext(dispatcherProvider.IO) {
```
LukasLechnerDev, line 41: dispatcherProvider.Default should be used in this case I think because it's not really IO related work that is performed but CPU heavy work. However, the Dispatcher switch should happen in CclJsonFunctions.evaluateFunctions() so that not all callers of this functions need to switch to the default dispatcher before.

mtwalli: 1- if you dig deeper in `writeTree()` you will see that it is actually doing `IO` operations with streams
2- I did not get what you mean by `the Dispatcher switch should happen within CclJsonFunctions.evaluateFunction() so that not all callers of this functions need to switch to the default dispatcher before.`
I think we should switch the dispatcher in a common place where you know actually what context you need. the caller of this function is not aware of that

LukasLechnerDev: 1 - Ah ok, I see - yes then the mapper can work on the IO Dispatcher. However the evaluation with JsonFunctions should happen on the Default Dispatcher IMO. 

2 - I mean that within `CCLJsonFunction` the evaluateFunction looks like this: 

```kotlin

suspend fun evaluateFunction(
        functionName: String,
        parameters: JsonNode
    ) = withContext(dispatcherProvider.Default) {
        ...
    }

```

This way, `DccWalletInfoCalculation` doesn't need to perform the dispatcher switch. Also new Classes that call `evaluateFunction()` (like the system-time-dependent time formatter) don't need to perform that switch.

mtwalli: done

LukasLechnerDev (approved):
Thanks!

BMItr (approved):

PR #4778, opened 2022-02-02
Fix: Root dialog (EXPOSUREAPP-11676)

mtwalli:
How to test ?
- use emulator to show the root dialog  or rooted device 
- increase font size and display size to maximum 
- the dialog should be scrollable 

### Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11676


jurajkusnier (approved):

BMItr (approved):
tested on rooted Pixel 2 Device with largest type settings..
Works!

PR #4781, opened 2022-02-03
Fix QR scanner stuck after scanning manipulated QR code (EXPOSUREAPP-11474)

SamuraiKek:
[Denial of service in CBOR library](https://github.com/advisories/GHSA-fj2w-wfgv-mwq6). Updating to 4.5.1 fixes the issue.

To test, scan the QR attached in the [ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11474), you should see an invalid QR code dialog. Before the fix, the scanner would hang.


Corona-Warn-App/build.gradle
```diff
@@ -487,7 +487,7 @@ dependencies {
     implementation "com.airbnb.android:lottie:3.5.0"
 
     // HCert
-    implementation("com.upokecenter:cbor:4.5")
+    implementation("com.upokecenter:cbor:4.5.1")
```
mtwalli, line 490: this is not the latest , can we upgrade to the latest version [v4.5.2](https://github.com/peteroupc/CBOR-Java/releases/tag/v4.5.2)

mtwalli: latest version works fine too

SamuraiKek: Sure!

mtwalli (approved):
Works

jurajkusnier (approved):

PR #4783, opened 2022-02-03
Bind dccWalletInfo update trigger (EXPOSUREAPP-11481)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11481


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTask.kt
```diff
@@ -15,14 +14,13 @@ import javax.inject.Inject
...
-    private val dccWalletInfoCalculation: DccWalletInfoCalculation,
-    private val dccWalletInfoRepository: DccWalletInfoRepository
+    private val dccWalletInfoCalculationManager: DccWalletInfoCalculationManager,
 ) : Task<DefaultProgress, Task.Result> {
     private val taskProgress = MutableStateFlow<DefaultProgress>(Started)
     override val progress: Flow<DefaultProgress> = taskProgress
 
     override suspend fun run(arguments: Task.Arguments): Task.Result {
-        // Calculate and update DccWalletInfo
+        dccWalletInfoCalculationManager.triggerCalculation()
```
LukasLechnerDev, line 23: Do we really need to trigger the calculation for all persons every time? Can there be situations where we only need to trigger it for a single person?

mtwalli: Might be , but considering the re-grouping the next sprint , where a split could happen I think we should consider re-calculating it for all

LukasLechnerDev (approved):
LGTM 🔫

jurajkusnier (approved):

PR #4792, opened 2022-02-07
Fix language (DEV)

chiljamgossow:


mtwalli (dismissed):
lgtm

mtwalli (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclDefaultInputParameters.kt
```diff
@@ -42,11 +43,15 @@ private val supportedLanguages = arrayOf(
     "tr",
 )
 
-private fun getLanguage(): String {
-    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N) {
-        LocaleList.getDefault().getFirstMatch(supportedLanguages)?.language ?: Locale.getDefault().language
+val cclLanguage: String by lazy {
```
mtwalli, line 46: Could you please check if the language changes if the device language changes while using the App?
I suspect that the value is cached , because this `val`
is for all the whole App

chiljamgossow: Yes, you are right, it won't change. Getting it each time has impact on performance, so let me find a different solution.

chiljamgossow: I tried several things. The only thing that would work is a broadcast receiver for the locale change, but is it worth it? will be reset on next app start, and since we do not calculate the wallet info again, the new language may not be there right away anyway

mtwalli: using map as above did not wrok?

chiljamgossow: it won't work if the user changes the second language because Locale.getDefault().language won't change

chiljamgossow: Until walletInfo is recalculated, the new language is not available anyway, so it may even be worse to update immediately

mtwalli: you mean to update on locale change in this case ?

mtwalli (approved):
Let us merge it for now , We should test it when languages are there

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclDefaultInputParameters.kt
mtwalli: We should keep an eye on that

chiljamgossow: Yes, update the language and trigger a new calculation on locale change

LukasLechnerDev (approved):

PR #4796, opened 2022-02-08
 Update Text for 116117 info telephone

CV113:
The text which recommends to call 116117 for questions needs to be updated.

--> https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11708


KathrinLuecke (approved):
UA approved

mtwalli (approved):

PR #4801, opened 2022-02-08
FIX: Improve App Performance (EXPOSUREAPP-11707)

jurajkusnier:
### Testing
- App-Informationen > Fehlerberichte > Starten
- scan a large amount of certificates 20-30 for different server env. so you have some valid and some invalid certificates
- do some fast tab switches, scanning & recycling certificates
- app shouldn't crash and shouldn't show an ANR error

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11707


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/valueset/internal/DccValueSetMapper.kt
```diff
@@ -9,11 +9,39 @@ import timber.log.Timber
...
+            it.vaccinationValueSets.vp.items.size,
+            it.vaccinationValueSets.mp.items.size,
+            it.vaccinationValueSets.ma.items.size,
+            languageCode,
+            it.testCertificateValueSets.tg.items.size,
+            it.testCertificateValueSets.tt.items.size,
+            it.testCertificateValueSets.ma.items.size,
+            it.testCertificateValueSets.tr.items.size,
+        )
+    }
```
mtwalli, line 44: I think just simple message would be just fine ?

mtwalli (commented):
Thanks for investigating!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/valueset/internal/DccValueSetMapper.kt
jurajkusnier: yes, it could be enough, updated

mtwalli (approved):

chiljamgossow (approved):
LGTM

PR #4810, opened 2022-02-09
Restoring tests over active tests will show result screen immediately (EXPOSUREAPP-11629)

SamuraiKek:
Issue description:

If you tried to scan a test that exists in the recycle bin and you also had an active test of the same type, the app would move the active test into rb and try to restore the scanned test from the rb, but instead of showing the result of the restored test right away, the submission flow would start as if it was a newly scanned test, resulting in an error when completing the flow. 

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11629)

How to test:
- Scan a test and then delete it once you get the result
- Scan another test of the same type
- Scan the previous test. The app will ask you if you want to restore the test and then bring you to the test deletion screen. When pressing continue, the active test will get deleted and the old test result will show up right away. Navigating back will bring you to the main fragment.


LukasLechnerDev (approved):
I tested it and works fine. Code also looks ok.

I don't know if this fix can have unintended side-effects, since I am not familiar with this part of the app. Maybe someone who is more familiar with it can check.

jurajkusnier (approved):
LGTM

PR #4811, opened 2022-02-09
Data structures (EXPOSUREAPP-11749)

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccadmission/model/DccAdmissionCheckScenariosInput.kt
```diff
@@ -0,0 +1,15 @@
+package de.rki.coronawarnapp.ccl.dccadmission.model
```
mtwalli, line 1: please add this package to the `proguard` file

mtwalli (commented):
Looks good 👍🏼

mtwalli (approved):

BMItr (approved):

PR #4812, opened 2022-02-09
FIX: Export test results shown in contact journal (EXPOSUREAPP-11441)

jurajkusnier:
### Testing
- Scan some positive/negative PCR/RAT tests 
- Check tests in the contact journal
- Export Entries
- Tests should be visible in the exported data

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11441

### Contact Journal (Export of Data) Spec
https://github.com/corona-warn-app/cwa-app-tech-spec/blob/main/docs/spec/contact-diary.md#export-of-data


chiljamgossow (commented):
I tested, and it works fine if there are also other entries.
But with RAT tests only, there was no entry in the export. Is that intended?

jurajkusnier:
> I tested, and it works fine if there are also other entries.
> But with RAT tests only, there was no entry in the export. Is that intended?

Fixed and new test case was added

BMItr (approved):
lgtm.
tested on Samsung S20 Device.
image

mtwalli (approved):

DerVogel2020:
My browser cannot find the specs at

https://github.com/corona-warn-app/cwa-app-tech-spec/blob/main/docs/spec/contact-diary.md#export-of-data

Where can I find them?

PR #4813, opened 2022-02-09
Download ccl config and booster rules again if they failed earlier today (DEV)

LukasLechnerDev:
This PR fixes the issue that the `lastExecutionTime` of `CCLConfigurationUpdater` was set although there were failing network requests (either the one to ccl-config or booster rules or both). 

This meant that there were no new retries to get new data for the current day when the app went to foreground. 

This is fixed now. `lastExecutionTime` is now only set if both network requests were successful.

How to test:
- fresh install
- activate airplane mode
- start app
- see failing network requests 
- deactivate airplane mode
- bring app back to foreground
- network requests should be performed again

More Information: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/proposal/logic-engine/docs/spec/dcc-ccl.gen.md#api-for-updating-the-ccl-configuration-from-cdn


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdater.kt
```diff
@@ -57,9 +56,29 @@ class CCLConfigurationUpdater @Inject constructor(
...
+            val boosterRulesDeferred = async { boosterRulesRepository.update() }
+            val cclConfigDeferred = async { cclConfigurationRepository.updateCCLConfiguration() }
+
+            val boosterRulesResult = boosterRulesDeferred.await()
+            val cclConfigResult = cclConfigDeferred.await()
+
+            updateExecutionTimeOnSuccess(boosterRulesResult, cclConfigResult)
+
+            val newBoosterRules = (boosterRulesDeferred.await() == BoosterRulesRepository.UpdateResult.UPDATE)
+            val newCclConfig = (cclConfigDeferred.await() == CCLConfigurationRepository.UpdateResult.UPDATE)
```
mtwalli, line 68: You already have
```kotlin 
    val boosterRulesResult = boosterRulesDeferred.await()
    val cclConfigResult = cclConfigDeferred.await()
```
can we use them here ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/BoosterRulesRepository.kt
```diff
@@ -41,32 +43,46 @@ class BoosterRulesRepository @Inject constructor(
...
         }.let { boosterNotificationRules ->
             boosterNotificationRules.also { Timber.tag(TAG).d("Booster notification rules size=%s: %s", it.size, it) }
         }
 
-        return updated
+        return updateResult
+    }
+
+    enum class UpdateResult {
+        UPDATE, NO_UPDATE, FAIL
```
mtwalli, line 85: can we use one `UpdateResult`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/server/CCLConfigurationServer.kt
```diff
@@ -39,7 +39,7 @@ class CCLConfigurationServer @Inject constructor(
             }
         } catch (e: Exception) {
             Timber.tag(TAG).e(e, "Failed to get ccl configuration")
-            null
+            throw e
         }
     }.also { Timber.tag(TAG).d("Returning %s", it) }
```
jurajkusnier, line 44: Using `ByteArray` parameter in the log seems to be useless as it's not translated into a string. It could be removed, not necessary in this PR, as this is about something else.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdater.kt
LukasLechnerDev: Oh... fixed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/server/CCLConfigurationServer.kt
LukasLechnerDev: You are right. I can remove it with this PR. Is that fine for you @mtwalli ?

mtwalli: Yes, makes sense

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/BoosterRulesRepository.kt
LukasLechnerDev: Yes. Created an enum in `utils.repositories`. I hope this is fine for you?

mtwalli (approved):

jurajkusnier (approved):

PR #4814, opened 2022-02-09
Calculation (EXPOSUREAPP-11757)

chiljamgossow:


mtwalli (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/calculation/CclDefaultInputParameters.kt
```diff
@@ -8,20 +8,20 @@ import org.joda.time.format.ISODateTimeFormat
 import timber.log.Timber
 import java.util.Locale
 
-internal fun getDefaultInputParameters(
+fun getDefaultInputParameters(
```
LukasLechnerDev, line 11: What about moving this file from the `dccwalletinfo` package into the `ccl` root package, since its now not only used by `dccwalletinfo` anymore?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccadmission/calculation/DccAdmissionCheckScenariosTest.kt
```diff
@@ -0,0 +1,53 @@
...
+    private val dateTime = DateTime.parse("2021-12-30T10:00:00.897+01:00")
+    private lateinit var instance: DccAdmissionCheckScenariosCalculation
+
+    @BeforeEach
+    fun setup() {
+        MockKAnnotations.init(this)
+        coEvery { cclJsonFunctions.evaluateFunction("getDCCAdmissionCheckScenarios", NullNode.instance) } returns
+            NullNode.instance
+        every { mapper.treeToValue(any(), DccAdmissionCheckScenarios::class.java) } returns dccAdmissionCheckScenarios
+        every { mapper.valueToTree<JsonNode>(any()) } returns NullNode.instance
```
LukasLechnerDev, line 35: What about returning a mock of `DccAdmissionCheckScenariosInput` here instead of `NullNode`? Would make the test a bit easier to understand.

LukasLechnerDev (approved):

mtwalli (approved):

PR #4826, opened 2022-02-11
Avoid logging the entire ccl config (EXPOSUREAPP-11788)

LukasLechnerDev:
This PR adjusts the logs so that only the CCLConfig Versions and Identifiers are logged instead of the whole config (which is very large). Also one bytestring is not logged anymore, since its not very useful. 

How to test: 
- Perform Force-Update of CCL Configuration in the Dev Menu and check if only the versions and identifiers are logged, and not the whole config.


BMItr (approved):
lgtm!

mtwalli (approved):
LGTM

PR #4830, opened 2022-02-14
Merge default and downloaded config (EXPOSUREAPP-11750)

LukasLechnerDev:
Merges default and downloaded CCL configuration.

No new default config available yet - Check tests for verification.

Details: https://github.com/corona-warn-app/cwa-app-tech-spec/blob/4e297fe221d3524214c94c528def000ff89455f5/docs/spec/dcc-ccl.gen.md#calling-a-ccl-api-with-jsonfunctions


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/storage/CCLConfigurationRepository.kt
```diff
@@ -43,30 +44,30 @@ class CCLConfigurationRepository @Inject constructor(
...
+                Timber.tag(TAG).d("Nothing to update. Keeping old ccl config list")
+                return@updateBlocking this
+            }
 
-            if (rawData != null && newConfig == null) {
-                // parsing failed
+            val downloadedConfigList = downloadedConfigListRaw.tryParseCCLConfigurations() ?: run {
                 updateResult = UpdateResult.FAIL
+                return@updateBlocking this
             }
```
LukasLechnerDev, line 60: Not related to the actual merging. Just some refactoring.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/storage/CCLConfigurationRepository.kt
```diff
@@ -43,30 +44,30 @@ class CCLConfigurationRepository @Inject constructor(
...
-                    newConfig
-                }
-
-                false -> {
-                    Timber.tag(TAG).d("Nothing to update. Keeping old ccl config list")
-                    this
-                }
-            }
+            Timber.tag(TAG).d("Saving new config data")
+            downloadedCclConfigurationStorage.save(rawData = downloadedConfigListRaw)
```
LukasLechnerDev, line 63: We still only store the downloaded config, but we update the `HotDataFlow` with the merged config (see below)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/storage/DownloadedCCLConfigurationStorage.kt
```diff
@@ -8,7 +8,7 @@ import timber.log.Timber
 import java.io.File
 import javax.inject.Inject
 
-class CCLConfigurationStorage @Inject constructor(
+class DownloadedCCLConfigurationStorage @Inject constructor(
```
LukasLechnerDev, line 11: Renamed to `DownloadedCCLConfigurationStorage` to make it clear that only the downloaded config is stored, not the merged one.

mtwalli (approved):
LGTM

chiljamgossow (commented):
Nice test!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/storage/DownloadedCCLConfigurationStorage.kt
LukasLechnerDev: You are right, it should actuallly be camelCase. However we are using the non-camle case CLL everywhere. I would suggest that if we want to change it, we should open another refactoring PR. What do you think @mtwalli

mtwalli: Ye, @chiljamgossow  is right, we should follow camelCase

mtwalli: See https://google.github.io/styleguide/javaguide.html#s5.3-camel-case

LukasLechnerDev: Ok. But let's change it for all classes in a different PR after this one is merged...

chiljamgossow (approved):

PR #4831, opened 2022-02-14
FIX: Wong Output, when accessing OS push notification after pending test results turned positive (EXPOSUREAPP-11776)

jurajkusnier:
This PR fixes `SubmissionTask`. When the user has both PCR and RAT tests waiting for submission and the user performs manual submission on one of these tests, `SubmissionTask` picks only the first test regardless of user selection. The second issue is when the test is in state (`isSubmitted == true`) we display the test result to the user in the `HomeFragment` even when we should display the test in `TestResultReady` state 

### TESTING
- scan pending PCR  and RAT tests without OS-Consent and without warning others
- change tests to positive 
- open later scanned test and perform the submission
- everything should work as expected
- if you have time check if notifications work as expected too, follow the steps in the bug ticket

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11776


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/task/SubmissionTask.kt
```diff
@@ -125,10 +125,12 @@ class SubmissionTask @Inject constructor(
         }
     }
 
-    private suspend fun performSubmission(): Result {
+    private suspend fun performSubmission(testType: CoronaTest.Type?): Result {
         val availableTests = coronaTestRepository.coronaTests.first()
         Timber.tag(TAG).v("Available tests: %s", availableTests)
-        val coronaTest = availableTests.firstOrNull { it.isSubmissionAllowed }
+        val coronaTest = availableTests
+            .filter { testType == null || it.type == testType }
```
LukasLechnerDev, line 132: In which case can `testType` be null here?

jurajkusnier: when automatic submission is scheduled on the background

LukasLechnerDev: I see. Maybe add a comment or have a second function `performSubmissionInBackground()` that is called when the `testType` is `null` ?

jurajkusnier: I've added a comment

mtwalli: all tests have a `type` if you filter with `testType == null` the filter result will empty, right? and this will lead to 
`?: throw IllegalStateException("No valid test available to authorize submission")`

LukasLechnerDev (approved):
Tested it and LGTM

BMItr (commented):
code lgtm,
will approve after second test

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/task/SubmissionTask.kt
mtwalli: um, I see

mtwalli (approved):

PR #4832, opened 2022-02-14
Admission scenarios screen (EXPOSUREAPP-11751)

mtwalli:
## Changes 
- Admission scenarios screen 
- Update admission scenarios and selectedScenario Id after selection
- Trigger DccWalletInfo calculation
- Suppress Error per person calculation to continue for other persons
- Screenshots
- With `2.18` bug fix , UpdateTask became kind of Obsolete -> 🗑️ 
- Reaming 
```diff
-  suspend fun triggerCalculationAfterConfigChange(
+  suspend fun triggerAfterConfigChange(

- suspend fun triggerCalculationAfterCertificateChange(
+ suspend fun triggerNow(
```

## TODO
- [x] unit tests 

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11751


KathrinLuecke (dismissed):
UA approved

chiljamgossow (dismissed):
Looks good to, tested it
The popup seems to be dismissed right away as the calculation is fast. Maybe we can drop it if displaying it is slower than the calculation, but that would be an enhancement

mtwalli:
> Looks good to, tested it
> The popup seems to be dismissed right away as the calculation is fast. Maybe we can drop it if displaying it is slower than the calculation, but that would be an enhancement

That is right for a few person certificates ,but in case there are many persons we need the indicator and maybe in the future Configuration will change and do more work. let us keep as in Figma

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTrigger.kt
```diff
@@ -22,66 +21,65 @@ import javax.inject.Singleton
...
-            DefaultTaskRequest(
-                type = DccWalletInfoUpdateTask::class,
-                arguments = DccWalletInfoUpdateTask.Arguments(
-                    dccWalletInfoUpdateTriggerType = TriggeredAfterConfigUpdate(
-                        configurationChanged
-                    ),
-                    admissionScenarioId = getAdmissionScenarioId()
-                ),
-                originTag = TAG
-            )
```
mtwalli, line 64: Task removed

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModelTest.kt
```diff
@@ -259,6 +260,28 @@ class PersonOverviewViewModelTest : BaseTest() {
         }
     }
 
+    @Test
+    fun `openAdmissionScenarioScreen - success`() {
```
LukasLechnerDev, line 264: These 2 tests are fine. However, I like it a bit better if we would call `triggerNow()` in these tests and verify if `dccWalletInfoCalculationManager.triggerNow()` was called and the correct scenarioId was passed. Then we wouldn't need "@VisibleForTesting" for `admissionScenarioId()` .

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/ui/admission/AdmissionScenariosViewModelTest.kt
```diff
@@ -0,0 +1,95 @@
...
+                subtitle shouldBe "Für dieses Bundesland liegen momentan keine Regeln vor"
+                enabled shouldBe false
+            }
+        }
+    }
+
+    @Test
+    fun selectScenario() {
+        instance().selectScenario("admissionScenarioId")
+        coVerifySequence {
```
LukasLechnerDev, line 81: Maybe also verify here if the correct events were posted? (`Calculation` and `CalculationDone`)?

LukasLechnerDev (dismissed):
Great work! 2 optional comments.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModelTest.kt
mtwalli: this function is not triggering DccWalletInfo ,it calculates the scenarios, but I will verify that :)

mtwalli: updated tests

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/ui/admission/AdmissionScenariosViewModelTest.kt
mtwalli: I tried ,but SingleLiveData with two emissions is really flaky in this test , no clue what is wrong

LukasLechnerDev: Ok. Then let's leave it as is.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/ui/overview/PersonOverviewViewModelTest.kt
LukasLechnerDev: Thanks!

LukasLechnerDev (approved):

chiljamgossow (approved):

PR #4833, opened 2022-02-14
Repository (EXPOSUREAPP-17753)

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccadmission/storage/DccAdmissionCheckScenariosRepository.kt
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+
+    suspend fun save(scenarios: DccAdmissionCheckScenarios) {
+        val json = mapper.writeValueAsString(scenarios)
+        cclSettings.setAdmissionCheckScenarios(json)
+    }
+
+    suspend fun clear() {
+        cclSettings.setAdmissionCheckScenarios("")
```
LukasLechnerDev, line 39: Shouldn't `null` be passed when clearing them instead of `""`?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccadmission/DccAdmissionCheckScenariosTestData.kt
```diff
@@ -1,27 +1,79 @@
-package de.rki.coronawarnapp.ccl.dccadmission.model.storage
+package de.rki.coronawarnapp.ccl.dccadmission.model
```
LukasLechnerDev, line 1: Android Studio tells me that the package name is not right.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccadmission/storage/DccAdmissionCheckScenariosRepository.kt
chiljamgossow: changed it to null, but the data store seems to only accept non-null, so then there it will be an empty string

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccadmission/storage/DccAdmissionCheckScenariosRepository.kt
```diff
@@ -0,0 +1,41 @@
...
+        }
+
+    suspend fun save(scenarios: DccAdmissionCheckScenarios) {
+        val json = mapper.writeValueAsString(scenarios)
+        cclSettings.setAdmissionCheckScenarios(json)
+    }
+
+    suspend fun clear() {
+        cclSettings.setAdmissionCheckScenarios(null)
+    }
```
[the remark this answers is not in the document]
chiljamgossow: When the calculation fails (eg. after a config change)? Save does not accept null

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccadmission/storage/DccAdmissionCheckScenariosRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+
+@Suppress("BlockingMethodInNonBlockingContext")
+class DccAdmissionCheckScenariosRepository @Inject constructor(
+    private val cclSettings: CCLSettings,
+    @BaseJackson private val mapper: ObjectMapper
+) {
+
+    val admissionCheckScenarios: Flow<DccAdmissionCheckScenarios?> =
+        cclSettings.admissionCheckScenarios.map {
+            it?.let { json ->
```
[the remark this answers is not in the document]
chiljamgossow: changed the logging, clear is also gone

LukasLechnerDev (approved):
LGTM

mtwalli (approved):

PR #4838, opened 2022-02-15
Fix booster notification background trigger (EXPOSUREAPP-11804)

LukasLechnerDev:
This PR fixes the issue that the booster notifications are not triggered by our background worker. 

Because of a copy-paste mistake of mine (in `WorkerBinder.kt`), the background worker did actually never run...

How to test: 
- Fresh install
- Start app
- Switch to WRU and set "fake correct device time" in Remote Config Data Test Menu (Not sure if this switch is really necessary to set
- Close app
- Set device time to 2022-02-10 12:30
- Open App
- Force ccl config update (config from WRU and INT are different)
- Scan certificate from the ticket
- See that booster notification shows up
- Close app and click on notification
- Close app
- Set device time 24h into the future (2022-02-11 12:30)
- Without opening the app, the worker should start its work and the booster notificaiton should appear

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdateScheduler.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdateScheduler.kt
@@ -1,13 +1,12 @@
 package de.rki.coronawarnapp.ccl.configuration.update
 
-import androidx.work.Constraints
 import androidx.work.ExistingPeriodicWorkPolicy
-import androidx.work.NetworkType
 import androidx.work.PeriodicWorkRequest
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import de.rki.coronawarnapp.util.coroutine.AppScope
 import de.rki.coronawarnapp.util.device.ForegroundState
+import de.rki.coronawarnapp.worker.BackgroundConstants
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filter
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdateScheduler.kt
```diff
@@ -48,14 +47,13 @@ class CCLConfigurationUpdateScheduler @Inject constructor(
...
     private fun buildWorkRequest(): PeriodicWorkRequest {
-        val constraints = Constraints.Builder()
-            .setRequiredNetworkType(NetworkType.CONNECTED)
-            .build()
-
         return PeriodicWorkRequestBuilder<CCLConfigurationUpdateWorker>(
             repeatInterval = 24,
             repeatIntervalTimeUnit = TimeUnit.HOURS
-        ).setConstraints(constraints).build()
+        ).setInitialDelay(
```
LukasLechnerDev, line 53: Not related to the bug, but I added a 1 minute delay for the worker so that the app and the worker won't both update the config after app start at the same time.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdateScheduler.kt
```diff
@@ -48,14 +47,13 @@ class CCLConfigurationUpdateScheduler @Inject constructor(
     }
 
     private fun buildWorkRequest(): PeriodicWorkRequest {
-        val constraints = Constraints.Builder()
```
LukasLechnerDev, line 51: We shouldn't only run the worker when we have network connection, since the dccwallet recalculation should run daily even if we can't get the config from the network.

mtwalli (approved):
LGTM

jurajkusnier (approved):
LGTM

PR #4844, opened 2022-02-16
CCL not recalculated base on validuntil (EXPOSUREAPP-11850)

LukasLechnerDev:
Our implementation did the `validUntil` check with the same cadence as the ccl-config / boosterRule download (once a day).

However, it is specified that this check should happen every time the app gets into the foreground. This is now fixed with this PR. 

How to test:
- "triggerDccWalletInfoUpdateAfterConfigUpdate()" should now be logged every time the app gets into the foreground. 
- If you have time and know how to replace the ccl-config attached to the Jira ticket with the mock server, it would be great to test it as described in ticket

### Jira ticket: 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11850


mtwalli (approved):
LGTM

jurajkusnier (approved):
LGTM

PR #4846, opened 2022-02-17
Read appconfig flag to control admission scenarios (EXPOSUREAPP-11801)

LukasLechnerDev:
The new admission scenario tile is now enabled / disabled depending on the `dcc-admission-check-scenarios-disabled` appconfig flag.

Details: https://github.com/corona-warn-app/cwa-app-tech-spec/pull/157/commits/c0cb645ae4dfddc2a6e4e441248b7e6cd397b729

I also did some refactorings in `CWAConfigMapper` its associated test.

Flag can be enabled/disabled via Mock-Server:

grafik

in text, thanks to Safari:
```js 
cwa app-config set ccl-admission-check-scenarios-disabled 0 --android -e cf
cwa app-config set ccl-admission-check-scenarios-disabled 1 --android -e cf
```


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -33,34 +34,15 @@ class CWAConfigMapper @Inject constructor() : CWAConfig.Mapper {
...
-            (0 until appFeatures.appFeaturesCount)
-                .map { appFeatures.getAppFeatures(it) }
-                .firstOrNull { it.label == "unencrypted-checkins-enabled" }
-                ?.let { it.value == 1 }
-                ?: false
-        } catch (e: Exception) {
-            Timber.e(e, "Failed to map `unencrypted-checkins-enabled` from %s", this)
-            false
-        }
-    }
```
LukasLechnerDev, line 63: Created the new extension function `findBoolean()` to avoid duplicated code like this.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -103,6 +85,11 @@ class CWAConfigMapper @Inject constructor() : CWAConfig.Mapper {
         }
     }
 
+    private fun ApplicationConfigurationAndroid.dccAdmissionCheckScenariosDisabled() = findBoolean(
```
LukasLechnerDev, line 88: Use the new extension function `findBoolean()` for the new flag.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapperTest.kt
```diff
@@ -62,69 +62,96 @@ class CWAConfigMapperTest : BaseTest() {
...
-                        }.build()
-                    )
-                }
-            )
-            .build()
-        createInstance().map(rawConfig).apply {
-            isDeviceTimeCheckEnabled shouldBe false
-            isUnencryptedCheckInsEnabled shouldBe false
-        }
+    fun `findBoolean() should return true on 1`() {
```
LukasLechnerDev, line 65: Added some tests for `findBoolean()` which also cover the edge-cases when the flag contains an invalid value or an exception is thrown. These cases are already covered, so we don't need tests for these edge cases for every flag that uses `findBoolean()`.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapperTest.kt
```diff
@@ -237,34 +264,7 @@ class CWAConfigMapperTest : BaseTest() {
     }
 
     @Test
-    fun `feature, unencrypted checkins disabled`() {
```
LukasLechnerDev, line 240: I removed some test cases like this because they were covered by other tests above already.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -103,6 +84,31 @@ class CWAConfigMapper @Inject constructor() : CWAConfig.Mapper {
...
+        return try {
+            val value = (0 until appFeatures.appFeaturesCount)
+                .map { appFeatures.getAppFeatures(it) }
+                .firstOrNull { it.label == labelValue }
+                ?.value
+
+            return when (value) {
+                null -> defaultValue // no item found
+                1 -> true // found item has value '1'
+                else -> defaultValue // found item has value different to '1'
```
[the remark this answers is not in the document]
LukasLechnerDev: The code that you posted has a problem that it returns `false` for every `it.value` different to `1` but it should return `defaultValue` in this case.

This was a very small bug that we had, but it wasn't a problem, since every default value that we have is `false`.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/CWAConfigMapper.kt
```diff
@@ -103,6 +84,31 @@ class CWAConfigMapper @Inject constructor() : CWAConfig.Mapper {
         }
     }
 
+    private fun ApplicationConfigurationAndroid.dccAdmissionCheckScenariosDisabled() = findBoolean(
+        labelValue = "dcc-admission-check-scenarios-disabled",
+        defaultValue = DCC_ADMISSION_CHECK_SCENARIOS_DISABLED
```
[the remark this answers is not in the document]
LukasLechnerDev: Removed it completely, see other comment.

mtwalli:
@LukasLechnerDev  did you push your latest changes ? I don't see them here yet

LukasLechnerDev:
@mtwalli Ah, you are right. I forgot to push the "Add default value" commit. Its now pushed.

mtwalli (approved):
Thanks, LGTM

chiljamgossow (approved):
Looks good, thanks for refactoring 👍

PR #4848, opened 2022-02-17
Adapt privacy notice for 2.19 (EXPOSUREAPP-11616)

LukasLechnerDev:
How to test:
- Check if privacy notice in german, english and tr works fine


vaubaehn:
Hi @LukasLechnerDev , all the text in the privacy note concerning the reissuance API is a draft and inserted as a placeholder, isn't it?

LukasLechnerDev:
Hi @vaubaehn ! Thanks for finding this! 👍

Yes the texts still contain placeholders so I will update them as soon as we get new ones.

vaubaehn:
> Hi @vaubaehn ! Thanks for finding this! 👍
> 
> Yes the texts still contain placeholders so I will update them as soon as we get new ones.

@LukasLechnerDev Great! Yes, there are placeholders _in_ the text. But the text _itself_, that's not the final version from the legal department, is it?

mtwalli (dismissed):

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1484,6 +1526,12 @@ <h3>
     b. Daten auf Serversystemen
 
 </h3>
+<p>
+    Im Falle der Aktualisierung eines digitalen COVID-Zertifikats (siehe Abschnitt 6.o) werden die
+    für die Aktualisierung an das Serversystem der App übermittelten digitalen COVID-Zertifikate
+    nach Abschluss des Aktualisierungsvorgangs wieder gelöscht. Der Hash-Wert der elektronischen
+    Signatur des aktualisierten digitalen COVID-Zertifikats wird nach X ZEITABSCHNITTE gelöscht.
```
vaubaehn, line 1533: Hi  @LukasLechnerDev ,
in this line there is still the "365 days" missing.

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1716,6 +1764,16 @@ <h3>
...
 </p>
+<h3>
+    i. Einverständnis „Aktualisierungsfunktion für COVID-Zertifikate“
+</h3>
+<p>
+    Da die Aktualisierung eines digitalen COVID-Zertifikats nach der Erteilung Ihres
+    Einverständnisses sofort durchgeführt wird, kann das Einverständnis nicht mehr zurückgenommen
+    werden. Wenn die Aktualisierung nicht erfolgreich war oder Sie ein weiteres digitales
+    COVID-Zertifikat aktualisieren möchten, müssen Sie für jeden einzelnen Aktualisierungsvorgang
+    ein gesondertes Einverständnis erteilen.
```
vaubaehn, line 1775: Just as a technical remark:
It depends on _when_ in the process Ubirch stores the hash:
If they store it right at the beginning of the reissuance process, and there is a loss of internet connection before the new certificate can be sent to the client, then the user won't be able to trigger the reissuance again (fyi @mlenkeit ) as the already stored hash then will prevent a new reissuance (as an anti-measure for (D)DoSing).
If this is the case, UX/UI should have some words for clarification (e.g., "Due to an error, your DCC can't be reissued. Please see your pharmacist for further support.) It would be interesting, whether in pharmacies then reissuing would also be blocked due to the stored hash...

If Ubirch stores the hash _after_ the **successful** submission to the client, then DDoSing with spoofed sender IPs would still be possible.

If the 365-days period is rather to avoid abuse, then it could be good to discuss to shorten that period, e.g. to 24 hours, what should be enough to prevent (D)DoSing.
Unless Ubirch or other stakeholders don't charge the BMI for any re-issued certificate (per certificate), I don't see any downside to lower the storage period to 24 hours.
Probably the most common/likely case for re-issuance will be, when the DCC is experied from the period of _technical_ validity (COSE header). DCCs with expired technical validity could _always_ be reissued.

cc: @thinkberg

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1762,6 +1820,11 @@ <h2>
     Ihnen bereitgestellten Informationen zu Ihrer Identität in der Regel nicht
     erfüllt werden können.
 </p>
+<p>
+    Wenn anlässlich der Aktualisierung eines digitalen COVID-Zertifikats der Hash-Wert der
+    elektronischen Signatur vorübergehend gespeichert wird, ermöglicht es dieser dem RKI nicht, die
```
vaubaehn, line 1825: "vorübergehend" is a bit misleading when it's actually 365 days. Yes, one year is also "temporary" in a way, but the concrete storage period would be better in this place.
If the legal department has some capacities to have a look into it, maybe they can agree to state "365 days" here also.

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1264,6 +1268,44 @@ <h3>
...
+    auf ein vorhandenes digitales COVID-Zertifikat anzuwenden.
+</p>
+<p>
+    Dazu lädt die App im Hintergrund die aktuellen inhaltlichen und technischen Spezifikationen für
+    die digitalen COVID-Zertifikate vom Serversystem der App herunter. Die App überprüft dann anhand
+    der in den von Ihnen gespeicherten digitalen COVID-Zertifikaten enthaltenen Daten, ob eine
+    Aktualisierung eines gespeicherten digitalen COVID-Zertifikats über die App möglich ist. Sie
+    werden dann auf eine mögliche Aktualisierung hingewiesen. Für die Aktualisierung wird mit Ihrem
+    Einverständnis das zu aktualisierende digitale COVID-Zertifikat von der App verschlüsselt an das
+    Serversystem der App übermittelt. Wenn es sich bei dem zu aktualisierenden digitalen
```
vaubaehn, line 1288: This part reads, as if the reissuance will only be triggered automatically by the app, the user will be notified and then the reissuing can be done once - because after receiving a reissued certificate, any additional reissuing won't be necessary anymore.
See next comment where I see a small confusion.

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1264,6 +1268,44 @@ <h3>
...
+<p>
+    Um eine missbräuchliche mehrfache Verwendung der Aktualisierungsfunktion verhindern zu können,
+    wird die elektronische Signatur des aktualisierten digitalen COVID-Zertifikats in Form einer
+    speziellen Einwegverschlüsselung (sog. Hash-Wert) vorübergehend auf dem Serversystem der App
+    gespeichert. Das bedeutet, dass die elektronische Signatur nach einem bestimmten mathematischen
+    Verfahren so verfremdet wird, dass sie nicht mehr erkennbar ist. Aus dem Hash-Wert kann nicht
+    auf die Daten im betreffenden digitalen COVID-Zertifikat oder andere Angaben zu Ihrer Person
+    geschlossen werden. Die eindeutige Zuordnung des Hash-Werts zu dem aktualisierten digitalen
+    COVID-Zertifikat ist aber weiterhin möglich. Nach 365 Tagen wird der Hash-Wert wieder gelöscht.
+</p>
```
vaubaehn, line 1308: This part reads, as if the user would be able to initiate a reissuing at any time, in contrast to what is stated above.
Yes, in the end, both paragraphs can be true at the same time, independently from how the reissuance procedure is implemented into the app. But maybe a better clarification could be good here. E.g.,
"Um eine missbräuchliche mehrfache Verwendung der Aktualisierungsfunktion durch automatisierte System verhindern zu können, ..."
Then even if reissuing is implemented in the way, that "app decides whether to offer reissuing to the user or not, and to present a "temporary button" for reissuing", then it gets clear to the user, why at all it is necessary to store the DCC hash(es).

vaubaehn (changes requested):
Hi @LukasLechnerDev ,
thanks for your kind feedback. At one place there is still a placeholder.
For the other comments: @mlenkeit should decide whether further discussion with Ubirch is necessary to decide if UI (information to the user) should be extended about hash storing times, and possible surprises when a DCC can't be reissued again in a second try, in case during the first try the re-issued DCC could not be submitted to the client successfully.
Also, there are some other parts, where the legal department could have a look, whether some enhancements could be applied to the text.
Thank you!

FYI: @Ein-Tim

Corona-Warn-App/src/main/assets/privacy_de.html
Ein-Tim:
```suggestion
    Signatur des aktualisierten digitalen COVID-Zertifikats wird nach 365 Tagen gelöscht.
```

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1264,6 +1268,44 @@ <h3>
...
+    der App übermittelt. Dort wird dann die Gültigkeit der übermittelten digitalen COVID-Zertifikate
+    geprüft (siehe Abschnitt 6.h.). Anhand der Daten aus den übermittelten digitalen
+    COVID-Zertifikaten wird eine neue Version des COVID-Impfzertifikats für die Auffrischungsimpfung
+    generiert. Bei der Aktualisierung werden die Daten in den übermittelten COVID-Zertifikaten
+    (siehe Abschnitte 5.h und 5.i) kurzzeitig verarbeitet. Das aktualisierte COVID-Zertifikat wird
+    sodann an Ihre App zurückübermittelt.
+</p>
+<p>
+    Um eine missbräuchliche mehrfache Verwendung der Aktualisierungsfunktion verhindern zu können,
+    wird die elektronische Signatur des aktualisierten digitalen COVID-Zertifikats in Form einer
```
vaubaehn, line 1301: There is another point, that @timokoenig was stumbling upon:
Is it really true that the hash of the **updated** DCC is stored despite the hash of the DCC **to be updated**?
We wonder, if the statement is true, what would be the positive effect?
See also: https://github.com/Digitaler-Impfnachweis/certification-apis/issues/231#issuecomment-1056815201

LukasLechnerDev:
@vaubaehn Thanks for your feedback. We got a new version from our lawyers.

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1298,13 +1298,14 @@ <h3>
 </p>
 <p>
     Um eine missbräuchliche mehrfache Verwendung der Aktualisierungsfunktion verhindern zu können,
-    wird die elektronische Signatur des aktualisierten digitalen COVID-Zertifikats in Form einer
-    speziellen Einwegverschlüsselung (sog. Hash-Wert) vorübergehend auf dem Serversystem der App
+    werden die elektronischen Signaturen des aktualisierten und des neuen digitalen
```
vaubaehn, line 1301: This is even more confusing than the first version:
In the first version for us it's not clear why the hash of the updated DCC is stored, instead of the hash of the DCC to be updated. Now, the text handles an updated and additionally a new DCC?

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -1529,8 +1530,9 @@ <h3>
 <p>
     Im Falle der Aktualisierung eines digitalen COVID-Zertifikats (siehe Abschnitt 6.o) werden die
     für die Aktualisierung an das Serversystem der App übermittelten digitalen COVID-Zertifikate
-    nach Abschluss des Aktualisierungsvorgangs wieder gelöscht. Der Hash-Wert der elektronischen
-    Signatur des aktualisierten digitalen COVID-Zertifikats wird nach X ZEITABSCHNITTE gelöscht.
+    nach Abschluss des Aktualisierungsvorgangs wieder gelöscht. Die Hash-Werte der elektronischen
+    Signaturen des aktualisierten und des neuen digitalen COVID-Zertifikats werden nach 365 Tagen
```
vaubaehn, line 1534: same like above

vaubaehn (commented):
Hi @LukasLechnerDev ,

thanks for your notification of the  updated texts.
I couldn't find any forgotton placeholder anymore.
I see, storage time of hashes is still 365 days. This leaves open the question, whether it is misleading to name storing period "vorübergehend"/temporary on other places.
The question, whether the hash of the DCC to be updated or actually the hash of the updated DCC has to be stored is now even more confusing, as the text handles stored hashes of an "updated DCC" and additionally a "new DCC"?
I (and probably others) still do not understand why the hash of the updated DCC shall be stored instead of the DCC to be updated? There is no difference in terms of abuse by multi-usage of the same DCC in the wild, but a difference in mitigating (D)DoS, which seems to be the reason why storing of hashes had been introduced at all - according to the rest of the text.
@mlenkeit Could you help out here and describe in short, what is the reasoning of storing the hash of the **updated*** DCC (instead of to be updated DCC) and what is the "new certificate" in addition to the updated certificate?
Thanks a lot!

FYI: @timokoenig @Ein-Tim

Corona-Warn-App/src/main/assets/privacy_de.html
Ein-Tim: In my understanding: aktualisiertes Zertifikat = neues Zertifikat

vaubaehn: Yes, and I don't see a good reason to introduce a redundancy here.
My gut feeling is, that legal department has problems to understand what is the subject actually about. Maybe there was a talk with them that it should be "certificate to be updated" instead of "updated certificate", but they didn't understand correctly and added a "new certificate"? 🤔

timokoenig: I think I have read somewhere that there are two cases that could explain this naming. In some cases extending a certificate is handled as "UPDATE" and wrongly issued certificates will create "NEW" certificates. Even if this is right, this would noch explain the reason for the confusion about the hashing

timokoenig: Or maybe a possible PCR test would create a new recovery certificate.

(these are all just wild guesses 😅)

mlenkeit:
@vaubaehn @timokoenig the wording - looking at it in isolation - is a but confusing, I agree. As far as I know, the hash of the certificate that is to be replaced is stored. But that's something that Ubirch would need to confirm, see https://github.com/Digitaler-Impfnachweis/certification-apis/issues/231.

We're checking with legal if we need to adjust the texts.

vaubaehn:
Hi @mlenkeit , thanks for your feedback,

> As far as I know, the hash of the certificate that is to be replaced is stored.

That is how we also think it would be correct - but then the related text in the privacy note is wrong. It would then need to change it to "zu aktualisierendes Zertifikat" / "certificate to be updated".

> But that's something that Ubirch would need to confirm, see [Digitaler-Impfnachweis/certification-apis#231](https://github.com/Digitaler-Impfnachweis/certification-apis/issues/231).

I guess you're in touch with them - but would it help you to place another ping in that issue?

Corona-Warn-App/src/main/assets/privacy_de.html
```diff
@@ -113,17 +117,17 @@ <h2>
...
+    App hinzufügen (siehe hierzu Punkt 6.h.) oder aktualisieren (siehe hierzu Punkt 6.o) oder im
+    Rahmen einer Ticketbuchung überprüfen lassen (siehe hierzu Punkt 6.k.).
 </p>
 <p>
     Die App verzichtet standardmäßig auf die Auswertung Ihres Nutzungsverhaltens durch
     Analyse-Tools. Nur wenn Sie ausdrücklich der freiwilligen Datenspende zustimmen oder einen
     Fehlerbericht aufzeichnen und mit dem RKI teilen, werden bestimmte Daten über Ihre Nutzung der
-    App an das RKI übermittelt (siehe hierzu Punkt 5 k. und Punkt 5 m.).
+    App an das RKI übermittelt (siehe hierzu Punkt 5 j. und Punkt 5 m.).
 </p>
```
chiljamgossow, line 131: Should this also be 5.j and 5.m as above?

Corona-Warn-App/src/main/assets/privacy_en.html
```diff
@@ -117,24 +120,17 @@ <h2>
...
-    Section 6 j.).
-</p>
-<p>
-    Exceptions apply to the feature for proving a rapid test result, which allows you to display a
-    confirmation issued in your name for negative rapid test results (see Section 6 c.), the feature
-    for creating a rapid test profile, which allows you to provide a testing point with the data
-    required to perform a rapid test (see Section 6 d.) and if you add digital COVID certificates in
-    the app.
+    certificates in the app (see Section 6 h.) or update them (see Section 6 o.) or have these
+    verified when booking tickets (see Section 6 k.).
```
chiljamgossow, line 128: same here 6.h and 6.k

Corona-Warn-App/src/main/assets/privacy_en.html
```diff
@@ -117,24 +120,17 @@ <h2>
...
 </p>
 <p>
     The app refrains by default from using analysis tools to evaluate the way you use it. Only if
     you expressly agree to voluntarily share data or to record an error report and share it with the
-    RKI (see Sections 5 k. and 5 m.), will certain data about your use of the app be transmitted to
+    RKI (see Sections 5 j. and 5 m.), will certain data about your use of the app be transmitted to
     the RKI.
```
chiljamgossow, line 134: here as well

Corona-Warn-App/src/main/assets/privacy_en.html
```diff
@@ -1248,6 +1244,44 @@ <h3>
...
+    possible update. For the update, with your consent, the digital COVID certificate to be updated
+    will be transmitted from the app to the app’s server system in encrypted form. If the digital
+    COVID certificate to be updated is a COVID vaccination certificate proving a booster
+    vaccination, then the digital COVID certificates proving the basic immunisation or recovery
+    prior to the booster vaccination will also be transmitted to the app’s server system in the
+    process. The validity of the transmitted digital COVID certificates will then be checked there
+    (see Section 6 h.). A new version of the COVID vaccination certificate for the booster
+    vaccination will be generated using the data from the transmitted digital COVID certificates.
+    During the update, the data in the transmitted COVID certificates (see Sections 5 h. and 5 i.)
+    will be processed for a short time. The updated COVID certificate will then be sent back to your
```
chiljamgossow, line 1271: ...and here

Corona-Warn-App/src/main/assets/privacy_en.html
```diff
@@ -1755,6 +1806,16 @@ <h2>
...
+<p>
+    If the hash value of the electronic signature is temporarily stored when a digital COVID
+    certificate is updated, this does not enable the RKI to determine the identity of certificate
+    holders (see Section 6 o.).
+</p>
+<p>
+    If the hash values of the electronic signatures are temporarily stored when a digital COVID
+    certificate is updated, this does not enable the RKI to determine the identity of certificate
+    holders (see Section 6 o.).
+</p>
```
chiljamgossow, line 1818: and here

chiljamgossow (approved):
Spotted some inconsistencies, but not a blocker

mtwalli:
@vaubaehn we have to include this PR in the next RC, please file a separate issue to avoid loosing the comments here ?

mtwalli (approved):

vaubaehn:
Hi @mtwalli ,

> @vaubaehn we have to include this PR in the next RC, please file a separate issue to avoid losing the comments here ?

Thanks for the notification.

Hi @dsarkar , I'm a bit short in time currently, could you please support here and open an issue in the documentation repo?
The review comments (except the ones from Chilja) in here would need to be copied to there.
They are also in context with questions in this issue: https://github.com/Digitaler-Impfnachweis/certification-apis/issues/231
So a link to there could be good.
I'd come back later to the new issue in docu repo and have a look whether everything is complete.
Thanks in advance!

dsarkar:
@vaubaehn Thanks, will do.

PR #4851, opened 2022-02-17
"Impfstatus" shows wrong infos (EXPOSUREAPP-11833)

BMItr:
This PR addresses [EXPOSUREAPP-11833](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11833)

- Delay helpers have been removed
- DCCWalletInfo is triggered if there were changes that lead to `DccWalletInfo` calculation such as:
   - Register Dcc
   - Recycle Dcc
   - Restore Dcc
   - Retrieve Dcc
   
- unification of some structures
- potentially better performance

## How test ?
- Scan certificates for many persons as many as possible and scan certificates for each person as many as possible 
- Scan certificate from the ticket and follow the steps that triggers change in G-status
- Make sure that changes that are not intended to trigger` DccWalletInfo` update such as `badge / and validity state changes` don't trigger dcc wallet info calculation


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/ccl/CCLTestViewModel.kt
```diff
@@ -27,20 +22,16 @@ class CCLTestViewModel @AssistedInject constructor(
...
+    val personIdentifiers = certificateProvider.certificateContainer
+        .map { certificateContainer ->
+            certificateContainer.allCwaCertificates.map { it.personIdentifier }
+                .distinct()
+                .map { PersonIdentifierSelection.Selected(it) }
+                .toMutableList<PersonIdentifierSelection>()
+                .apply { add(0, PersonIdentifierSelection.All) }
+                .toList()
+        }
+        .asLiveData2()
```
mtwalli, line 34: this has been removed in `2.19` please keep change minimal that address the bug ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTrigger.kt
```diff
@@ -2,16 +2,41 @@ package de.rki.coronawarnapp.ccl.dccwalletinfo.update
...
 class DccWalletInfoUpdateTrigger @Inject constructor(
-    private val taskController: TaskController
+    private val taskController: TaskController,
+    certificateProvider: CertificateProvider,
+    @AppScope appScope: CoroutineScope,
+    dispatcherProvider: DispatcherProvider
 ) {
 
+    init {
+        certificateProvider.certificateContainer
```
mtwalli, line 30: I don't think that:
1- This is the right provider it should `PersonCertificatesProvider` , not `CertificatesProvider` we should calculate the DccWalletInfo after grouping the certificates. Relying on the certificates repos to trigger calculation is not right unless we do the re-grouping here too without waiting for `PersonCertificatesProvider` to do that 

2- You do trigger the calculation on `EVERY` change in the flow (or latest since you are using `conflate` ) ,but that is still not right , calculation should be triggered on changes for the following events:
   - Register a Dcc
   - Restore a Dcc
   - Recycle a Dcc
   - Retrieve a Dcc
   
   and in `2.19` after new re-grouping

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
```diff
@@ -2,28 +2,74 @@ package de.rki.coronawarnapp.covidcertificate.common.certificate
...
-            vcRepo.cwaCertificates.first()
-
+        val certificates = certificateContainer.first().allCwaCertificates
         return certificates.find { it.containerId == containerId }!! // Must be a certificate
     }
+
+    data class CertificateContainer(
+        val recoveryCertificates: Set<RecoveryCertificateWrapper>,
+        val testCertificates: Set<TestCertificateWrapper>,
+        val vaccinationInfos: Set<VaccinatedPerson>
```
mtwalli, line 56: I don't think we need a container for all certificates in the app which use cases are using this ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
```diff
@@ -2,28 +2,74 @@ package de.rki.coronawarnapp.covidcertificate.common.certificate
...
+            testCertificates.mapNotNull { it.testCertificate }.toSet()
+        }
+
+        val vaccinationCwaCertificates: Set<VaccinationCertificate> by lazy {
+            vaccinationInfos.flatMap { it.vaccinationCertificates }.toSet()
+        }
+
+        val allCwaCertificates by lazy {
+            recoveryCwaCertificates + testCwaCertificates + vaccinationCwaCertificates
+        }
```
mtwalli, line 73: be careful with using `val` once a value is assigned it won't change , does it update its value when certificates  changes like adding / removing ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesProvider.kt
```diff
@@ -33,16 +28,15 @@ class PersonCertificatesProvider @Inject constructor(
...
-        testCertificateRepository.cwaCertificates,
-        recoveryCertificateRepository.cwaCertificates,
+        certificatesProvider.certificateContainer,
         personCertificatesSettings.currentCwaUser.flow,
         dccWalletInfoRepository.personWallets
-    ) { vaccPersons, tests, recoveries, cwaUser, personWallets ->
+    ) { certificateContainer, cwaUser, personWallets ->
+
+        val allCerts = certificateContainer.allCwaCertificates
+        val vaccPersons = certificateContainer.vaccinationInfos
```
mtwalli, line 37: I see we still need to use the vaccination repo even though certificates have been provided by `certificateContainer` I prefer to keep it as before

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTrigger.kt
BMItr: is the _PersonCertificatesProvider_ doing the regrouping in 2.19 ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
BMItr: The use case which I replaced with the container?
Most use-cases only need the cwa certificates, so I could make the rest private. But there are some special cases like the PersonCertificatesProvider, which also needs the VaccinationInfos. So why shouldn't we keep them accessible?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
BMItr: I don't get your point here. The container is immutable. It is not possible to add or remove certificates to/from the container. If the certificates in a repo change, a new container gets created..

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesProvider.kt
BMItr: please,  isnt it cleaner to have a single source instead of importing every cert repo and compare their flows everytime?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTrigger.kt
mtwalli: Yes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/configuration/update/CCLConfigurationUpdater.kt
```diff
@@ -47,7 +47,7 @@ class CCLConfigurationUpdater @Inject constructor(
         triggerRecalculation(configurationChanged = updated)
     }
 
-    private fun triggerRecalculation(configurationChanged: Boolean) {
+    private suspend fun triggerRecalculation(configurationChanged: Boolean) {
```
LukasLechnerDev, line 50: `suspend` modifier is not necessary here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTask.kt
```diff
@@ -36,20 +34,19 @@ class DccWalletInfoUpdateTask @Inject constructor(
...
-
         return object : Task.Result {}
     }
 
     override suspend fun cancel() = Unit
 
     data class Arguments(
         val dccWalletInfoUpdateTriggerType: DccWalletInfoUpdateTriggerType,
-        val startDelay: Long = 1_000L
+        val admissionScenarioId: String = ""
```
LukasLechnerDev, line 44: `admissionScenarioId` can be removed.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTrigger.kt
```diff
@@ -1,17 +1,48 @@
...
+    private val taskController: TaskController,
+    private val dccWalletInfoCalculationManager: DccWalletInfoCalculationManager,
+    private val dccWalletInfoCleaner: DccWalletInfoCleaner
 ) {
 
+    init {
+        appScope.launch {
+            personCertificateProvider.personCertificates
+                .drop(1) // Drop first value on App start that causes unnecessary calculation
+                .distinctUntilChangedBy { it.sortedQrCodeHashSet }
```
LukasLechnerDev, line 34: Maybe a short comment about which certificate changes will lead to a recalculation would be helpful here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTask.kt
```diff
@@ -36,20 +34,18 @@ class DccWalletInfoUpdateTask @Inject constructor(
...
 
     data class Arguments(
         val dccWalletInfoUpdateTriggerType: DccWalletInfoUpdateTriggerType,
-        val startDelay: Long = 1_000L
     ) : Task.Arguments
 
-    sealed class DccWalletInfoUpdateTriggerType {
-        object TriggeredAfterCertificateChange : DccWalletInfoUpdateTriggerType()
-        data class TriggeredAfterConfigUpdate(val configurationChanged: Boolean) : DccWalletInfoUpdateTriggerType()
+    sealed interface DccWalletInfoUpdateTriggerType {
```
LukasLechnerDev, line 46: `TriggerType` is now also not needed anymore. We currently only use this Tasks for Config Updates. But we can do this clean up in a subsequent PR as well.

LukasLechnerDev (approved):
Tested it with many certificates. `DccWalletInfo` is only recalculated when it should. 

Very nice work @BMItr and @mtwalli . That's a big improvement 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTask.kt
mtwalli: Yes, but I want to keep the changes minimal, as I mentioned in the daily I would like to merge this merge and do the other changes in another PR ,since it is not really to fixing / optimising the performance

LukasLechnerDev: fine for me!

mtwalli (approved):

PR #4857, opened 2022-02-18
Fix: App crashes when certificate is scanned with malformed data (EXPOSUREAPP-11924)

jurajkusnier:
Parsing `LocalDate` from `String` could end up with an exception when the text is not in the expected format. We should at least catch this exception and don't let the app crash. 

### Testing
- change server environment to MOCK-CLOUD
- scan malformed certificate

image

- App shouldn't crash
- Scan more VC/TC/RC to be sure that everything works as before

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11924


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/VaccinationCertificateCard.kt
```diff
@@ -37,7 +37,7 @@ class VaccinationCertificateCard(parent: ViewGroup) :
 
         certificateDate.text = context.getString(
             R.string.vaccination_certificate_vaccinated_on,
-            certificate.vaccinatedOn.toShortDayFormat()
+            certificate.vaccinatedOn?.toShortDayFormat()
```
mtwalli, line 40: this will display `null` right ? is this what is required ?

jurajkusnier: yes, it displays "null", required is to not let the app crash

chiljamgossow: Certificates used to be rejected when the dates where malformed -> no scan possible

jurajkusnier: Updated according to the spec: 
`If any error occurs during formatting, the raw value shall be displayed.`

chiljamgossow (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/VaccinationCertificateCard.kt
mtwalli: But you are not providing the raw value here 
should not it be ?
```kotlin
certificate.vaccinatedOn?.toShortDayFormat() ?: certificate.rawCertificate.vaccination.dt
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesExtensions.kt
```diff
@@ -20,15 +20,15 @@ fun Collection<CwaCovidCertificate>.toCertificateSortOrder(): List<CwaCovidCerti
...
+                    is TestCertificate -> it.sampleCollectedAt?.toLocalDateUserTz()
                     is RecoveryCertificate -> it.validFrom
                     else -> throw IllegalStateException("Can't sort $it")
                 }
             },
             {
                 when (it) {
                     is VaccinationCertificate -> it.headerIssuedAt.toLocalDateUserTz()
-                    is TestCertificate -> it.sampleCollectedAt.toLocalDateUserTz()
+                    is TestCertificate -> it.sampleCollectedAt?.toLocalDateUserTz()
```
mtwalli, line 31: do we need to provide `default`  dates ?

jurajkusnier: I added a unit test covering this scenario, a test without a date is moved to the end of the order

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/VaccinationCertificateCard.kt
jurajkusnier: sorry I somehow missed this one, fixed

mtwalli (approved):
LGTM, Thanks

BMItr (approved):
lgtm

PR #4863, opened 2022-02-22
Refactor person settings (EXPOSUREAPP-11989)

mtwalli:
- Migrate to `DataStore` in `PersonSettings`
- Due to regrouping and Dcc Reissuance , a new way to save settings per person is needed independently from how certificate are saved. HERE you go 
- Provided a way to show / dismiss Dcc Reissuance badge
- Provided a way to show / dismiss Dcc Booster badge (Should be used as part of regrouping migration)

## TODO
- [x] Unit tests

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11989


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonSettings.kt
```diff
@@ -0,0 +1,18 @@
...
+
+data class PersonSettings(
+    @JsonProperty("lastSeenBoosterRuleIdentifier")
+    val lastSeenBoosterRuleIdentifier: String? = null,
+
+    @JsonProperty("lastBoosterNotifiedAt")
+    val lastBoosterNotifiedAt: Instant? = null,
+
+    @JsonProperty("showDccReissuanceBadge")
+    val showDccReissuanceBadge: Boolean = false,
```
LukasLechnerDev, line 14: Is it enough to store a `Boolean` here or do we need to store the information about which certificate can be re-issued? Because I think it is similar to the Booster Rules - if there is a new certificate that can be replaced we need to show the badge and in order to know if it is a new certificate, we have to store the old one.

LukasLechnerDev:
@mtwalli Will the properties in `VaccinatedPersonData` that are now in `PersonSettings` will be removed in a different PR?

mtwalli:
> @mtwalli Will the properties in `VaccinatedPersonData` that are now in `PersonSettings` will be removed in a different PR?

I think it should be part of the re-grouping migration,
@jurajkusnier ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonSettings.kt
mtwalli: I thought the same, but the specs is clear on that if the there is reissuance info in the dcc wallet info and it did not exist before that notification and therefore the badge should be shown, it did not state if they were different as in booster case .

/cc @mlenkeit fyi

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesStorageTest.kt
```diff
@@ -1,69 +1,89 @@
...
 
     @Test
     fun `clearing deletes all data`() {
-        mockPreferences.edit {
-            putString("deleteme", "test")
+        createInstance().apply {
+            setCurrentCwaUser(personIdentifier1)
+            setBoosterNotifiedAt(personIdentifier1)
+            setDccReissuanceNotifiedAt(personIdentifier1)
+            clear()
```
LukasLechnerDev, line 66: Would be nice to additionally check before calling `clear()` that these fields are non-null: 

```kotlin
        fakeDataStore[CURRENT_PERSON_KEY] shouldNotBe null
        fakeDataStore[PERSONS_SETTINGS_MAP] shouldNotBe null
```

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonSettings.kt
mtwalli: in any case I think flag is `enough`

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/covidcertificate/person/core/PersonCertificatesStorageTest.kt
mtwalli: done

jurajkusnier (approved):
LGTM

LukasLechnerDev (approved):

PR #4868, opened 2022-02-23
Fix: Scanner is not opening after time travel to the past (DEV)

mtwalli:
- Scanner from tab menu was not opening when time travel to the past 

How to test?
- Open scanner once 
- Time travel while the app Is open to the past 
- Tab the scanner again, it should open


BMItr (approved):
Small change, Big Result. Thx

chiljamgossow (approved):

PR #4875, opened 2022-02-25
Change app config parameter for admission scenarios (EXPOSUREAPP-12044)

LukasLechnerDev:
### TechSpec
https://github.com/corona-warn-app/cwa-app-tech-spec/pull/162/files#diff-25c8b7188514d7cdfb430f7f8f963fed63bca279fd74753bc6cefe29dd599b55

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12044

### Open

- [x] It should be possible to enable and disable this flag on `MOCK-CLOUD` with `cwa app-config set dcc-admission-check-scenarios-enabled 1 -e cf`, but that didn't work for me. We should get that working and then test the implementation to see if the admission scenario tile gets displayed or hidden. => Max deployed a fix, it is working now!


mtwalli (approved):
LGTM

BMItr (approved):

PR #4882, opened 2022-03-01
Fix: New line separators are not working as expected (DEV)

mtwalli:
@LukasLechnerDev  found out that `\n` is not recognised in `TextView`. `\n` should be escaped to work properly

| Before        | After           |
| ------------- |:-------------:|
|  |  |


LukasLechnerDev (approved):
Thanks for the fix!

marcauberer (approved):

PR #4889, opened 2022-03-02
Dismiss dcc reissuance badge automatically (DEV)

mtwalli:
### How to reproduce?

1. Scan one of dcc reissuance presets
2. Request new Dcc
3. Now Delete the `new ` requested Dcc
4. Restore old recycled Dcc
5. The badge should be shown 
6. Restore the new recycled Dcc from step 3
7.  the badge should dismiss (after the fix), (before the fix badge is not dismissible)

### DIFF
--- a/Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccreissuance/notification/DccReissuanceNotificationServiceTest.kt
+++ b/Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccreissuance/notification/DccReissuanceNotificationServiceTest.kt
@@ -34,6 +34,7 @@ internal class DccReissuanceNotificationServiceTest : BaseTest() {
     fun setUp() {
         MockKAnnotations.init(this)
         coEvery { personCertificatesSettings.setDccReissuanceNotifiedAt(any(), any()) } just Runs
+        coEvery { personCertificatesSettings.dismissReissuanceBadge(any()) } just Runs
         every { personNotificationSender.showNotification(any(), any()) } just Runs
         every { dccWalletInfo.certificateReissuance } returns mockk()
     }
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccreissuance/notification/DccReissuanceNotificationServiceTest.kt
```diff
@@ -71,4 +72,25 @@ internal class DccReissuanceNotificationServiceTest : BaseTest() {
...
+            oldWalletInfo = dccWalletInfo,
+            newWalletInfo = mockk<DccWalletInfo>().apply { every { certificateReissuance } returns null }
+        )
+
+        coEvery {
+            personCertificatesSettings.dismissReissuanceBadge(personIdentifier)
+        }
+
+        coVerify(exactly = 0) {
+            personNotificationSender.showNotification(personIdentifier, R.string.notification_body_certificate)
```
LukasLechnerDev, line 92: I think we could improve the test by using 
```kotlin
coverify { personNotificationSender wasNot Called }
```

so that we verify that there was no action at all with the sender.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/dccreissuance/notification/DccReissuanceNotificationServiceTest.kt
```diff
@@ -71,4 +72,25 @@ internal class DccReissuanceNotificationServiceTest : BaseTest() {
...
+            personCertificatesSettings.setDccReissuanceNotifiedAt(personIdentifier, any())
```
LukasLechnerDev, line 93: Maybe change to 

```kotlin
personCertificatesSettings.setDccReissuanceNotifiedAt(any(), any())
```

so that we verify that `setDccReissuanceNotified` wasn't called no matter what the arguments are?

LukasLechnerDev (approved):
Tested and LGTM. 

2 optional proposals to improve the test.

BMItr (approved):

mtwalli:
@LukasLechnerDev, Thanks I will keep that in mind, I will improve it another PR. As it is already approved I will merge it for RC

PR #4893, opened 2022-03-03
[Feature] Red risk card for positive test result (EXPOSUREAPP-12059)

mtwalli:
- Show high risk card for positive test results and all risk card states after the test is beyond quarantine days

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12059


Ein-Tim:
Could we get a short explanation what is implemented here?

mtwalli:
@Ein-Tim  
- The risk calculation is not stopped anymore if a positive test result is available
- The risk card is only `hidden` if
    - The risk card would show no risk or a low risk (green card) AND
    - There is at least one positive test which is not older than n days (n = config value)
      - The age of a PCR test is calculated from the test registration time
      - The age of a RAT test is calculated from the test execution time

AlexanderAlferov (approved):
LGTM 🚀 
Tested with both RAT and PCR.

BMItr (approved):
nothing to complain :ok_hand:

PR #4897, opened 2022-03-03
Repositories improvements (DEV)

jurajkusnier:
Sync `RecoveryCertificateRepository`, `TestCertificateRepository` and `VaccinationCertificateRepository` to use  `Map<containerId,container>`  structure to represent `internalData`.
With Map, we can be sure that there will be always only 1 certificate for any QR code.
This is not the case for Set, where the collection cannot contain duplicate elements, but as the container consists not only of QR code but also other data, we can't be sure there are no containers with the same QR code.


chiljamgossow (approved):
Looks much better now, thanks!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/storage/VaccinationStorage.kt
```diff
@@ -37,8 +37,6 @@ class VaccinationStorage @Inject constructor(
                 prefs.getString(PKEY_VACCINATION_CERT, null) ?: return emptySet(),
                 TYPE_TOKEN
             )
-            .associateBy { it.vaccinationQrCode }
-            .map { it.value }
             .toSet()
```
LukasLechnerDev, line 40: `.toSet()` could be removed here, or?

LukasLechnerDev (approved):
👍 only one comment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/core/repository/storage/VaccinationStorage.kt
jurajkusnier: could be

PR #4904, opened 2022-03-04
Risk encounters in contact journal shouldn't be clickable (EXPOSUREAPP-12058)

mtwalli:
- Scan tests , use test menu to have fake risk encounters
- Check contact journal entries from step 1, they should NOT be clickable 
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12058


BMItr (approved):
LGTM.. tested on oneplus 8 device.

image

SamuraiKek (approved):

PR #4906, opened 2022-03-04
Filter checkIns by max age (EXPOSUREAPP-12120)

chiljamgossow:
https://github.com/corona-warn-app/cwa-app-tech-spec/pull/164/files
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12120

Unfortunately, plenty of changes were needed to support different time periods for calculation, contact journal, retention etc. 

Contact diary needs to reflect the risk calculation. That means risk encounters will not appear that were not included in the current risk result. However checkins are still there, but won't be have a high risk note if not included any more.

Also includes an additional filter for exposure windows to make sure lastRiskEncounter and contact journal are correct.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/risk/storage/PresenceTracingRiskDaos.kt
```diff
@@ -0,0 +1,146 @@
...
+
+@Entity
+data class TraceTimeIntervalMatchEntity(
+    @PrimaryKey(autoGenerate = true) @ColumnInfo(name = "id") val id: Long? = null,
+    @ForeignKey(
+        entity = TraceLocationCheckInEntity::class,
+        parentColumns = ["id"],
+        childColumns = ["checkInId"],
+        onDelete = ForeignKey.CASCADE
+    )
```
mtwalli, line 51: As we are touching this anyway , can we change who the `ForeignKey` is referenced here? it would help upgrade to the latest version easily

chiljamgossow: How?

mtwalli:
``` kotlin
@Entity(foreignKeys = [ForeignKey(entity = ParentClass::class,
                    parentColumns = arrayOf("id"),
                    childColumns = arrayOf("checkInId"),
                    onDelete = ForeignKey.CASCADE))]
```

chiljamgossow: Changing this creates problems, so I would won't change it in this PR (I only moved the code here)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/ExposureWindowsFilter.kt
```diff
@@ -0,0 +1,60 @@
...
+        return list.filter { ewDayRisk ->
+            !ewDayRisk.localDateUtc.isBefore(deadline)
+        }
+    }
+
+    private fun ExposureWindowRiskCalculationConfig.getDeadline(nowUtc: Instant): Instant =
+        nowUtc.minusDays(getMaxEwAgeInDays()).toInstant()
+
+    private fun ExposureWindowRiskCalculationConfig.getMaxEwAgeInDays() =
+        if (maxEncounterAgeInDays > 0) maxEncounterAgeInDays else DEFAULT_EW_AGE_IN_DAYS
```
LukasLechnerDev, line 44: What do you think about having this default value logic in `ExposureWindowRiskCalculationConfigMapper` in case we want to use  `maxEncounterAgeInDays` in other places too?

Corona-Warn-App/schemas/de.rki.coronawarnapp.presencetracing.risk.storage.PresenceTracingRiskDatabase/1.json
```diff
@@ -69,6 +69,12 @@
             "columnName": "riskStateCode",
             "affinity": "INTEGER",
             "notNull": true
+          },
+          {
+            "fieldPath": "calculatedFromMillis",
```
LukasLechnerDev, line 74: Do we need the changes here from `1.json`? We are migrating to `2.json` anyway?

chiljamgossow: No this file was generated after I added the field and before I changed the version. I will remove it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/ExposureWindowsFilter.kt
chiljamgossow: moved the logic to the config mapping as UI should also use the default

LukasLechnerDev (approved):
LGTM 🚀

jurajkusnier (approved):
LGTM

PR #4909, opened 2022-03-07
Align date formatting (EXPOSUREAPP-11285)

mtwalli:
- dob formatting in the person details card was different from other dates formatting in the screen 
- Now , it will follow the same formatting if POSSIBLE otherwise original text 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11285


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/CwaUserCard.kt
```diff
@@ -43,15 +43,9 @@ class CwaUserCard(parent: ViewGroup) :
 
     private fun formatBirthDate(dateOfBirthFormatted: String): String =
```
chiljamgossow, line 44: I believe the old logic was there to respect the cases where only a year is provided. Is that still the case? So a birth year of 1980 will still only be a year, and not 1980-01-01

mtwalli: Yes, if dob=`1980` or `1980-01` it won't pass the date parsing and the original value will be displayed

SamuraiKek (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/CwaUserCard.kt
chiljamgossow: the standard parsing will accept a year as a valid date and simply set it to 1.1.YYYY. If you checked that then it is fine

mtwalli: Added unit tests, see https://github.com/corona-warn-app/cwa-app-android/pull/4909/commits/a6486fc631380d7a438b3a633bde054db254ac1e

SamuraiKek (approved):

chiljamgossow (approved):

PR #4913, opened 2022-03-08
Addition to "Überblick" > "Infektionsrisiko" (EXPOSUREAPP-11246)

KathrinLuecke:
Added that check-ins are taken into account for calculating the risk of infection
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11246


CV113 (approved):
UA approved

jurajkusnier (approved):

mtwalli (approved):

PR #4915, opened 2022-03-08
Reissuance - additonal error handling (EXPOSUREAPP-11618)

BMItr:
This PR addresses [Exposureapp-11618.](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12156)

Note:
TechSpec will be updated soon. The property is called **`"error"` and not `"errorCode"`**


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccreissuance/core/server/data/DccReissuanceErrorResponse.kt
```diff
@@ -0,0 +1,8 @@
+package de.rki.coronawarnapp.dccreissuance.core.server.data
+
+import com.google.gson.annotations.SerializedName
+
+data class DccReissuanceErrorResponse(
+    @SerializedName("error") val error: String,
```
mtwalli, line 6: `errorCode` as per [specs]( https://github.com/corona-warn-app/cwa-app-tech-spec/pull/154/commits/e0302a180bfa9aad8693f3835171d711f3318898#diff-b503751806bd910ffe10c62a514f15f69ed3be758d95ecff6e8700e1cab5b5cbR183)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccreissuance/core/server/DccReissuanceServer.kt
```diff
@@ -61,16 +62,24 @@ class DccReissuanceServer @Inject constructor(
...
-            400 -> ErrorCode.DCC_RI_400
-            401 -> ErrorCode.DCC_RI_401
-            403 -> ErrorCode.DCC_RI_403
-            406 -> ErrorCode.DCC_RI_406
-            429 -> ErrorCode.DCC_RI_429
-            500 -> ErrorCode.DCC_RI_500
-            in 400..499 -> ErrorCode.DCC_RI_CLIENT_ERR
-            else -> ErrorCode.DCC_RI_SERVER_ERR
-        }.also { throw DccReissuanceException(errorCode = it) }
+        val serverErrorCode = tryGetServerErrorCode()
```
mtwalli, line 65: I think better naming would be `val severError = tryGetServerError()` ?
Here we are not just getting the code , but the whole `error (code + message)`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccreissuance/core/server/data/DccReissuanceErrorResponse.kt
mtwalli: small unit test for parsing the error would be appreciated

BMItr: Please read the description.
There is a missmatch between the errorCodes & and the Codes in the Spec. Max will provide an updated one and then the spec will match to this PR.

mtwalli: Ok 👍

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccreissuance/core/server/DccReissuanceServer.kt
BMItr: 👌 adjusted

mtwalli (approved):

mtwalli:
Thx!

chiljamgossow (approved):
LGTM

PR #4917, opened 2022-03-08
Change number of days for which the red risk card is shown after an exposure: UI (EXPOSUREAPP-12121)

jurajkusnier:
The configured number of days is changed dynamically in the following texts (see Figma)

- "Period Logged" on the red risk card
- "Period Logged" on the green risk card
- "Period Logged" on the exposure logging screen
- "Risk of Infection" on the app overview screen
- On the initial onboarding screen 
- The information popup "Information zur Funktionsweise der Risiko-Ermittlung"

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12065

PR #4921, opened 2022-03-10
Show additional high risk notifications (EXPOSUREAPP-12064)

LukasLechnerDev:
### Description

New notifications for additional high risks. When the user clicks on the notification, or the next time the app is opened, a new dialog should be displayed: [screenshot of new dialog](https://user-images.githubusercontent.com/10398034/157615373-e9767765-2f7d-4051-8f62-6007a7a3db11.png). Once the user clicks "OK" on the dialog, it won't show up on subsequent app opens. 

Covers all subtasks of this story: https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12064

### How to test
- Set device time to real time
- Set radio box `INCREASED_RISK_DEFAULT` in Test Menu
- Click `CALCULATE RISK LEVEL`
- User should now get to the "initial high risk" state

Then there are 2 options to continue

_Option 1:_
- Leave the app in this state over night
- A risk calculation in the background (with an additional high risk) on the next day should trigger a new notification.

_Option 2:_
- Set the device time to one day in the future
- Open the app
- The risk calculation will be triggered, resulting in an "additional high risk"
- The dialog should be shown on the HomeFragment
- A notification should be triggered


Ein-Tim:
Why does the notification talk about the last 10 days?
Will the contact journal be reduced to only keep the entries of the last 10 days?

CV113 (dismissed):
UA approved

LukasLechnerDev:
> Why does the notification talk about the last 10 days? Will the contact journal be reduced to only keep the entries of the last 10 days?

@Ein-Tim Yes, this value will change from 14 to 10 days. Also see this PR: https://github.com/corona-warn-app/cwa-app-android/pull/4917

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/changedetection/CombinedRiskLevelChangeDetector.kt
```diff
@@ -50,10 +53,20 @@ class CombinedRiskLevelChangeDetector @Inject constructor(
...
+                Timber.v("Checking for low-to-high or high-to-low risklevel change.")
                 results.checkForRiskLevelChanges()
             }
-            .catch { Timber.e(it, "App config change checks failed.") }
+            .catch { Timber.e(it, "RiskLevel checks failed.") }
+            .launchIn(appScope)
+
+        // send notifications for additional HIGH risks after an initial HIGH risk
+        riskLevelStorage.latestAndLastSuccessfulCombinedEwPtRiskLevelResult
+            .map { it.lastSuccessfullyCalculated }
```
chiljamgossow, line 64: I gave you the wrong information. You need to use lastCalculated instead (data is overwritten even if it fails, so the available data for risk encounters belongs to the last calculation)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -281,6 +281,24 @@ class HomeFragmentViewModel @AssistedInject constructor(
             .asLiveData(context = dispatcherProvider.Default)
     }
 
+    data class ShowAdditionalHighRiskLevelDialogEvent(
+        val show: Boolean,
+        val maxEncounterAgeInDays: Int
+    )
+
+    val showAdditionalHighRiskLevelDialog: LiveData<ShowAdditionalHighRiskLevelDialogEvent> by lazy {
```
BMItr, line 289: Could this lead to situations where the same dialog is created multiple times and overlays other instances?
E.g. when the livedata reconnects to the flow, like when the user puts the app into background for more than 5 seconds while the dialog is shown.

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -350,6 +350,17 @@
...
     <string name="risk_lowered_dialog_button_confirm">"OK"</string>
 
+    <!-- ####################################
+       Additional High Risk Dialog
+    ###################################### -->
+
+    <!-- XHED: additional high risk - dialog headline -->
+    <string name="additional_high_risk_dialog_headline">Neue Risikobegegnung</string>
+    <!-- XTXT: additional high risk - dialog content -->
+    <string name="additional_high_risk_dialog_body">Sie hatten eine erneute Begegnung mit erhöhtem Risiko.\n\nAlle Risikobegegnungen der vergangenen %d Tage finden Sie in Ihrem Kontakttagebuch.</string>
```
BMItr, line 360: I guess %1$d

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/changedetection/CombinedRiskLevelChangeDetector.kt
LukasLechnerDev: Changed it. Thanks!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -281,6 +281,13 @@ class HomeFragmentViewModel @AssistedInject constructor(
             .asLiveData(context = dispatcherProvider.Default)
     }
 
+    val showAdditionalHighRiskLevelDialog: LiveData<Boolean> by lazy {
```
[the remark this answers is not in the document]
LukasLechnerDev: copy-paste relict 😅 not necessary

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
LukasLechnerDev: You are right! Thanks for pointing this out. I am using `SingleLiveEvent` now.

Corona-Warn-App/src/main/res/values-de/strings.xml
LukasLechnerDev: Changed, thanks

AlexanderAlferov (approved):
Great work! Tested with option 1.

BMItr (approved):
👌

CV113 (approved):
UA approved

PR #4922, opened 2022-03-10
Requesting DCC are not deletable (EXPOSUREAPP-12091)

jurajkusnier:
### Testing

- add negative RAT/PCR test with certificate
- immediately open the certificate tab during cert. loading
- turn off wifi
- loading state should be replaced with an error state

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12091

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalWorker.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalWorker.kt
@@ -9,6 +9,8 @@ import dagger.assisted.AssistedInject
 import de.rki.coronawarnapp.covidcertificate.test.core.TestCertificateRepository
 import de.rki.coronawarnapp.util.worker.InjectedWorkerFactory
 import de.rki.coronawarnapp.worker.BackgroundConstants
+import kotlinx.coroutines.CancellationException
+
 import timber.log.Timber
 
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/execution/TestCertificateRetrievalWorker.kt
```diff
@@ -38,6 +40,9 @@ class TestCertificateRetrievalWorker @AssistedInject constructor(
             }
         } catch (e: Exception) {
             Timber.tag(TAG).e(e, "Test result retrieval worker failed.")
+            if (e is CancellationException) {
```
LukasLechnerDev, line 43: I think we should surround the whole `doWork()` function body with try-catch. Otherwise this wouldn't work if the Worker is called in the beginning?

LukasLechnerDev: Maybe allso call `refreshCleanup()` when the worker fails in Line `27`?

LukasLechnerDev: I see. Thanks!

LukasLechnerDev (approved):
LGTM!

mtwalli (approved):
LGTM

PR #4929, opened 2022-03-14
[FEATURE] ENF delta (EXPOSURAPP-12210)

chiljamgossow:
How to test:
- check database 'keyfiles' and storage data -> data -> de.coronawarnapp.test -> cache -> diagnosis-keys
- entries marked as checked should disappear on the next app start
- triggering key download should not result in download of files that are marked as checked
- only unmarked files are provided to ENF, marked after successful submission

currently targeting 2.19

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheDatabase.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheDatabase.kt
@@ -45,6 +45,9 @@ abstract class KeyCacheDatabase : RoomDatabase() {
 
         @Update(entity = CachedKeyInfo::class)
         suspend fun updateDownloadState(update: CachedKeyInfo.DownloadUpdate)
+
+        @Query("UPDATE keyfiles SET checkedForExposures = 1 WHERE id = :id")
+        suspend fun setChecked(id: String)
     }
 
     class Factory @Inject constructor(@AppContext private val context: Context) {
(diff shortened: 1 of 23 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheDatabase.kt
```diff
@@ -45,6 +45,9 @@ abstract class KeyCacheDatabase : RoomDatabase() {
 
         @Update(entity = CachedKeyInfo::class)
         suspend fun updateDownloadState(update: CachedKeyInfo.DownloadUpdate)
+
```
BMItr, line 48: Use the generic @Update, so you can pass in a list of objects and make a batch update, which would be faster. Update the checked flag in the repo.

jurajkusnier (approved):
LGTM

AlexanderAlferov (approved):
LGTM 🚀

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheDatabase.kt
chiljamgossow: I do not want to risk overwriting other fields and performance is not an issue (would only be less than 10 files)

MikeMcC399:
Will you consider creating and using a `Fix 2.19.1` label, since 2.19.0 has already been released?

PR #4930, opened 2022-03-14
Delta key file download and submission (EXPOSUREAPP-12213)

chiljamgossow:
- download only files that are missing and have not been checked
- mark key files as checked after successful submission
- submit only delta files

- [X] unit test delta download
- [X] unit test update checked flag

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
@@ -97,9 +97,10 @@ open class BaseKeyPackageSyncTool(
             val exists = key.path.exists()
             val checked = key.info.checkedForExposures
             if (complete && !exists && !checked) {
+                // marked complete, but does not exist nor has been checked for exposures -> attempt new download
                 Timber.tag(tag).v("Incomplete download, will overwrite: %s", key)
             }
-            // We overwrite incomplete ones that have not been checked
+            // files that have been downloaded already
             complete && (exists || checked)
         }
 
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
@@ -97,8 +97,8 @@ open class BaseKeyPackageSyncTool(
             val exists = key.path.exists()
             val checked = key.info.checkedForExposures
             if (complete && !exists && !checked) {
-                // marked complete, but does not exist nor has been checked for exposures -> attempt new download
-                Timber.tag(tag).v("Incomplete download, will overwrite: %s", key)
+                // marked complete, but does not exist nor has been checked for exposures -> counts as missing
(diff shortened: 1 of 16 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/BaseKeyPackageSyncTool.kt
```diff
@@ -87,19 +87,20 @@ open class BaseKeyPackageSyncTool(
...
             val exists = key.path.exists()
-            if (complete && !exists) {
+            val checked = key.info.checkedForExposures
+            if (complete && !exists && !checked) {
                 Timber.tag(tag).v("Incomplete download, will overwrite: %s", key)
             }
-            // We overwrite not completed ones
-            complete && exists
+            // We overwrite incomplete ones that have not been checked
+            complete && (exists || checked)
```
mtwalli, line 103: These 4 lines are confusing, in `if` checks for `!` existing `!` checked keys and logs call them incomplete:
```kotlin
if (complete && !exists && !checked) {
   Timber.tag(tag).v("Incomplete download, will overwrite: %s", key)
}
```
BUT in here, we return complete and (existing or checked)  and the comment says incomplete too:
```kotlin 
// We overwrite incomplete ones that have not been checked
complete && (exists || checked)
```
Could you please elaborate more ? or correct the log / comment statements?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DayPackageSyncTool.kt
```diff
@@ -92,25 +92,25 @@ class DayPackageSyncTool @Inject constructor(
...
-        val staleDays = cachedDays.findStaleData(listOf(availableDays))
-
-        if (staleDays.isNotEmpty()) {
-            Timber.tag(TAG).d("Deleting stale days (loation=%s): %s", location, staleDays)
-            keyCache.delete(staleDays.map { it.info })
+        // remove files that are no longer available on the server
+        val staleKeys = cachedKeys.findStaleData(listOf(availableDays))
+        if (staleKeys.isNotEmpty()) {
+            Timber.tag(TAG).d("Deleting stale days (loation=%s): %s", location, staleKeys)
+            keyCache.delete(staleKeys.map { it.info })
```
mtwalli, line 109: - is `checkedForSubmission` Keys part of this? , if so, we should `NOT` delete the `meta-data` as per specs , but we SHOULD delete the `files`
- I don't see where actually the `files` are deleted. Do we delete them anywhere?

thomasaugsten (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/download/DayPackageSyncTool.kt
chiljamgossow: Stale Keys are those which have been removed from the server. They cannot be downloaded anymore and should also be removed from the device.
Delete will remove the meta data and the file itself.

mtwalli: ok, where do we delete checkedForSubmission files?

chiljamgossow: next PR

mtwalli (approved):
Thanks!

PR #4931, opened 2022-03-15
Delete checked files (EXPOSUREAPP-12214)

chiljamgossow:
house keeping:
prevent deletion of meta data entries for checked files
delete files that are marked as checked

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheRepository.kt
@@ -58,11 +58,19 @@ class KeyCacheRepository @Inject constructor(
     }
 
     private suspend fun doHouseKeeping() {
+        // delete files that have been marked as complete, but do not exists nor have been checked
         val dirtyInfos = getAllCachedKeys().filter {
-            it.info.isDownloadComplete && !it.path.exists()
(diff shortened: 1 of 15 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheRepository.kt
```diff
@@ -138,18 +148,22 @@ class KeyCacheRepository @Inject constructor(
...
             Timber.v("Deleted %s", key)
-            val path = getPathForKey(key)
-            if (path.delete()) Timber.v("Deleted cache key file at %s", path)
+            deleteFile(key)
         }
     }
 
+    private fun deleteFile(keyInfo: CachedKeyInfo) {
+        val path = getPathForKey(keyInfo)
+        if (path.delete()) Timber.v("Deleted cache key file at %s", path)
```
mtwalli, line 162: should catch the error here when deleting? if a file deletion failed for any reason then the others (loop) will continue to be deleted

mtwalli (approved):
LGTM

AlexanderAlferov (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/diagnosiskeys/storage/KeyCacheRepository.kt
chiljamgossow: The file may already be deleted from the cache by the OS. We want the others to be deleted anyway

PR #4938, opened 2022-03-16
Don't trigger high-to-low-risk notification if the low-risk card is not displayed (EXPOSUREAPP-12244)

LukasLechnerDev:
### Jira Ticket 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12244

### Info
There were some refactorings: I created a new class `RiskCardDisplayInfo`, which has a `shouldShowRiskCard()`function that determines if the risk card should be displayed on the Home Screen. This function is also called before we check if we should trigger the high-to-low-risk notification (only do so when the low risk is also shown). 

### How to test
Scan a positive test and check if the low-risk card is still hidden and only the high-risk is shown in this case. If something here is not working anymore, then I introduced some issues with my refactoring. 

Unfortunately I don't know how to test if the notification isn't triggered. We would need to trigger a low-risk calculation after we got a high-risk-card. Does anybody know how to trigger that? Played around in the TestMenu, but wasn't able to achieve that.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -195,25 +197,17 @@ class HomeFragmentViewModel @AssistedInject constructor(
...
-                    } // else -> // Don't show risk card
-                }
 
-                // Otherwise show risk card
-                else -> add(tracingItem)
+            val currentRiskState = when (tracingItem) {
+                is IncreasedRiskCard.Item -> RiskState.INCREASED_RISK
+                is LowRiskCard.Item -> RiskState.LOW_RISK
+                is TracingFailedCard.Item -> RiskState.CALCULATION_FAILED
+                else -> null // tracing is disabled or calculation is currently in progress
```
mtwalli, line 206: In that case it should show up , right ?

mtwalli: I would not max the state and the test checks , please keep the logic separate, I see that you are passing `RiskState.LOW_RISK` all the time in the detector

mtwalli (dismissed):

BMItr (dismissed):
:ok_hand:

BMItr (approved):

mtwalli (approved):

PR #4943, opened 2022-03-16
Round `normalizedTime` (EXPOSUREAPP-12211)

### LINKED ISSUE #4925: PRIO: Twitter Reports: Risk status changes without new checks carried out
## Avoid duplicates

## Technical details

- Device name: ZTE A2020G Pro
- Android version: Android 10
- App version:  2.18.1

## Describe the bug

I have a Twitter user reporting the following: She received an increased risk encounter for the 04.03.22 two days ago. Since she received the encounter, the app switches between increased risk and low risk.

See these screenshots from today, both were taken with the last risk check carried out at 04:22h:

| 06:28h | 08:10h |
|---|---|
| mxo40akY jpg-small | e5ZfHQVR jpg-small |

I'm not sure whether this happens with all risk encounters at the moment or what exactly could be done to reproduce this. I did not identify any special things while writing with the user on Twitter. 

## Expected behaviour

The app should **never** switch between different risk statuses without carrying out a new check. 

## Additional context
1. Here is the EN log of the user: [all-exposure-checks.txt](https://github.com/corona-warn-app/cwa-app-android/files/8231986/all-exposure-checks.txt)
2. The user does not have any check ins in her CWA.
3. Report on Twitter: https://twitter.com/mimihasstnazis/status/1502180895104684032?s=20&t=4HWEXQmOxeZk4DfSutuE5Q

## Pings

@thomasaugsten Please take a look & let me know if you want a CWA error log. 
@dsarkar, @larswmh or @svengabr Please mirror this to JIRA ASAP.
FYI @vaubaehn @mtwalli

---
Internal Tracking ID: [EXPOSUREAPP-12211](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12211)

mtwalli:
Read Max's comment here for more information https://github.com/corona-warn-app/cwa-app-android/issues/4925#issuecomment-1069156806

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12211


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/math/Math.kt
```diff
@@ -0,0 +1,9 @@
+package de.rki.coronawarnapp.util.math
+
+import kotlin.math.roundToInt
+
+/**
+ * Round a fraction to a decimal number with the required decimal places
+ * @param denominator Power of 10
+ */
+fun Double.roundToDecimal(denominator: Double = 100.0): Double = (this * denominator).roundToInt() / denominator
```
mlenkeit, line 9: I think `denominator` should not have a default, because any default always depends on the context.

Second, the current default is technically not correct. It correctly rounds 3.599 to 3.6 (actually 3.60, but 0 is suppressed) but it would round 3.559 to 3.56.

Instead, I would suggest to accept an (unsigned) int as number decimal places and calculate the denominator as 10**decimalPlaces.

mtwalli: Adjusted as per suggestion

mlenkeit:
@mtwalli related PR - https://github.com/corona-warn-app/cwa-app-android/pull/4946

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/risk/DefaultRiskLevels.kt
```diff
@@ -254,7 +255,7 @@ class DefaultRiskLevels @Inject constructor() : RiskLevels {
 
         // 2. Determine `Normalized Time per Date`
         val normalizedTime = exposureWindowsAndResultForDate.values
-            .sumOf { it.normalizedTime }
+            .sumOf { it.normalizedTime.roundToDecimal(decimalPlacesNumber = 1u) }
```
mlenkeit, line 258: I think it might also be more appropriate to only round the sum and not individual operands.

AlexanderAlferov (approved):

mlenkeit (approved):

BMItr (approved):
:ok_hand:

PR #4946, opened 2022-03-17
Update common EW risk calculation test cases

mlenkeit:
This PR updates the common exposure window risk calculation test cases and specifically adds a new test case for https://github.com/corona-warn-app/cwa-app-android/pull/4943 / EXPOSUREAPP-12211 / https://github.com/corona-warn-app/cwa-app-android/issues/4925.


mlenkeit:
I had to update all test cases to reflect the most recent configuration, so I hope the only test case that breaks is the one regarding EXPOSUREAPP-12211

PR #4949, opened 2022-03-17
PPA: do not overwrite test result received on subsequent polling (EXPOSUREAPP-12185)

chiljamgossow:


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/keysubmission/AnalyticsKeySubmissionCollector.kt
```diff
@@ -30,6 +30,7 @@ class AnalyticsKeySubmissionCollector @Inject constructor(
 
     fun reportPositiveTestResultReceived(type: CoronaTest.Type) {
         if (disabled) return
+        if (type.storage.testResultReceivedAt.value > 0) return
```
LukasLechnerDev, line 33: Suggestion: Maybe do something like the code below to make it a bit more understandable? Or add a comment?

```kotlin
val alreadySet = type.storage.testResultReceivedAt.value != -1
if(!alreadySet){
    type.storage.testResultReceivedAt.update { timeStamper.nowUTC.millis }
}

```

LukasLechnerDev:
Found some "pollServer()" lines in the logs after the first time the positive result was received, so I think this will fix this bug 👍. It also explained why it didn't happened in my time-travel test.

LukasLechnerDev (dismissed):
1 Optional comment

mtwalli (approved):

jurajkusnier (approved):

PR #4961, opened 2022-03-18
Add family test tile on homescreen (EXPOSUREAPP-12314)

SamuraiKek:
Added a new tile for the family test. 
Tried my best to center the text inside the badge, but it still looks off. Any suggestions are welcome.

Screen Shot 2022-03-18 at 17 10 32


CV113 (dismissed):
UA approved

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -258,6 +258,9 @@ class HomeFragment : Fragment(R.layout.home_fragment_layout), AutoInject {
                 HomeFragmentDirections.actionMainFragmentToFederalStateSelectionFragment()
             )
             is HomeFragmentEvents.DeleteOutdatedRAT -> viewModel.deleteCoronaTest(event.identifier)
+            is HomeFragmentEvents.GoToFamilyTests -> {
+                // TODO: add wiring to family test results screen
```
mtwalli, line 262: I think you can do that now, a blank screen is there ,but it should help in this case

mtwalli (approved):
LGTM

chiljamgossow (approved):
LGTM

PR #4966, opened 2022-03-21
Fix G status badge dismiss (EXPOSUREAPP-12359)

SamuraiKek:
Badge would sometimes not dismissing when pressing the close button because `popBackStack()` would execute before dismissing the badge. 
Also added badge dismissal on every navigation away from the details screen.

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12359)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsFragment.kt
```diff
@@ -59,8 +59,7 @@ class PersonDetailsFragment : Fragment(R.layout.person_details_fragment), AutoIn
         binding.apply {
             root.transitionName = args.personCode
             toolbar.setNavigationOnClickListener {
-                viewModel.dismissAdmissionStateBadge()
-                popBackStack()
+                viewModel.dismissAdmissionStateBadge(true)
```
mtwalli, line 62: Do you know why `onStop()` would not work here ?

SamuraiKek: I'm not really sure. I think it's because the lifecycle of this fragment is tied to the lifecycle of the main activity so `onStop()` will trigger when the `onStop()` of the main activity actually gets triggered.
 I've also tried using `navController.addOnDestinationChangedListener()` but it still wasn't behaving as expected in all instances so I've just reverted to calling this dismiss badge function on all the places where we travel to and from the person details fragment as it's been the most reliable way so far.

jurajkusnier (approved):
LGTM

mtwalli (approved):

PR #4970, opened 2022-03-21
CoronaTest refactoring (EXPOSUREAPP-12330)

chiljamgossow:
Split interface to facilitate differentiation between personal and family tests 

changes come from renaming, logic stays the same

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepository.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepository.kt
@@ -57,19 +57,19 @@ class CoronaTestRepository @Inject constructor(
     }
 
     /**
-     * Returns a flow with an unfiltered set of [CoronaTest]
+     * Returns a flow with an unfiltered set of [PersonalCoronaTest]
      */
     val allCoronaTests: Flow<Set<PersonalCoronaTest>> = internalData.data.map { it.values.toSet() }
 
     /**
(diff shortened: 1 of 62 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestRepository.kt
```diff
@@ -59,19 +59,19 @@ class CoronaTestRepository @Inject constructor(
     /**
      * Returns a flow with an unfiltered set of [CoronaTest]
```
DerVogel2020, line 60: The docs should be changed as well.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/CoronaTest.kt
```diff
@@ -49,10 +29,16 @@ interface CoronaTest : Recyclable {
     // Has the corresponding entry been created in the test certificate storage
     val isDccDataSetCreated: Boolean
 
-    //  The ID of the lab that uploaded the test result
-    val labId: String?
-
     val qrCodeHash: String?
+}
+
+interface KeySubmission {
```
mtwalli, line 35: Maybe we should be more specific here as it is for test `CoronaTestKeySubmission` or `CoronaTestSubmission` ?

mtwalli (approved):
Nice 👌, one comment , which can be addressed in followup PR

SamuraiKek (approved):
LGTM.

PR #4971, opened 2022-03-21
Deeplinking/QRHandler - Single source of truth (EXPOSUREAPP-11884)

BMItr:
This PR addresses Exposureapp-11884.
Deeplinking and QRHandling was handled in different places.
Refactored to single class usage.

How to test:

- Generate some RAT and R-PCR qr codes
- Scan them or use deeplinking
- Test different scenarios, like a test is in recycle bin or already scanned in
- Scanning and deeplinking should behave equally

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
@@ -189,32 +189,39 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
             showRestoreCoronaTestConfirmation(coronaTestResult.recycledCoronaTest)
             null
         }
-        is CoronaTestResult.RestoreDuplicateTest -> NavGraphDirections.actionToSubmissionDeletionWarningFragment(
-            coronaTestResult.restoreRecycledTestRequest
-        )
-        is CoronaTestResult.TestInvalid -> NavGraphDirections.actionGlobalSubmissionTestResultInvalidFragment(
-            testType = coronaTestResult.test.type,
-            testIdentifier = coronaTestResult.test.identifier
-        )
-        is CoronaTestResult.TestNegative -> NavGraphDirections.actionGlobalSubmissionTestResultNegativeFragment(
-            testType = coronaTestResult.test.type,
-            testIdentifier = coronaTestResult.test.identifier
-        )
-        is CoronaTestResult.TestPending -> NavGraphDirections.actionSubmissionTestResultPendingFragment(
-            testType = coronaTestResult.test.type,
-            testIdentifier = coronaTestResult.test.identifier,
-            forceTestResultUpdate = true
-        )
-        is CoronaTestResult.TestPositive -> NavGraphDirections.actionGlobalSubmissionTestResultKeysSharedFragment(
-            testType = coronaTestResult.test.type,
-            testIdentifier = coronaTestResult.test.identifier
-        )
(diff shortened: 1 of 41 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -171,11 +175,58 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
...
+                testType = coronaTestResult.test.type,
+                testIdentifier = coronaTestResult.test.identifier,
+                forceTestResultUpdate = true
+            )
+        is CoronaTestResult.TestPositive ->
+            NavGraphDirections.actionGlobalSubmissionTestResultKeysSharedFragment(
+                testType = coronaTestResult.test.type,
+                testIdentifier = coronaTestResult.test.identifier
+            )
+        is CoronaTestResult.TestRegistrationSelection ->
```
mtwalli, line 217: I think this the only case that has to handled here. 
- in case the test needs to be restored i.e `RestoreDuplicateTest` -> then test selection screen has to be shown -> the rest of the flow depending on whether it is personal or family test  
- In case the test is new, or existing i.e  `TestRegistrationSelection` ->  then test selection screen has to be shown

any other case, the card is displayed in HomeScreen anyway and `no need` to open the respective result screen

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivityViewModel.kt
```diff
@@ -145,16 +150,18 @@ class MainActivityViewModel @AssistedInject constructor(
 
     private suspend fun QrCodeExtractor<CoronaTestQRCode>.handleCoronaTestQr(uriString: String) = try {
         val qrCode = extract(rawString = uriString)
-        val test = submissionRepository.testForType(qrCode.type).first()
-        when {
-            test != null -> event.postValue(MainActivityEvent.GoToDeletionScreen(qrCode))
-            else -> event.postValue(MainActivityEvent.GoToSubmissionConsentFragment(qrCode))
```
mtwalli, line 151: you need to remove those events from `MainActivityEvent`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivityViewModel.kt
```diff
@@ -145,16 +150,18 @@ class MainActivityViewModel @AssistedInject constructor(
...
-        }
+        val coronaTestResult = coronaTestQRCodeHandler.handleQrCode(qrCode)
+        mutableCoronaTestResult.postValue(coronaTestResult)
     } catch (e: Exception) {
         Timber.w(e, "onNavigationUri failed")
         event.postValue(MainActivityEvent.Error(e))
     }
 
+    fun restoreCoronaTest(recycledCoronaTest: CoronaTest) = launch {
```
mtwalli, line 160: As I mentioned above the restoration should happen probably after test selection screen anyway. So it is no needed here

mtwalli:
I don't see any changes related specifically to `Android 12` in this case , I suggest you rebase to `2.21.x` directly

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivityViewModel.kt
BMItr: done

BMItr:
followUp PR opened.
Closing this.

PR #4979, opened 2022-03-22
Deeplinking/QRHandler - Single source of truth (EXPOSUREAPP-11884, 12331)

BMItr:
Follow Up PR for #4971 
Rebased to 2.21

This PR addresses Exposureapp-[11884](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11884) & Exposureapp-[12331](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12331), [12326](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12326)

- [x] some adjustments needed / WIP.
- [x] additional tests 

This PR addresses Exposureapp-11884.
Deeplinking and QRHandling was handled in different places.
Refactored to single class usage.

How to test:

Generate some RAT and R-PCR qr codes
Scan them or use deeplinking
Test different scenarios, like a test is in recycle bin or already scanned in
Scanning and deeplinking should behave equally


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt
```diff
@@ -239,38 +241,13 @@ class QrCodeScannerFragment : Fragment(R.layout.fragment_qrcode_scanner), AutoIn
                 QrCodeScannerFragmentDirections.actionUniversalScannerToSubmissionDeletionWarningFragment(
                     scannerResult.restoreRecycledTestRequest
                 )
-            is CoronaTestResult.TestPending ->
```
mtwalli, line 242: We need to keep this here!
The idea is that `CoronaTestQRCodeHandler` should still provide all possible cases ,but the consumer of these results such as `Recycle-Bin` , `MainActivity`(Deep-link) or `Scanner` they decide what to do with these results

BMItr: Did you comment on the wrong class?
But I can add the restored test to the Event 'RestoredTest', so the consumer can use this to decide what to do next. I guess this would be a little bit better than creating an event for every test result...

mtwalli (approved):
Works nicely 👌 _nothing to complain_

jurajkusnier (approved):

PR #4996, opened 2022-03-25
Revocation - API, Repository, Provider (EXPOSUREAPP-12303, 12310)

BMItr:
This PR addresses EXPOSUREAPP-12303, 12310.

Story: Support Revocation of certificates from an issuer (on invalidation rules).

Implemented:
- Invalidation rules API/Repo
- Provide invalidation rules to Ccl calculation (coming in follow up PR)
- Tests

See ticket for details.
Info: Remaining Story adjustments will come in a separte PR.


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/configuration/update/CclConfigurationUpdaterTest.kt
```diff
@@ -41,19 +43,21 @@ internal class CclConfigurationUpdaterTest : BaseTest() {
 
         coEvery { boosterRulesRepository.update() } returns UpdateResult.UPDATE
         coEvery { cclConfigurationRepository.updateCclConfiguration() } returns UpdateResult.NO_UPDATE
+        coEvery { dccValidationRepository.updateInvalidationRules() } returns UpdateResult.NO_UPDATE
```
LukasLechnerDev, line 46: I think the function names in this Test Class should now also in include ` ... and invalidation rules...`. Currently, they are something like `updateIfRequired() should update booster rules and ccl configuration if required`()`

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/configuration/update/CclConfigurationUpdaterTest.kt
```diff
@@ -73,46 +77,63 @@ internal class CclConfigurationUpdaterTest : BaseTest() {
         }
 
     @Test
-    fun `updateConfiguration() should return true when new booster rules or new configuration was downloaded or false otherwise`() =
+    fun `updateConfiguration() should return true if any item was downloaded or false otherwise`() =
```
LukasLechnerDev, line 80: I think test above called `updateIfRequired() should NOT update booster rules and ccl configuration if NOT required but should trigger DccWalletInfo recalculation` should also verify that the `dccValidationRepository` was not called.

LukasLechnerDev (approved):
LGTM ✅

jurajkusnier (approved):

PR #4999, opened 2022-03-29
Adapt test result screens to include family tests (EXPOSUREAPP-12318,12319,12320)

SamuraiKek:
Text changes and conditions for displaying test results for family tests. The cards displayed in the result screens will be adapted in a different PR.

How to test:

When scanning a pcr/rat test, the result screen will display the right information depending on which type of test you selected after scanning the QR (For me/For others). If you choose to scan the test for others, the test result screen will show up with the name of the person the test belongs to and a few text changes. When returning back to the home screen, the tile for family tests will be visible (selecting it will lead to a blank screen for now).


mtwalli:
I think we need to modify the logic a bit, not just xml changes. Currently view models for test results screens rely on test type, which is not suitable any longer in this case, I think we need to use test identifier and look for the test in both repos

CV113 (dismissed):
UA approved

mtwalli:
@SamuraiKek Did you update the screenshots ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+import timber.log.Timber
+import javax.inject.Inject
+
+class CoronaTestProvider @Inject constructor(
+    private val coronaTestRepository: CoronaTestRepository,
+    private val familyTestRepository: FamilyTestRepository,
+    private val testResultAvailableNotificationService: PCRTestResultAvailableNotificationService,
+) {
+
+    private val allTests: Flow<Set<BaseCoronaTest>> = combine(
```
BMItr, line 21: Add shareLatest

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestProvider.kt
```diff
@@ -0,0 +1,63 @@
...
+        coronaTestRepository.coronaTests,
+        familyTestRepository.familyTests
+    ) { personalCoronaTests, familyCoronaTests ->
+        personalCoronaTests.plus(familyCoronaTests)
+    }
+
+    fun findTestById(testIdentifier: TestIdentifier): Flow<BaseCoronaTest?> =
```
BMItr, line 28: What about renaming this fun like testForIdentifier?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ui/submission/testavailable/SubmissionTestResultAvailableViewModelTest.kt
```diff
@@ -121,7 +121,7 @@ class SubmissionTestResultAvailableViewModelTest : BaseTest() {
     }
 
     @Test
-    fun `go to test result without updating TEK history if NO consent is given`() {
+    suspend fun `go to test result without updating TEK history if NO consent is given`() {
```
BMItr, line 124: RunBlockingTest instead making it a suspend fun

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestProvider.kt
mtwalli: is is private , used internally , no need to share , I actually I added it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/CoronaTestProvider.kt
mtwalli: what difference would it make ? `findTestById` is self explanatory or? you think it is confusing

BMItr (dismissed):
lgtm! please check tests. (will reapprove after update)

mtwalli (approved):

PR #5005, opened 2022-03-30
FIX: Dcc reissuance url (DEV)

mtwalli:
👆


jurajkusnier (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/ui/text/CclTextFormatter.kt
```diff
@@ -89,9 +89,7 @@ class CclTextFormatter @Inject constructor(
         return text.cleanText().format(*parameters.convertValues(locale))
     }
 
-    private fun String.cleanText() = this
-        .replace("%@", "%s")
-        .replace("\\n", "\n")
+    private fun String.cleanText() = this.replace("%@", "%s")
```
mtwalli, line 92: should be fixed by CCL , it affects iOS too

SamuraiKek (approved):

BMItr (approved):

PR #5019, opened 2022-03-31
Duplicate warning screen back navigation (EXPOSUREAPP-12325)

SamuraiKek:
When trying to register another personal test, the duplicate warning screen back button will now take you back to the test registration selection screen instead of returning to the home screen. The button text has also changed.


BMItr (approved):

mtwalli (approved):

KathrinLuecke (commented):
UA approved

PR #5036, opened 2022-04-04
[Feature] Revocation of single certs (EXPOSUREAPP-12455)

BMItr:
Feature Branch for the Revocation of single Certificates.


KathrinLuecke:
Hi @BMItr, I found an error in the family_tests_strings.xml. How should I proceed to correct it?

PR #5039, opened 2022-04-05
Fix: `UK` lang (EXPOSUREAPP-12625)

mtwalli:
🇺🇦

## Ticket 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12625

### DIFF
--- a/Corona-Warn-App/build.gradle
+++ b/Corona-Warn-App/build.gradle
@@ -43,7 +43,7 @@ android {
         println("Used versionName: $versionName")
 
         testInstrumentationRunner "testhelpers.TestApplicationUIRunner"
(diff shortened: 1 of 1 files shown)


chiljamgossow (commented):
There are also supported languages for ccl in de.rki.coronawarnapp.ccl.configuration.model.CclDefaultInputParameters.kt. Should it be added there too?

BMItr (approved):

mtwalli:
> There are also supported languages for ccl in de.rki.coronawarnapp.ccl.configuration.model.CclDefaultInputParameters.kt. Should it be added there too?

Yes, that is a good point, do we have some other places in the app that use config like that ?

mtwalli:
Will provide a new one targeting `2.20`

PR #5041, opened 2022-04-05
Fix: Missing Ids (EXPOSUREAPP-12545)

schauersbergern:
The current Appium tests are relying on ids in the header itself, containing NavigationButton and containing title TextViews.
When the included header layout was replaced with MaterialToolbar in release 2.21, the ids got lost, causing the Appium tests to fail.
This pull requests fixes the issue

Changes Made:
Change toolbar ids
Add BindingAdapter to set ids for MaterialToolbar children
Add resource file for holding ids


jurajkusnier (commented):
It would be much cleaner to make changes on the tester's side, please discuss with the testers if it's possible

Corona-Warn-App/src/main/res/layout/fragment_information_privacy.xml
```diff
@@ -25,7 +26,9 @@
             android:layout_width="match_parent"
             android:layout_height="wrap_content"
             android:focusable="true"
-            app:title="@string/information_privacy_title" />
+            app:title="@string/information_privacy_title"
+            bind:navigationButtonId="@{@id/header_button_back}"
+            bind:titleTextViewId="@{@id/information_privacy_header}" />
```
mtwalli, line 31: maybe call it `automation:*` so we know it is just for automation  , same for other layouts

mtwalli, line 31: maybe call it `automation:*` so we know it is just for automation  , same for other layouts

mtwalli:
> It would be much cleaner to make changes on the tester's side, please discuss with the testers if it's possible



> It would be much cleaner to make changes on the tester's side, please discuss with the testers if it's possible

That is not possible , the idea is to run the automation against our releases, we had a meeting with them already. We agreed to fix it for old screens that has been working ,for new screens , they should find their own solution

mtwalli:
> It would be much cleaner to make changes on the tester's side, please discuss with the testers if it's possible



> It would be much cleaner to make changes on the tester's side, please discuss with the testers if it's possible

That is not possible , the idea is to run the automation against our releases, we had a meeting with them already. We agreed to fix it for old screens that has been working ,for new screens , they should find their own solution

PR #5045, opened 2022-04-05
Update rapid test profile card (EXPOSUREAPP-12564)

SamuraiKek:
Replace the two rapid test profile cards with a new card.

Screen Shot 2022-04-05 at 12 40 01


KathrinLuecke (approved):
UA approved

Corona-Warn-App/src/main/res/layout/include_dispatcher_card.xml
```diff
@@ -22,13 +22,25 @@
             name="icon"
             type="android.graphics.drawable.Drawable" />
 
+        <variable
```
chiljamgossow, line 25: Would be nice to convert this into a custom view class so the preview actually shows the cards layout

SamuraiKek: Yes, I thought about it too, but I would propose we do it after the last sprint when we start refactoring and bug fixing anyway.

chiljamgossow (approved):

mtwalli (approved):

PR #5050, opened 2022-04-05
Fix: Support `UK` 🇺🇦  lang (EXPOSUREAPP-12625)

mtwalli:
- Build a `release` APK and check UK locale exists 🛑 
- Check Ccl Text is also provided
- Gradle will assemble an array to be used in the App to have once unified place to update `supportedLocales`
- Added unit test to check resources agains what we support in case a new locale has been added from example `es` an error message should be reported 
```js
New Locale detected in `res` directory!. Make sure to update the following:
- `supportedLocales` array in build.gradle file
- `CclDefaultInputParameters`
- `CWADateTimeFormatPatternFactory` 
- `CWADateTimeFormatPatternFactoryTest`
- `NewReleaseInfoFragmentTest`
- run the App and check newly added localization and ccl localization ==> iterable contents differ at index [3], expected: <pl> but was: <es>
Expected :pl
Actual   :es
```

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12625

### DIFF
--- a/Corona-Warn-App/build.gradle
+++ b/Corona-Warn-App/build.gradle
@@ -12,6 +12,8 @@ def environmentExtractor = { File path ->
     return "\"${escapedJson}\""
 }
 
+def supportedLocales = ['de', 'en', 'tr', 'bg', 'pl', 'ro', 'uk']
+
 jacoco {
     toolVersion = "0.8.7"
 }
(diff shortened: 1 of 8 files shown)


BMItr (approved):
I did some testing with the release APK..
everything looks fine as far as I can understand :-)

image

chiljamgossow (approved):
Tested and looks fine
Screenshot_20220405_184437

PR #5061, opened 2022-04-07
FIX: Navigation library (EXPOSUREAPP-12571)

jurajkusnier:
Our nav graph is not compatible with the new nav library `2.4.1` that we need to use because we target Android 12. 
See https://issuetracker.google.com/issues/228201897

As a quick fix, I downgraded the nav library back to version`2.3.5`. However, this library is not compatible with Android 12 so I introduced `SafeNavDeepLinkBuilder` which is just a copy of `NavDeepLinkBuilder` but uses `FLAG_IMMUTABLE` when creating `PendingIntent` to make it compatible with Android 12. This is just a dirty fix and hopefully, we will have time to properly fix it later.

### Testing
- test bottom navigation + notifications (possibly all many notifications as possible)
- eg.: scan VC that triggers a notification, close the app, and click on the notification, then you should see certificate detail and when you close it you should be able to navigate to all fragments (Status, Certificates, Check-In, Journal)
notification

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12571


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningFragment.kt
```diff
@@ -27,16 +27,7 @@ class SubmissionDeletionWarningFragment : Fragment(R.layout.fragment_submission_
...
-        NavOptions.Builder().setPopUpTo(
-            destinationId = when {
-                args.popToTestCategorySelection -> R.id.testRegistrationSelectionFragment
-                else -> R.id.submissionDeletionWarningFragment
-            },
-            inclusive = true
-        )
-            .build()
-    }
+    private val navOptions = NavOptions.Builder().setPopUpTo(R.id.submissionDeletionWarningFragment, true).build()
```
chiljamgossow, line 30: Is it on purpose that the when clause is gone?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/notification/DigitalCovidCertificateNotifications.kt
```diff
@@ -50,7 +51,7 @@ class DigitalCovidCertificateNotifications @Inject constructor(
             setSmallIcon(R.drawable.ic_notification_icon_default_small)
             priority = NotificationCompat.PRIORITY_DEFAULT
 
-            val pendingIntent = NavDeepLinkBuilder(context)
+            val pendingIntent = navDeepLinkBuilderFactory.create(context)
                 .setGraph(R.navigation.trace_location_attendee_nav_graph)
                 .setComponentName(MainActivity::class.java)
                 .setDestination(R.id.checkInsFragment)
```
chiljamgossow, line 57: Just noticed this here, although you did not change this here. Seems wrong to me that the DCC notification goes to the checkInsFragment??

chiljamgossow: The destination is overwritten in code, but the graph is wrong, strange...

SamuraiKek (approved):
Tested on device. Everything seems to be working as expected.

chiljamgossow (approved):
Tested check in and the certificate, all worked fine.

PR #5062, opened 2022-04-07
Notification trigger adjustments (EXPOSUREAPP-12484)

BMItr:
This PR addresses [EXPOSUREAPP-12484](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12484)
Trigger behaviour for notfications has been optimizied.

How to test:
- scan invalid, blocked, expired certificates
- check the notifications (see ticket for details and cases)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/expiration/DccExpirationChangeObserver.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        certificateProvider.certificateContainer
+            .onStart { Timber.tag(TAG).d("Started monitoring certs for state changes") }
+            .map { certificateContainer ->
+                certificateContainer.allCwaCertificates
+                    .filterNot { it.getState() is CwaCovidCertificate.State.Valid }
+                    .associate { it.uniqueCertificateIdentifier to it.getState() }
+            }
+            .distinctUntilChanged()
+            .drop(1) // Drop initial emission to avoid unnecessary check after app start
```
mtwalli, line 38: we should actually do it when app comes to foreground see `DccStateCheckScheduler`, maybe we should have it in one place. I suggest we remove it from `DccStateCheckScheduler` and let `DccExpirationChangeObserver` handle state notifications. BUT this observer can run in the background also which is fine as the specs states that notifications should be triggered from background too see https://github.com/corona-warn-app/cwa-app-tech-spec/blob/4fa9158c9c6215630b5b1563a45a55c37122f068/docs/spec/dgc-overview-client.md#creating-local-notifications-for-validity-changes-of-digital-green-certificates

mtwalli (approved):

chiljamgossow (commented):
I tested with the certificate in the ticket on WRU, but got no notification. It should come immediately, correct? Have you tested successfully?

chiljamgossow (approved):

PR #5064, opened 2022-04-07
Revocation - data structures & error codes (EXPOSUREAPP-12552)

BMItr:
This PR addresses EXPOSUREAPP-12552(data), 12559 (errors).
DataStructures & Errorcodes Subtask for Revocation of Single Certificates.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/model/RevocationEntryCoordinates.kt
```diff
@@ -1,11 +1,18 @@
...
+
+    /** See [RevocationHashType] */
+    @SerializedName("type") val type: RevocationHashType,
+
+    /** A hex-encoded byte representing the first byte of the hash */
+    @SerializedName("x") val x: ByteString,
+
+    /** A hex-encoded byte representing the second byte of the hash */
+    @SerializedName("y") val y: ByteString
+)
```
mtwalli, line 18: this is calculated from a DCC, I think we can calculate it on the fly and it does not have to be persisted see [specs](https://github.com/corona-warn-app/cwa-app-tech-spec/blob/abd486e134328fe8484cabd78f65657c3d26366d/docs/spec/dcc-revocation-client.md#data-structure-for-revocation-entry-coordinates)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/model/RevocationChunk.kt
```diff
@@ -1,4 +1,16 @@
...
+data class RevocationChunk(
+    @SerializedName("hashes") val hashes: List<ByteString>
+)
+
+data class CachedRevocationChunk(
+    @SerializedName("kid") val kid: ByteString,
+    @SerializedName("hashType") val hashType: RevocationHashType,
+    @SerializedName("x") val x: ByteString,
+    @SerializedName("y") val y: ByteString,
+    @SerializedName("revocationChunk") val revocationChunk: RevocationChunk
```
mtwalli, line 15: Those data structures are actually proto-buffers , right ? what about avoiding converting them to JSON and save them directly as Proto-buffers. I saw in the skeleton PR https://github.com/corona-warn-app/cwa-app-android/pull/5037#discussion_r844007581 that you plan to use DataStore for persisting it which offers Proto-buff support, can we use that? , or do you see any issues? see https://developer.android.com/topic/libraries/architecture/datastore#proto-datastore

BMItr: First of all, initially the datastore was only planned to be used to store the _execution time_ of the updater, which should only run _once a day_.
Then regarding the _protobufs_. We are not going to save them directly. Only the cached data structures will be stored and they are not defined in the protobufs, so I would have to define new protobufs. -> I guess it would be easier to save them as JSON. I'm also not sure if the datastore supports to save protobufs directly.

mtwalli: Ok, I thought it would be better because we can save proto-buffs with datastore , in general it is more efficient and lesser in size than JSON

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/model/RevocationEntryCoordinates.kt
mtwalli: So please remove SerializedName I think it is not needed than ;)

mtwalli (approved):
👌🏼

PR #5068, opened 2022-04-08
FIX: Broken TAN flow (EXPOSUREAPP-12699)

jurajkusnier:
This PR fixes missing `testIdentifier` in `SubmissionTanFragment`

### Testing
- Sie lassen sich testen? > TAN für PCR-Test eingeben > enter TAN
- You should be able to register test using TAN

You can generate a new TAN on INT env. or use any valid TAN on the local mock server.
If you want to work on a local mock server don't forget to update `.server-config.yml` to look as follows:

```yml
endpoints:
      - name: /version/v1/registrationToken
        responseId: '201'
      - name: /version/v1/tan
        responseId: '201'
      - name: /version/v1/testresult
        responseId: 200+2
```
then start the cwa server and use any valid TAN eg: "9A3B578UMG", "DEU7TKSV3H", "PTPHM35RP4"

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12699


SamuraiKek (approved):
Tested on device with TAN on WRU. Result showed up immediately.

mtwalli (approved):
lgtm

PR #5069, opened 2022-04-08
Travelbutton wrongly active (EXPOSUREAPP-12659)

schauersbergern:
Setting the isLoading Property makes the button active so its only set when the certificate is not blocked

Testing Tip:
set this in CwaCovidCertificate.kt
val isNotBlocked get() = false


schauersbergern:
accidently added wrong commit

PR #5071, opened 2022-04-08
Travelbutton wrongly active (EXPOSUREAPP-12659)

schauersbergern:
Setting the isLoading Property makes the button active so its only set when the certificate is not blocked

Testing Tip:
set this in CwaCovidCertificate.kt
val isNotBlocked get() = false

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12659


mtwalli (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/items/RecoveryCertificateCard.kt
```diff
@@ -62,10 +62,12 @@ class RecoveryCertificateCard(parent: ViewGroup) :
         startValidationCheckButton.apply {
             defaultButton.isEnabled = certificate.isNotBlocked
             isEnabled = certificate.isNotBlocked
-            isLoading = curItem.isLoading
             defaultButton.setOnClickListener {
                 curItem.validateCertificate(certificate.containerId)
             }
+            if (isEnabled) {
+                isLoading = curItem.isLoading
+            }
```
chiljamgossow, line 70: isLoading = isEnabled && curItem.isLoading

if isEnabled is false, is loading is not set in the current implementation

schauersbergern: Yes, thats intended, because if isLoading is set on a disabled button it gets enabled, what should be prevented in this ticket

schauersbergern: Changed condition to certificate.isNotBlocked

mtwalli: I think what Chilja means that you can a short syntax without if  just like:
```kotlin
isLoading = certificate.isNotBlocked && curItem.isLoading
```

schauersbergern: This will not work, because if isLoading is set (whether true or false) the button will activate. 
I also think when the button is not active, setting a loading state is redundant

chiljamgossow: Not sure what you mean, with && both needs to be true to result in true
We should not rely on the default value for it to work, so it should be set either way

chiljamgossow: you can also just add an else branch

mtwalli (approved):
LGTM

chiljamgossow (approved):

PR #5078, opened 2022-04-12
Fix blocked certificate accepted (EXPOSUREAPP-12707)

schauersbergern:
As blocked certificates check changed from uniqueCertificateIdentifier to QR Code Hash validation
old certificates from old versions where accepted as valid.
A re-add of the uniqueCertificateIdentifier check fixes this issue


mtwalli:
@schauersbergern  we should not use AppConfig again, I'm not sure if this issue would need a fix actually, next CCL calculation is fixed. By the time 2.21 is rolled out and all devices have latest CCL version it the certificate will be blocked. It does not make sense to have a fix in 2.22.x in my opinion

PR #5082, opened 2022-04-13
Profile navigation (EXPOSUREAPP-12683)

chiljamgossow:
name of family test is preset if coming from a profile

How to test: creates some profiles, modify, and scan a test qr code from the profile. If you choose family test, the name should be filled already on the consent screen


SamuraiKek (approved):
Works great!

schauersbergern (approved):

PR #5084, opened 2022-04-13
Fix Menu color (EXPOSUREAPP-11515)

schauersbergern:
Add style OverflowMenu for setting the menu color to design requirements

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11515


SamuraiKek (approved):
LGTM.

jurajkusnier (commented):
I see no difference in the design after applying these changes.
I think the bug has already been fixed in 2.21 by the Android 12 redesign and should be set as obsolete.

schauersbergern:
> I see no difference in the design after applying these changes. I think the bug has already been fixed in 2.21 by the Android 12 redesign and should be set as obsolete.

Its difficult to see the difference between a very light gray and white, but when you switch to night mode and compare the share icon on the main fragment to the menu icon you will see the difference between this and 2.21 branch

chiljamgossow (approved):

PR #5086, opened 2022-04-14
Update scheduler (EXPOSUREAPP-12556)

mtwalli:
- [x] Add unit tests
- [x] Add test menu entry

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12556


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
```diff
@@ -30,12 +30,24 @@ class CertificateProvider @Inject constructor(
     dispatcherProvider: DispatcherProvider
 ) {
 
+    /**
+     * All certificates in the App size recycled or not
+     */
```
BMItr, line 35: the certificate container contains all certs... use him to calculate the size?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/update/RevocationListUpdater.kt
```diff
@@ -0,0 +1,74 @@
...
+    private val timeStamper: TimeStamper,
+    private val certificatesProvider: CertificateProvider,
+    private val revocationUpdateSettings: RevocationUpdateSettings,
+    private val revocationRepository: RevocationRepository,
+) {
+
+    private val mutex = Mutex()
+
+    init {
+        appScope.launch {
```
BMItr, line 34: the forced update shall only be triggered when a new cert gets registered.
Is it possible to use distingUntilChanged {} with some little custom logic to compare the size of the emission?

BMItr:
👌 looks Good to me!
two little comments.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
mtwalli: glad you asked 😀, container 🫙 provide active certificates only. I need the  count of all certificates recycled or not so when certificate is moved to and from recycle bin does not count as new registration. see my Kotlin comment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/update/RevocationListUpdater.kt
mtwalli: This flow is a state flow, so I will quote your words. it has it be design 😉

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/common/certificate/CertificateProvider.kt
BMItr: Ahhh, I see now. Smart!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/update/RevocationListUpdater.kt
BMItr: Yes, stateful flow is distintUntilChanged by design but it relies on equalsTo for that. This would mean the update gets triggered when the count reduces, like when a cert gets deleted by the user or after 30 days in recycle bin. So my thought was to use something like distinctUntilChanged { newEmission > oldEmission }. Would that work?

BMItr (approved):

PR #5098, opened 2022-04-19
Profile censor (EXPOSUREAPP-12827)

chiljamgossow:
Create profiles and check the error log


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/profile/ProfileCensor.kt
```diff
@@ -0,0 +1,102 @@
...
+import kotlinx.coroutines.flow.filterNotNull
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import org.joda.time.format.DateTimeFormat
+import javax.inject.Inject
+
+@Reusable
+class ProfileCensor @Inject constructor(
```
BMItr, line 23: Is it really necessary to save all that stuff in here?
checkLog is suspending, so it would be possible to get the values from the flow directly

SamuraiKek (approved):
LGTM.

BMItr (approved):
lgtm

PR #5100, opened 2022-04-20
Fix booster badge for PL & UK languages (EXPOSUREAPP-12777)

SamuraiKek:
The booster badge can be pushed out of sight when text becomes too long (in this case with Polish and Ukrainian translations). Making the text a spannable string with the badge as an `imageSpan` ensures that the badge will always keep the right position no matter the length of text. 

To test, scan the two certificates from the [ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12777) and change the language to Polish or Ukrainian.


jurajkusnier (approved):
interesting solution
tested on Pixel 4 and it works

chiljamgossow (approved):
LGTM

PR #5101, opened 2022-04-20
Improve performance of certificate censor (EXPOSUREAPP-12459)

chiljamgossow:
Due to performance issues, the censoring needs to be improved.
The objective here is to reduce the number of checks that are performed on each log line by removing duplicates in censored items.


jurajkusnier (approved):

SamuraiKek (approved):
LGTM.

PR #5065, opened 2022-04-07
Adapt CollapsingToolbarLayout to end at half height (EXPOSUREAPP-12346)

schauersbergern:
Adapt CollapsingToolbarLayout to end at half height 
of the first element of the certificates list

PR #5103, opened 2022-04-21
Fix certificate deletion app crash (EXPOSUREAPP-12875)

SamuraiKek:
Sometimes when removing all the certificates of a person, the app would crash upon deleting the last remaining certificate. This fix skips over the `PersonDetailsFragment` entirely and goes straing to `PersonOverviewFragment` when deleting the last certificate of a person.

To test, scan one or more certificate for the same person and then delete them one by one.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/person/ui/details/PersonDetailsViewModel.kt
```diff
@@ -116,7 +116,8 @@ class PersonDetailsViewModel @AssistedInject constructor(
 
         return UiState(
```
BMItr, line 117: Handle the logic in the viewmodel and send an gobackEvent if the corresponding numberofCertificates is reached

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -149,6 +149,14 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
     private fun FragmentRecoveryCertificateDetailsBinding.onNavEvent(event: RecoveryCertificateDetailsNavigation) {
         when (event) {
             RecoveryCertificateDetailsNavigation.Back -> popBackStack()
+            RecoveryCertificateDetailsNavigation.ReturnToPersonDetailsAfterRecycling -> {
```
BMItr, line 152: That's acutally the responsibility of the PersonDetailsViewModel..

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/details/TestCertificateDetailsFragment.kt
```diff
@@ -181,6 +181,14 @@ class TestCertificateDetailsFragment : Fragment(R.layout.fragment_test_certifica
     private fun FragmentTestCertificateDetailsBinding.onNavEvent(event: TestCertificateDetailsNavigation) {
         when (event) {
             TestCertificateDetailsNavigation.Back -> popBackStack()
+            TestCertificateDetailsNavigation.ReturnToPersonDetailsAfterRecycling -> {
```
BMItr, line 184: same as above

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/vaccination/ui/details/VaccinationDetailsFragment.kt
```diff
@@ -125,6 +125,14 @@ class VaccinationDetailsFragment : Fragment(R.layout.fragment_vaccination_detail
             viewModel.events.observe(viewLifecycleOwner) { event ->
                 when (event) {
                     VaccinationDetailsNavigation.Back -> popBackStack()
+                    VaccinationDetailsNavigation.ReturnToPersonDetailsAfterRecycling -> {
```
BMItr, line 128: same as above

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
SamuraiKek: Yes, it was, but due to the way we calculate the blue background to always end at half the height of the first card from the `PersonDetailsFragment`, this bug started to show up. I wasn't able fix it without giving up on that new logic that was introduced by @schauersbergern in #5065. I ended up implementing this solution where we travel straight to the `PersonOverviewFragment` because that's where we want to end up anyway.

chiljamgossow (approved):
Tested, worked well

BMItr (approved):

PR #5104, opened 2022-04-21
Revocation of single certs - storage, repo, updateservice (EXPOSUREAPP-12555, 12826)

BMItr:
This PR adds 
- the revocation update service (EXPOSUREAPP-12826, Subtask)
- storage and repo (EXPOSUREAPP-12555, Subtask)
- Date Reset handling

Note: 
- some missing tests will be added in a separte PR
- scheduler update adjustments will be added in a separte PR

Draft PR.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/storage/RevocationRepository.kt
```diff
@@ -1,39 +1,75 @@
...
 import timber.log.Timber
+import java.io.IOException
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class RevocationRepository @Inject constructor(
     @AppScope appScope: CoroutineScope,
-    private val revocationServer: RevocationServer
+    @BaseGson private val gson: Gson,
```
jurajkusnier, line 30: using Jackson would be nice as we want to get rid of Gson

jurajkusnier (approved):
Let's merge it to the feature branch and test everything together

PR #5125, opened 2022-04-27
Navigation on back press in submission flow (EXPOSUREAPP-12904)

chiljamgossow:
How to test:
open the qr code scanner from different paths (eg. the profile, deep link etc.). Once you reach a submission screen (eg. selection, or result) the X or back press should land on the home screen

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12904


schauersbergern (approved):
Works fine

jurajkusnier (approved):

PR #5137, opened 2022-04-29
 FIX: global layout listener in PersonDetailsFragment (DEV)

jurajkusnier:
### Testing
- scan any certificate, for example
image

- check if the toolbar in the person overview screen has the right height
- remove the certificate, restore the certificate
- remove the certificate > restore the certificate (don't click on the notification) > remove certificate > click on the notification
- check if the layout is still the same, the app shouldn't crash


SamuraiKek (approved):
Looks like it's working nicely.

schauersbergern (approved):
Nice solution, works like a charm

PR #5153, opened 2022-05-04
Change padding info button (EXPOSUREAPP-11516)

schauersbergern:
image
General Card Padding is 8dp, to have the figma 16/14 padding, change top padding to 8dp and end padding to 6dp


Corona-Warn-App/src/main/res/values/styles.xml
```diff
@@ -606,7 +606,8 @@
     <style name="CertificateCardInfoButton" parent="buttonIcon">
         <item name="android:height">@dimen/button_icon</item>
         <item name="android:width">@dimen/button_icon</item>
-        <item name="android:padding">14dp</item>
+        <item name="android:padding">8dp</item>
```
chiljamgossow, line 609: reducing the padding makes the clickable are smaller and harder to hit, especially for impaired users

SamuraiKek: Yes, I remember we actually had a fix a while back that introduced a bigger padding exactly for this purpose.

schauersbergern: Do you have a ticket link, then I can close the current ticket with this reference

chiljamgossow: You can change the position of the icon without changing its padding

schauersbergern: Fixed position

chiljamgossow: negative margins seem hacky. You can remove the padding of the card, there are guidelines already.

schauersbergern: removed negative margin, but had also to remove the style because of having padding bottom, without further hacks

SamuraiKek (dismissed):
LGTM.

chiljamgossow (approved):

SamuraiKek (approved):

PR #5159, opened 2022-05-09
Submit analytics date conversion (EXPOSUREAPP-12616)

schauersbergern:
Using the device Timezone for toDateTimeAtStartOfDay results in a date two hours before midnight the day before.  
Using UTC Timezone results in midnight of the same day


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/exposureriskmetadata/ExposureRiskMetadataDonor.kt
```diff
@@ -74,7 +75,7 @@ class ExposureRiskMetadataDonor @Inject constructor(
         val riskLevelPtForMetadata = lastPTRiskResult.toMetadataRiskLevel()
         val mostRecentDateAtPtRiskLevel = lastPTRiskResult
             .mostRecentDateAtRiskState
-            ?.toDateTimeAtStartOfDay()
+            ?.toDateTimeAtStartOfDay(DateTimeZone.UTC)
             ?.toInstant()
```
chiljamgossow, line 79: Can you add a unit test for that?

chiljamgossow: Thanks for the test. Can you add one for the donor to make sure this is and stays fixed?

schauersbergern: Changed present tests to use toDateTimeAtStartOfDayUtc

chiljamgossow (approved):

SamuraiKek (approved):

Issue #5130, opened 2022-04-28
Phone number in RAT Profile not saved (no message, just gone) when entered in a particular format


jkrwdf:
## Avoid duplicates
* [X] Bug is not mentioned in the [FAQ](https://www.coronawarn.app/en/faq/)
* [X] Bug is specific for Android only, for general issues / questions that apply to iOS and Android please raise them in the [documentation repository](https://github.com/corona-warn-app/cwa-documentation)  I DO NOT KNOW, HAVE NO IOS DEVICE WITH CWA
* [X] Bug is not already reported in another issue

## Technical details

- Device name: Google Pixel 4a (5G)
- Android version: 12
- App version: 2.21.1

## Describe the bug

Some phone numbers in quick test profiles are not saved and there is no message or error shown, the data is just lost unnoticeably.

## Steps to reproduce the issue

Create a quick test profile or update an existing quick test profile. 

Important for robust reproduction: There shall be no phone number entered yet. 

If you open an existing profile with number, clear the number first and save before you continue, otherwise it might be that you cannot reproduce the effect.

Enter this phone number:
`+49 172 123 45678`
(the spaces are important) and save. 
--> The number is not saved.

Now edit again and enter this same number without spaces (while entering "the system" will insert some spaces for you and you will get)
`+49 172 12345678`
--> Then the number is saved. 

Now it gets really weird.

Edit again and just insert a space between the "123" and the "45678" so you get to my first example
--> Now the number is saved in the first format. 

## Expected behaviour

Phone number is always saved. 

If CWA however accepts only certain formats and will not save others, an error message must be sent. 

---
Internal Tracking ID: [EXPOSUREAPP-12964](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12964)

MikeMcC399:
@jkrwdf 

I can reproduce this issue.

MikeMcC399:
It seems that there are two issues:
1. The error message "Invalid phone number" is only shown if the focus leaves the field "Phone Number". The problem can only be reproduced if at least one other field on the Rapid Test Profile is populated, otherwise the dialog will not allow the profile to be saved.
2. Spaces should be allowed in telephone numbers, however the dialog treats a space as a phone number input error

To show a bit more how the view reacts to inputs:
1. Delete any existing Rapid Test Profile
3. Enter `+491` into the field "Phone Number"
4. Note that `+49 1` is displayed and the `Save` button is active
[...]

PR #5175, opened 2022-05-13
Phone number in rat profile not saving if validation does not pass (EXPOSUREAPP-12964)

SamuraiKek:
The current implementation of the validation for the phone number and email fields was causing the data to be deleted if the fields were edited and the validation didn't pass. Since these fields are not required for the profile creation I've removed the validation altogether. The appropriate keyboards show up when typing so the formatting of the text should still work.


mtwalli (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/profile/ui/create/ProfileCreateFragment.kt
```diff
@@ -82,39 +81,13 @@ class ProfileCreateFragment : Fragment(R.layout.profile_create_fragment), AutoIn
...
-                    phoneInputLayout.error = null
-                }
+                viewModel.phoneChanged(it.toString())
             }
             phoneInputEdit.addTextChangedListener(PhoneNumberFormattingTextWatcher())
 
             // E-mail
             emailInputEdit.addEmojiFilter().doAfterTextChanged {
-                // Propagate email to view model if it matches the pattern
-                if (Patterns.EMAIL_ADDRESS.matcher(it.toString()).matches()) {
```
schauersbergern, line 105: Shouldn't we keep the e-mail validation at least, to have a valid e-mail? When I scan the QR from the profile it also displays the E-Mail field. Maybe hospitals/officials would use this

MikeMcC399:
I successfully validated against the steps to reproduce in #5130:

## Invalid phone number not saved

1. Click on Manage Your Tests
2. Scroll down and click on Rapid Test Profiles
3. Click "+ RAPID TEST PROFILE"
4. Enter "A" in "First Name" then "9" in "Phone Number"
5. Click "Save"
6. "9" is stored and shown

## Trailing space makes phone number invalid

1. Click on Manage Your Tests
2. Scroll down and click on Rapid Test Profiles
3. Click "+ RAPID TEST PROFILE"
4. Enter a valid telephone number, such as "08007540001" in "Phone Number" then add a space after the digits
5. Click "Save" button
6. The phone number "08007540001" is shown


An enhancement to this UI would be to strip off any trailing blanks. This isn't strictly necessary though if the field verification has been removed. This PR does resolve #5130 in its current form.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/profile/ui/create/ProfileCreateFragment.kt
SamuraiKek: Alright, I will keep the validation for the email.

SamuraiKek: Updated the validation. In the end I also added it for the phone number as it wasn't much trouble.

SamuraiKek:
@MikeMcC399, I've updated the fix to actually not let you save the phone number & email until it's formatted correctly (this also takes care of any trailing spaces).

MikeMcC399:
@SamuraiKek 

> I've updated the fix to actually not let you save the phone number & email until it's formatted correctly (this also takes care of any trailing spaces).

If the user enters a phone number, actively entering `<space>` characters, for example:
`+49<space>172<space>12345678`
there is still an unexpected trailing space left in the field causing "Invalid phone number". Since the trailing space is not immediately visible unless the user moves the cursor to the end of the field, it is difficult to spot the problem.

It is however a definite improvement from the original condition!

schauersbergern (approved):

mtwalli (approved):

PR #5186, opened 2022-05-16
Top margin certificate details (EXPOSUREAPP-13101)

schauersbergern:
How to test:
- Scan invalid certificate (e.g.  cwa vc gen --exp 2021-07-22T08:57:25Z -e mock-cloud)
- Certificate detail view has top margin

Before:

Screenshot_20220516-111633_CWA TEST

After:

Screenshot_20220516-111344_CWA TEST

In Ticket [EXPOSUREAPP-11516](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11516) padding was removed from the card.

This works as long the certificate is valid and the covpass info title and button is shown, since the padding is applied

to the info title.

When the certificate is invalid, covpass info title and button is not shown, so padding has to be applied


SamuraiKek:
It looks funky in dark mode. Couldn't this be done from the layout itself?
Screen Shot 2022-05-16 at 13 14 38

jurajkusnier (dismissed):
Doesn't work in dark mode as @SamuraiKek noticed
Using something like 
`app:layout_goneMarginTop="@dimen/guideline_card"` in `include_certificate_qrcode_card.xml` looks much cleaner to me and it's just one change in one file

schauersbergern:
> Doesn't work in dark mode as @SamuraiKek noticed Using something like `app:layout_goneMarginTop="@dimen/guideline_card"` in `include_certificate_qrcode_card.xml` looks much cleaner to me and it's just one change in one file

Changed to layout

SamuraiKek (approved):
Looks good now 👍

jurajkusnier (approved):

PR #5193, opened 2022-05-17
Family cards (EXPOSUREAPP-12614)

chiljamgossow:
reduce redundancy
fix layout (cause was wrap_content instead of 0dp for the name field)

How to test:
scan different tests for family members
enter a really long name
check the cards

redeemed state is reached when the test is 21 days old and not available on the server, so you can change the environment and time-travel to simulate


mtwalli (dismissed):
Nice refactoring 👍 
Looks nice now:
Screenshot 2022-05-23 at 11 33 01

mtwalli (approved):

schauersbergern (approved):
Looks very good on my Sony Xperia and Pixel 6

PR #5129, opened 2022-04-27
Broken UI after deleted test (EXPOSUREAPP-12938)

BMItr:
This PR addresses Exposureapp-12938.
Bugfix for Notification-Destination after a test has been deleted.

**How to test:**
- scan in a VC or RC
- trigger the revocation of certs (cli tool) (also possible with revocation via invalidation rules)
- delete the test
- follow the notification which should be still there
(steps/details also in the ticket)

PR #5198, opened 2022-05-18
Certificate details screen after scan (EXPOSUREAPP-12991)

SamuraiKek:
Sometimes after scanning a certificate, we would arrive to the certificate details screen too quickly and the certificates flow from the repository wouldn't give us the latest value in time, causing an immediate return to the person details screen. 

To test bug [12991](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12991) simply try scanning many certificates by generating them with the cli tool `cwa vc/tc/rc gen -e mock-cloud`. You should be taken to the certificate details screen each time you scan a new certificate.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/ui/details/TestCertificateDetailsFragment.kt
```diff
@@ -259,15 +250,17 @@ class TestCertificateDetailsFragment : Fragment(R.layout.fragment_test_certifica
     }
 
     private fun showCertificateDeletionRequest() {
```
BMItr, line 252: ⬇️ 
This should not be part of a Fragment..
We already have a DccValidityStateChangeObserver. You can use that and update the notification accordingly if needed

SamuraiKek: Yes, but I don't want to update the notification when the state of the DCC changes. I might have seen the notification already and dismissed it, and then the state changes and I could trigger the notification again. I just want to check if the notification is already emitted, and if it is, then update it when I'm deleting the DCC so it will no longer point to the certificate details screen.

Why shouldn't this be part of a fragment?

mtwalli: I just don't get what is wrong with the previous approach? We always look for the certificate by container ID and  if it is not there or recycled simply move back.

SamuraiKek: The "old" approach was [introduced](https://github.com/corona-warn-app/cwa-app-android/pull/5129) a few weeks ago. Then this bug [12991](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12991) showed up and I think the reason is because sometimes, after scanning a new cert, we get to the details screen too quickly, and the repo didn't update yet, so we called the `popBackstack()` function that lead us away from the screen. 

Why do we keep returning to this approach of Go to screen ->  check if anything can be displayed -> go to other screen if there's nothing to be displayed. We've already had a few bugs because of this, and even though the calculation happens fast, sometimes you can still see the screen for a fraction of a second before reaching the right screen. If we can determine the final destination before travelling to another screen, why try to jump through other screens?

BMItr: My idea was to use _DccValidityStateChangeObserver_ as a trigger for a function which handles the logic to check if a notification needs to be updated.
Putting such logic inside a Fragment has the downsides that it is hard to test and you have to repeat it multiple times, like we see in this PR. Also this code only gets executed if the Fragment is displayed. Are those the only situations where a certificate can get removed? What happens if a certificate gets replaced by reissuance?

SamuraiKek: Alright, I understand now. This makes sense. I will try to do something like this to decouple the logic from the fragments.

SamuraiKek: Gave up on trying to update the notification. The fix now is to simply create a separate flow with the certificates that will give us the latest value more quickly in the details screen.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -158,14 +148,14 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
 
     private fun FragmentRecoveryCertificateDetailsBinding.onNavEvent(event: RecoveryCertificateDetailsNavigation) {
         when (event) {
-            RecoveryCertificateDetailsNavigation.Back -> goBack()
+            RecoveryCertificateDetailsNavigation.Back -> popBackStack()
```
mtwalli, line 151: In VM I don't see where `Back` is triggered

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/test/core/TestCertificateRepository.kt
```diff
@@ -129,6 +130,17 @@ class TestCertificateRepository @Inject constructor(
...
+    fun findCertificateDetails(containerId: TestCertificateContainerId): Flow<TestCertificate?> =
+        internalData.data.map { map ->
+            val container = map[containerId]
+            if (container?.isNotRecycled == true) {
+                container.toTestCertificateWrapper(
+                    valueSetsRepository.latestTestCertificateValueSets.first(),
+                    dccValidityMeasuresObserver.dccValidityMeasures.first()
+                )?.testCertificate
+            } else null
+        }
```
mtwalli, line 142: I think this can be simplified to 
```kotlin
            map[containerId].takeIf { it?.isNotRecycled == true }?.toVaccinationCertificateWrapper(
                valueSetsRepository.latestVaccinationValueSets.first(),
                dccValidityMeasuresObserver.dccValidityMeasures.first()
            )?.vaccinationCertificate
```

mtwalli: and small unit test please ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
mtwalli: `onClose` is called when `<-` is pressed. My understanding is it that it is required to navigate `Back` when a notification lead to the details screen and the certificate is not existing, right? If yes, I don't see this is happening anywhere now. I assumed you want to get the certificate in  reliable way and if it does not exist go Back

SamuraiKek: Ah, yes, I added it back in yesterday. It might not have been there when you reviewed but it should be in there now.

mtwalli (approved):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/core/RecoveryCertificateRepository.kt
```diff
@@ -114,6 +115,14 @@ class RecoveryCertificateRepository @Inject constructor(
             scope = appScope
         )
```
BMItr, line 117: _allcertficates_ is already doing this mapping, why do it here again 😸 ?

mtwalli: when we register a certificate `internalData` flows report that the certificate is registered and returns its `containerId` and this container ID is used to navigate to the details. `allcertficates` is flow based on another flow and some others flows if many certificates are already in the App it takes time to emit the latest certificates while we are sure that the certificate is just registered, `allcertficates` does not have it yet. 

TL;DR
If we are rely on `internalData` flow to know if the certificate is registered, `internalData` should be the source to get the Certificate from

BMItr: all right

BMItr (approved):
LGTM 👌

PR #5199, opened 2022-05-18
replace JodaTime in analytics (EXPOSUREAPP-13114)

jurajkusnier:
Get rid of JodaTime step by step, this time in the bug-reporting

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13114


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
```diff
@@ -70,12 +71,13 @@ class Surveys @Inject constructor(
...
         Timber.v("Requested survey: %s", config)
-        val now = timeStamper.nowUTC
+        val now = timeStamper.nowJavaUTC
 
         oneTimePasswordRepo.otpAuthorizationResult?.apply {
-            if (authorized && redeemedAt.toDateTime().monthOfYear() == now.toDateTime().monthOfYear()) {
+            if (authorized && redeemedAt.toLocalDateUtc().month == now.toLocalDateUtc().month) {
                 throw SurveyException(SurveyException.Type.ALREADY_PARTICIPATED_THIS_MONTH)
             }
+            val a: org.joda.time.Instant
```
mtwalli, line 80: not used ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
```diff
@@ -70,12 +71,13 @@ class Surveys @Inject constructor(
     suspend fun requestDetails(type: Type): Survey {
         val config = appConfigProvider.getAppConfig().survey
         Timber.v("Requested survey: %s", config)
-        val now = timeStamper.nowUTC
+        val now = timeStamper.nowJavaUTC
 
         oneTimePasswordRepo.otpAuthorizationResult?.apply {
-            if (authorized && redeemedAt.toDateTime().monthOfYear() == now.toDateTime().monthOfYear()) {
+            if (authorized && redeemedAt.toLocalDateUtc().month == now.toLocalDateUtc().month) {
```
mtwalli, line 77: Before change DateTime was used, not LocalDate, I think `ZonedDateTime ` should be used here. it is true that month is referenced here , but if the time is at end of month I think it would be problematic , wouldn’t be ?

mtwalli: @chiljamgossow  can you please check that?

mtwalli (dismissed):
LGTM

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
mtwalli: what about ` YearMonth.from(redeemedAt) == YearMonth.from(now)` ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/common/Calculations.kt
```diff
@@ -14,37 +15,37 @@ fun calculateDaysSinceMostRecentDateAtRiskLevelAtTestRegistration(
     testRegisteredAt ?: return -1
     lastDateAtRiskLevel ?: return -1
     if (lastDateAtRiskLevel.isAfter(testRegisteredAt)) return -1
-    return Days.daysBetween(
-        lastDateAtRiskLevel,
-        testRegisteredAt
-    ).days
+    return Duration.between(
```
chiljamgossow, line 18: This should be date changes, so it should be a calculation based on local dates, not on 24h intervals

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
chiljamgossow: The logic before was already flawed, as Mohamed mentioned there is an issue with comparing only the month but not the year. 
Juraj is right too, toLocalDateUtc() is actually better than it was before, as device time zone can change and we use utc in many places for that reason. So add a year check as well? (could also fix it separately as is a bug that was there before)

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/datadonation/analytics/modules/testresult/AnalyticsTestResultCollectorTest.kt
```diff
@@ -59,7 +59,7 @@ class AnalyticsTestResultCollectorTest : BaseTest() {
     fun setup() {
         MockKAnnotations.init(this)
 
-        every { timeStamper.nowUTC } returns Instant.parse("2021-03-02T09:57:11+01:00")
```
chiljamgossow, line 62: Can we keep the time zones here to not spoil the test cases?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/common/Calculations.kt
jurajkusnier: I think the implementation is the same as before, this should also be covered by `PpaDataExtensionsTest`
Can you show me where the new implementation could get different results from the previous one?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/survey/Surveys.kt
jurajkusnier: I've added a year check

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/common/Calculations.kt
chiljamgossow: Some days are only 23 hours when changing to summer time. toDays() will divide by 24h to get the number of days. Two dates both with time start of the day, say 27.3.2022 and 31.3.2022 have four date changes, but only 3 with your solution (given the 27.3. has only 23 h)

jurajkusnier: ok got it, I change the implementation using `ChronoUnit`

chiljamgossow (approved):
LGTM, thanks

mtwalli (approved):

PR #5205, opened 2022-05-19
Hyphenation (EXPOSUREAPP-11250)

schauersbergern:
How to test:
Settings -> Display -> Font Size -> Set Biggest
Settings -> set german language

CWA:
Einstellungen -> Risiko-Ermittlung -> Tap "Länderübergreifende Risiko-Ermittlung" ->
Headline "Länderübergreifende" is broken into "Länder- \n übergreifende"


KathrinLuecke (dismissed):
UA approved

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1745,7 +1745,7 @@ Sollten Sie den Test in der App gelöscht haben, können Sie ihn aus dem Papierk
     <string name="settings_interoperability_subtitle">Teilnehmende Länder</string>
 
     <!-- XHED: Header of interoperability information/configuration view -->
-    <string name="interoperability_configuration_title">Länderübergreifende Risiko-Ermittlung</string>
+    <string name="interoperability_configuration_title">Länder\u00ADübergreifende Risiko-Ermittlung</string>
```
mtwalli, line 1748: what is `\u00AD` ? and why?

SamuraiKek: @mtwalli it's a [soft hyphen](https://www.fileformat.info/info/unicode/char/00ad/index.htm) and it "indicates word division when a word is broken at the end of a line.", according to[ this website](https://jkorpela.fi/shy.html#:~:text=The%20first%2C%20called%20the%20'hard,the%20end%20of%20a%20line.) Looking at the bug, it seems to be used correctly here?

mtwalli: I think ` android:hyphenationFrequency="full"` is enough? I used it before in one of the check-ins feature and afaik it works, why here we need the soft hyphen additionally ?

mtwalli: Without the soft hyphen , it is exactly the same 
Screenshot 2022-05-19 at 10 00 26
Screenshot 2022-05-19 at 10 01 30

mtwalli:
I found another issue , which is kind of related ,the layout is not right , can you please check it ?
Screenshot 2022-05-19 at 10 03 41

schauersbergern:
> I found another issue , which is kind of related ,the layout is not right , can you please check it ? Screenshot 2022-05-19 at 10 03 41

This is adressed in the next pr which follows soon with Ticket https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11531

Corona-Warn-App/src/main/res/values-de/strings.xml
schauersbergern: With the soft hyphen it breaks always at this suggested placement:
Länder-
übergreifende
Risikoermittlung

With NO soft hyphen it breaks always at overflowing placement:
e.g.:
Länderübergre-
ifende Risiko-
ermittlung

@KathrinLuecke what do you prefer? Soft hyphen, or NO Shoft hyphen?

schauersbergern: Thanks, I'm fully convinced now, removed the hyphen

mtwalli (approved):

marcauberer (approved):

PR #5209, opened 2022-05-19
Fix standard header bg (EXPOSUREAPP-13116)

schauersbergern:
scan e.g. cwa vc gen  -e wru


schauersbergern:
> 

yes

mtwalli (approved):

SamuraiKek:
Shouldn't this fix apply not only to Vaccination certificates, but also to Test & Recovery certificates?

schauersbergern:
> Shouldn't this fix apply not only to Vaccination certificates, but also to Test & Recovery certificates?

Yes, only to Vaccination certificates, in Test and Recovery this happens at
```
expandedImage.setImageResource(certificate.expendedImageResource(args.colorShade))
```

SamuraiKek (approved):

PR #5210, opened 2022-05-19
 replace JodaTime in AppConfig (EXPOSUREAPP-13141)

jurajkusnier:
Get rid of JodaTime step by step, this time in the bug-reporting

### Jira Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13141


Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainerTest.kt
```diff
@@ -17,96 +17,96 @@ class ConfigDataContainerTest : BaseTest() {
...
         }
 
-        forOffset(Duration.standardHours(1)).isDeviceTimeCorrect shouldBe true
+        forOffset(Duration.ofHours(1)).isDeviceTimeCorrect shouldBe true
 
         forOffset(Duration.ZERO).isDeviceTimeCorrect shouldBe true
 
-        forOffset(Duration.standardHours(2).minus(1)).isDeviceTimeCorrect shouldBe true
-        forOffset(Duration.standardHours(-2).plus(1)).isDeviceTimeCorrect shouldBe true
+        forOffset(Duration.ofHours(1)).isDeviceTimeCorrect shouldBe true
```
chiljamgossow, line 103: For what reason has the setup changed here?

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/diagnosiskeys/download/HourPackageSyncToolTest.kt
```diff
@@ -88,8 +90,8 @@ class HourPackageSyncToolTest : CommonSyncToolTest() {
 
         every { downloadConfig.revokedHourPackages } returns listOf(
             RevokedKeyPackage.Hour(
-                day = invalidHour.info.day,
-                hour = invalidHour.info.hour!!,
+                day = invalidHour.info.day.toJavaTime(),
+                hour = LocalTime.of(invalidHour.info.hour!!.hourOfDay, 0),
```
chiljamgossow, line 94: Wouldn't it be easier to transform to Java time directly from the data base? We store integers like 23

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/sources/remote/AppConfigServer.kt
```diff
@@ -102,7 +104,7 @@ class AppConfigServer @Inject constructor(
         val rawDate = headers()["Date"] ?: throw IllegalArgumentException(
             "Server date unavailable: ${headers()}"
         )
-        Instant.parse(rawDate, DATE_FORMAT)
+        LocalDateTime.parse(rawDate, DATE_FORMAT).toInstant(ZoneOffset.UTC)
     } catch (e: Exception) {
```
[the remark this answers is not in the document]
jurajkusnier: It's actually not necessary, updated

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/appconfig/internal/ConfigDataContainerTest.kt
jurajkusnier: It was done as part of some bulk find & replace update and it worked. But I updated the test now to have the same values as before

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/diagnosiskeys/download/HourPackageSyncToolTest.kt
jurajkusnier: this is only the temporal solution until we use both joda and java time, in the following PR the implementation should look cleaner

chiljamgossow (approved):

schauersbergern (approved):

PR #5211, opened 2022-05-19
Fix expired certificate icons (EXPOSUREAPP-13128)

chiljamgossow:
scan an expired certificate and check reissuance screen
Screenshot_20220519_181522


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccreissuance/ui/consent/DccReissuanceCertificateCard.kt
```diff
@@ -68,31 +68,62 @@ class DccReissuanceCertificateCard(parent: ViewGroup) :
...
         dccReissuanceCertificateIcon.setImageDrawable(
-            AppCompatResources.getDrawable(context, vaccinationIcon)
+            AppCompatResources.getDrawable(context, expiredIcon)
         )
+        dccReissuanceCertificateBg.setImageDrawable(
+            AppCompatResources.getDrawable(context, expiredBackground)
+        )
+    }
+
+    private fun DccReissuanceCertificateCardBinding.setVaccination(
```
mtwalli, line 85: I think this case be reusable , the icon and title can be passed ?

mtwalli (approved):

SamuraiKek (approved):
LGTM.

PR #5216, opened 2022-05-23
GH actions (EXPOSUREAPP-13085)

mtwalli:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13085)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/playbook/OrganizerPlaybook.kt
```diff
@@ -175,13 +174,8 @@ class OrganizerPlaybook @Inject constructor(
             ErrorType.TAN -> OrganizerSubmissionException.ErrorCode.TAN_OB_CLIENT_ERROR
             ErrorType.SUBMISSION -> OrganizerSubmissionException.ErrorCode.SUBMISSION_OB_CLIENT_ERROR
         }
-        // HTTP status code 5XX
-        is CwaServerError -> when (type) {
```
chiljamgossow, line 179: Why is this removed?

chiljamgossow (commented):
Nice clean up!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/playbook/OrganizerPlaybook.kt
mtwalli: it is the same as `else` , it was duplicate

schauersbergern (approved):
The technical debt subtasks could have their own pull requests. Besides of that it looks fine

chiljamgossow (approved):

PR #5208, opened 2022-05-19
Add back missing title and badge in certificate details (EXPOSUREAPP-13108)

SamuraiKek:
[TIcket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13108)

To test, scan an invalid or soon expiring certificate. The title and a badge should be visible on the certificate details screen.
A valid certificate will not display these two views.

PR #5218, opened 2022-05-23
 Add title and badge to revoked certs (EXPOSUREAPP-13108)

SamuraiKek:
A continuation of #5208, adding the title and badge when the certificate is also revoked.


mtwalli (approved):

BMItr (approved):
wonderful

PR #5219, opened 2022-05-23
Fix censor (EXPOSUREAPP-13127)

chiljamgossow:
adds check to make sure there is no replacement of short names


mtwalli (approved):
LGTM

jurajkusnier (approved):

schauersbergern (approved):

PR #5221, opened 2022-05-23
Use DataStore Approach - Part #1 (EXPOSUREAPP-13157)

BMItr:
Exposureapp-13157
This PR refactores the usage of SharedPreferences and removes FlowPrefs etc.
Part #1.

Info: Do not merge yet.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/upload/history/LogUpload.kt
```diff
@@ -1,8 +1,9 @@
 package de.rki.coronawarnapp.bugreporting.debuglog.upload.history
 
+import com.fasterxml.jackson.annotation.JsonProperty
 import java.time.Instant
 
 data class LogUpload(
```
mtwalli, line 6: I think data models needs to be excluded in proguard rules. See proguard files end

BMItr: LogUpload excluded?
please give some details :-)

mtwalli: In release build the annotations is not kept, Please see our progaurd file end. For more information read caveats section here https://github.com/FasterXML/jackson-module-kotlin/blob/2.14/README.md#caveats

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/preferences/FlowPreference.kt
```diff
@@ -109,6 +111,7 @@ inline fun <reified T : Any?> SharedPreferences.createFlowPreference(
     writer = FlowPreference.basicWriter()
 )
 
+@Deprecated("Switch to DataStore")
```
mtwalli, line 114: I think that is not necessary , it makes sonar complains and we are anyway removing it soon, Please remove it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/preferences/SharedPreferenceExtensions.kt
```diff
@@ -4,6 +4,7 @@ import android.content.SharedPreferences
 import androidx.core.content.edit
 import timber.log.Timber
 
+@Deprecated("Switch to DataStore")
```
mtwalli, line 7: same here

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/upload/history/LogUpload.kt
BMItr: I still think that it is not neccessary but will add it.
Maybe we should add the following if we observe problems with jackson and proguard:
```
-keepclassmembers public class com.mypackage.** {
    public synthetic <methods>;
}
```

mtwalli (dismissed):

jurajkusnier (dismissed):

mtwalli (approved):

PR #5223, opened 2022-05-24
Fetching state for test on home screen (EXPOSUREAPP-10746)

chiljamgossow:
display test result immediately if it is not pending
removed unused method onRemove


SamuraiKek (approved):
LGTM.

jurajkusnier (approved):

PR #5231, opened 2022-05-25
Provide notification type to notification sender (EXPOSUREAPP-13183)

mtwalli:
- Scan certificate from the ticket
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13183

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/BoosterNotificationService.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/BoosterNotificationService.kt
@@ -17,6 +17,9 @@ class BoosterNotificationService @Inject constructor(
     private val timeStamper: TimeStamper,
     private val personCertificatesSettings: PersonCertificatesSettings,
 ) : DccWalletInfoNotificationService {
+
+    override val notificationSenderType: Int = 0xB00CC
+
     override suspend fun notifyIfNecessary(
         personIdentifier: CertificatePersonIdentifier,
         oldWalletInfo: DccWalletInfo?,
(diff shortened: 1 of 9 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/BoosterNotificationService.kt
```diff
@@ -17,6 +17,9 @@ class BoosterNotificationService @Inject constructor(
     private val timeStamper: TimeStamper,
     private val personCertificatesSettings: PersonCertificatesSettings,
 ) : DccWalletInfoNotificationService {
+
+    override val notificationSenderType: Int = 0xB00CC
```
chiljamgossow, line 21: How did you come with these numbers?

chiljamgossow:
Can you provide a litte explanation? Does this mean there can be one notification per person and type at the same time?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/booster/BoosterNotificationService.kt
mtwalli: it can be integer , it can be 0,1,2 for different notification services , I choose hex integer to avoid repeating them accidentally

mtwalli:
> Can you provide a litte explanation? Does this mean there can be one notification per person and type at the same time?
Person notification sender uses PersonIdentifier hash to show a notification  if the same person have multiple notifications
last will override the others, to keep them identifiable I Introduced sender type , So one person can have booster , dcc reissuance and G status notifications at the same time , and of course this will be person

chiljamgossow (approved):

jurajkusnier (approved):

PR #5233, opened 2022-05-25
FIX: Fix time issue (DEV)

jurajkusnier:
### Testing
- install CWA 2.22 set env. to mock-cloud
- App Information > Error Reports > Start > Send Error Report
- install CWA from this PR
- App Information > Error Reports > Start > Send Error Report
- ID History should contain previous and new error reports and app shouldn't crash


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/debuglog/ui/DebugLogFragment.kt
```diff
@@ -142,7 +142,7 @@ class DebugLogFragment : Fragment(R.layout.bugreporting_debuglog_fragment), Auto
 
             val now = Instant.now()
 
-            if (lastLog != null && Duration.between(lastLog, now).toSeconds() < 3) {
+            if (lastLog != null && Duration.between(lastLog, now).seconds < 3) {
```
mtwalli, line 145: what is the difference between `toSeconds` and `seconds`?

BMItr (approved):

mtwalli (approved):

PR #5236, opened 2022-05-25
Use DataStore Approach - Part #2 (EXPOSUREAPP-13158)

BMItr:
[Exposureapp-13158](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13158)
This PR refactores the usage of SharedPreferences and removes FlowPrefs etc.
Part 2


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
```diff
@@ -1,18 +1,56 @@
...
-    fun bindResettableTracingSettings(resettable: TracingSettings): Resettable
+    @Module
+    internal interface ResetModule {
+        @Binds
+        @IntoSet
+        fun bindResettableStorageReset(resettable: StorageReset): Resettable
+    }
 }
+
+private const val STORAGE_DATASTORE_NAME = "storage_localdata"
```
mtwalli, line 44: please be specific ? "storage_localdata" is very generic

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageReset.kt
```diff
@@ -0,0 +1,18 @@
...
+import androidx.datastore.core.DataStore
+import androidx.datastore.preferences.core.Preferences
+import de.rki.coronawarnapp.util.datastore.clear
+import de.rki.coronawarnapp.util.reset.Resettable
+import timber.log.Timber
+import javax.inject.Inject
+
+class StorageReset @Inject constructor(
+    @StorageDataStore private val dataStore: DataStore<Preferences>
+) : Resettable {
```
mtwalli, line 12: what is the purpose of this class? can't we just clear `dataStore` directly ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/encryptionmigration/EncryptedPreferencesMigration.kt
```diff
@@ -47,7 +48,7 @@ class EncryptedPreferencesMigration @Inject constructor(
         Timber.d("Migration finish")
     }
 
-    private fun copyData(encryptedSharedPreferences: SharedPreferences) {
+    private fun copyData(encryptedSharedPreferences: SharedPreferences) = runBlocking {
```
mtwalli, line 51: Uhm  😕 , this is called from the App `onCreate` , please use app scope instead

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/AppConfigModule.kt
```diff
@@ -52,7 +50,6 @@ object AppConfigModule {
         return Retrofit.Builder()
             .client(configHttpClient)
             .baseUrl(url)
-            .addConverterFactory(gsonConverterFactory)
```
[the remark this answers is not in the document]
BMItr: was not used.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/OnboardingSettings.kt
```diff
@@ -1,63 +1,63 @@
...
 import kotlinx.coroutines.flow.map
 import java.time.Instant
-import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class OnboardingSettings @Inject constructor(
-    @AppContext private val context: Context
-) : Resettable {
```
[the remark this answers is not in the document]
BMItr: I clear the datastore entirely in _StorageReset_

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
```diff
@@ -1,18 +1,56 @@
...
+    @Provides
+    @StorageDataStore
+    fun provideDataStore(
+        @AppContext context: Context,
+        @AppScope appScope: CoroutineScope,
+        dispatcherProvider: DispatcherProvider
+    ) = PreferenceDataStoreFactory.create(
+        scope = appScope + dispatcherProvider.IO,
+        produceFile = { context.preferencesDataStoreFile(STORAGE_DATASTORE_NAME) },
+        migrations = LEGACY_SHARED_PREFS.map { SharedPreferencesMigration(context, it) }
```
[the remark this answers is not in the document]
BMItr: What is the benefit in doing so?
The separation of the content is done by corresponding setting classes. I think that is enough.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageReset.kt
mtwalli: ok, but you would need to create a class for every DataStore then , right ? 
if takes specific datastore because of the the annotation

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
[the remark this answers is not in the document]
mtwalli: It is a principal "separation of concerns", three different settings for three different features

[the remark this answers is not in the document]
BMItr: yes and the separation of concern is done by the distinct setting classes. It is not the concern of the setting classes how the data gets stored.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageReset.kt
BMItr: So, your idea is to create a single class where every datastore gets injected and resetted?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
BMItr: I will make it "storage_settings_localdata" !?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageReset.kt
mtwalli: No, you can directly make the respective DataStore resttable and just do the logic there, instead of creating a specific class for each datastore to reset other the datastore, maybe you can highlight why the approach you follow here is better ?

BMItr: how can I do that?

mtwalli: I see why you introduced this class, it is because you want to use the same Datastore for three settings ,so instead of resetting each and evert settings which at the end resets the same DataStore.  It is is confusing and I think you should be keep them saved in different files which is exactly my point above https://github.com/corona-warn-app/cwa-app-android/pull/5236#discussion_r884651509

BMItr: In my opinion it's not really confusing and actually simpler than creating a distinct datastore for every setting which is part of the storage module and then make every setting resettable to clear their own datastore.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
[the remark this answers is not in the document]
schauersbergern: Removed StorageDataStore and implemented resetting for every DataStore implementation, but migrating can only happen in this module where all former SharedPreference implementations are migrated at once

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
schauersbergern: Renamed to "shared_settings_storage" because this is a storage which stores settings which are shared between multiple packages

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageReset.kt
schauersbergern: See https://github.com/corona-warn-app/cwa-app-android/pull/5236#discussion_r893471922

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/encryptionmigration/EncryptedPreferencesMigration.kt
schauersbergern: Changed to run in coroutine scope. If this isn't sufficient, would have to add mutex to all datastore props?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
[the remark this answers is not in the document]
mtwalli: can you please elaborate more why all of them has to be migrated once and in one  datastore?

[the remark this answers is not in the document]
schauersbergern: The combined datastore is now separated into own datastores

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageModule.kt
schauersbergern: See https://github.com/corona-warn-app/cwa-app-android/pull/5236#discussion_r894230617

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/StorageReset.kt
schauersbergern: See https://github.com/corona-warn-app/cwa-app-android/pull/5236#discussion_r894230617

mtwalli:
@schauersbergern yes, that looks good

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/deltaonboarding/ui/DeltaOnboardingFragmentViewModel.kt
```diff
@@ -18,14 +19,17 @@ class DeltaOnboardingFragmentViewModel @AssistedInject constructor(
     private val settings: CWASettings,
     private val analyticsSettings: AnalyticsSettings,
     private val traceLocationSettings: TraceLocationSettings,
-    private val contactDiarySettings: ContactDiarySettings,
+    private val contactDiaryUiSettings: ContactDiaryUiSettings,
```
chiljamgossow, line 22: Can use use the term 'contact journal' everywhere (contact diary is not proper English)

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/deltaonboarding/ui/DeltaOnboardingFragmentViewModel.kt
```diff
@@ -18,14 +19,17 @@ class DeltaOnboardingFragmentViewModel @AssistedInject constructor(
...
     private val covidCertificateSettings: CovidCertificateSettings,
     dispatcherProvider: DispatcherProvider
 ) : CWAViewModel(dispatcherProvider = dispatcherProvider) {
 
     val changelogVersion: LiveData<Long> =
         settings.lastChangelogVersion.flow.asLiveData(context = dispatcherProvider.Default)
 
+    val isOnboardingDone: LiveData<Boolean> = contactDiaryUiSettings.isOnboardingDone
```
chiljamgossow, line 30: can you be mor specific, eg. isContactJournalOnboardingDone

Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/deltaonboarding/ui/DeltaOnboardingFragmentViewModel.kt
schauersbergern: No, I wouldn't do the contactDiary -> contactJournal renaming in this PR, this should be done in its own ticket.
contactDiary naming spreads around the whole codebase and just changing it here would leave ambiguous naming.

chiljamgossow (approved):

mtwalli (approved):

PR #5244, opened 2022-05-30
Fix tests for screenshots (EXPOSUREAPP-11295)

schauersbergern:
Fix tests. This PR can be reviewed and merged


mtwalli (dismissed):

SamuraiKek:
LGTM so far, but have a look at the latest comments in the ticket? Looks like some screen is not scrolled all the way down. Will approve once that's clarified.

BMItr (dismissed):

SamuraiKek (dismissed):

.circleci/config.yml
```diff
@@ -492,6 +492,7 @@ workflows:
           requires:
             - ktlint_check
             - detekt
+      - firebase_screenshots
```
mtwalli, line 495: should be removed

schauersbergern: It's removed now. It just was present because I had to regenerate the screenshots. For that cause I also added the "do not merge" label again. But now it can be merged

mtwalli (approved):

PR #5256, opened 2022-06-03
Reissuance certificate pinning (EXPOSUREAPP-13207)

chiljamgossow:
Checking the server certificate after the request is too late, DCCs have already been transmitted to a malicious party -> needs to happen after the handshake and lead to abort

How to test:
Check that reissuance still works

You might get the error DCC_RI_PIN_MISMATCH when changing the environment and the config has not been updated due to the different server certificate hash


SamuraiKek (approved):
Tested on device. Still working.

mtwalli (approved):

PR #5271, opened 2022-06-09
Refactor exporting individual certificates to use SVG approach (EXPOSUREAPP-13341)

mtwalli:
- Use SVG templates which have better quality and low size in the final PDF
- Introduce new message when no certificates are available to export such all expired or blocked or revoked 
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13341)


KathrinLuecke (approved):
UA approved

SamuraiKek:
I really like the new approach, but for me the dialog when there's no exportable certificates in not working reliably. Sometimes I get the dialog and can't generate the pdf, sometimes, it generates fine.

mtwalli:
> I really like the new approach, but for me the dialog when there's no exportable certificates in not working reliably. Sometimes I get the dialog and can't generate the pdf, sometimes, it generates fine. I tried it with expired certificates and with invalid certificates.

I think it depends on what certificates you would like to export, in general expired, expiring soon , valid and invalid certificates are exportable. Test certificates should be less than 72 hours. You can have a look at https://github.com/corona-warn-app/cwa-app-tech-spec/blob/35d5b1332cba74a1854f6cd2c6394852f203678c/docs/spec/dgc-overview-client.md#displaying-all-dccs-as-pdf for more information

SamuraiKek:
> I think it depends on what certificates you would like to export, in general expired, expiring soon , valid and invalid certificates are exportable. Test certificates should be less than 72 hours. You can have a look at https://github.com/corona-warn-app/cwa-app-tech-spec/blob/35d5b1332cba74a1854f6cd2c6394852f203678c/docs/spec/dgc-overview-client.md#displaying-all-dccs-as-pdf for more information


Ok, I've tried it with revoked certs and it works.

SamuraiKek (approved):
LGTM. Nice improvement 🚀

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/WebView.kt
```diff
@@ -0,0 +1,23 @@
+package de.rki.coronawarnapp.covidcertificate.pdf.ui
+
+import android.webkit.WebSettings
+import android.webkit.WebView
+import android.webkit.WebViewClient
+
+fun WebView.setupWebView(onPageFinished: (view: WebView) -> Unit) {
```
schauersbergern, line 7: I would name this more specific, because this cannot be applied to any webview

SamuraiKek: Let's change this in another PR so we can get @mtwalli PR in now? We have the approval from UA, pushing some changes now would dismiss all approvals and we'd have to wait some more.

schauersbergern (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/WebView.kt
mtwalli: @schauersbergern if you want, rename it the name you like ?

PR #5285, opened 2022-06-14
Fix order of certificates in person details screen (EXPOSUREAPP-13186)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13186)

Explanation of the bug: After scanning a certificate, we look in the personCertificates to check if there's a change and we trigger a wallet recalculation if we find said change. But closing the app right after scanning a certificate will stop the process of recalculating the wallet before it ends. The next time we open the app, it's business as usual, it looks like no change happened so no recalculation is needed. In the end, the priority of the certificates is not getting recalculated either, leading to the bug described in the ticket.

Testing:

Scan many certificates for the same person. Close the app right after scanning a certificate, open it again and check the certificates in the person details screen (follow the video in the ticket).

use `cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=1 v.0.mp='EU/1/20/1528' v.0.dt='2022-06-14' -e wru` and change the dn and sd parameters incrementally each time. Make sure they both have the same value 1,1; 2,2; 3,3; 4,4; 5,5... so on


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTrigger.kt
```diff
@@ -35,8 +34,6 @@ class DccWalletInfoUpdateTrigger @Inject constructor(
     init {
         appScope.launch {
             personCertificateProvider.personCertificates
-                .drop(1) // Drop first value on App start that causes unnecessary calculation
```
SamuraiKek, line 38: Not entirely unnecessary. Yes, removing this line will cause a wallet recalculation on each app start, but will prevent any failed recalculations from closing the app too early. This was the most non-intrussive fix I could think of.

Corona-Warn-App/src/test/java/de/rki/coronawarnapp/ccl/dccwalletinfo/update/DccWalletInfoUpdateTriggerTest.kt
```diff
@@ -102,24 +102,6 @@ internal class DccWalletInfoUpdateTriggerTest : BaseTest() {
         }
     }
 
-    @Test
-    fun `update is not triggered on same certificates change`() = runTest2(true) {
```
chiljamgossow, line 106: Would it make sense to keep this test as it tests something else? I suppose the calculation must run now once

SamuraiKek: Yes, didn't think of that. I will add it back.

chiljamgossow (approved):

PR #4788, opened 2022-02-07
Lifecycle owner crash (DEV)

mtwalli:
Trying to fix this crash:
```console 
java.lang.IllegalStateException: 
  at androidx.fragment.app.Fragment.getViewLifecycleOwner (Fragment.java:2)
  at de.rki.coronawarnapp.util.ui.ViewBindingExtensionsKt$viewBinding$2.invoke (ViewBindingExtensions.kt:3)
  at de.rki.coronawarnapp.util.ui.ViewBindingProperty.getValue (ViewBindingExtensions.kt:9)
  at de.rki.coronawarnapp.qrcode.ui.QrCodeScannerFragment.getBinding (QrCodeScannerFragment.kt:1)
  at de.rki.coronawarnapp.qrcode.ui.QrCodeScannerFragment.startDecode (QrCodeScannerFragment.kt:1)
  at de.rki.coronawarnapp.qrcode.ui.QrCodeScannerFragment$$ExternalSyntheticLambda12.onDismiss (Unknown Source:2)
  at android.app.Dialog$ListenersHandler.handleMessage (Dialog.java:1586)
  at android.os.Handler.dispatchMessage (Handler.java:106)
  at android.os.Looper.loop (Looper.java:214)
  at android.app.ActivityThread.main (ActivityThread.java:7050)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:494)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:965)
```

my assumption, it is a bug in the `viewBinding` lazy that is trying to access the lifecycle owner  while fragment is closing

PR #5290, opened 2022-06-20
Adjust scanner binding to prevent crash (EXPOSUREAPP-13396)

SamuraiKek:
Trying the same method applied in #4788.

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13396)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/warn/qrcode/OrganizerWarnQrCodeScannerFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/warn/qrcode/OrganizerWarnQrCodeScannerFragment.kt
@@ -30,6 +30,8 @@ class OrganizerWarnQrCodeScannerFragment : Fragment(R.layout.fragment_qrcode_sca
     @Inject lateinit var viewModelFactory: CWAViewModelFactoryProvider.Factory
     private val viewModel: OrganizerWarnQrCodeScannerViewModel by cwaViewModels { viewModelFactory }
 
+    // This type of binding initialization should not be followed elsewhere.
+    // Please use lazy initialization wherever possible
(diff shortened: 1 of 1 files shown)


chiljamgossow (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/organizer/warn/qrcode/OrganizerWarnQrCodeScannerFragment.kt
```diff
@@ -31,7 +30,8 @@ class OrganizerWarnQrCodeScannerFragment : Fragment(R.layout.fragment_qrcode_sca
     @Inject lateinit var viewModelFactory: CWAViewModelFactoryProvider.Factory
     private val viewModel: OrganizerWarnQrCodeScannerViewModel by cwaViewModels { viewModelFactory }
 
-    private val binding: FragmentQrcodeScannerBinding by viewBinding()
```
schauersbergern, line 34: We use this delegate for the binding everywhere, why doesn't this work here?

SamuraiKek: Honestly I'm not sure why. All I know is we are using the same layout for the universal scanner and for this `OrganizerWarnOthers` scanner. The bug is related to the binding of the layout and it works with the universal scanner but doesn't work in the organizer one. The universal scanner had this binding initialization change #4788 so I'm trying to see if this will fix the bug once an RC is built.

schauersbergern: Please add a comment that this type of binding initialisation doesn't get removed accidently

schauersbergern (approved):

PR #5291, opened 2022-06-21
Certificates PDF Export is not displayed properly (EXPOSUREAPP-13370)

schauersbergern:
Add h6 and alter style above to prevent Android from inserting roboto font which can't be displayed on other platforms

Testing:
Export Certificates per mail, open file and check if its displayed right in acrobat reader, preview, etc


chiljamgossow (approved):
It is still working on my test device (did not have the issue before)

SamuraiKek (approved):
Works for me.

PR #5299, opened 2022-06-24
Adjust Navigation for PDF Export (EXPOSUREAPP-13438, EXPOSUREAPP-13439)

schauersbergern:
Quitting Export, Export errors and back navigation should all go back to personOverviewFragment
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13438)
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13439)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/exportAll/DccExportAllOverviewFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/exportAll/DccExportAllOverviewFragment.kt
@@ -5,6 +5,8 @@ import androidx.fragment.app.Fragment
 import android.view.View
 import androidx.core.view.isEmpty
 import androidx.core.view.isVisible
+import androidx.navigation.NavOptions
+import androidx.navigation.fragment.findNavController
 import com.google.android.material.dialog.MaterialAlertDialogBuilder
 import de.rki.coronawarnapp.R
 import de.rki.coronawarnapp.covidcertificate.pdf.ui.exportAll.DccExportAllOverviewViewModel.EmptyResult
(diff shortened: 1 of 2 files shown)


chiljamgossow (commented):
Navigation from the dialogs are ok.
When pressing back from the pdf overview I would expect to see the info screen again, not the overview. Was that requested?

SamuraiKek:
> Navigation from the dialogs are ok. When pressing back from the pdf overview I would expect to see the info screen again, not the overview. Was that requested?

I agree. That's how the flow works for single certs exports. I'm not sure whether this was requested or not, but either way, both cases should behave the same imo.

schauersbergern:
> Navigation from the dialogs are ok. When pressing back from the pdf overview I would expect to see the info screen again, not the overview. Was that requested?

Talked with Kerstin Jumpertz and implemented it accordingly to keep backstack on backnav

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/exportAll/DccExportAllOverviewFragment.kt
```diff
@@ -30,13 +32,16 @@ class DccExportAllOverviewFragment : Fragment(R.layout.fragment_dcc_export_all_o
     @Inject
     lateinit var viewModelFactory: CWAViewModelFactoryProvider.Factory
     private val viewModel by cwaViewModels<DccExportAllOverviewViewModel> { viewModelFactory }
+    private val navOptions = NavOptions.Builder().setPopUpTo(R.id.personOverviewFragment, true).build()
```
chiljamgossow, line 35: You can move this into the method as well and only use a local variable

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/exportAll/DccExportAllOverviewFragment.kt
```diff
@@ -92,17 +97,21 @@ class DccExportAllOverviewFragment : Fragment(R.layout.fragment_dcc_export_all_o
             .setMessage(R.string.export_all_error_message)
             .setNeutralButton(R.string.export_all_error_faq) { _, _ ->
                 openUrl(R.string.certificate_export_all_error_dialog_faq_link)
-            }.setPositiveButton(android.R.string.ok) { _, _ -> }
-            .setOnDismissListener { popBackStack() }
+            }.setPositiveButton(android.R.string.ok) { _, _ -> popBackStack() }
```
chiljamgossow, line 100: Is this necessary?

chiljamgossow (commented):
looks fine in general

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/ui/exportAll/DccExportAllOverviewFragment.kt
schauersbergern: No, thanks for pointing out. Removed this call.

SamuraiKek (approved):
LGTM.

PR #5303, opened 2022-06-27
pdf scale issues (EXPOSUREAPP-13440)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13440)
Remove paddings and margins from user agent style sheet


chiljamgossow (approved):
Looks much better, thanks

SamuraiKek (approved):

PR #5306, opened 2022-06-28
Fix back navigation when coming from validity check (EXPOSUREAPP-13428)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13428)

When checking the validity we are either coming from the person details screen or the certificate details screen. When we go back from the result screen of the validity check, we need to skip over the validity check screen and land where we started from, hence the double popBackStack(). I've tried multiple methods but simply calling this function twice seems to work best, althogh I know it's not ideal.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/failed/DccValidationFailedFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/failed/DccValidationFailedFragment.kt
@@ -2,6 +2,7 @@ package de.rki.coronawarnapp.covidcertificate.validation.ui.validationresult.fai
 
 import android.os.Bundle
 import android.view.View
+import androidx.activity.addCallback
 import androidx.fragment.app.Fragment
 import androidx.navigation.fragment.navArgs
 import de.rki.coronawarnapp.R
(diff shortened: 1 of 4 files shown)


schauersbergern (commented):
You could also implement your custom (back)navigation in each fragment. In my opinion it would be more straightforward.
You could implement it like e.g.:

```
findNavController().navigate(R.id.covid_certificates_graph, null, 
NavOptions.Builder().setPopUpTo(R.id.personOverviewFragment, true).build()
)
```
and change the action or destination id like you would need per fragment.
So you would also prevent to multiply the `popBackStackTwice()` function which basically does the same

SamuraiKek:
> You could also implement your custom (back)navigation in each fragment. In my opinion it would be more straightforward. You could implement it like e.g.:
> 
> ```
> findNavController().navigate(R.id.covid_certificates_graph, null, 
> NavOptions.Builder().setPopUpTo(R.id.personOverviewFragment, true).build()
> )
> ```
> 
> and change the action or destination id like you would need per fragment. So you would also prevent to multiply the `popBackStackTwice()` function which basically does the same

But I'm not sure this would be more straightforward. In each fragment I need to return to either `PersonDetailsFragment` or to `CertificateDetailsFragment`, so I'd need to pass around the id of the correct fragment where I want to land in the `setPopUpTo()` function. 

The best method would have been to implement something like this in each fragment:
`parentFragmentManager.popBackStack(parentFragmentManager.getBackStackEntryAt(parentFragmentManager.backStackEntryCount-2).id, FragmentManager.POP_BACK_STACK_INCLUSIVE )` 

This actually takes me back to where I want to, but after I reach the desired screen, if I press the back button 2 times, nothing happens. The third time it actually works, like the backstack is not cleared, even though I can see the correct fragment. From what I understand, the flag `POP_BACK_STACK_INCLUSIVE` should take care of this but I couldn't get it to work...

chiljamgossow (approved):
Works fine. I like simple solutions

schauersbergern (approved):

PR #5313, opened 2022-06-30
Update certificate validity when app comes in foreground (EXPOSUREAPP-13124)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13214)

The validity state will update when the app comes back into foreground.

To test scan a valid 1/1 and an expiring soon 2/2 certificate. Put the app in the background but don't close it and wait for the 2/2 cert to expire. Bring the app back into foreground and check the state of the certificates. 

Testing:
1/1 certificate:
cwa vc gen --pii 42 -s v.0.dn=1 v.0.sd=1 v.0.mp='EU/1/20/1525' v.0.dt='2022-04-04' -e mock-cloud
2/2 certificate:
cwa vc gen --pii 42 -s v.0.dn=2 v.0.sd=2 v.0.mp='EU/1/20/1528' v.0.dt='2022-05-04' --exp 1minutes -e mock-cloud


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/revocation/update/DccRevocationUpdateScheduler.kt
```diff
@@ -41,11 +40,6 @@ class DccRevocationUpdateScheduler @Inject constructor(
             .launchIn(scope = appScope)
     }
 
-    fun forceUpdate() = appScope.launch {
```
SamuraiKek, line 44: Removed because it wasn't being used anywhere.

schauersbergern (dismissed):
Works like requested

schauersbergern (approved):

PR #5330, opened 2022-07-07
Fix notification spam (EXPOSUREAPP-13503)

SamuraiKek:
Spam was caused by wrong grouping of certificates where some users had slight name variations and had certificates that were eligible for booster. The previous grouping caused the `dccWalletInfo` to always be null, triggering a recalculation of the wallet each time the app started or the background worker was performing it's task, subsequently triggering the booster notification over and over again.


schauersbergern (approved):
Works like expected

PR #5335, opened 2022-07-11
Reuse tan if submission fails (EXPOSUREAPP-13477)

chiljamgossow:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13477

If the registration token has already been used to receive a tan, and the submission fails, any retry must use the same tan as the registration token cannot be reused. Therefore the tan aka auth code is persisted with the test.

Also removes BackgroundNoise as the probability was set to 0

How to test: add some positive tests or TANs and warn others. You should not see an error saying that the TAN has been used.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/coronatest/type/rapidantigen/RATestProcessor.kt
```diff
@@ -165,6 +165,13 @@ class RATestProcessor @Inject constructor(
...
 
+    override suspend fun updateAuthCode(test: PersonalCoronaTest, authCode: String): PersonalCoronaTest {
+        Timber.tag(TAG).v("updateAuthCode(test=%s)", test.identifier)
+        test as RACoronaTest
+
+        return test.copy(authCode = authCode)
+    }
+
     override suspend fun markSubmitted(test: PersonalCoronaTest): RACoronaTest {
         Timber.tag(TAG).d("markSubmitted(test=%s)", test)
```
mtwalli, line 176: can you please log the identifier here too similar to `PCRTestProcessor`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/submission/task/SubmissionTask.kt
```diff
@@ -162,8 +163,24 @@ class SubmissionTask @Inject constructor(
...
         Timber.tag(TAG).d("Transformed CheckIns from: %s to: %s", checkIns, checkInsReport)
 
+        val authCode: String
+        try {
+            authCode = playbook.retrieveTan(
+                registrationToken = coronaTest.registrationToken,
+                authCode = coronaTest.authCode
+            )
+        } catch (e: Exception) {
+            playbook.submit(null)
```
mtwalli, line 173: what about making it clear that in this case it submits fake data?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/playbook/Playbook.kt
```diff
@@ -19,23 +32,210 @@ import de.rki.coronawarnapp.server.protocols.internal.pt.CheckInOuterClass
...
+    companion object {
+        const val MIN_NUMBER_OF_SEQUENTIAL_PLAYBOOKS = 1
+        const val MAX_NUMBER_OF_SEQUENTIAL_PLAYBOOKS = 1
+        const val MIN_DELAY_BETWEEN_SEQUENTIAL_PLAYBOOKS = 0
+        const val MAX_DELAY_BETWEEN_SEQUENTIAL_PLAYBOOKS = 0
+    }
 
     data class SubmissionData(
         val registrationToken: String,
+        val authCode: String? = null,
```
mtwalli, line 233: I think this should not be nullable , it is required for submission otherwise we don't submit

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/playbook/Playbook.kt
```diff
@@ -19,23 +32,210 @@ import de.rki.coronawarnapp.server.protocols.internal.pt.CheckInOuterClass
...
+            newAuthCode
+        } else {
+            // fake request
+            verificationServer.retrieveTanFake()
+            authCode
+        }
+    }
+
+    suspend fun submit(
+        data: SubmissionData?
```
mtwalli, line 118: I would split this into two methods one that submits a non nullable data and one for fake submission

mtwalli (commented):
Please make sure it is OK to:
1- persist the auth code per test 
2- remove the background noise

cc / @mlenkeit fyi

chiljamgossow:
> Please make sure it is OK to: 1- persist the auth code per test 2- remove the background noise
> 
> cc / @mlenkeit fyi

1. I asked Max, we need to persist the auth code, otherwise it is not possible to retry in background 2. The background noise on app start was removed as it was not running anyway. The probability was set to 0, so it would never run (there is still background noise triggered by a worker)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/playbook/Playbook.kt
chiljamgossow: It is null until we were successful with exchanging it for the registration token, so it has to be nullable

chiljamgossow: Ah, you are right, I confused with the test data, will adapt it

mtwalli (dismissed):
Thanks for refactoring it!
Tested with several positive QRCodes and TANs

mtwalli (approved):

SamuraiKek (approved):
Lgtm, tried breaking the submission and didn't see any error message.

PR #5343, opened 2022-07-12
Negative test result detail page pending state (EXPOSUREAPP-13465)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13465)
As certificate infos also shouldn't be present on pending state they are completely removed
and logic for hiding/displaying last step is simplified


KathrinLuecke (commented):
UA approved

SamuraiKek (approved):

PR #5349, opened 2022-07-14
Refactor `ProgressLoadingButton` (EXPOSUREAPP-12993)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12993)

- Add isActive state to ProgressLoadingButton
- Replace MaterialButtons with ProgressLoadingButtons on Screens which have fullscreen loading indicator to match CWA iOS app


SamuraiKek:
Nice improvement. Code looks good but I think on some of the screens the width of the button has shrinked. Can you check with the designs please? Ex:
Screen Shot 2022-07-15 at 13 13 52

schauersbergern:
> Nice improvement. Code looks good but I think on some of the screens the width of the button has shrinked. Can you check with the designs please? Ex: Screen Shot 2022-07-15 at 13 13 52

@SamuraiKek Adapted Buttons to current design now

SamuraiKek (dismissed):
LGTM. Buttons have correct width now.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/ProgressLoadingButton.kt
```diff
@@ -54,27 +53,46 @@ class ProgressLoadingButton @JvmOverloads constructor(
...
+                    defaultButton.text = if (value) "" else defaultText
+                    progressIndicator.isVisible = value
+                }
+                field = value
+            }
+        }
+
+    var isActive: Boolean = true
+        set(value) {
+            if (!value) {
```
chiljamgossow, line 82: How come you do not set the field first? Just wandering as you set isLoading here which checks the value of isActive, which has not been updated at this point

mtwalli (approved):

PR #5358, opened 2022-07-19
Upgrade libs (DEV)

mtwalli:
- Upgrade dependencies to the latest versions 
- Upgrade to Kotlin `v1.7.10`
- Upgrade to coroutines `v1.6.2`
- Upgrade other AndroidX libraries 

Note : 
 - There is desugaring new version ,but it requires the AGP plugin 7.3-beta or later  https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md
 - mockK new version causes failure in testing packages see https://github.com/mockk/mockk/issues/819

Test:
- Build release APK and go through it make sure it works as before


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/cards/OccupiedIntensiveCareCard.kt
```diff
@@ -27,7 +28,8 @@ class OccupiedIntensiveCareCard(parent: ViewGroup) :
         )
     }
 
-    private val currentSelectedLocale = ConfigurationCompat.getLocales(resources.configuration).get(0)
+    private val currentSelectedLocale =
+        ConfigurationCompat.getLocales(resources.configuration).get(0) ?: Locale.getDefault()
```
mtwalli, line 32:
```kolin
 ConfigurationCompat.getLocales(resources.configuration).get(0) is nullable now 
```

SamuraiKek (approved):

schauersbergern (approved):

chiljamgossow (approved):
Smoke tested

PR #5361, opened 2022-07-19
Update dates in journal (EXPOSUREAPP-13578)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13578)

If the app is open in background for a few number of days, and the user tries to access the journal, the list of dates will now update to reflect the real date of the device without having to restart the application.

To test, got to the journal overview screen, go to settings, time travel to the future/past, return to the app and check it. The dates should update accordingly.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/overview/ContactDiaryOverviewViewModel.kt
```diff
@@ -70,14 +68,6 @@ class ContactDiaryOverviewViewModel @AssistedInject constructor(
...
     private val datesFlow = MutableStateFlow(dates())
 
-    private val reloadDatesMidnightTimer = fixedRateTimer(
-        name = "Reload-contact-journal-dates-timer-thread",
-        daemon = true,
-        startAt = timeStamper.localDate().plusDays(1).toDate(),
-        period = Days.ONE.toStandardDuration().millis,
-        action = { datesFlow.value = dates() }
-    )
-
```
mtwalli, line 80: Looking at https://github.com/corona-warn-app/cwa-app-android/pull/2835 where this timer was introduced , the changes you made here would fix the issue reported by the respective ticket ,but will introduce the other issue back , please keep the timer as before and may be just update the dates when screen is opened again too.

SamuraiKek: This code isn't working at all. I'm not sure why, but sometime between 2.1 and now, something changed that made the code non functional. When it worked, it was only updating the list at 12 a.m if you happened to be on this screen which is a very limited use case...
My fix is also covering the bug that was fixed by #2835 as long as you refresh the screen or put the app from background to foreground which is much more likely to occur in a real world situation than staying all day on this screen and waiting for 12:00 to show up.

mtwalli: it is a real case scenario reported by users see https://github.com/corona-warn-app/cwa-app-android/issues/2740

SamuraiKek: #2740 reports that the day doesn't show up even after a few hours since 00:00. It looks like traveling from and to the screen or resetarting the app also didn't work. Have you tested the PR?

SamuraiKek: > your solution would only work if the screen is resumed again, but this does not happen if you stay on the screen , you would need an automatic timer as before , and the code just executes every day the duration is set to one day

Exaclty, you would need to resume the screen which actually happens in >99% of real world cases, and if you're part of that 0.1% who stays all day on this screen waiting to update the journal at 00:01 and you see no change, you could just go back to a different screen and open the journal again. As I've said, the previous code isn't working anyway.

mtwalli: I tested it , in your PR automatic addition for new date doesn't happen while in 2.26.0 it does. Please keep the timer as it was before and just add the update onResume to fix the issue in here ?

SamuraiKek: Ok, for me the automatic addition is not working in 2.26, but I'll add back the code.

mtwalli (approved):

PR #5371, opened 2022-07-21
Change duration calculation (EXPOSUREAPP-11549)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11549)

Testing -> see Ticket, intervals are in 10min steps instead of 15 min steps


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/checkout/ContactJournalCheckInEntryCreator.kt
```diff
@@ -61,7 +61,7 @@ class ContactJournalCheckInEntryCreator @Inject constructor(
         // and Check-in EndDate and rounding it to the closest 15-minute duration
```
mtwalli, line 61: can you pease update the comment here ? it is still mentioning 15 minutes

schauersbergern: Updated comment

mtwalli (approved):

PR #5374, opened 2022-07-21
Unresponsive UI after continuous certificate deletion (EXPOSUREAPP-13660)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13660)

How to test:

Using the certificates in the ticket:

1. Scan the first certificate, on the details page, click the three dot menu up top and delete the cert. 
2. Scan the second certificate (go straight to the scanner without traveling to any other screen between the first scan and the second one) -> go to the three dot menu up top and delete the cert -> you should land on the home fragment and the UI should still be responsive.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
```diff
@@ -23,6 +23,9 @@ fun Fragment.popBackStack(): Boolean {
         }
         return false
     }
+    if (findNavController().backQueue.isEmpty()) {
+        findNavController().navigate(R.id.launcherActivity)
+    }
     return findNavController().popBackStack()
```
chiljamgossow, line 29: Wouldn't if... else make sense here? Or does popBackStack still need to be executed?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
```diff
@@ -23,6 +23,9 @@ fun Fragment.popBackStack(): Boolean {
         }
         return false
     }
+    if (findNavController().backQueue.isEmpty()) {
```
mtwalli, line 26: Um I think the actual issue lays  in this part in details fragments 
```kotlin
                        if (args.numberOfCertificates == 1) {
                            doNavigate(
                                VaccinationDetailsFragmentDirections
                                    .actionVaccinationDetailsFragmentToPersonOverviewFragment()
                            )
                        } else popBackStack()

in this case the certificate count is `1` the moment the details screen is opened. the action `action_vaccinationDetailsFragment_to_personOverviewFragment` will execute  which also pops up to `personOverviewFragment` inclusively. using `popBackStack` would fix it ,but I'm not sure why the Navigation was done that way. Maybe you can ;)?
```

SamuraiKek: Yes, I think you're right. I will try and see if I can fix it somehow, I had introduced this piece of code when fixing a bug in the past where deleting the last certificate would sometimes cause the app to crash instead of going back to the overview screen, but I know we've made improvements to the certificates flow in the meantime and that might also fix that bug.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
SamuraiKek: Yes, I think it would.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
SamuraiKek: I've tested this and it's not the cause of the bug... I will get to addressing @chiljamgossow comment and I still think it's the only viable solution for now. Either way, I've only seen it happen if you do these steps that are almost never going to be done in real life. You would have to scan a certificate, delete it right away, and then once again scan a certificate and delete it right away...

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
SamuraiKek: It looks like this extension function is expecting a boolean return statement so I would leave it like this and not add an if/else in here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
mtwalli: Did you try to use `popBackStack` when number of certificates is 1 ? I did try it and did not encounter the issue anymore

SamuraiKek: But args.numberOfCertificates will only have the value 1 if you come from `PersonDetailsFragment` and there's one certificate left... If you come from the UQS to the details screen, number of certificates will be 0, even though you've scanned only 1 certificate.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/FragmentExtensions.kt
chiljamgossow: you can just return true in that case to indicate that navigation has happened

mtwalli (approved):

chiljamgossow (approved):

PR #5379, opened 2022-07-25
Toggle button design (EXPOSUREAPP-13220)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13220)

Check first toggle button


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/CertificateStateHelper.kt
```diff
@@ -177,7 +177,7 @@ fun PersonOverviewItemBinding.setUIState(
                 thirdCertificate?.cwaCertificate
             )
         }
-        certificateToggleGroup.check(0)
+        certificateToggleGroup.check(firstCertificateButton.id)
```
mtwalli, line 180: should we change the id based on what is selected ? there could be 3 certificate buttons ,right?

schauersbergern: This is just for initialisation, the MaterialButtonToggleGroup handles checking the button by itself, so it is not needed to additionally handle checking in click listener

mtwalli: 👍🏼

mtwalli (approved):

PR #5405, opened 2022-08-02
Submission flow symptoms screens (EXPOSUREAPP-13178)

chiljamgossow:
What has been changed:
Navigate directly to thank-you or main screen after submission has been triggered
removed loading dialog as it is now obsolete 

https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13178

Background: 
If the submission takes a while or runs into timeout, users are confused. As we trigger autosubmission anyway in case of failure, there is no need to keep the user waiting.

How to test: check out the submission flow with symptoms

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
@@ -1,7 +1,5 @@
 package de.rki.coronawarnapp.ui.submission.symptoms.calendar
 
-import androidx.lifecycle.LiveData
-import androidx.lifecycle.MediatorLiveData
 import androidx.lifecycle.asLiveData
 import androidx.navigation.NavDirections
 import dagger.assisted.Assisted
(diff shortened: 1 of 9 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
```diff
@@ -92,29 +83,29 @@ class SubmissionSymptomCalendarViewModel @AssistedInject constructor(
...
+        )
     }
 
     fun onCancelConfirmed() {
         Timber.d("onCancelConfirmed() clicked on calendar screen.")
         performSubmission {
             analyticsKeySubmissionCollector.reportSubmittedAfterCancel(testType)
         }
+        routeToScreen.postValue(
+            SubmissionSymptomCalendarFragmentDirections.actionSubmissionSymptomCalendarFragmentToMainFragment()
```
SamuraiKek, line 98: This should not return to the main fragment, but to the tab from where the UQS was opened. I've also opened a PR that should fix this navigation issue within the submission flow: #5408

chiljamgossow: We may not have passed the UQS, eg. if we enter the submission flow from home screen on the TAN flow

SamuraiKek: Yep, let's just merge this one and I'll address it in my PR.

SamuraiKek (approved):

schauersbergern (approved):

mtwalli (approved):
LGTM, tested with TAN and Qrcode Tests

PR #5408, opened 2022-08-02
Fix navigation after test scanning (EXPOSUREAPP-12985)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12985)

~~Scan pending/positive/negative pcr/rat test from the certificates/check-in/journal tabs. 
Go through the submission process and you should land on the same tab from where you opened the universal scanner.
For positive tests, canceling the submission process before it's finished will also land you on the tab from where you opened the universal scanner.~~

The navigation through the submission flow needed to be changed more extensively to match figma and iOS.

When scanning a test (or registaring it through TAN), every operation throughout the submission flow needs to end where the flow was started with the exception of scanning and TAN registering from the dispatcher screen, in which case, the flow should always end on the home fragment. 

What to test:

Scan a new PCR/RAT certificate starting from:

1. Status Tab
2. Certificates Tab
3. Check-In tab
4. Journal Tab
5. Manage Your Tests screen

All types of tests must be tested: negative; pending; invalid and positive
For positive tests, canceling the registration, not entering symptoms/warning others and interrupting the flow in general, should always land you on the screen from where the scanning was started.
      - also try the os back navigation while registering a positive test
      - also try going through the positive test registration with and without giving consent

Try replacing the tests with new ones -> you should go through all/most steps above, especially the positive test registration flow with and without consent.

Try restoring an already deleted test -> again pay attention to positive test registrations, especially ones where the consent was not given and/or others weren't warned.

Try restoring an already deleted test while having a registered test already in the app so that the restored certificate will replace the current one and you have to go through the test deletion screen.

Try registering a test with a TAN.

Try replacing an already registered test by registering a new one with a TAN.

[Figma](https://www.figma.com/file/KuDEcMSREg3VhZRhC3FXTY/COVID19_Master_2.25?node-id=752%3A23612)

P.S. I don't know what to do about detekt's `LongParameterList` error. The params are part of the constructor, can we supress this error somehow?


Corona-Warn-App/src/main/res/navigation/nav_graph.xml
```diff
@@ -358,8 +353,8 @@
         <action
             android:id="@+id/action_submissionSymptomCalendarFragment_to_submissionDoneFragment"
             app:destination="@id/submissionDoneFragment"
-            app:popUpTo="@id/status_nav_graph"
-            app:popUpToInclusive="false" />
+            app:popUpTo="@id/submissionTestResultConsentGivenFragment"
```
chiljamgossow, line 356: We do not pass this screen if we use the TAN flow

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/warnothers/SubmissionResultPositiveOtherWarningNoConsentViewModel.kt
```diff
@@ -37,7 +37,8 @@ class SubmissionResultPositiveOtherWarningNoConsentViewModel @AssistedInject con
     private val coronaTestProvider: CoronaTestProvider,
     private val checkInRepository: CheckInRepository,
     private val analyticsKeySubmissionCollector: AnalyticsKeySubmissionCollector,
-    @Assisted private val testIdentifier: TestIdentifier
+    @Assisted private val testIdentifier: TestIdentifier,
+    @Assisted private val comesFromDispatcherFragment: Boolean
```
schauersbergern, line 41: Just add @Suppress("LongParameterList") above class, its frequently used in codebase for exactly this problem

SamuraiKek: Thanks <3

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningFragment.kt
```diff
@@ -86,24 +86,32 @@ class SubmissionDeletionWarningFragment : Fragment(R.layout.fragment_submission_
...
                 }.also { findNavController().navigate(it, navOptions) }
             }
 
             viewModel.routeToScreen.observe2(this) { event ->
                 Timber.d("Navigating to %s", event)
                 when (event) {
-                    DuplicateWarningEvent.Back -> popBackStack()
+                    DuplicateWarningEvent.Back -> {
+                        if (args.comesFromDispatcherFragment) {
+                            SubmissionDeletionWarningFragmentDirections
```
chiljamgossow, line 111: Did you forget something here?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/positive/SubmissionTestResultKeysSharedFragment.kt
```diff
@@ -75,8 +75,13 @@ class SubmissionTestResultKeysSharedFragment :
...
         }
 
-        viewModel.routeToScreen.observe2(this) { navDirections ->
-            navDirections?.let { doNavigate(it) } ?: popBackStack()
+        viewModel.routeToScreen.observe2(this) {
+            if (navArgs.comesFromDispatcherFragment) {
+                doNavigate(
+                    SubmissionTestResultKeysSharedFragmentDirections
+                        .actionGlobalToSubmissionTestResultKeysSharedFragmentToMainFragment()
+                )
```
chiljamgossow, line 83: I think the first To does not belong there, meaning it should be actionGlobalSubmissionTestResultKeysSharedFragmentToMainFragment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/deletionwarning/SubmissionDeletionWarningFragment.kt
SamuraiKek: Yes, apparently :)

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/positive/SubmissionTestResultKeysSharedFragment.kt
SamuraiKek: Yes, good catch, it should be just `actionSubmissionTestResultKeysSharedFragmentToMainFragment`

chiljamgossow (dismissed):
Tested some paths with TAN and QR codes and worked as expected

chiljamgossow (commented):
While testing for another bug I noticed: When I use the TAN flow and choose No on the symptoms introduction screen I go back to the test result screen, from where sharing fails as it has happened already. Can you check?

Just popping the back stack on the thank you screen is not sufficient. the back stack needs to be cleared when navigating to the thank you screen

SamuraiKek:
> While testing for another bug I noticed: When I use the TAN flow and choose No on the symptoms introduction screen I go back to the test result screen, from where sharing fails as it has happened already. Can you check?
> 
> Just popping the back stack on the thank you screen is not sufficient. the back stack needs to be cleared when navigating to the thank you screen

Alright, I'll have a look. Thanks!

SamuraiKek:
> While testing for another bug I noticed: When I use the TAN flow and choose No on the symptoms introduction screen I go back to the test result screen, from where sharing fails as it has happened already. Can you check?
> 
> Just popping the back stack on the thank you screen is not sufficient. the back stack needs to be cleared when navigating to the thank you screen

I've found the issue and pushed a fix. I've also found a missing dialog which I've added now.

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1181,6 +1181,14 @@ Sollten Sie den Test in der App gelöscht haben, können Sie ihn aus dem Papierk
     <string name="submission_accept_button">"Einverstanden"</string>
     <!-- XACT: other warning - illustration description, explanation image -->
     <string name="submission_positive_other_illustration_description">"Ein Smartphone übermittelt einen positiven Testbefund verschlüsselt ins System."</string>
+    <!-- XHED: CLose screen popup title for no consent given cancel dialog -->
+    <string name="submission_positive_other_warning_dialog_title">"Zufallskennungen mit der App „Corona-Warn“ teilen?"</string>
```
KathrinLuecke, line 1185: Throughout the app, we use "Zufalls-IDs" and not "Zufallskennungen", this should be changed.
Also, I think it would be better to rephrase the title as: "Zufalls-IDs über Corona-Warn-App teilen?"

Corona-Warn-App/src/main/res/values-de/strings.xml
```diff
@@ -1181,6 +1181,14 @@ Sollten Sie den Test in der App gelöscht haben, können Sie ihn aus dem Papierk
...
+    <!-- XHED: Close screen popup body for no consent given cancel dialog -->
+    <string name="submission_positive_other_warning_dialog_body">"Deine Zufallskennungen der letzten 14 Tage werden dazu genutzt, andere darüber zu benachrichtigen, dass du in ihrer Nähe gewesen bist und sie daher möglicherweise in Kontakt mit einer Person gekommen sind, die positiv auf COVID-19 getestet worden ist. Deine Identität wird niemals mit anderen Personen geteilt."</string>
```
KathrinLuecke, line 1187: We never use the informal "Du" in the app, only "Sie". So this should be corrected:
```suggestion
    <string name="submission_positive_other_warning_dialog_body">"Ihre Zufalls-IDs der letzten 14 Tage werden dazu genutzt, andere darüber zu benachrichtigen, dass Sie in ihrer Nähe gewesen sind und sie daher möglicherweise in Kontakt mit einer Person gekommen sind, die positiv auf COVID-19 getestet worden ist. Ihre Identität wird niemals mit anderen Personen geteilt."</string>
```

Corona-Warn-App/src/main/res/values-de/strings.xml
SamuraiKek: Alright, I've updated it with the new text.

chiljamgossow (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/familytest/ui/consent/FamilyTestConsentFragment.kt
```diff
@@ -65,7 +65,10 @@ class FamilyTestConsentFragment : Fragment(R.layout.fragment_family_test_consent
                 }
                 is FamilyTestConsentNavigationEvents.NavigateClose -> {
                     binding.root.hideKeyboard()
-                    popBackStack()
+                    if (navArgs.comesFromDispatcherFragment) {
+                        FamilyTestConsentFragmentDirections
+                            .actionGlobalMainFragment()
```
mtwalli, line 70: shouldn't `doNavigate` be used here?

mtwalli (commented):
Please fix code smells https://sonarcloud.io/project/issues?id=corona-warn-app_cwa-app-android&pullRequest=5408&resolved=false&types=CODE_SMELL

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/familytest/ui/consent/FamilyTestConsentFragment.kt
SamuraiKek: Yes it should

chiljamgossow (dismissed):

mtwalli (approved):

chiljamgossow (approved):

PR #5412, opened 2022-08-03
Test certificate sort order on pdf export (EXPOSUREAPP-13708)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13708)

The sampleCollectedAt date for test certificates also has hour and minute formatting so the instant looks like 2022-08-01T07:00:00.000Z
When this is converted to javaTime the hour and minute information gets lost resulting in 2022-08-01
As sorting also works with instant this conversion is removed to maintain hour and minute information

Testing: Scan the certificates in the ticket and check that they are sorted properly now


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/pdf/core/DccExportFilter.kt
```diff
@@ -30,7 +29,7 @@ internal fun List<CwaCovidCertificate>.sort(): List<CwaCovidCertificate> = sorte
         { it.fullName },
         {
             when (it) {
-                is TestCertificate -> it.sampleCollectedAt?.toLocalDateUserTz()?.toJavaTime()
+                is TestCertificate -> it.sampleCollectedAt
```
mtwalli, line 32: this is not going to work properly and may cause a crash 💥. 
The change you made compare an `Instant` to `LocaleDate`s. `vaccinatedOn` and `testedPositiveOn` are `LocaleDate`s while `sampleCollectedAt ` is an `Instant`

mtwalli: if time is crucial in comparison as issue reports, maybe we need to convert all dates here into `Instant`s

schauersbergern: Converted all dates to instant

mtwalli: looks good. lint is failing

mtwalli (approved):

PR #5418, opened 2022-08-04
Remove time from location start and end dates in diary (EXPOSUREAPP-6443)

mtwalli:
Screenshot 2022-08-04 at 11 48 33

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6443)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInExtension.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInExtension.kt
@@ -14,8 +14,8 @@ val CheckIn.locationName: String
             traceLocationStart.millis > 0 &&
             traceLocationEnd.millis > 0
         ) {
-            val formattedStartDate = traceLocationStart.toUserTimeZone().toString(DateTimeFormat.shortDateTime())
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/presencetracing/checkins/common/CheckInExtension.kt
```diff
@@ -14,8 +14,8 @@ val CheckIn.locationName: String
             traceLocationStart.millis > 0 &&
             traceLocationEnd.millis > 0
         ) {
-            val formattedStartDate = traceLocationStart.toUserTimeZone().toString(DateTimeFormat.shortDateTime())
-            val formattedEndDate = traceLocationEnd.toUserTimeZone().toString(DateTimeFormat.shortDateTime())
+            val formattedStartDate = traceLocationStart.toUserTimeZone().toString(DateTimeFormat.shortDate())
```
schauersbergern, line 17: Shouldn't be the two dates removed completely?
image

mtwalli: According to the ticket , dates without times should appear , see screenshot in the ticket?
There two kind of events , event that has start and end date and one that does not , I think you choose the last category when you created your event

schauersbergern (approved):

PR #5425, opened 2022-08-05
Fix: re-grouping schema 1.3.2 (EXPOSUREAPP-13724, 13723)

mtwalli:
## Test
- Scan the attached certificates group from the ticket
- There should be two groups each and the app should not crash 

[Updated specs
](https://github.com/corona-warn-app/cwa-app-tech-spec/pull/191/commits/33e97a146240d2f9a015955e9a3937ab9b065789)
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13724) , [Ticket2](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13723)


SamuraiKek (approved):
LGTM.

PR #5430, opened 2022-08-08
Use certificate pinner DCC Ticketing (EXPOSUREAPP-13331)

chiljamgossow:
This is a proof of concept, testing whether the certificate pinner can be used in this specific scenario.
In this case, the list of hosts and corresponding sha256 fingerprints is retrieved from the server, so the http client needs to be set up on the go and cannot be injected in advance. I'm not sure if it is possible to use a list of allowed pairs the way I did or at all. This PR was meant to test this approach.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/common/DccJWKConverter.kt
```diff
@@ -37,4 +40,17 @@ class DccJWKConverter @Inject constructor() {
      * @return [PublicKey]
      */
     fun createPublicKey(jwk: DccJWK): PublicKey = createX509Certificate(jwk).publicKey
+
+    fun getCertificatePinner(
+        url: String,
+        jwkSet: Set<DccJWK>
+    ): CertificatePinner {
```
chiljamgossow, line 47: This is where the list of allowed pairs is added, may work this way or not, never tested

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/server/DccTicketingApiProvider.kt
```diff
@@ -0,0 +1,39 @@
...
+    private val gsonConverterFactory: GsonConverterFactory,
+    private val dccJWKConverter: DccJWKConverter,
+    private val dccTicketingApiV1Lazy: Lazy<DccTicketingApiV1>,
+) {
+    private val _dccTicketingApiV1: DccTicketingApiV1
+        get() = dccTicketingApiV1Lazy.get()
+
+    fun getDccTicketingApiV1() = _dccTicketingApiV1
+
+    fun getDccTicketingApiV1(
```
chiljamgossow, line 23: The idea here was to create the client on the fly, to add the certificate pinner with the allow list

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/dccticketing/core/check/DccTicketingServerCertificateChecker.kt
```diff
@@ -15,64 +13,7 @@ import java.security.cert.Certificate
...
-    /**
-     * Checks the server certificate against a set of [DccJWK]
-     *
-     * @throws [DccTicketingServerCertificateCheckException] if the check fails
-     *
-     * Note that the absence of an error code indicates a successful check
-     */
-    @Throws(DccTicketingServerCertificateCheckException::class)
-    fun checkCertificate(certificateChain: List<Certificate>, jwkSet: Set<DccJWK>) = try {
-        Timber.tag(TAG).d("checkCertificate(certificateChain=%s, jwkSet=%s)", certificateChain, jwkSet)
```
chiljamgossow, line 44: The plan was to replace this logic with feeding the allowed list of hosts and fingerprints to the certificate pinner, which would cause an abort at the handshake rather than afterwards

PR #5444, opened 2022-08-09
Share data talkback toggle (EXPOSUREAPP-11544)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-11544)

To test, enable talkback, open cwa settings, go to share data and tap on the toggle switch. It should say "Share data on/off" without being followed by "true/false" at the end.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/MoreInformationView.kt
```diff
@@ -69,7 +69,7 @@ class MoreInformationView @JvmOverloads constructor(
         val subtitle = binding.subtitleElement.text
         val isToggleVisible = binding.toggle.isVisible
         contentDescription = if (isToggleVisible) {
-            "$title $subtitle $isChecked"
+            "$title $subtitle"
```
schauersbergern, line 72: now as both branches of if statement do the same it could be removed from code and set in xml

SamuraiKek: Yes, that's right. I'll change it.

mtwalli (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/MoreInformationView.kt
```diff
@@ -68,11 +68,7 @@ class MoreInformationView @JvmOverloads constructor(
         val title = binding.titleElement.text
         val subtitle = binding.subtitleElement.text
         val isToggleVisible = binding.toggle.isVisible
```
mtwalli, line 70: [Code smell] `isToggleVisible` is not used anymore , should be removed as well from the function

schauersbergern (approved):

mtwalli (approved):

PR #5451, opened 2022-08-11
Stistics accessibility focus talkback on active statistics tyle when scrolling (EXPOSUREAPP-5136)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-5136)

To test, enable talkback and scroll down to the accessibility tiles. Swipe left/right on the tiles. Talkback should automatically focus tiles after swipe and start reading out the details.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/statistics/ui/homecards/StatisticsHomeCard.kt
```diff
@@ -51,6 +53,23 @@ class StatisticsHomeCard(
...
+                                viewHolder?.itemView?.requestFocus()
+                                viewHolder?.itemView?.sendAccessibilityEvent(AccessibilityEvent.TYPE_VIEW_FOCUSED)
+                                viewHolder?.itemView?.announceForAccessibility(
+                                    "$visibleItem of ${statisticsCardAdapter.itemCount}"
+                                )
+                            },
+                            1000
+                        )
+                    }
+                })
```
mtwalli, line 72: these another KTX for postDelayed , could you please use it , it is more Kotlin friendly 
```kotlin
recyclerView.postDelayed(1_000L) {
 // Code 
}
```

SamuraiKek: I've removed `postDelayed` as I've realised it wasn't actually needed at all.

mtwalli (dismissed):

KathrinLuecke (approved):
UA approved

mtwalli (approved):

PR #5459, opened 2022-08-12
Restrict minimum typical stay of Events to `15m` (EXPOSUREAPP-6764)

### LINKED ISSUE #2950: End time not used for automatic checkout when typical stay is 0 minutes (instead: 15 minutes)
## Avoid duplicates
* [X] Bug is specific for Android only (? cannot check, no iOS)

## Technical details

- Device name: Google Pixel 3 XL
- Android version: 11
- App version: 2.0.4

## Describe the bug

An event of type "Event" mandates that the owner sets a "Start" and an "End" date, otherwise it cannot be saved. However, the "Typical Stay" value is pre-set with "00:00 hrs" and can remain on this value. 

image

When now a guest checks in to this Event, then the information about the "End" of the event is completely unused, instead an automatic checkout time of 15 minutes (sic!) is proposed:

image

Due to this, the clients are automatically checked out after this time.

image

Guests can of course later on adapt this time, but I fear this will happen only seldom. 

Therefore I consider this as a bug, not as a Wishlist item. It will cause unwanted early check-outs and therefore missed notification targets. 

Ups, already given above.

## Expected behaviour

I propose that when the owner of the "Event" specified no "Typical Stay" value (which is currently the default), then the guest CWA shall use the "End" time of the event for automatic check out. 

To be honest, this is IMHO the only reason why the "End" date is maintained at all (the "Start" time at least has the effect that guests, when checking-in earlier, get a message "The event has not yet started, do you want to check in anyway", or similar). 

And maybe "Event"s shall not have a "Typical Stay" at all. For them, we already have events of type "Place".

----
Internal Tracking-ID: [EXPOSUREAPP-6764](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6764)

mtwalli:
- While creating a new event minimum stay should be at least 15m 
close #2950
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-6764)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/durationpicker/DurationPicker.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/durationpicker/DurationPicker.kt
@@ -18,7 +18,13 @@ class DurationPicker : DialogFragment(R.layout.duration_picker) {
         fun onChange(duration: Duration)
     }
 
-    private var onChangeListener: OnChangeListener? = null
+    fun interface OnDurationValueChangeListener {
+        fun onDurationChange(duration: Duration)
+    }
+
+    private var onChangeListener = OnChangeListener {}
+    private var onDurationValueChangeListener = OnDurationValueChangeListener { }
+
     private val hoursArray by lazy { requireArguments().getStringArray(HOURS_KEY).orEmpty() }
     private val title by lazy { requireArguments().getString(TITLE_KEY).orEmpty() }
     private val minutesArray by lazy { requireArguments().getStringArray(MINUTES_KEY).orEmpty() }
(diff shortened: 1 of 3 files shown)


chiljamgossow (commented):
The ticket says : 'When the owner of the "Event" specified no "Typical Stay" value (which is currently the default), then the guest CWA shall use the "End" time of the event for automatic check out.'
You have changed the default duration to 15 min, which seems not to be what the ticket asked for, or was there some further discussion?

mtwalli:
> The ticket says : 'When the owner of the "Event" specified no "Typical Stay" value (which is currently the default), then the guest CWA shall use the "End" time of the event for automatic check out.' You have changed the default duration to 15 min, which seems not to be what the ticket asked for, or was there some further discussion?

Please follow the discussion on the Github issue linked above specially https://github.com/corona-warn-app/cwa-app-android/issues/2950#issuecomment-826147376
on iOS this seem to be the behaviour

chiljamgossow (approved):

SamuraiKek (approved):

PR #5462, opened 2022-08-16
Refine setting of toolbar menu ids (EXPOSUREAPP-13750)

schauersbergern:
ToolbarUtils.getActionMenuView just got us the wrapping layout, which test automation could not refer to,
so this change just searches for the button to set the id directly.
Also set the same Id to all certificate detail views

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13750)


mtwalli (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ui/MaterialToolbarExtensions.kt
```diff
@@ -1,12 +1,17 @@
...
 import android.annotation.SuppressLint
+import androidx.core.view.children
 import com.google.android.material.appbar.MaterialToolbar
 import com.google.android.material.internal.ToolbarUtils
 
 @SuppressLint("RestrictedApi")
 fun MaterialToolbar.addMenuId(id: Int) {
-    ToolbarUtils.getActionMenuView(this)?.id = id
+    ToolbarUtils.getActionMenuView(this)?.children?.forEach {
+        if (it.javaClass.name == "androidx.appcompat.widget.ActionMenuPresenter\$OverflowMenuButton") {
```
mtwalli, line 11: does this mean that all views in the menu will have the same ID? is that intended

chiljamgossow: This seems very specific. Do we need this in generic method?

schauersbergern: If an action button is added its class is androidx.appcompat.view.menu.ActionMenuItemView, so only the Menu Button gets set with this id. 
Before the whole layout which could contain multiple actionbuttons and one menu button was set with the id, so test automation couldn't click the specific Menu button

PR #5475, opened 2022-08-19
RAT result text no certificate (EXPOSUREAPP-8544)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8544)

Use figma text when no RAT test certificate is available or requested

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/negative/SubmissionTestResultNegativeFragment.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/testresult/negative/SubmissionTestResultNegativeFragment.kt
@@ -143,6 +143,16 @@ class SubmissionTestResultNegativeFragment : Fragment(R.layout.fragment_submissi
                 R.string.test_certificate_sampled_on,
                 certificate?.sampleCollectedAt?.toUserTimeZone()?.toDayFormat()
             )
+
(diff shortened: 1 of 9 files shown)


KathrinLuecke (commented):
I am not sure whether this is correct and it somehow doesn't reflect the issue described in the Jira ticket. In the ticket it was requested to show the sentence "Ein offizielles digitales COVID-Testzertifikat kann nicht ausgestellt werden, da diese Teststelle die Ausstellung der Testzertifikate nicht unterstützt." also **after** the test result is available. The issue was that this sentences was shown as long as the result was pending, but not afterwards.
And the sentence that you now added ""Sie können den hier angezeigten Befund auch als Nachweis für das Vorliegen eines negativen Schnelltest-Ergebnisses verwenden....." doesn't sound right to me. I could not find it either for i/OS. I think this needs to be double-checked.

KathrinLuecke:
Martin Hermes confirmed that "Sie können den hier angezeigten Befund auch als Nachweis für das Vorliegen eines negativen Schnelltest-Ergebnisses verwenden....." is outdated. 
In addition, he said that https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8544 should not be considered a bug, but a user story, as we then need to distinguish two cases (user has not requested digital certificate, user has requested digital certificate + test site doesn't support it), and we need to consider i/OS then as well. 
So this PR is obsolete and can be reverted.

PR #5476, opened 2022-08-19
DccValueSets language changes with system language (EXPOSUREAPP-8864)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8684)

Some text fields were not changing when switching system language.

To test, check the validity of a certificate and try to land on the "Certificate not valid page". Go to system settings and change the language. All text should be in the system language set, with the exception of the affected fields that can't be transalted and are only available in German/English. 
Also the country name should be displayed in the language you switched to.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/businessrule/BusinessRuleVH.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/businessrule/BusinessRuleVH.kt
@@ -3,13 +3,15 @@ package de.rki.coronawarnapp.covidcertificate.validation.ui.validationresult.com
 import android.view.ViewGroup
 import androidx.annotation.DrawableRes
 import de.rki.coronawarnapp.R
+import de.rki.coronawarnapp.covidcertificate.validation.core.rule.DccValidationRule
 import de.rki.coronawarnapp.covidcertificate.validation.ui.validationresult.common.listitem.BaseValidationResultVH
 import de.rki.coronawarnapp.covidcertificate.validation.ui.validationresult.common.listitem.EvaluatedField
 import de.rki.coronawarnapp.covidcertificate.validation.ui.validationresult.common.listitem.ValidationResultItem
 import de.rki.coronawarnapp.databinding.CovidCertificateValidationResultRuleItemBinding
 import de.rki.coronawarnapp.util.lists.diffutil.HasPayloadDiffer
 import de.rki.coronawarnapp.util.lists.diffutil.update
-import de.rki.coronawarnapp.util.ui.LazyString
+import de.rki.coronawarnapp.util.ui.toResolvingString
+import java.util.Locale
 
 class BusinessRuleVH(
     parent: ViewGroup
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/businessrule/BusinessRuleVH.kt
```diff
@@ -42,10 +46,22 @@ class BusinessRuleVH(
...
+        val descItem = description.find { it.languageCode == currentLocaleCode }
+            ?: description.find { it.languageCode == "en" } ?: description.firstOrNull()
+        return descItem?.description ?: identifier
+    }
+
     data class Item(
         @DrawableRes val ruleIconRes: Int,
-        val ruleDescriptionText: LazyString,
-        val countryInformationText: LazyString,
+        val ruleDescriptionText: DccValidationRule,
```
mtwalli, line 63: can you pleas rename it to `rule`?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/businessrule/BusinessRuleVH.kt
```diff
@@ -42,10 +46,22 @@ class BusinessRuleVH(
...
+        val countryInformationText: Pair<Int, String?>,
```
mtwalli, line 64: since you are already passing the `rule` do you still need to  this `pair` ?  I suggest you do the following check in `VH` directly.

```kotlin
when(item.rule.type){
  DccValidationRule.Type.ACCEPTANCE ->
            context.getString(R.string.validation_rules_acceptance_country,
            country)
   DccValidationRule.Type.INVALIDATION -> 
          context.getString( R.string.validation_rules_invalidation_country)
}           
```

what do you think?

SamuraiKek: Yes, I was thinking about that too. I'll push an update with the code moved.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/businessrule/BusinessRuleVH.kt
SamuraiKek: Yes.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/validation/ui/validationresult/common/listitem/businessrule/BusinessRuleVH.kt
```diff
@@ -42,10 +52,21 @@ class BusinessRuleVH(
         }
     }
 
+    private fun countryName(countryCode: String?, userLocale: Locale = Locale.getDefault()): String =
+        if (countryCode != null) {
```
mtwalli, line 56: `countryCode is not nullable anymore, condition can be removed?

mtwalli (approved):
Lgtm

chiljamgossow (approved):

PR #5487, opened 2022-08-23
Add plurals handling (EXPOSUREAPP-13427)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13427)

### DIFF
--- a/Corona-Warn-App/src/main/res/values-bg/covid_certificate_strings.xml
+++ b/Corona-Warn-App/src/main/res/values-bg/covid_certificate_strings.xml
@@ -233,12 +233,6 @@
     <string name="validation_rules_result_valid_header">"Сертификатът е валиден"</string>
     <!-- XTXT: No rules available text for the validation passed screen  -->
     <string name="validation_no_rules_available_valid_text">"В момента няма правила относно влизането в избраната държава в посоченото време за този вид сертификат. Моля, обърнете внимание на информацията по-долу."</string>
-    <!-- XTXT: Rule validation text for the validation passed screen  -->
-    <plurals name="validation_rules_result_valid_rule_text">
-        <item quantity="one"></item>
-        <item quantity="other">"Успешно преразгледани правила за държавата на местоназначение: %1$d."</item>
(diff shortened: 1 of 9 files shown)


Corona-Warn-App/src/main/res/values-bg/covid_certificate_strings.xml
```diff
@@ -234,7 +234,11 @@
     <!-- XTXT: No rules available text for the validation passed screen  -->
     <string name="validation_no_rules_available_valid_text">"В момента няма правила относно влизането в избраната държава в посоченото време за този вид сертификат. Моля, обърнете внимание на информацията по-долу."</string>
     <!-- XTXT: Rule validation text for the validation passed screen  -->
-    <string name="validation_rules_result_valid_rule_text">"Успешно преразгледани правила за държавата на местоназначение: %1$d."</string>
+    <plurals name="validation_rules_result_valid_rule_text">
+        <item quantity="one"></item>
+        <item quantity="other">"Успешно преразгледани правила за държавата на местоназначение: %1$d."</item>
+        <item quantity="zero">"В момента няма правила относно влизането в избраната държава в посоченото време за този вид сертификат. Моля, обърнете внимание на информацията по-долу."</item>
+    </plurals>
```
mtwalli, line 241: please add DE and EN only ? to avoid any possible issues with the translation tool.
I guess you had to do this because of the resource type change. I suggest in this case to remove the old key from
DE file `- validation_rules_result_valid_rule_text ` and add new resource in DE and EN only 

`+ validation_rules_result_valid_text`

the rest will be taken care by Translation team, they will remove old one from other files and add new ones to other files. Keep in mind that DE is the file reference for Translation. when adding a new key it is necessary to add it in EN file so Android build won't fail

schauersbergern: Removed other plurals than en and de

KathrinLuecke (approved):
UA approved

mtwalli (approved):

Issue #3166, opened 2021-05-12
Privacy policy not displayed correctly on old devices


fynngodau:
## Technical details

- Android version: API 23 / Android 6
- App version: 2.2.x

## Describe the bug

The ordered list is not parsed correctly by the platform HTML parser on old Android versions.

Result on API 23 | Result on API 28
---|---
![Screenshot_1620857867](https://user-images.githubusercontent.com/16943720/118051419-4681d900-b381-11eb-9bd8-b872f32b1312.png) | ![Screenshot_1620857853](https://user-images.githubusercontent.com/16943720/118051421-471a6f80-b381-11eb-97f2-1bfa158db199.png)

## Steps to reproduce the issue

View privacy policy on affected platform versions.

## Expected behaviour

Display text correctly. Possibly also show numbers in front of the *ordered* list?

## Possible fixes

* Remove ordered list
* Use custom tag handler such as this one: https://bitbucket.org/Kuitsi/android-textview-html-list/src/master/app/src/main/java/fi/iki/kuitsi/listtest/MyTagHandler.java

## Additional context

Regression in 2.2.x due to a change in the privacy policy that added an ordered list instead of `p` tags.

----
Internal Tracking-ID: [EXPOSUREAPP-7268](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7268)

heinezen:
@fynngodau 

Thank you. The issue has been reported to the devs (EXPOSUREAPP-7268).

----
Corona-Warn-App Open Source Team

MikeMcC399:
Is this supposed to have been fixed in the release/2.3.x branch or is it still a work-in-progress? The changes which have been made so far in the release/2.3.x branch don't solve the issue well and have not returned the display of the privacy policy to the clean status it had in the 2.1 release.

Using release/2.3.x to view the English language privacy policy:
- on API 23 there are no line breaks between each entry in the table of contents list
- on API 30 there is an extra bullet before the number of each entry in the table of contents list

Perhaps it is time to revert to 2.1.x formatting?

---
release/2.3.x on API 23

![image](https://user-images.githubusercontent.com/66998419/119685541-be520800-be45-11eb-9db0-6b59a7235c3a.png)

---
release/2.3.x on API 30

![image](https://user-images.githubusercontent.com/66998419/119686704-e55d0980-be46-11eb-874f-e4037e6e2f0e.png)

---
released version 2.1.2 on API 23

[...]

PR #5498, opened 2022-08-25
Privacy policy not displayed correctly on old devices (EXPOSUREAPP-7268)

mtwalli:
Please make sure that all App Information sub screens are working as before + on Android 6 have right formatting

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-7268

close https://github.com/corona-warn-app/cwa-app-android/issues/3166

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DataBindingAdapters.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/DataBindingAdapters.kt
@@ -3,6 +3,7 @@ package de.rki.coronawarnapp.util
 import android.content.res.ColorStateList
 import android.graphics.PorterDuff
 import android.graphics.PorterDuffColorFilter
+import android.webkit.WebView
 import android.widget.ImageView
 import androidx.core.widget.ImageViewCompat
 import androidx.databinding.BindingAdapter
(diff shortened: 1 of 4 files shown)


MikeMcC399:
This PR causes some longer URLs not to be completely displayed any more.

On Android 8, using 2.25.0 and selecting App Information > Data Privacy then URLs are fully visible. See especially the URLs in Section 7:

Android 8 2 25

With this PR, the URLs still work as hyperlinks however they are truncated when displayed:

Android 8 issue with fix

mtwalli:
> This PR causes some longer URLs not to be completely displayed any more.
> 
> On Android 8, using 2.25.0 and selecting App Information > Data Privacy, then URLs are fully visible. See especially the URLs in Section 7:
> 
> Android 8 2 25
> 
> With this PR, the URLs still work as hyperlinks however they are truncated when displayed:
> 
> Android 8 issue with fix
> 
> (Same problem on Android 6 and 12.)

@MikeMcC399  it is not truncated, you just need to scroll to the right. I will try to make content fit in the screen width.

MikeMcC399:
@mtwalli 

> it is not truncated, you just need to scroll to the right. 

You are right. Since there is no scroll-bar shown, I didn't think of trying to scroll to the right.

> I will try to make content fit in the screen width.

Thank you!

I noticed two other unwanted side-effects of this change on Android 8:

1. When the privacy document is shown during onboarding, it displays as literal HTML text instead of being interpreted as HTML.
2. Links in the Terms of Use (App Information > Terms of Use) are no longer clickable.

mtwalli:
Looks like using WebView comes with other issues ,despite text should be in HTML format , it appears to have some formatting issues. Closing this PR in favour of https://github.com/corona-warn-app/cwa-app-android/pull/5502

PR #5500, opened 2022-08-26
Terms of use hyperlinks (EXPOSUREAPP-8848)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-8848)

### DIFF
--- a/Corona-Warn-App/src/main/assets/terms_de.html
+++ b/Corona-Warn-App/src/main/assets/terms_de.html
@@ -73,8 +73,8 @@
     können diese Nutzungsbedingungen jederzeit in der App unter
     „App-Informationen“ und unter https://www.coronawarn.app/assets/documents/cwa-eula-de.pdf
     abrufen. Die separate Datenschutzerklärung (siehe unter „App-Informationen“
-    und unter https://www.coronawarn.app/assets/documents/cwa-privacy-notice-de.pdf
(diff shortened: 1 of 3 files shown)


SamuraiKek (approved):

MikeMcC399:
- I noticed a side effect in the open PR https://github.com/corona-warn-app/cwa-app-android/pull/5498#issuecomment-1228445813 which would make the links in the Terms of Use not clickable. I can see this on Android 8 and 12.
- I wonder also why the Terms of Use links are plain text whereas the Privacy links use HTML tags.

mtwalli:
> * I noticed a side effect in the open PR [Privacy policy not displayed correctly on old devices (EXPOSUREAPP-7268) #5498 (comment)](https://github.com/corona-warn-app/cwa-app-android/pull/5498#issuecomment-1228445813) which would make the links in the Terms of Use not clickable. I can see this on Android 8 and 12.
> 
>     * I wonder also why the Terms of Use links are plain text whereas the Privacy links use HTML tags.

@MikeMcC399  #5498 is not merged yet, I think Terms of use links were broken even before . this PR is fixing them which is independent from the other PR

MikeMcC399:
@mtwalli 

- Thanks for the feedback! Now that this PR for Terms of Use is merged, we can continue the discussion only in #5498.

PR #5501, opened 2022-08-26
Text space in invalid test result (EXPOSUREAPP-12698)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12698)

Added space after fullstop in Bulgarian translation.


schauersbergern (approved):

mtwalli:
I think you should inform Albert about it , as long as it is not in the DE file it will be overridden again by translation tool

PR #5509, opened 2022-08-29
Separate date and time with commas (EXPOSUREAPP-13709)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13709)


mtwalli (dismissed):

Corona-Warn-App/src/main/res/values-de/event_registration_strings.xml
```diff
@@ -193,10 +193,10 @@
     <string name="trace_location_organiser_list_delete_all_popup_positive_button">"Löschen"</string>
 
     <!-- XTXT: Event organiser list item: duration-->
-    <string name="trace_location_organizer_list_item_duration">"%1$s - %2$s Uhr"</string>
+    <string name="trace_location_organizer_list_item_duration">"%1$s, - %2$s Uhr"</string>
```
mtwalli, line 196: I think there is something wrong here how `, -` be part of the text after each other? btw I think I had a look at this issue before , we use  `DateTimeForamtter.shortDateTime()` in some places which by default does not add `comma`. if you a comma is needed I believe we need to construct the string using  `DateTimeForamtter.shortDate() and DateTimeForamtter.shorTime()` separately

SamuraiKek: Ah, you are right. I thought it was like the others.

KathrinLuecke (dismissed):
UA approved

mtwalli (dismissed):

PR #5510, opened 2022-08-29
Adjust foreground risk calc (EXPOSUREAPP-13155)

chiljamgossow:
Before: Risk calculation was only triggered on app start (not if app remains in background), ew download on every resume when online and no recent download. -> no failed risk calc card when offline and app has been alive in background

Now: starts ew and pt calculations when app comes into foreground and onboarded, at most once per hour (unless force closed). Latest run is checked in the ew download task anyway, so if there was a recent download the task is aborted.
Home card UI adjusted to prevent frequent resizing.

- [X] unit test


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
```diff
@@ -90,69 +76,12 @@ class TracingRepository @Inject constructor(
...
         exposureWindowRiskWorkScheduler.runRiskTasksNow(sourceTag = TAG)
         presenceTracingRiskWorkScheduler.runRiskTaskNow(sourceTag = TAG)
     }
 
-    /**
-     * Launches the RetrieveDiagnosisKeysTransaction and RiskLevelTransaction in the viewModel scope
-     */
-    suspend fun refreshRiskLevel() {
-        // check if the network is enabled to make the server fetch
-        val isNetworkEnabled = networkStateProvider.networkState.first().isInternetAvailable
```
mtwalli, line 105: where network is checked now ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ForegroundRiskCalculationService.kt
```diff
@@ -35,13 +47,24 @@ class WatchdogService @Inject constructor(
...
-            return
+        foregroundState.isInForeground
+            .distinctUntilChanged()
+            .filter { it }
+            .onEach { runRiskCalculations() }
+            .launchIn(appScope)
+    }
+
+    private suspend fun runRiskCalculations() = mutex.withLock {
+        if (Duration.between(latestRunTimeStamp, timeStamper.nowJavaUTC) < minTimeBetweenRuns) {
```
mtwalli, line 63: 1- this could problematic in time travel cases, maybe you consider calculating `abs`tract value of time difference ?
2- As you already mentioned above `Task` already taking care of aborting if redundant calculation is requested , is that still needed here ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ForegroundRiskCalculationService.kt
```diff
@@ -75,7 +92,8 @@ class WatchdogService @Inject constructor(
         .apply { acquire() }
 
     companion object {
-        private const val TAG = "WatchdogService"
+        private const val TAG = "ForegroundRiskCalc"
         private const val TEN_MINUTE_TIMEOUT_IN_MS = 10 * 60 * 1000L
+        private val minTimeBetweenRuns = Duration.ofMinutes(60)
```
mtwalli, line 97: could this cause an issue with ENF quota which is 6 times per day ? should we increase this duration

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/storage/TracingRepository.kt
chiljamgossow: It is not on purpose - that was part of the issue. It will fail with download but run the calculation to show the failed risk card.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ForegroundRiskCalculationService.kt
chiljamgossow: yes, it will abort but the UI would show the in progress state each time the user opens the home screen which is distraction

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/util/ForegroundRiskCalculationService.kt
chiljamgossow: no, that is checked in the task itself

mtwalli:
There is a noticeable delay in calculation after fresh install:

## PR
https://user-images.githubusercontent.com/25054729/187912414-eea27ce1-50ff-49f2-b275-4f8b77b5634f.mp4

## 2.27
https://user-images.githubusercontent.com/25054729/187912554-0aeed343-93ac-42a7-b0bf-79940dcc78bb.mp4

chiljamgossow:
I have added a onboarding check so it behaves the same as before on fresh install.

mtwalli (approved):

PR #5515, opened 2022-08-30
Blog link tr (EXPOSUREAPP-9807)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-9807)


mtwalli:
@schauersbergern please inform translation team about it ? if it is not the German file , it will be reverted back in the next translation delivery 😕

PR #5536, opened 2022-09-05
Sort certs for wallet assignment (EXPOSUREAPP-13503)

chiljamgossow:
to make sure we use the same order when assigning and storing the wallet

(I do not know if this is causing issues right now)


SamuraiKek (approved):
Looks like a pontetial solution. The oldWalletInfo becomes null when we update the wallet info for a person and it might be caused by a different orderd of certificates when we do the grouping.

PR #5522, opened 2022-08-31
Main fragment/ overflow menu text style (EXPOSUREAPP-12611)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12611)

Aligns overflow menu text spacing with design specs.

PR #5537, opened 2022-09-05
Fix inactive menu color (DEV)

SamuraiKek:
After changing the text style  for the menu items in #5522, @schauersbergern noticed that disabled menu items had the same color as enabled items. This is a fix that changes the disabled menu items back to a grey color.


schauersbergern (approved):
Looks right now

PR #5538, opened 2022-09-06
Red negative buttons (EXPOSUREAPP-13864)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13864)

## delete: Sie planen eine Veranstaltung? Meine QR Codes

Delete one QR-Code:

Delete all QR-Codes:

## delete: Schnelltestprofil
Currently no screenshot, but changed in code
## delete: Tagebuch - Personen und Orte

Delete one location

Delete all locations

Delete one person

Delete all persons

## deny: Kamerazugriff

This cannot be edited as this is a system dialog, but permission explanation dialog:

PR #5567, opened 2022-09-15
Reusable dialog fragment template (EXPOSUREAPP-12789)

SamuraiKek:
First PR in the dialog refactoring [story](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-12789), wrapping all dialogs in a dialog fragment to prevent potential memory leaks and save dialog state if the view gets destroyed when the app is put in background.

I've kept thinking of how I would like to use this template and I've come up with this `showDialog` extension function for fragments because it feels easy to use and there's not much boilerplate code that needs to be written but I'm open to change this up if you prefer another method.

Another approach  would be to create sepparate files where we store all dialogs for a specific fragment/entire feature like we currently do for example in the `RecycleBinDialogType`, but I'm not sold on the idea of writing boilerplate code for each dialog type. We would need an `XDialogHelper` file and an `XDialogHelperType` class for each fragment/feature.
Please let me know what your preference would be and if this approach gets more votes, I'll implement it this way.

I'm not passing the `resId` of the strings and instead I'm using the `getString()` function to pass the string directly to the template because there are instances where we are using string interpolation to display a dialog message so a `resId` woudln't suffice. 

I've started changing the dialogs from the QR Scanner fragment as an example of how this works. I haven't made any changes to the Error dialogs yet as they require more work and I'd do them at a later time once we get this template in.


mtwalli:
@SamuraiKek Thanks for the nice description. I have some ideas (maybe to explore)
- Instead of having once function with many params , could we use a builder pattern. Dialogs can vary from different cases
I think Builder pattern would be a good fit in here.
- I think we can use Kotlin DSL here? https://kotlinlang.org/docs/type-safe-builders.html 
- Maybe through this building pattern we could expose what `MateriaDailogAlertBuilder` offer anyway and just display it in `DialogFragment` and this would offer flexibility to custom the dialog from once screen to another without having to provide so many args. and would extend the pattern to offer new functionally such Delete Dialog red button 

Let me what are your thoughts on that ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
```diff
@@ -0,0 +1,82 @@
...
+        Dismissed
+    }
+
+    @Parcelize
+    data class Config(
+        val titleRes: String,
+        val messageRes: String,
+        val positiveButtonRes: String,
+        val negativeButtonRes: String?,
+        val neutralButtonRes: String?,
```
chiljamgossow, line 67: You call it ...Res, but it seems to be a string, I would expect an int here.
Will it work with changing languages?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
```diff
@@ -0,0 +1,82 @@
...
+        val isDeleteDialog: Boolean = false,
```
chiljamgossow, line 68: This seems to be used to set the color of the positive button to red. I would prefer a more obvious and direct way, eg. color of positive button, so I can use orange or color a dialog button although it is not for deletion.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplateHelper.kt
```diff
@@ -0,0 +1,59 @@
...
+
+import androidx.fragment.app.Fragment
+
+typealias DialogAction = () -> Unit
+
+@Suppress("LongParameterList")
+fun Fragment.showDialog(
+    title: String,
+    message: String,
+    positiveButton: String,
```
chiljamgossow, line 11: You could use a data class here Button(text, action, color?) to reduce the number of parameters

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: I've removed them completely. You can now call the `MaterialAlertDialogBuilder` functions directly.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: I'm not sure if I'd change this one. I agree that it'd be nice to use it like you described if we would truly have different colors for dialog buttons, but since it's only red, I'd keep it like this. We didn't have any differently colored dialog buttons up until recently #5538 and other than the red negative buttons, I doubt we'll have other different colors in the future. I'd prefer to just pass a `true` when creating a dialog button than having to pass the res id of the red color with the hope we'll need to use different colors in the future which is highly unlikely to change.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplateHelper.kt
SamuraiKek: I've tried @mtwalli's approach of creating a DSL. We already had the builder from the `MaterialAlertDialogBuilder` and I've just exposed it and we can now use it's own functions for setting up the dialog. The number of params has now reduced to an acceptable level where `detekt` isn't complaining anymore.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
```diff
@@ -0,0 +1,41 @@
...
+        if (isDeleteDialog) {
+            (dialog as AlertDialog).getButton(DialogInterface.BUTTON_POSITIVE)
+                ?.setTextColor(requireContext().getColorCompat(R.color.colorTextDeleteButtonDialog))
+        }
+    }
+
+    override fun onDismiss(dialog: DialogInterface) {
+        dismissAction()
+        super.onDismiss(dialog)
+    }
```
mtwalli, line 36: I think  whole `onDismiss` is not needed and `dismissAction()` can be removed and `MaterialAlertDialogBuilder.setOnDismissListener` can be used as other functions

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplateHelper.kt
```diff
@@ -0,0 +1,17 @@
+package de.rki.coronawarnapp.ui.dialog
+
+import androidx.fragment.app.Fragment
+import com.google.android.material.dialog.MaterialAlertDialogBuilder
+
+@SuppressWarnings("LongParameterList")
```
mtwalli, line 6: can be removed ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
```diff
@@ -0,0 +1,41 @@
...
+    private val cancelable: Boolean = true,
+    private val isDeleteDialog: Boolean = false,
+    private val dismissAction: () -> Unit = { },
+    private val materialDialog: MaterialAlertDialogBuilder? = null,
+    private val config: MaterialAlertDialogBuilder.() -> Unit
+) : DialogFragment() {
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        isCancelable = cancelable
+        return materialDialog?.create() ?: MaterialAlertDialogBuilder(requireContext()).apply(config).create()
```
mtwalli, line 22: `config` need to be applied to `materialDialog?.create()`  too? just incase some other extension need to be added it can have also effect on the provided builder

mtwalli:
Nice work!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
```diff
@@ -0,0 +1,41 @@
...
+import de.rki.coronawarnapp.R
+import de.rki.coronawarnapp.util.ContextExtensions.getColorCompat
+
+class DialogFragmentTemplate(
+    private val cancelable: Boolean = true,
+    private val isDeleteDialog: Boolean = false,
+    private val dismissAction: () -> Unit = { },
+    private val materialDialog: MaterialAlertDialogBuilder? = null,
+    private val config: MaterialAlertDialogBuilder.() -> Unit
+) : DialogFragment() {
```
mtwalli, line 18: I'm not sure about passing arguments to fragments from the constructor, I think the best way is to use a Bundle 
https://developer.android.com/reference/android/app/DialogFragment

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplateHelper.kt
SamuraiKek: Yep, forgot about it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: I will try it again to make sure, but it didn't work for me. From reading up, it seems that when using dialogs inside a `DialogFragment`, this action is being handled by the `DialogFragment` itself 🤷 https://stackoverflow.com/a/18269965/13296736

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: Wouldn't applying the config here override what's already configured in the `materialDialog`?
I was thinking we are either passing a fully configured `MaterialAlertDialogBuilder`, or we're passing the config for a `MaterialAlertDialogBuilder` and we apply it here.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: Tried it again, it dosen't work.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: Neverming, I see what you mean now. I've applied the config to the `materialDIalog` as well. Very nice :D

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
SamuraiKek: Ok,  I've removed the constructor params and created a parcelized data class to pass them as a bundle like we do with the `RecycleBinDialogFragment`

chiljamgossow (dismissed):
Nicely done!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/DialogFragmentTemplate.kt
mtwalli: I did try it and it worked 😄 , it might be because you were not applying the `config` when `dialog` is passed?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt
```diff
@@ -350,70 +342,69 @@ class QrCodeScannerFragment : Fragment(R.layout.fragment_qrcode_scanner), AutoIn
...
-        MaterialAlertDialogBuilder(requireContext())
-            .setTitle(R.string.qr_code_error_max_person_max_title)
-            .setCancelable(false)
-            .setMessage(getString(R.string.qr_code_error_max_person_max_body, max))
-            .setOnDismissListener { popBackStack() }
-            .setPositiveButton(R.string.qr_code_error_max_person_covpasscheck_button) { _, _ ->
+    private fun showMaxPersonExceedsMaxResult(max: Int) =
+        displayDialog(cancelable = false) {
+            setTitle(R.string.qr_code_error_max_person_max_title)
+            setMessage(R.string.qr_code_error_max_person_covpasscheck_button)
```
mtwalli, line 381: message here is wrong

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt
```diff
@@ -350,70 +342,69 @@ class QrCodeScannerFragment : Fragment(R.layout.fragment_qrcode_scanner), AutoIn
...
+            setNegativeButton(R.string.qr_code_error_max_person_faq_button) { _, _ ->
                 openUrl(R.string.qr_code_error_max_person_faq_link)
             }
-            .setNeutralButton(android.R.string.ok) { _, _ -> }
-            .show()
-    }
+            setNeutralButton(android.R.string.ok) { _, _ ->
+                findNavController().navigate(deeplink, navOptions)
+            }
+        }
```
mtwalli, line 376: `setOnDismissListener` is missing after refactoring it

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt
SamuraiKek: Ok, fixed it.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/qrcode/ui/QrCodeScannerFragment.kt
SamuraiKek: Fixed it.

mtwalli (approved):

PR #5571, opened 2022-09-16
Onboarding exposure notification popup (EXPOSUREAPP-13903)

schauersbergern:
While onboarding, in OnboardingTracingFragment the tracing was disabled onResume. This would cause race conditions if you quickly navigate to the next fragment and move back, so the dialog was not always shown when tapping "enable" button.
I checked with Martin Hermes on iOS behaviour, which revealed that iOS only shows the dialog if tracing is not activated and doesn't show the dialog again when navigating back and hitting the enable button again, so I also disabled the automatic tracing disabling.

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13903


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingTracingFragment.kt
```diff
@@ -70,12 +70,6 @@ class OnboardingTracingFragment : Fragment(R.layout.fragment_onboarding_tracing)
         }
     }
 
-    override fun onResume() {
-        super.onResume()
-        binding.onboardingTracingContainer.sendAccessibilityEvent(AccessibilityEvent.TYPE_ANNOUNCEMENT)
```
mtwalli, line 75: is this container also removed ? why you don't send AccessibilityEvent now?

schauersbergern: No, I re-added the AccessibilityEvent

mtwalli (approved):
Works fine

PR #5589, opened 2022-09-26
Process rebirth on environment change (EXPOSUREAPP-13973)

mtwalli:
Background, switching from environment to another led sometimes to issues such as:
1- AppConfig is not updated
2- DSC List 
3- CCL 
and even when the respective data is cleaned up the URLs are not reflected unless app is killed and opened again.
- Now after switching env the app will restart again automatically.

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13973


chiljamgossow (approved):
Worked nicely for me. Maybe a button in the dialog to confirm that you really want to restart now in case it was accidental

mtwalli:
> Worked nicely for me. Maybe a button in the dialog to confirm that you really want to restart now in case it was accidental
I will consider that, let us see wait for testers feedback

PR #5604, opened 2022-09-29
Replace dialogs in MainActivity, HomeFragment (EXPOUSREAPP-13994)

SamuraiKek:
Continuation of the dialog refactoring ticket.


CV113 (dismissed):
UA approved

Corona-Warn-App/src/main/res/values/strings.xml
```diff
@@ -301,8 +302,9 @@
     <string name="risk_details_headline_period_logged">"Period Logged"</string>
     <!-- XHED: risk details - infection period logged headling, below behaviors -->
     <string name="risk_details_subtitle_period_logged">"This period is included in the calculation."</string>
+    <!-- XTXT: Tracing explanation dialog message -->
+    <string name="tracing_explanation_dialog_message">"Die Berechnung des Infektionsrisikos kann nur für die Zeiträume erfolgen, an denen die Risiko-Ermittlung aktiv war. Die Risiko-Ermittlung sollte daher dauerhaft aktiv sein. Für Ihre Risiko-Ermittlung wird der Zeitraum der letzten %d Tage betrachtet.\nÄltere Tage werden automatisch gelöscht, da sie aus Sicht des Infektionsschutzes nicht mehr relevant sind.\nWeitere Informationen finden Sie in den FAQ."</string>
```
schauersbergern, line 306: Why german? Shouldn't this stay empty for translation?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
```diff
@@ -167,7 +166,9 @@ class MainActivity : AppCompatActivity(), HasAndroidInjector {
                 is MainActivityEvent.GoToCheckInsFragment -> navController.navigate(
                     CheckInsFragment.createDeepLink(event.uriString)
                 )
-                is MainActivityEvent.Error -> event.error.toErrorDialogBuilder(this).show()
+                is MainActivityEvent.Error -> displayDialog {
+                    event.error.toErrorDialogBuilder(context)
```
mtwalli, line 170: this should be passed as `dialog=event.error.toErrorDialogBuilder(context)` otherwise it does not have an effect , right?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -88,37 +77,43 @@ class HomeFragment : Fragment(R.layout.home_fragment_layout), AutoInject {
...
+            }
         }
         viewModel.coronaTestErrors.observe2(this) { tests ->
             tests.forEach { test ->
-                test.lastError?.toErrorDialogBuilder(requireContext())?.apply {
-                    val testName = when (test.type) {
-                        BaseCoronaTest.Type.PCR -> R.string.ag_homescreen_card_pcr_title
-                        BaseCoronaTest.Type.RAPID_ANTIGEN -> R.string.ag_homescreen_card_rapidtest_title
+                displayDialog {
+                    test.lastError?.toErrorDialogBuilder(requireContext())?.apply {
```
mtwalli, line 109: same here

Corona-Warn-App/src/main/res/values/strings.xml
SamuraiKek: I've created a new string, and it needs to be present in the default strings file as well. The actual text will get overwritten by the translation anyway.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/MainActivity.kt
SamuraiKek: Ah yes, you're right. I'll update it.

SamuraiKek: Not sure about the refactoring. We'd go back to a constructor with many params. I like the lambda way of configuring the dialog. We just need to make sure to not put anything else in here, same as with any lambda.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -88,37 +77,43 @@ class HomeFragment : Fragment(R.layout.home_fragment_layout), AutoInject {
...
+                Timber.e(e, "Listener has been triggered after onDestroyView()")
             }
-        )
+        }
 
         viewModel.showPopUps()
         viewModel.events.observe2(this) { event -> navigate(event) }
         viewModel.homeItems.observe2(this) { homeAdapter.update(it) }
-        viewModel.errorEvent.observe2(this) { it.toErrorDialogBuilder(requireContext()).show() }
+        viewModel.errorEvent.observe2(this) { displayDialog { it.toErrorDialogBuilder(context) } }
```
mtwalli, line 95: This is still problematic, I'm not sure if the refactoring is not needed , you yourself the designer of the API keeps forgetting it, you can imagine others who don't know how this works actually.

SamuraiKek: I didn't keep forgetting it. It's not like I've done it again, this change was made with those other two in the beginning. I still wouldn't refactor it as this argument could be said about every lamda function in this app.

mtwalli (approved):

schauersbergern (approved):

PR #5617, opened 2022-10-05
`Feature` Pandemic  Radar (EXPOSUREAPP-14073)

mtwalli:
Feature

## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14073


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentEvents.kt
```diff
@@ -20,6 +20,7 @@ sealed class HomeFragmentEvents {
     ) : HomeFragmentEvents()
 
     object GoToStatisticsExplanation : HomeFragmentEvents()
+    data class OpenExternalLink(val url: String) : HomeFragmentEvents()
```
SamuraiKek, line 23: I'd name this `OpenLinkCardUrl` so we know it's the link card one. `OpenExternalLink` feels too generic.

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragmentViewModel.kt
```diff
@@ -191,13 +195,28 @@ class HomeFragmentViewModel @AssistedInject constructor(
...
+        val ordering = Ordering.explicit(cardIdSequence.toList())
+        val stats = localStatsData.items.plus(statsData.items)
+            .filterIsInstance<StatsSequenceItem>()
+            .filter { it.cardType.id in cardIdSequence }
+            .sortedWith { a, b ->
+                ordering.compare(a.cardType.id, b.cardType.id)
+            }
+
+        val addStatsItems = when {
+            !cardIdSequence.contains(StatsType.LOCAL_INCIDENCE.id) -> emptySet<StatsItem>()
```
SamuraiKek, line 209: I'm not sure if this is alright. I agree that it's probably better to not show the statistics at all if we don't get them from the server, but the current version will display a card that tells the user that statistics are not available due to no internet connection, or something along those lines. I think this should be clarified with UX.

PR #5632, opened 2022-10-10
Get rid of `doNavigate` (EXPOSUREAPP-14017)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14017


SamuraiKek (approved):
Very nice!

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/UIExtensions.kt
```diff
@@ -15,17 +14,6 @@ import de.rki.coronawarnapp.databinding.ActivityMainBinding
...
 import java.lang.ref.WeakReference
 
-/**
- * Extends NavController to prevent navigation error when the user clicks on two buttons at almost
- * the exact time.
- *
- * @see [NavController]
- */
-fun NavController.doNavigate(direction: NavDirections) {
-    currentDestination?.getAction(direction.actionId)
```
schauersbergern, line 25: After removing doNavigate, how do we check now if the action isn't null?
findNavController().navigate will crash, if the action is null

schauersbergern (approved):

PR #5633, opened 2022-10-10
Fix certificate deletion flow (EXPOSUREAPP-13987)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13987)

What caused it: When the fragment is created, we are getting the certificate that needs to be displayed, if we can't get it we do a `popBackStack()`. Whenever we deleted the certificate and the dialog closed, we would first try and get the certificate again but this time it would be null because we deleted it resulting in`popBackStack()` being called. After that, we called `popBackStack()` again because the delete button of the dialog was pressed resulting in double `popBackStack()` and getting to the `PersonOverview` screen instead of `PersonDetails`.

P.S. Hi @harambasicluka, good to see you dropping in from time to time 🚀


schauersbergern (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -80,7 +81,7 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
 
         viewModel.errors.observe(viewLifecycleOwner) { onError(it) }
         viewModel.events.observe(viewLifecycleOwner) { onNavEvent(it) }
-        viewModel.recoveryCertificate.observe(viewLifecycleOwner) {
+        viewModel.recoveryCertificate.observeOnce(viewLifecycleOwner) {
```
mtwalli, line 84: observe once would prevent any changes that happens in the certificate such as `State` from showing up while you are on the screen for long time , or App is put to background and then foreground again. It looks like if you remove `popBackStack` from `ReturnToPersonDetailsAfterRecycling` handling  is fine? if the deleted certificate is not there anymore ,the screen will navigate back anyway. any side effects  you see?

SamuraiKek: I've also thought about this. I chose the `observeOnce` route becuase I'm not sure leaving `observe` will result in a null vaccination certificate every time after deletion. There is a chance that the flow woudln't update right away and by removing `popBackStack` the user would have to manually go back. Also the deleted certificate would still be visible which could be confusing to the user who just deleted the certificate.

Not sure what would be more affecting, not getting null every time after deletion and not popping backstack, or not receiveing a UI update in the slight chance the state of the certificate changes while you're on the screen.

mtwalli: Any side effect could really happen on bigger scale , when millions of users are using the app. I fine with any solution that does not have side effects.

SamuraiKek: Ok, I brought back the `observe` and just included an extra fail safe. We should be popping the backstack when the certificate is null only when opening this screen from an old notification. That's why this check was included in the first place. So in theory, any future null value of the certificate shouldn't result in a `popBackStack`. I think this has us covered. It fixes the bug and it also lets the UI update when the state of the cert changes.

mtwalli: What about having stateFlow for events in the viewModel.  if event `Back` is triggered twice it won't emit to the UI because it is the same.

We would need to  move the logic of these navigation logic from the Fragment and control it from the viewModel

SamuraiKek: Ok, that's more elegant 😄 I'll give it a shot

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/covidcertificate/recovery/ui/details/RecoveryCertificateDetailsFragment.kt
```diff
@@ -84,7 +84,7 @@ class RecoveryCertificateDetailsFragment : Fragment(R.layout.fragment_recovery_c
                 true -> onCertificateReady(it)
                 false -> {
                     Timber.tag(TAG).d("Certificate is null. Closing %s", TAG)
-                    popBackStack()
+                    viewModel.onClose()
```
schauersbergern, line 87: As the event triggered here is named `Back` we should use the naming `goBack() ` like in Tests also in Recovery and Vaccination Fragments.
Besides of that its working fine

schauersbergern (approved):

PR #5647, opened 2022-10-14
Fix cancel button (EXPOSUREAPP-14137)

mtwalli:
Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14137


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/presencetracing/attendee/checkins/CheckInsFragment.kt
```diff
@@ -164,19 +164,20 @@ class CheckInsFragment : Fragment(R.layout.trace_location_attendee_checkins_frag
     }
 
     private fun showRemovalConfirmation(checkIn: CheckIn?, position: Int?) =
-        DialogHelper.showDialog(
```
mtwalli, line 167: it is not reported in the ticket, but it had the same issue, Also used the new dialog way to avoid re-working it

schauersbergern (approved):

PR #5660, opened 2022-10-21
Fix crash  / remove the faulty dialog (EXPOSUREAPP-14192)

mtwalli:
Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14192


SamuraiKek (approved):
image

schauersbergern (approved):
App doesn't crash anymore when returning from background with active dialogs

PR #5662, opened 2022-10-26
Fault Tolerance - BNR triggered again because of TC (EXPOSUREAPP-14218)

mtwalli:
- Check the ticket and follow instructions on how to re-produce the issue 
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14218

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -179,7 +179,7 @@ jobs:
           path: Corona-Warn-App/build/test-results
 
       - name: Publish Test Results - Device Release
-        uses: dorny/test-reporter@v1.5.0
+        uses: dorny/test-reporter@v1.6.0
         if: always()
         with:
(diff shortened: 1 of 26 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -179,7 +179,7 @@ jobs:
           path: Corona-Warn-App/build/test-results
 
       - name: Publish Test Results - Device Release
-        uses: dorny/test-reporter@v1.5.0
+        uses: dorny/test-reporter@v1.6.0
```
mtwalli, line 182: Test report was broken 😞

SamuraiKek (approved):
Tested on device and it's working nicely. I think this could also hopefully fix the notification spam that some people are still reporting with boosters.

schauersbergern (approved):

PR #5656, opened 2022-10-20
Update remaining dialogs (EXPOSUREAPP-14000)

SamuraiKek:
This task updates the remaining dialogs in the app to the new dialog fragment approach.

PR #5663, opened 2022-10-26
Alternative to the current dialog builder using a DSL builder (DEV)

SamuraiKek:
alwayshasbeen

Unfortunately extending the `MaterialAlertDialogBuilder` does not fully work because the callbacks for the dialog buttons that are passed to the `DialogFragment` as lambdas in a bundle will cause the app to crash when the app is put in background because lambdas cannot be serialized propperly through a bundle [¯\_(ツ)_/¯](https://stackoverflow.com/a/48870902)

This "new new builder" DSL passes only the texts and the options of a dialog through the bundle and the actions for the dialogs are communicated sepparetely. 

I've included a small example of the usage in the HomeFragment with a comparison against the "old new builder". You can modify and test it to see that it's not crashing anymore. Also, let me know if you would like to use this DSL in a different way. 

If I get the green light, I will clean the code and merge it into the main dialog refactoring PR #5656 and change the dialogs to use this builder.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialog.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialog.kt
@@ -10,49 +10,29 @@ annotation class CwaDialogDsl
 
 typealias DialogAction = () -> Unit
 
-data class CwaDialog(
-    val texts: CwaDialogTexts,
-    val actions: CwaDialogActions,
-    val options: CwaDialogOptions
-)
-
-@CwaDialogDsl
-class CwaDialogBuilder {
-    private var config = CwaDialogTexts()
-    private var actions = CwaDialogActions()
-    private var options = CwaDialogOptions()
-
-    fun configureTexts(lambda: CwaDialogTextsBuilder.() -> Unit) {
-        config = CwaDialogTextsBuilder().apply(lambda).build()
-    }
-
-    fun configureActions(lambda: CwaDialogActionsBuilder.() -> Unit) {
-        actions = CwaDialogActionsBuilder().apply(lambda).build()
-    }
-
-    fun configureOptions(lambda: CwaDialogOptionsBuilder.() -> Unit) {
-        options = CwaDialogOptionsBuilder().apply(lambda).build()
-    }
-
-    fun build() = CwaDialog(config, actions, options)
-}
-
 @Parcelize
 data class CwaDialogTexts(
     val title: @RawValue IntOrString? = null,
     val message: @RawValue IntOrString? = null,
-    val positiveButton: @RawValue IntOrString? = null,
-    val negativeButton: @RawValue IntOrString? = null,
-    val neutralButton: @RawValue IntOrString? = null,
+    val positiveButtonText: @RawValue IntOrString? = null,
+    val negativeButtonText: @RawValue IntOrString? = null,
+    val neutralButtonText: @RawValue IntOrString? = null,
 ) : Parcelable
 
 @CwaDialogDsl
-class CwaDialogTextsBuilder {
+class CwaDialogConfigBuilder {
     private var title = IntOrString()
     private var message = IntOrString()
-    private var positiveButton = IntOrString()
-    private var negativeButton = IntOrString()
-    private var neutralButton = IntOrString()
+    private var positiveButtonText = IntOrString()
+    private var positiveButtonAction: DialogAction = { }
+    private var negativeButtonText = IntOrString()
+    private var negativeButtonAction: DialogAction = { }
+    private var neutralButtonText = IntOrString()
+    private var neutralButtonAction: DialogAction = { }
+    private var dismissAction: DialogAction = { }
+    private var isCancelable: Boolean = true
+    private var isDeleteDialog: Boolean = false
+    private var customView: Int? = null
 
     @OptIn(ExperimentalTypeInference::class)
     @OverloadResolutionByLambdaReturnType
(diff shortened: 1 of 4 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialog.kt
```diff
@@ -0,0 +1,201 @@
...
+
+@CwaDialogDsl
+class CwaDialogTextsBuilder {
+    private var title = IntOrString()
+    private var message = IntOrString()
+    private var positiveButton = IntOrString()
+    private var negativeButton = IntOrString()
+    private var neutralButton = IntOrString()
+
+    @OptIn(ExperimentalTypeInference::class)
```
SamuraiKek, line 57: Unfotunately you can't overload functions with lambda params of different types in Kotlin by default. I had to opt to use this annotation. https://youtrack.jetbrains.com/issue/KT-11265/Overload-resolution-by-lambda-return-type#focus=Comments-27-4303340.0-0

I understand if you don't want to include experimental features in. I can make the DSL usable without it, but it won't look as nice and tidy :/

SamuraiKek: Removed it.

schauersbergern (approved):
I like this new approach

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
```diff
@@ -118,6 +119,16 @@ class HomeFragment : Fragment(R.layout.home_fragment_layout), AutoInject {
             }
         }
 
+        // Usage example of the new new builder
+        createDialog {
+            title("New implementation")
+            message("Putting the app in background won't cause a crash with the dialog open")
+            positiveButton("Ok") { dummyDialog1() }
+            negativeButton(R.id.button_done)
```
schauersbergern, line 127: You mean R.string... though?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialog.kt
```diff
@@ -0,0 +1,130 @@
...
+
+    fun setCancelable(cancelable: Boolean) {
+        isCancelable = cancelable
+    }
+
+    fun setDeleteDialog(isDelete: Boolean) {
+        isDeleteDialog = isDelete
+    }
+
+    fun customView(view: Int) {
```
mtwalli, line 94: `@LayoutRes`

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialog.kt
```diff
@@ -0,0 +1,130 @@
...
+@Parcelize
+open class IntOrString : Parcelable {
+    @Parcelize
+    data class StringRes(
+        val stringResource: String
+    ) : Parcelable, IntOrString()
+
+    @Parcelize
+    data class IntRes(
+        val intResource: Int
```
mtwalli, line 128: @StringRes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
SamuraiKek: Ah yes, I just needed an id of something real quick :D

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialogHelper.kt
```diff
@@ -0,0 +1,53 @@
...
+ * Helper function that creates a CwaDialog and displays it.
+ * Usage:
+ *
+ * ```kotlin
+ * createDialog {
+ *     title { "Title example" }
+ *     message { "Message example" }
+ *     positiveButton("Positive button text") { action { foo1() }
+ *     negativeButton("Negative button text") { action { foo2() }
+ *     neutralButton("Neutral button text") { action { foo3() }
```
mtwalli, line 15: I think the doc here needs an update ?

SamuraiKek: Yes id tdoes

mtwalli: `message` and `title` don't require lambdas too

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialog.kt
mtwalli: this is still missing

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/main/home/HomeFragment.kt
mtwalli: After adding the annotation this display an error to indicate it is not a string resource 👍🏼

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/dialog/CwaDialogHelper.kt
```diff
@@ -0,0 +1,53 @@
...
+ * createDialog {
+ *     title { "Title example" }
+ *     message { "Message example" }
+ *     positiveButton("Positive button text") { foo1() }
+ *     negativeButton("Negative button text") { foo2() }
+ *     neutralButton("Neutral button text") { foo3() }
+ *     dismissAction { foo4() }
+ *     setCancelable(false)
+ *     setDeleteDialog(true)
+ *     setCustomView(R.id.dialog_view)
```
mtwalli, line 19: `R.layout.dialog_view`?

mtwalli (commented):
LGTM, tested device rotation , putting app to background, I guess current places where we have a builder would still show as such , right ?

Now we can't really pass a builder as before

mtwalli (approved):

mtwalli:
Please test is also on older android versions while refactoring ?

SamuraiKek:
> LGTM, tested device rotation , putting app to background, I guess current places where we have a builder would still show as such , right ?
> 
> Now we can't really pass a builder as before

Yes, but I have a fix ready for my other PR that removes the need to rely on other builders.

PR #5665, opened 2022-10-28
Fix: Installation time (EXPOSUREAPP-14249)

mtwalli:
-  Days since installation note  in the risk card should not show up in the app if it is installed more than 14 days
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14249

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/installTime/InstallTimeProvider.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/installTime/InstallTimeProvider.kt
@@ -1,30 +1,25 @@
 package de.rki.coronawarnapp.installTime
 
-import android.content.Context
-import de.rki.coronawarnapp.util.di.AppContext
+import de.rki.coronawarnapp.util.TimeStamper
+import de.rki.coronawarnapp.util.di.AppInstallTime
 import de.rki.coronawarnapp.util.toLocalDateUserTz
 import java.time.Instant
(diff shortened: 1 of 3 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/installTime/InstallTimeProvider.kt
```diff
@@ -1,30 +1,25 @@
...
-        .toLocalDateUserTz()
+    private val dayOfInstallation = installTime.toLocalDateUserTz()
 
     val today: LocalDate
-        get() = Instant.now().toLocalDateUserTz()
+        get() = timeStamper.nowUTC.toLocalDateUserTz()
 
     val daysSinceInstallation: Int
-        get() = Period.between(dayOfInstallation, today).days
+        get() = ChronoUnit.DAYS.between(dayOfInstallation, today).toInt()
```
mtwalli, line 24: Core change `Period.between(dayOfInstallation, today).days` gets the days part only from the period

SamuraiKek (approved):

schauersbergern (approved):

Ein-Tim:
Thanks, I have a twitter user reporting this bug. Will this be fixed in a hot fix?

mtwalli:
> Thanks, I have a twitter user reporting this bug. Will this be fixed in a hot fix?

Yes, in progress

PR #5681, opened 2022-11-02
Self Report Submission config param (EXPOSUREAPP-14187)

mtwalli:
Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14187

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SelfReportSubmissionConfigMapper.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SelfReportSubmissionConfigMapper.kt
@@ -1,0 +1,68 @@
+package de.rki.coronawarnapp.appconfig.mapping
+
+import dagger.Reusable
+import de.rki.coronawarnapp.appconfig.SafetyNetRequirementsContainer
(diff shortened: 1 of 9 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/appconfig/mapping/SelfReportSubmissionConfigMapper.kt
```diff
@@ -0,0 +1,68 @@
...
+                SelfReportSubmissionConfigContainer.DEFAULT
+            }
+            else -> rawConfig.selfReportParameters.map()
+        }
+    } catch (e: Exception) {
+        Timber.d(e, "SelfReportSubmissionConfigMapper failed -> returning default")
+        SelfReportSubmissionConfigContainer.DEFAULT
+    }
+
+    fun PpddSrsParameters.PPDDSelfReportSubmissionParametersAndroid.map(): SelfReportSubmissionConfigContainer {
```
SamuraiKek, line 30: `Ppdd` vs `PPDD`. Can you choose either lower case or upper case to have consistencty?

mtwalli: it is autogenerated from Proto-Buff

PR #5683, opened 2022-11-03
Update Register Test Card in Home Fragment for SRS (EXPOSUREAPP-14173)

SamuraiKek:
Screenshot 2022-11-03 at 09 37 26


Corona-Warn-App/src/main/res/drawable-night/register_test_card_background.xml
```diff
@@ -0,0 +1,24 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:aapt="http://schemas.android.com/aapt"
+    android:width="336dp"
+    android:height="243dp"
+    android:viewportWidth="336"
+    android:viewportHeight="243">
+    <path android:pathData="M0,4C0,1.79 1.79,0 4,0H332C334.21,0 336,1.79 336,4V243H0V4Z">
+        <aapt:attr name="android:fillColor">
```
mtwalli, line 8: I think this will cause a crash on Android 6, `fillColor` is supported from API 24. could you please double check it?

Corona-Warn-App/src/main/res/drawable/register_test_card_background.xml
```diff
@@ -0,0 +1,27 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:aapt="http://schemas.android.com/aapt"
+    android:width="336dp"
+    android:height="243dp"
+    android:viewportWidth="336"
+    android:viewportHeight="243">
+    <path android:pathData="M0,4C0,1.79 1.79,0 4,0H332C334.21,0 336,1.79 336,4V243H0V4Z">
+        <aapt:attr name="android:fillColor">
```
mtwalli, line 8: same here

Corona-Warn-App/src/main/res/values/srs_submission_strings.xml
```diff
@@ -1,4 +1,14 @@
 <?xml version="1.0" encoding="utf-8"?>
 <resources xmlns:tools="http://schemas.android.com/tools" tools:ignore="MissingTranslation">
 
+    <!-- Home screen register test card -->
```
mtwalli, line 4: are the old strings still used? can we delete them?

mtwalli (changes requested):

Corona-Warn-App/src/main/res/values/srs_submission_strings.xml
SamuraiKek: Yes

Corona-Warn-App/src/main/res/drawable-night/register_test_card_background.xml
SamuraiKek: So how should I import  them then? I've used the same method up until now.

SamuraiKek: I remember you used a tool to reduce the size of these drawables. Could that actually remove the fillColor completely?

CV113 (approved):
UA approved

Corona-Warn-App/src/main/res/drawable-night/register_test_card_background.xml
mtwalli: You need to ask Anja to export it differently, the tool won't fix it

SamuraiKek: I did a quick check to see if this is used anywhere else and it's found in many drawables. I guess if it caused a crash it would have been reported by now.

mtwalli: Ok, please check it on Android 6 if it works ?

SamuraiKek: Ok, it doesn't work when using the resource to set the background but it works fine for the icon... I'll ask Anja

SamuraiKek: UX was unable to produce an asset without this property but I pushed a workaround where instead of setting the drawable as the background of the `ConstraintLayout` I simply created an `ImageView` and set it to span accross the whole top part of the card with `srcCompat` and that seems to prevent the app from crashing on 6.0

mtwalli: works

mtwalli (approved):

PR #5689, opened 2022-11-08
Update test dispatcher screen (EXPOSUREAPP-14174)

SamuraiKek:
[Figma](https://www.figma.com/file/KuDEcMSREg3VhZRhC3FXTY/COVID19_Master_2.25?node-id=58210%3A51790)

 - Add the two new cards
 - Update light theme background (grey)
 - Hide TAN related Cards
 - Check the screenshots 
 - use id for the text element to make them testable for automation test

Created a custom view for the dispatcher cards as per @chiljamgossow 's request in this ticket https://github.com/corona-warn-app/cwa-app-android/pull/5045#discussion_r843565326

PR #5692, opened 2022-11-09
Remove tan tiles from dispatcher screen (DEV)

SamuraiKek:
Got the go to remove the TAN tiles. 
I haven't removed them completely, just made them gone for now.


Corona-Warn-App/src/main/res/layout/fragment_submission_dispatcher.xml
```diff
@@ -141,7 +141,7 @@
                     android:layout_height="wrap_content"
                     android:layout_marginTop="@dimen/spacing_medium"
                     android:text="@string/submission_dispatcher_already_positive_subheadline"
-                    android:visibility="visible"
```
schauersbergern, line 144: Why was this visibility attribute present? Is the visibility toggled from code? If not, hiding here is good I think

SamuraiKek: When I added the changes to the dispatcher screen #5689, I already made these tiles gone, but then @mtwalli asked me to add them back because we didn't have the decision to remove them final at the time. So I just used the visiblity attribute here to set them to visible for the time being.

schauersbergern (approved):

PR #5693, opened 2022-11-09
Submissions in  journal (EXPOSUREAPP-14181)

mtwalli:
- SRS submission should add entry in Journal 
- Regular Submission should add also an entry in Journal (multiple submissions on same day should be displayed as one)

Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14181


InkenTownsend (approved):
UA Approved

SamuraiKek (approved):
LGTM. Also tested on device and submission shows up in the journal as expected.

PR #5677, opened 2022-11-02
`Feature` SRS (EXPOSUREAPP-13454)

mtwalli:
👷🏼‍♂️ 🚧 
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-13454

PR #5700, opened 2022-11-11
Update thank you screen (EXPOSUREAPP-14180)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14180)

Screenshot_20221111-142858
Screenshot_20221111-142903
Screenshot_20221111-142906


Ein-Tim:
@UA-Team I don't think the note "Sie werden eventuell von Ihrem Gesundheitsamt [...]" fits if the user reported a positive self test, as the health authorities don't get notified if someone has a positive self test. This could actually lead to users thinking that the CWA informed the health authority about their positive test.

mtwalli:
@InkenTownsend  Please have a look?

InkenTownsend:
Thank you for the feedback @Ein-Tim. Let me take this back to the team and I will get back to you.

mtwalli (approved):

Ein-Tim:
@mtwalli You merged this PR, although @InkenTownsend wrote that she will get feedback from the team and let us know if the text is fine to stay here. Will @InkenTownsend provide the feedback in the main PR https://github.com/corona-warn-app/cwa-app-android/pull/5677?

InkenTownsend:
Hello @Ein-Tim, sorry that I didn't reply on Friday anymore. I checked and as the screen is not specific to the one use case but is shared, we have to leave it as it is. The wording with "eventuell" covers that the contact is only a maybe. I am sorry about that. Thank you again for your very appreciated input.

Ein-Tim:
@InkenTownsend 

Alright, thanks for checking!

PR #5708, opened 2022-11-14
Add otp censor (EXPOSUREAPP-14190)

schauersbergern:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14190)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/bugreporting/censors/submission/OtpCensor.kt
```diff
@@ -0,0 +1,29 @@
...
+
+@Reusable
+class OtpCensor @Inject constructor(
+    private val srsSubmissionSettings: SrsSubmissionSettings,
+) : BugCensor {
+
+    override suspend fun checkLog(message: String): BugCensor.CensorContainer? {
+        var container = BugCensor.CensorContainer(message)
+
+        val otp = srsSubmissionSettings.otp.last()
```
mtwalli, line 17: this would censor only OTP once it is saved in the repo, but before than I won't be censored

schauersbergern: censor otp on fetch and renewal

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/srs/core/repository/SrsSubmissionRepository.kt
```diff
@@ -115,7 +116,9 @@ class SrsSubmissionRepository @Inject constructor(
             srsOtp
         } else {
             Timber.tag(TAG).d("SRS otp is not valid -> new otp generated")
-            SrsOtp()
+            SrsOtp().also {
+                OtpCensor.otp = it
+            }
```
mtwalli, line 121: you already doing this above , it should cover this case too , right?

schauersbergern: yes, removed

mtwalli (approved):

PR #5746, opened 2022-12-06
Pandemic radar url language handling (EXPOSUREAPP-14375)

SamuraiKek:
DE & TR will have their own specific urls while all the other languages will open the EN url.


schauersbergern (approved):

MikeMcC399:
https://github.com/corona-warn-app/cwa-app-android/blob/000078652baf56d677b026ea8b19730d4a5947fc/Corona-Warn-App/src/main/res/values/strings.xml#L1524

lists also Russian as language, which is not mentioned or covered in this PR.

PR #5748, opened 2022-12-07
Checkin not possible (EXPOSUREAPP-14387, EXPOSUREAPP-14388) 

schauersbergern:
Fixes both issues:
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14387
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14388


SamuraiKek:
It works now but there's just one thing left to be done. When checking in to an event, you will always see the oboarding consent screen. Even after you have viewed it. It should not show up after the first time.

PR #5753, opened 2022-12-13
Fix: srs error(EXPOSURE-APP 14395)

mtwalli:
Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14395


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/srs/core/server/SrsAuthorizationServer.kt
```diff
@@ -134,33 +136,35 @@ class SrsAuthorizationServer @Inject constructor(
...
-                .currentConfig
-                .first()
-                .selfReportSubmission
-                .common
-                .timeBetweenSubmissionsInDays
-                .toDays()
-        )
-
-        ErrorCode.TIME_SINCE_ONBOARDING_UNVERIFIED,
-        ErrorCode.MIN_TIME_SINCE_ONBOARDING -> {
```
SamuraiKek, line 149: Are these going to be added back when you remove HTTP logging?

mtwalli: this just a refactoring , it is a common function `ErrorCode.errorArgs()`

mtwalli: see https://github.com/corona-warn-app/cwa-app-android/pull/5753/files#diff-0d3cedb31b14d57883b52c3cf3df13de9dad38979bcc8a32626aee0755f1ca48R152

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/srs/core/server/SrsSubmissionApi.kt
```diff
@@ -12,7 +12,7 @@ import retrofit2.http.POST
 interface SrsSubmissionApi {
     @Throws(CwaWebException::class)
     @POST("version/v1/diagnosis-keys")
-    @Headers("Content-Type: application/x-protobuf")
+    @Headers("Content-Type: application/x-protobuf", "cwa-fake: 0")
```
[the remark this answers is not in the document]
mtwalli: Not sure  ,but it is also exist for usual key submission so added it here also for consistency

SamuraiKek (approved):

PR #5758, opened 2022-12-15
Reset otp after submission (DEV)

SamuraiKek:
Resetting the valid otp after a successful srs submission.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/srs/core/server/SrsSubmissionServer.kt
```diff
@@ -23,6 +24,7 @@ import javax.inject.Inject
 class SrsSubmissionServer @Inject constructor(
     srsSubmissionApi: Lazy<SrsSubmissionApi>,
     private val paddingTool: PaddingTool,
+    private val srsSubmissionSettings: SrsSubmissionSettings,
```
mtwalli, line 27: I believe you can do it in `SrsSubmissionRepository` where `srsSubmissionSettings` is already available 
Screenshot 2022-12-15 at 12 58 14

SamuraiKek: Ok.

schauersbergern (dismissed):

mtwalli (dismissed):

PR #5770, opened 2022-12-20
SRS : Teks on fly (EXPOSUREAPP-14489)

mtwalli:
Pass TEKs though srs flow without caching them at all
## How to test?
- Try to use SRS to submit as usual
- Check cached keys storage , it should not have any keys cached from SRS 
- Make sure to stop in the middle of the flow and put the app in background , and use "Don't Keep Activities" in Developer options to make sure that keys are bound correctly to lifecycle 

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14489)


SamuraiKek (approved):
Teks are cached when view gets destroyed.

schauersbergern (approved):

PR #5781, opened 2022-12-27
Update dimensions (EXPOSUREAPP-12796)

SamuraiKek:
Use standardized dimensions everywhere.


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/hometestcards/ui/HomeTestCardsFragment.kt
```diff
@@ -34,7 +34,7 @@ class HomeTestCardsFragment : Fragment(R.layout.fragment_test_home_test_cards_la
         binding.recyclerView.apply {
             layoutManager = LinearLayoutManager(requireContext())
             itemAnimator = DefaultItemAnimator()
-            addItemDecoration(TopBottomPaddingDecorator(topPadding = R.dimen.spacing_tiny))
+            addItemDecoration(TopBottomPaddingDecorator(topPadding = R.dimen.standard_8))
```
mtwalli, line 37: Opinion: I don’t think that we should specify the actual size in the name. `standard_tiny` would be easy to change the actual size if needed and the name is still meaningful. Also if there is a need to override the actual size for different device categories such as phones or tablets
```xml
# Phones 
<dimen name"standard_tiny">8dp</dimen>
# Tablets 
<dimen name"standard_tiny">8dp</dimen>
```
thoughts ?

SamuraiKek: Yes, I get it, but when creating new screens my number 1 gripe with the dimensions namings was that I always had to open the dimensions file to see which corresponded to which.

On some of them I ended up hard coding the value because it was more convenient than having to find the right dimension name. And I've noticed this happen not only with myself but also our collegues who have hardcoded the values for probably the same reason. 

Having the size in the name means I don't have to go looking for the right one in the dimension file and will actually make me want to use the dimensions now instead of hardcoding them.

mtwalli (approved):
let us keep it as is for now

PR #5796, opened 2023-01-10
Extend Ppa for Srs (EXPOSUREAPP-14321)

mtwalli:
### How to test?
- Test analytics before SRS submission -> no contribution 
- Test analytics after SRS submission -> contribution

[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14321)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/keysubmission/srs/AnalyticsSrsKeySubmissionStorage.kt
```diff
@@ -0,0 +1,52 @@
...
+import timber.log.Timber
+import java.io.IOException
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class AnalyticsSrsKeySubmissionStorage @Inject constructor(
+    @AnalyticsSrsKeySubmissionDataStore private val dataStore: DataStore<Preferences>
+) {
+    private val dataStoreFlow = dataStore.data
```
schauersbergern, line 21: could be lazy

mtwalli: yes, it could. why?

schauersbergern: Performance 😁

mtwalli: 😄 I know what `lazy` is for, I meant why here specifically. You refactored the datastores non of them is using lazy when accessing the flow.

schauersbergern: Yes, you are right, didn't realised beforehand, but this would be one additional refactoring which could make the datastore approach a tiny bit more efficient.

mtwalli: It depends, it not necessarily true that using `lazy` will improve the performance. it could make it worse. see https://stackoverflow.com/questions/65543747/why-does-using-by-lazy-make-my-kotlin-code-slower

schauersbergern (approved):

PR #5812, opened 2023-01-20
Dialog causes crash when put in background (EXPOSUREAPP-14569)

SamuraiKek:
[Ticket.](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14569)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/srs/core/repository/SrsSubmissionRepository.kt
```diff
@@ -164,8 +164,7 @@ class SrsSubmissionRepository @Inject constructor(
                 val errorCode = e.type.toSrsErrorType()
                 SrsSubmissionException(
                     errorCode = errorCode,
-                    errorArgs = errorCode.errorArgs(appConfig.selfReportSubmission),
-                    cause = e
```
SamuraiKek, line 168: Passing the cause will lead to `SrsSubmissionException` parcelization errors. We aren't using the cause to display error dialogs anyway.

mtwalli: would `@IgnoredOnParcel` prevent the crash?

SamuraiKek: I'll give it a try

SamuraiKek: Update: nope, still crashing with `@IgnoredOnParcel`

mtwalli (approved):

PR #5821, opened 2023-01-31
Add new Pins (EXPOSUREAPP-14648)

mtwalli:
Add new certificates Pins
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14648)


Corona-Warn-App/src/main/res/xml/network_security_config.xml
```diff
@@ -8,6 +8,8 @@
             expiration="2024-02-12"
             tools:ignore="MissingBackupPin">
```
SamuraiKek, line 9: Can this be removed now that there are more pins?

mtwalli: done

SamuraiKek (approved):

schauersbergern (approved):

PR #5825, opened 2023-02-03
Remove comma at end of journal text (EXPOSUREAPP-14677)

SamuraiKek:
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14677)
Sepparator comma is deleted from the end of composed string in contact journal.

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/overview/adapter/day/contact/ContactAdapter.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/overview/adapter/day/contact/ContactAdapter.kt
@@ -45,7 +45,7 @@ class ContactAdapter : BaseAdapter<ContactAdapter.ContactItemViewHolder>() {
                 ContactItem.Type.LOCATION -> context.getString(R.string.accessibility_location, key.name)
                 ContactItem.Type.PERSON -> context.getString(R.string.accessibility_person, key.name)
             }
-            val attributes = getAttributes(key.duration, key.attributes, key.circumstances)
+            val attributes = getAttributes(key.duration, key.attributes, key.circumstances).dropLast(2)
             if (attributes.isNotEmpty()) {
                 contactDiaryOverviewElementAttributes.text = attributes
             } else contactDiaryOverviewElementAttributes.visibility = View.GONE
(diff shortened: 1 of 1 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/contactdiary/ui/overview/adapter/day/contact/ContactAdapter.kt
```diff
@@ -45,7 +45,7 @@ class ContactAdapter : BaseAdapter<ContactAdapter.ContactItemViewHolder>() {
                 ContactItem.Type.LOCATION -> context.getString(R.string.accessibility_location, key.name)
                 ContactItem.Type.PERSON -> context.getString(R.string.accessibility_person, key.name)
             }
-            val attributes = getAttributes(key.duration, key.attributes, key.circumstances)
+            val attributes = getAttributes(key.duration, key.attributes, key.circumstances).dropLast(2)
```
schauersbergern, line 48: I think this only works when circumstances are null. If circumstances are appended it would chop off the last chars of them

SamuraiKek: Well, that's the fix :) By dropping off the last two chars, it will always eliminate the comma and the space at the end.

schauersbergern: Yes, but if key.circumstances is not null e.g. it is "some circumstances", `getAttributes` wont append `, ` and so the result would be "some circumstanc"

mtwalli: maybe `joinToString` instead of `forEach` is better to avoid the superfluous `,`

SamuraiKek: @schauersbergern, I still don't understand what you mean. `getAttributes` returns a spannable string and inside it, it appends the `", "` (comma + space) sepparator, but it also does this for the last attribute as well. Calling `.dropLast(2)` here will only cut the last `", "` that shouldn't be there in the first place. Have you tried running it on device?

schauersbergern: @mtwalli tested all cases now no superfluous comma present

schauersbergern (approved):

mtwalli (approved):

PR #5829, opened 2023-02-07
Ppa Analytics (EXPOSUREAPP-14695)

mtwalli:
###
[Ticket](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14695)

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/keysubmission/AnalyticsKeySubmissionStorage.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/keysubmission/AnalyticsKeySubmissionStorage.kt
@@ -225,9 +225,6 @@ open class AnalyticsKeySubmissionStorage(
             }
         }
         dataStore.edit { prefs -> keys.forEach { key -> prefs.remove(key) } }
-        Timber.d(
-            dataStore.data.first().asMap().keys.toString()
-        )
     }
 }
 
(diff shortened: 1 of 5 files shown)


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/server/DataDonationAnalyticsServer.kt
```diff
@@ -19,14 +19,15 @@ class DataDonationAnalyticsServer @Inject constructor(
             Timber.d("Response code: %d", it)
         }
 
-        return when (code) {
-            204 -> Timber.d("Analytics upload completed successfully")
-            400, 401, 403 -> {
+        return when {
+            response.isSuccessful -> Timber.d("Analytics upload completed successfully")
+            code in listOf(400, 401, 403) -> {
```
mtwalli, line 24: this is different issue, response code is 200 now, I use isSuccessful instead to cover all possible success codes

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/datadonation/analytics/modules/keysubmission/AnalyticsKeySubmissionStorage.kt
```diff
@@ -216,7 +217,17 @@ open class AnalyticsKeySubmissionStorage(
...
+            if (sharedPrefKeySuffix.isEmpty()) {
+                it.name.endsWith(AnalyticsRAKeySubmissionStorage.sharedPrefKeySuffix).not()
+            } else {
+                it.name.endsWith(AnalyticsRAKeySubmissionStorage.sharedPrefKeySuffix)
+            }
+        }
+        dataStore.edit { prefs -> keys.forEach { key -> prefs.remove(key) } }
+        Timber.d(
+            dataStore.data.first().asMap().keys.toString()
+        )
```
mtwalli, line 230: clear before was clearing all keys regardless of which Test Type storage triggered `clear` and this lead to what ever test is registered last it will clear pervious test collected data not only the relevant test type

schauersbergern (approved):

PR #5837, opened 2023-02-10
Use appScope (14703)

mtwalli:
Report flag in appScope , vm scope was cancelled after navigating from screen 🤷‍♂️ 
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14703

### DIFF
--- a/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
+++ b/Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
@@ -89,7 +89,7 @@ class SubmissionSymptomCalendarViewModel @AssistedInject constructor(
             )
         }
         performSubmission {
-            launch {
+            appScope.launch {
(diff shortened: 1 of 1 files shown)


schauersbergern (approved):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/submission/symptoms/calendar/SubmissionSymptomCalendarViewModel.kt
```diff
@@ -89,7 +89,7 @@ class SubmissionSymptomCalendarViewModel @AssistedInject constructor(
             )
         }
         performSubmission {
-            launch {
+            appScope.launch {
```
schauersbergern, line 92: Could this also lead to problems elsewhere? vm scope is often used after navigation

mtwalli: the navigation still performed from vm scope. the issue here is that reporting the flag was also done from vm scope and we navigate immediately which cancels vm scope as we leave the screen and there was no chance to update the flag

PR #5851, opened 2023-02-20
EOL - onboarding (EXPOSUREAPP-14788)

SamuraiKek:
[Subtask](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14788)

Changes to the onboarding flow after EOL, skipping most of the screens. 
For testing, ~~changing the EOL date from the test manu won't work as the onboarding flow doesn't have access to the test menu. You'll have to change the hardcoded EOL date from the `EolSettings` class.~~ go into the test menu, change the EOL date to sometime in the past and reset the app from the settings menu.


Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingPrivacyViewModel.kt
```diff
@@ -1,16 +1,28 @@
...
 import de.rki.coronawarnapp.util.viewmodel.CWAViewModel
 import de.rki.coronawarnapp.util.viewmodel.SimpleCWAViewModelFactory
 
-class OnboardingPrivacyViewModel @AssistedInject constructor() : CWAViewModel() {
+class OnboardingPrivacyViewModel @AssistedInject constructor(
+    eol: AppEol,
+    dispatcherProvider: DispatcherProvider = DefaultDispatcherProvider(),
+) : CWAViewModel() {
     val routeToScreen: SingleLiveEvent<OnboardingNavigationEvents> = SingleLiveEvent()
+    val isEol = eol.isEol.asLiveData(context = dispatcherProvider.Default)
```
mtwalli, line 18: You are exposing `isEol` from the view model to the fragment which keeps track of the is flag and then when a button is clicked you pass the flag back to the view model. why don’t you just check in the viewmodel upon button click ?
```kotlin
fun onNextButtonClick() = launch{
 routeToScreen.postValue( 
    if(eol.isEol.first()) NavigateToMainActivity else NavigateToOnboardingTracing
 )
}
```

mtwalli:
> [Subtask](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14788)
> 
> Changes to the onboarding flow after EOL, skipping most of the screens. For testing, changing the EOL date from the test manu won't work as the onboarding flow doesn't have access to the test menu. You'll have to change the hardcoded EOL date from the `EolSettings` class.

I just tested it using my approach and I can see the right flow based on EolSettings from DEV menu. Please double check it ?

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/onboarding/OnboardingPrivacyViewModel.kt
SamuraiKek: Yes, true. I'll update it.

SamuraiKek:
> > [Subtask](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14788)
> > Changes to the onboarding flow after EOL, skipping most of the screens. For testing, changing the EOL date from the test manu won't work as the onboarding flow doesn't have access to the test menu. You'll have to change the hardcoded EOL date from the `EolSettings` class.
> 
> I just tested it using my approach and I can see the right flow based on EolSettings from DEV menu. Please double check it ?

Yes, it works.

mtwalli (approved):

PR #5861, opened 2023-02-23
EOL- Test Menu - allow logger to work after EOL (EXPOSUREAPP-14849)

mtwalli:
- [x] Allow logger after EOL for testers ONLY
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14849


Ein-Tim:
Is this related to the error logging feature in the app, which the user can access? Or is "logger" another logger in this case? :D

mtwalli:
> Is this related to the error logging feature in the app, which the user can access? Or is "logger" another logger in this case? :D

No, it is for testing purposes

schauersbergern (approved):

PR #5864, opened 2023-02-23
EOL - default stats (EXPOSUREAPP-14836)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14836


Corona-Warn-App/src/main/assets/default_stats/default_stats.bin
```diff
@@ -0,0 +1,3 @@
+
+1
+�ҟ�%https://corona-pandemieradar.de/:lang
```
SamuraiKek, line 3: what's up with the weird characters before the url?

mtwalli: it is a `bin` file when it is decoded it works fine

SamuraiKek (approved):

PR #5875, opened 2023-02-28
Fix : close X button in dcc validation screens (EXPOSUREAPP-14870)

mtwalli:
## Ticket
https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14870


SamuraiKek (dismissed):

Corona-Warn-App/src/main/java/de/rki/coronawarnapp/ui/view/CovidCertificateValidationResultView.kt
```diff
@@ -82,4 +82,8 @@ class CovidCertificateValidationResultView @JvmOverloads constructor(
     override fun setOnClickListener(l: OnClickListener?) {
```
SamuraiKek, line 82: Do we need this one then? I don't think it's used anywhere else and it's basically replaced by `setNavigationOnClickListener`

SamuraiKek (approved):

PR #5906, opened 2023-03-21
Update privacy notice for EOL (EXPOSUREAPP-14955)

SamuraiKek:
[Ticket.](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14955)

Please compare the contents of the word documents found in the ticket with the changes made to the privacy files.

PR #5907, opened 2023-03-22
Fix privacy notes (DEV)

SamuraiKek:
Addressing the comments from #5906 
Please have another look.

### DIFF
--- a/Corona-Warn-App/src/main/assets/privacy_de.html
+++ b/Corona-Warn-App/src/main/assets/privacy_de.html
@@ -1,5 +1,5 @@
 <h1>
-    Hinweise zur Beendigung der Verarbeitung personenbezogener
+    Hinweise zur Beendigung der Verarbeitung personenbezogener Daten
 </h1>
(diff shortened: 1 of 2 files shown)


Corona-Warn-App/src/main/assets/privacy_de.html
MikeMcC399, line 1: In line 1
"Hinweise zur Beendigung der Verarbeitung personenbezogener"

is incomplete. I would expect "Daten" at the end.

MikeMcC399 (changes requested):
Word missing at end of `h1` at top of document.

image

Corona-Warn-App/src/main/assets/privacy_de.html
SamuraiKek: Thanks, I will update it.

MikeMcC399 (approved):
Changed heading (DE) looks fine now. 👍🏻

OLVI5 (approved):

PR #5912, opened 2023-03-29
Fix invalid certs wru z (EXPOSUREAPP-14998)

SamuraiKek:
[Ticket.](https://jira-ibs.wbs.net.sap/browse/EXPOSUREAPP-14998)

Possible fix for testers to keep the certificates valid when switching to WRU-Z after EOL.


Corona-Warn-App/src/deviceForTesters/java/de/rki/coronawarnapp/test/debugoptions/ui/DebugOptionsFragmentViewModel.kt
```diff
@@ -52,7 +55,9 @@ class DebugOptionsFragmentViewModel @AssistedInject constructor(
     }
 
     private fun cleanCachedData() = appScope.launch {
-        environmentSunset.reset()
+        if (!eol.isEol.first()) {
```
schauersbergern, line 58: Better add this check before calling cleanCachedData() or rename cleanCachedData() to cleanCachedDataIFEol()

SamuraiKek: Ok, I'll rename it. I added it here because we were already doing `.launch{}` so I could read the value from the flow.

schauersbergern (approved):

